PR #67, opened 2021-11-20
Fix a few URLs to the right website

mwestphal:


mwestphal:
@Meakk please quick review

GENERATE.md
```diff
@@ -1,3 +1,4 @@
+### OUT OF DATE ###
```
michaelmigliore, line 1: Then remove the file?

mwestphal: it still contains relevant information for website generation, but updating it does not make much sense. This will be removed once we stop using kw-doc

michaelmigliore (approved):

PR #76, opened 2021-11-21
Improve CI

mwestphal:
- Build and Test VTK 9.0.0, as we claim to support it in the ReadMe, which requires libXt in the docker image
- Use short SHA for VTK version

### DIFF
--- a/.github/docker/Dockerfile
+++ b/.github/docker/Dockerfile
@@ -12,6 +12,7 @@ RUN apt update && apt install -y \
     help2man \
     lcov \
     libx11-dev \
+    libxt-dev \
     openssh-client \
     software-properties-common \
     wget \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,11 +116,11 @@ jobs:
         path: './dependencies/vtk'
         ref: ${{env.VTK_SHA}}
 
-    - name: Patch VTK 9.0.1 source macos
+    - name: Patch VTK 9.0 source macos
       if: |
         steps.cache-vtk.outputs.cache-hit != 'true' &&
         matrix.os == 'macos-latest' &&
-        matrix.vtk_version == 'v9.0.1'
+        matrix.vtk_version == 'v9.0.0'
       working-directory: ${{github.workspace}}/dependencies/vtk
       run: |
         sed -i '' '57i\'$'\n''herr_t H5O__fsinfo_set_version(H5F_t *f, H5O_fsinfo_t *fsinfo);' ThirdParty/hdf5/vtkhdf5/src/H5Fsuper.c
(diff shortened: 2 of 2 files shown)


mwestphal:
@Meakk : I tried to separate the big job unto multiple jobs, that requires to duplicate the build matrix unless doing some crazy workarounds like this: https://github.community/t/how-to-share-matrix-between-jobs/128595/8 and does not give the ability to run a single job: https://github.community/t/ability-to-rerun-just-a-single-job-in-a-workflow/17234/80

Definitelly not worth it, lets keep as it is.

mwestphal:
@Meakk : I've been able to use short hash, but that requires to clone bare VTK first, which take 20s. Acceptable imo.

mwestphal:
I took a look into adding raytracing to coverage report. That require to add ospray AND a VTK with Raytracing enabled to the CI, increasing complexity and file length for little benefit. not worth it, it should be done once we figure out how to package ospray proper. @Meakk

.github/workflows/ci.yml
```diff
@@ -87,20 +87,26 @@ jobs:
...
+        key: vtk-${{matrix.vtk_version}}-${{matrix.os}}
 
-    - name: Setup VTK
+    - name: Setup VTK and clone bare
       if: steps.cache-vtk.outputs.cache-hit != 'true'
       working-directory: ${{github.workspace}}/dependencies
       run: |
         mkdir vtk_build
         mkdir vtk_install
+        git clone --bare https://github.com/Kitware/VTK.git bare
```
Meakk, line 103: meh, keeping the long sha is better than doing extra work imho.

.github/docker/Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN apt update && apt install -y \
     help2man \
     lcov \
     libx11-dev \
+    libxt-dev \
```
Meakk, line 15: why is it needed?

.github/workflows/ci.yml
mwestphal: this extra work is done only when building VTK, not when using the cache.

mwestphal: Still not worth it ?

.github/docker/Dockerfile
mwestphal: vtk 9.0.0 needs it

.github/workflows/ci.yml
Meakk: It adds complexity to the CI. You think it worth it?

.github/docker/Dockerfile
Meakk: ok

.github/workflows/ci.yml
mwestphal: I think this is fine to add if that helps us managing the CI

Meakk: ok then

Meakk (approved):

PR #104, opened 2021-12-05
Better codecov conf

mwestphal:
- wait all 3 build before reporting
- color range between 70 and 100
- round down with 2 precision
- treshold of 0.1% for success in ci
- require coverange change for commenting


Meakk (approved):

mwestphal:
@Meakk : this does not seems to work as expected, investigation needed.

PR #132, opened 2021-12-09
DRAFT: Add 2D interactor and image readers

### LINKED ISSUE #49: Consider adding a 2D interactor
If an image is 2D, it can be useful to have a 2D interactor (I'm thinking of DXF or medical images).
And tbh, if we implement it, then it makes sense to supports JPG/PNG/TIFF/HDR/TGA/... too since it is basically free.
I know this is not the main purpose of f3d but since f3d is GPU accelerated, it may outperform other image previewers with large images.

jpouderoux:
Fixes #49 

- [x] Add 2D interactor
- [x] Allow to switch between 2D and 3D interactors with `2` and `3` keys
- [x] Set the 2D interactor automatically for 2D data
- [x] Disable lighthing for 2D image data
- [ ] Set image scalars visibility by default for 2D image data
- [ ] Tests

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -67,10 +67,12 @@ declare_f3d_reader(F3D3DSReader)
 declare_f3d_reader(F3DCityGMLReader)
 declare_f3d_reader(F3DDICOMReader)
 declare_f3d_reader(F3DGLTFReader)
+declare_f3d_reader(F3DJPEGReader)
 declare_f3d_reader(F3DMetaImageReader)
 declare_f3d_reader(F3DNrrdReader)
 declare_f3d_reader(F3DOBJReader)
 declare_f3d_reader(F3DPLYReader)
+declare_f3d_reader(F3DPNGReader)
 declare_f3d_reader(F3DPTSReader)
 declare_f3d_reader(F3DSTLReader)
 declare_f3d_reader(F3DTIFFReader)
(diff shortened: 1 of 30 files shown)


mwestphal:
> Allow to switch between 2D and 3D interactors with 2 and 3 keys

There is already a hotkey to switch between interactors, why not cycle through available interactors ?

jpouderoux:
> > Allow to switch between 2D and 3D interactors with 2 and 3 keys
> 
> There is already a hotkey to switch between interactors, why not cycle through available interactors ?

Which key are you thinking about? `K` ?

mwestphal:
> > > Allow to switch between 2D and 3D interactors with 2 and 3 keys
> > 
> > 
> > There is already a hotkey to switch between interactors, why not cycle through available interactors ?
> 
> Which key are you thinking about? `K` ?

yes

jpouderoux:
> > > > Allow to switch between 2D and 3D interactors with 2 and 3 keys
> > > 
> > > 
> > > There is already a hotkey to switch between interactors, why not cycle through available interactors ?
> > 
> > 
> > Which key are you thinking about? `K` ?
> 
> yes

`K` enables trackball in 3D mode, looks a bit different to me and mixing the 2 concepts looks a bit confusing to me, don't you think?

mwestphal:
maybe we should discuss a bit about the right way to integrate this. There a few possibilities imo.

mwestphal:
> `K` enables trackball in 3D mode, looks a bit different to me and mixing the 2 concepts looks a bit confusing to me, don't you think?

Thinking about it, I agree

PR #133, opened 2021-12-09
Add a signal handler for winshellext

### LINKED ISSUE #134: explorer.exe broken after installing the thumbnail support
**Describe the bug**
After installing f3d on windows using the installer (or registering the shell extension), a bad file could segfault f3d and break the explorer completelly until unregistration of the winshellext

https://user-images.githubusercontent.com/3129530/145490946-792006de-fc46-4b64-b133-9623719c2ad5.mp4

**System Information:**
 - OS: Windos

mwestphal:
Fix #134

### DIFF
--- a/src/winshellext/F3DThumbnailProvider.cxx
+++ b/src/winshellext/F3DThumbnailProvider.cxx
@@ -96,7 +96,7 @@ HRESULT CreateBitmapFromBuffer(vtkImageData* image, HBITMAP* phbmp)
 }
 
 //------------------------------------------------------------------------------
-void SignalHandler(int signum)
+void SignalHandler(int)
 {
   // we cannot exit here as it will actually exit explorer.exe,
   // we just keep executing the software in case of signal.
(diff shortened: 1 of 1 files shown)


Meakk (approved):
That seems too easy ^^
Have you tested it with broken files?

mwestphal:
I have tested with a raise(SIGSEGV), please test the actual package from the ci with your file.

src/winshellext/F3DThumbnailProvider.cxx
```diff
@@ -93,6 +94,14 @@ HRESULT CreateBitmapFromBuffer(vtkImageData* image, HBITMAP* phbmp)
 
   return NOERROR;
 }
+
+//------------------------------------------------------------------------------
+void SignalHandler(int signum)
+{
+  std::cerr << "Interrupt signal (" << signum << ") received.\n";
```
jpouderoux, line 101: [Doc](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=msvc-170) says
"Do not issue low-level or STDIO.H I/O routines (for example, printf or fread)." so you should remove this (useless anyway) output.

mwestphal: good point

src/winshellext/F3DThumbnailProvider.cxx
```diff
@@ -160,6 +168,15 @@ IFACEMETHODIMP F3DThumbnailProvider::GetThumbnail(UINT cx, HBITMAP* phbmp, WTS_A
   char** argv;
   ::ConvertWCommandLineToArgs(warg, argc, argv);
 
+  // Shell extension CANNOT segfault, we must catch the signal
+  // or it can completelly break explorer.exe
+  signal(SIGTERM, ::SignalHandler);
```
jpouderoux, line 173: Shouldn't we put this in DLL initializing func or restore the default handlers at the end of this function call?

mwestphal: > Shouldn't we put this in DLL initializing func 

we could, but the code before cant segfault unless we made a big mistake.

> restore the default handlers at the end of this function call?

Not sure we can recover the default handlers

jpouderoux: See the doc example. The function signal returns the previous handler. Just restore it at the end.

mwestphal:
This approach is not going to work. We need something way more robust to not break users explorer.exe.
CreateProcess seems needed, which calls into question the whole libf3d stuff.

PR #137, opened 2021-12-12
FIX: fix potential Windows Explorer crash when trying to generate thumbnails for broken files

jpouderoux:
Instead of using f3d as a library and thumbnail image fetched in a buffer,
the F3D executable is ran in a separate process so if it crashes it does
not crash Explorer.
Note that this way of processing is way slower as:
 - a new process must created, and the VTK libraries must be loaded
   for each file/thumbnail;
 - the thumbnail image is created into a temporary PNG file by F3D,
   then the file is read back by the handler in order to pass the
   image buffer to the Explorer thumbnail manager;
 - the temporary file is deleted.

### DIFF
--- a/src/winshellext/CMakeLists.txt
+++ b/src/winshellext/CMakeLists.txt
@@ -1,4 +1,12 @@
+set(F3D_WINDOWS_THUMBNAIL_TIMEOUT 8000 CACHE STRING "Windows thumbnail creation timeout in milliseconds")
+
+configure_file(
+  "${CMAKE_CURRENT_SOURCE_DIR}/F3DThumbnailConfig.h.in"
+  "${CMAKE_CURRENT_BINARY_DIR}/F3DThumbnailConfig.h"
+  @ONLY)
+
 add_library(F3DShellExtension SHARED
+  "${CMAKE_CURRENT_BINARY_DIR}/F3DThumbnailConfig.h"
   F3DShellExtension.cxx
   F3DShellExtensionClassFactory.h
   F3DThumbnailProvider.h
@@ -8,9 +16,9 @@ add_library(F3DShellExtension SHARED
 
 target_compile_definitions(F3DShellExtension PUBLIC UNICODE)
 
-target_include_directories(F3DShellExtension PUBLIC ${CMAKE_BINARY_DIR})
+target_include_directories(F3DShellExtension PUBLIC ${CMAKE_BINARY_DIR} ${CMAKE_CURRENT_BINARY_DIR})
 
-target_link_libraries(F3DShellExtension PUBLIC libf3d PRIVATE pathcch shlwapi)
+target_link_libraries(F3DShellExtension PUBLIC libf3d PRIVATE pathcch shlwapi windowscodecs)
 
 set_target_properties(F3DShellExtension PROPERTIES
   RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin")
--- a/src/winshellext/F3DThumbnailProvider.cxx
+++ b/src/winshellext/F3DThumbnailProvider.cxx
@@ -2,6 +2,7 @@
 
 #include "F3DException.h"
 #include "F3DLoader.h"
+#include "F3DThumbnailConfig.h"
 #include "vtkF3DObjectFactory.h"
 
 #include <vtkImageData.h>
@@ -11,7 +12,10 @@
 #include <codecvt>
 #include <locale>
 #include <pathcch.h>
+#include <shlwapi.h>
 #include <sstream>
+#include <thumbcache.h>
+#include <wincodec.h>
 
 extern HINSTANCE g_hInst;
 extern long g_cDllRef;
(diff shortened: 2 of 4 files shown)


Meakk:
Awesome! Thanks.  
Looks good, I'll test the generated package tomorrow and do some test.  
Please fix the warning (you can set `F3D_STRICT_BUILD` to reproduce locally).

src/winshellext/F3DThumbnailProvider.cxx
```diff
@@ -151,42 +153,110 @@ IFACEMETHODIMP F3DThumbnailProvider::Initialize(LPCWSTR pszFilePath, DWORD)
...
-      ::CreateBitmapFromBuffer(image, phbmp);
-      *pdwAlpha = WTSAT_ARGB;
-      ret = NOERROR;
-    }
+  wcscat_s(image_filename, L".png");
+
+  // Create command to run
+  wchar_t command[MAX_PATH * 3 + 20];
+  swprintf_s(command, MAX_PATH * 3 + 20,
+    L"\"%s\" --input \"%s\" --output \"%s\" --dry-run -sta --no-background --resolution %d,%d\"",
```
mwestphal, line 176: probably need --quiet

jpouderoux: It was forgot in previous handler but yes, let's add it.

src/winshellext/F3DThumbnailProvider.cxx
```diff
@@ -11,87 +11,88 @@
...
 #include <pathcch.h>
+#include <shlwapi.h>
 #include <sstream>
+#include <thumbcache.h>
+#include <wincodec.h>
 
 extern HINSTANCE g_hInst;
 extern long g_cDllRef;
 
+
```
mwestphal, line 22: empty line

mwestphal:
> Note that this way of processing is way slower as

I agree but this issue is a must fix, it may be improved for the 1.3 or after with a smarter approach.

mwestphal:
Tested:

0. Issue IS fixed ! Thanks @jpouderoux 
1. Thumbnail generation is only slightly slower, definitelly not a deal breaker, fine to keep as is imo
2. There is a related issue : #139 , but out of scope of this pr

Meakk:
Thanks for testing @mwestphal. I agree, we should merge it in `release` first.

Meakk:
I have changed the target branch to `release` and triggered a new CI.

src/winshellext/F3DThumbnailProvider.cxx
```diff
@@ -151,42 +153,110 @@ IFACEMETHODIMP F3DThumbnailProvider::Initialize(LPCWSTR pszFilePath, DWORD)
...
+  si.cb = sizeof(STARTUPINFO);
+  si.dwFlags |= STARTF_USESTDHANDLES;
+
+  if (!CreateProcess(
+        m_f3dPath, command, nullptr, nullptr, TRUE, CREATE_NO_WINDOW, nullptr, nullptr, &si, &pi))
+  {
+    return HRESULT_FROM_WIN32(GetLastError());
+  }
+
+  WaitForSingleObject(pi.hProcess, INFINITE);
```
Meakk, line 194: Maybe we can add a reasonable timeout? If the user has a very big STEP file for instance, it will take forever to render, so maybe it's better to skip it?

jpouderoux: So we have to decide the timeout value :) 5 secs?

mwestphal: 5 secs seems a bit too much 2/3 secs seems more reasonnable.

Meakk: I would say the opposite, 10 secs :)

mwestphal: haha, @jpouderoux do you have an idea on how long WaterBottle takes ?

jpouderoux: I propose to set a 8sec timeout by default. It can be adjusted by a new CMake option.

Meakk: Sound good to me

mwestphal: Fair enough.

Meakk (approved):

src/winshellext/CMakeLists.txt
```diff
@@ -1,4 +1,13 @@
+set(F3D_WINDOWS_THUMBNAIL_TIMEOUT 8000 CACHE STRING "Windows thumbnail creation timeout in milliseconds")
+
+configure_file(
+  "${CMAKE_CURRENT_SOURCE_DIR}/F3DThumbnailConfig.h.in"
+  "${CMAKE_CURRENT_BINARY_DIR}/F3DThumbnailConfig.h"
+  @ONLY)
+
+
```
mwestphal, line 8: empty line

Meakk:
Can we move forward and merge it? I would like to release 1.2.1 as soon as possible.

PR #141, opened 2021-12-12
Improve quiet option

### LINKED ISSUE #139: --quiet is not quieting VTK warnings and errors on windows, yet it could
**Describe the bug**
--quiet is not quieting VTK warnings and errors on windows, yet it could

**Additional context**
While F3D cannot quiet VTK warnings and errors displaying directly into the terminal on unix, since windows has dedicated output window code, it is possible to quiet it, and it would be preferrable as, currently, the output window can pop up when showing thumbnails.

FYI @Meakk @jpouderoux

mwestphal:
Improve --quiet option so that even vtk warnings and errors are impacted.
Disable vtkLogger completely

Fix #139

### DIFF
--- a/src/F3DLoader.cxx
+++ b/src/F3DLoader.cxx
@@ -16,6 +16,7 @@
 #include <vtkCamera.h>
 #include <vtkDoubleArray.h>
 #include <vtkImageData.h>
+#include <vtkLogger.h>
 #include <vtkNew.h>
 #include <vtkPointGaussianMapper.h>
 #include <vtkPolyDataMapper.h>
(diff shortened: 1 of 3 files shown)


src/F3DLog.cxx
```diff
@@ -43,5 +36,6 @@ void F3DLog::SetUseColoring(bool use)
 //----------------------------------------------------------------------------
 void F3DLog::SetQuiet(bool quiet)
 {
-  F3DLog::Quiet = quiet;
```
Meakk, line 46: You probably want to remove the `Quiet` variable in the header too.

Meakk (changes requested):

src/F3DLoader.cxx
```diff
@@ -41,6 +42,12 @@ typedef struct ProgressDataStruct {
 //----------------------------------------------------------------------------
 F3DLoader::F3DLoader()
 {
+  // Disable vtkLogger in case VTK was built with log support
+  vtkLogger::SetStderrVerbosity(vtkLogger::VERBOSITY_OFF);
```
Meakk, line 46: not sure what it does.

mwestphal: // Disable vtkLogger in case VTK was built with log support 

^^

Meakk (approved):

PR #146, opened 2021-12-12
Added DEB to cPack generator

whiiiskyy:
DEB was added to cmake. That should generate a working DEB package to be installed in any DEB system like Debian, Mint, Ubuntu, etc.

### DIFF
--- a/cmake/packaging.cmake
+++ b/cmake/packaging.cmake
@@ -74,7 +74,7 @@ if(WIN32 AND NOT UNIX)
 
 else()
   if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
-    set(CPACK_GENERATOR TGZ TXZ)
+    set(CPACK_GENERATOR TGZ TXZ DEB)
   elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
     set(CPACK_GENERATOR DragNDrop)
     set(CPACK_PACKAGE_ICON "${CMAKE_SOURCE_DIR}/resources/logo.icns")
--- a/cmake/packaging.cmake
+++ b/cmake/packaging.cmake
@@ -74,6 +74,7 @@ if(WIN32 AND NOT UNIX)
 
 else()
   if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+    set(CPACK_PACKAGE_CONTACT "Whisley <whisley.santos@gmail.com>")
     set(CPACK_GENERATOR TGZ TXZ DEB)
   elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
     set(CPACK_GENERATOR DragNDrop)


mwestphal:
I think a bunch of cpack var **must** be set:

- CPACK_DEBIAN_PACKAGE_ARCHITECTURE
- CPACK_DEBIAN_PACKAGE_DEPENDS (should be minimal, but still)
- CPACK_PACKAGE_CONTACT

More info here: https://gitlab.kitware.com/cmake/community/-/wikis/doc/cpack/PackageGenerators

@Meakk : There is a .rpm packager as well, should we try it ?

@Meakk : This could go into release I think, worth it to improve F3D reach.

Meakk:
@whiiiskyy there is an error:

```
CMake Error at /usr/share/cmake-3.22/Modules/Internal/CPack/CPackDeb.cmake:517 (message):
CPack: Create package using DEB
  CPackDeb: Debian package requires a maintainer for a package, set
CPack: Install projects
  CPACK_PACKAGE_CONTACT or CPACK_DEBIAN_PACKAGE_MAINTAINER
```

You can set the variables just before the line you modified:  
```cmake
set(CPACK_PACKAGE_CONTACT "Your name <your@email.com>")
```

I think it should be enough, `CPACK_DEBIAN_PACKAGE_ARCHITECTURE` default value is fine and `CPACK_DEBIAN_PACKAGE_DEPENDS` should be empty I guess. We'll see what the resulting `.deb` file will look like.

@mwestphal 
Sure we should consider RPM later if someone is interested to implement and testing it.  
I agree it should target the `release` branch, I'll change it.

mwestphal:
@Meakk : CPACK_DEBIAN_PACKAGE_DEPENDS empty ? I think we do need to add some low level package like libc and such, unless they are considered always available. maybe @mzf-guest has inputs here.

mwestphal:
> Sure we should consider RPM later if someone is interested to implement and testing it.

Sounds good to me !

mwestphal:
> Sure we should consider RPM later if someone is interested to implement and testing it.

Sounds good to me !

whiiiskyy:
@Meakk I have made the change as requested.
Weirdly, the error I saw was regarding a `problem to compress the directory`.
Let's see how this go. I will pay more attention to it.

@mwestphal `CPACK_DEBIAN_PACKAGE_DEPENDS` has no dependencies, only a library under `/lib`
I am not sure if it is already counting on something installed in the system.

As far as CPack documentation goes, we should be good without it.

```
Sets the Debian dependencies of this package.

    Mandatory : NO
    Default :
```

Meakk:
@whiiiskyy the workflow succeeded, and you can download the `.deb` files here: https://github.com/f3d-app/f3d/actions/runs/1572064138
You probably want `f3d-ubuntu-latest-568b279cd2` archive.

Regarding the `problem to compress the directory` error, it's probably related to permission access. After the build, try to install F3D in a directory accessible for the user (set the `CMAKE_INSTALL_PREFIX` variable to something different from `/usr`).

For `CPACK_DEBIAN_PACKAGE_DEPENDS`, maybe we need OpenGL, X11, GLIBC... I'm not sure here.

whiiiskyy:
@Meakk It is working like a charm:

- no errors during the installation via the package manager
- F3D appears under the applications menu as expected
- thumbnails are working as expected. I deleted `.cache` to make sure it would load the thumbnails and it did
- f3d opens as expected
- double click on a STL opens f3d as expected
- I can call f3d via terminal and it opens as expected
- uninstall works as expected
- installing it via terminal `dpkg -i` works as expected

Is there anything else you would like me to check?

Thanks

whiiiskyy:
Regarding `After the build, try to install F3D in a directory accessible for the user (set the CMAKE_INSTALL_PREFIX variable to something different from /usr).`, you guys already cover this with `.tar.gz` and `.tag.xz` which allows me to run f3d from `~/Downloads` which doesn't require root access

It is worth it to mention tho, installing f3d as non-root will be as raw as it can get. You cannot get it 100% running as a system application like be able to search for it under the applications menu and those sort of things.

As far as I tested before contacting you guys, the thumbnails didn't work either. Only after I moved the folders to the right place under `/usr` as root and deleted the `.cache` folder.

When installing as root, you have the return below.
You cannot reach the same result running/installing it as a non-root user.
Linux users know of this "rule". 
Luckily, f3m works well without root access while many others will crash or require AppImage.

```
dpkg -i f3d-1.2.0-Linux.deb                                  ░▒▓ ✔ │ root@OptiPlex-7040 
Selecting previously unselected package f3d.
(Reading database ... 498177 files and directories currently installed.)
Preparing to unpack f3d-1.2.0-Linux.deb ...
Unpacking f3d (1.2.0) ...
Setting up f3d (1.2.0) ...
Processing triggers for gnome-menus (3.36.0-1ubuntu1) ...
Processing triggers for desktop-file-utils (0.24+linuxmint1) ...
Processing triggers for mime-support (3.64ubuntu1) ...
Processing triggers for hicolor-icon-theme (0.17-2) ...
Processing triggers for man-db (2.9.1-1) ...
Processing triggers for shared-mime-info (1.15-1) ...


```

mwestphal:
> You cannot get it 100% running as a system application like be able to search for it under the applications menu and those sort of things.

You can, just read the updated doc about DesktopIntegration : https://github.com/f3d-app/f3d#linux

Of course, installing .deb package requires root, which is expected. so all good on my side.

mwestphal:
> Is there anything else you would like me to check?

Drag and drop ? In any case, if it does not work, it is not related to the deb package but most likely to your file manager.

mwestphal (approved):

mwestphal:
@Meakk @whiiiskyy : fine to merge as his imo.

whiiiskyy:
> Drag and drop ? In any case, if it does not work, it is not related to the deb package but most likely to your file manager.

It works with files and with a folder using left/right to navigate.
Thanks for letting me know regarding the `.desktop`

Thank you and @Meakk so much for all the help and patience :)

Meakk:
This is good news! Thanks for your contribution @whiiiskyy.  
If you have time and motivation to try to implement and test the same thing for RPM packages, it would be great :)

whiiiskyy:
Sure thing. 

That will require a little more time to understand CentOS as a desktop instead of as a server.
That should cover the most common RPM desktop distro like CentOS, SUSE, Fedora and Mandriva.

I'll let you know.

Issue #108, opened 2021-12-05
Wrong computation of clip planes when specifying the camera position with certain datasets


mwestphal:
**Describe the bug**
Clipping planes are wrongly computed with large dataset when specyfying the camera positiong
![a](https://user-images.githubusercontent.com/3129530/144748193-9b5e2c19-f547-47ae-8780-e7a4342f9e94.png)

**To Reproduce**
 0. https://sketchfab.com/3d-models/sea-keep-lonely-watcher-09a15a0c14cb4accaf060a92bc70413d#download
 1. Open the file using `f3d source/Stronghold.fbx  --camera-position=7.32912,363.741,1003.36`
 2. nothing appears
 3. move the camera
 4. rendering appears

**Expected behavior**
Correct rendering, like this:
![a](https://user-images.githubusercontent.com/3129530/144748252-927a9cad-ce60-45b0-8929-81e023423db0.png)

**System Information:**
 - OS: ArchLinux
 - GPU and GPU driver: Intel chipset

**F3D Information**
Paste the content of `f3d --version`:

```
[glow@violet ~/Downloads/models/sea]$ ~/Downloads/f3d-1.2.0-Linux/bin/f3d --version                                                       
f3d 1.2.0

F3D - A fast and minimalist 3D viewer
Version: 1.2.0
Build date: 2021-12-04 19:23:10
System: Linux-5.11.0-1021-azure x86_64
Compiler: GNU 7.5.0
RayTracing module: OFF
Exodus module: ON
OpenCASCADE module: 7.6.0 (no metadata)
Assimp module: 5.1.0
VTK version: 9.1.20211124 (build 20211124)

Copyright (C) 2019-2021 Kitware SAS.
Copyright (C) 2021-2022 Michael Migliore, Mathieu Westphal.
License BSD-3-Clause.
Written by Michael Migliore, Mathieu Westphal and Joachim Pouderoux.
```

**Additional context**

mwestphal:
FYI @Meakk

schuhumi:
I have a problem with wrong clipping planes as well. This is the desired result, that shows up when I move the view or zoom in/out
![Bildschirmfoto vom 2021-12-13 23-43-28](https://user-images.githubusercontent.com/5825985/145901001-314a31bf-ba56-4c7c-9220-6ec6516bc388.png)

This is what it looks like upon opening though:
[...]

PR #164, opened 2021-12-14
fix clipping when setting camera position via arguments/options

schuhumi:
as per https://github.com/f3d-app/f3d/issues/108#issuecomment-993001731

Fix issue #108


Meakk:
Thanks! Is it fixing your issue?

schuhumi:
Yes, it is fixing it! Sorry I wrote that in issue 108 only

Meakk:
We'll need a test to ensure no regression, don't worry it's very simple with f3d, please see `cmake/testing.cmake`.  
- Add a line in `cmake/testing.cmake`, probably something like `f3d_test(NAME TestCameraClipping DATA erythrocyte2_.obj ARGS --camera-position=0,0,39 --camera-focal-point=0,0,0 --camera-view-up=1,0,0 --camera-view-angle=20)`  
- Copy `erythrocyte2_.obj` and `erythrocyte2_.mtl` to `data/testing` folder.
- Copy the expected resulting image in `data/baselines/TestCameraClipping.png`

If you do not want or are not allowed to include your data, please use any existing data in `data/testing` folder.

Meakk:
@schuhumi you can add `RESOLUTION 800,600` in your `f3d_test` declaration and remove it from the config file.

schuhumi:
@Meakk Ohh okay, is there a way to run test locally too? I'm not proficient with cmake and didn't understand how those work yet

Meakk:
Yes, you can test it locally. Just enable testing with CMake (`BUILD_TESTING` variable enabled with `ccmake` or `cmake-gui`), then build your project again. You can then run the test using `ctest -R TestCameraClipping`.

Meakk:
@schuhumi looks good. Let me know if we can merge it.

schuhumi:
Finally seems to work as intended. For documentation sake this is the object I created to make the issue more obvious, a checkerboard with green border:
TestCameraClipping
This is what it looks like when it gets clipped:
TestCameraClipping_fail

The clipping depends on both the camera settings and the size of the object, but I tried to make it prone to fail in case of clipping error:
- extend into depth from camera's point of view
- have the whole object visible, not the back covered by non-translucent front or something)

Meakk (approved):

Meakk:
Thanks @schuhumi !

mwestphal:
a license line should be added for the new file

Meakk:
> a license line should be added for the new data file

I'll take care of it.

schuhumi:
Oh I see, thank you for taking care of it. I created the data file solely for this test myself, so no strings attached.

mwestphal:
Thanks a lot for your contribution @schuhumi !

mwestphal:
@Meakk : file have not been uploaded as lfs for some reason, looking into it.

mwestphal:
https://github.com/f3d-app/f3d/pull/167

@schuhumi : you may have a git lfs issue, let us know if you need help with that.

schuhumi:
@mwestphal Oh I'm very sorry. I am familiar with lfs, but it didn't occur to me that it was used in the data folder. Since I was running my own test only and did't miss lfs data for that, no error made me aware of that. Actually I didn't even have lfs on my host installed, because the other projects I use it in are dockerized development environments.

In retrospect, considering how much q&a + work I caused you two with such a simple contribution, it might be worthwhile to write down a couple hints for non-regular contributors. I very much like how clean and organized the project is, this is not a source of confusion. It's just unfamiliar tooling, and the small things you overlook when trying to navigate a foreign code base. 

In the README.md at the "Build" section could be a good place to mention that git lfs is required for the data folder. After the "Configuration and building" subsection a short explanation how to run tests would have been helpful for me at least.

mwestphal:
@schuhumi : You are most definitely right. We are planning to rework the documentation and add a CONTRIBUTING.md that will cover all of this. README.md is getting way too long at this point.

PR #187, opened 2022-01-06
Split application/library folders

Meakk:


Meakk:
@mwestphal It's a simple change.  
I have created two folders `library` and `application` in `src` and moved the related code to each folder for `libf3d` and `f3d` target respectively

src/library/CMakeLists.txt
```diff
@@ -0,0 +1,157 @@
...
+    target_compile_options(libf3d PUBLIC /W4 /WX /wd4275)
+  else()
+    target_compile_options(libf3d PUBLIC -Wall -Wextra -Wshadow -Werror)
+  endif()
+endif()
+
+# coverage
+cmake_dependent_option(F3D_COVERAGE "Emit coverage files" OFF "UNIX" OFF)
+if(F3D_COVERAGE)
+  target_compile_options(libf3d PUBLIC -g -O0 --coverage)
```
mwestphal, line 155: maybe we need that for application too

mwestphal: that was not the case before, but we should think about it at some point.

mwestphal (approved):

PR #192, opened 2022-01-09
Fix reproducible build

mzf-guest:
This pull request fixes reproducible build issues, see [reproducible build website](https://reproducible-builds.org/) for the motivations.

 - Recorded timestamp through CMake should not depend on time zone (more info [here](https://reproducible-builds.org/docs/source-date-epoch/))
 - Platform version (for example kernel version on Linux platform) usually does not affect the generated binaries and should not be recorded (more info [here](https://reproducible-builds.org/docs/recording/))

The issues were detected by the Debian reprotest tool. You can look at [diffoscope_result.zip](https://github.com/f3d-app/f3d/files/7834236/diffoscope_result.zip) for the diff between two build environments that should produce the same binaries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ project(f3d
   DESCRIPTION "F3D - A fast and minimalist 3D viewer"
   LANGUAGES C CXX)
 
-string(TIMESTAMP F3D_BUILD_DATE "%Y-%m-%d %H:%M:%S")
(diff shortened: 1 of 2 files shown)


mwestphal (approved):

mwestphal:
Thanks @mzf-guest !

mzf-guest:
Thanks for the merge!

I take this opportunity to suggest to add [reprotest](https://salsa.debian.org/reproducible-builds/reprotest) to your CI pipeline :wink:

mwestphal:
Added reprotest to #194

PR #211, opened 2022-01-16
Add PDAL formats

Meakk:


Meakk:
GDAL and GeoTIFF is required to build PDAL. Setting this PR as draft, I'll do it later.

PR #231, opened 2022-02-21
Options initilization and check

mwestphal:
This initialize options in the constructor and also add a check flag when using set/get method

### DIFF
--- a/src/library/F3DOptions.cxx
+++ b/src/library/F3DOptions.cxx
@@ -730,6 +730,8 @@ F3DOptions F3DOptionsParser::GetOptionsFromCommandLine(std::vector<std::string>&
 //----------------------------------------------------------------------------
 void F3DOptionsParser::ConvertToNewAPI(const F3DOptions& oldOptions, f3d::options* newOptions)
 {
+  f3d::options::initialize(*newOptions);
+
   // TODO rework option names
   // TODO remove some options that do not exist in libf3d (like no-render)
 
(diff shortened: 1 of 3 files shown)


src/library/F3DOptions.cxx
```diff
@@ -730,6 +730,8 @@ F3DOptions F3DOptionsParser::GetOptionsFromCommandLine(std::vector<std::string>&
 //----------------------------------------------------------------------------
 void F3DOptionsParser::ConvertToNewAPI(const F3DOptions& oldOptions, f3d::options* newOptions)
 {
+  f3d::options::initialize(*newOptions);
```
Meakk, line 733: It should be in `options::F3DInternals` constructor I think. Any reason to call it explicitely?

src/library/f3d_options.cxx
```diff
@@ -83,4 +84,72 @@ template double options::get<>(const std::string& name) const;
...
+  // General
+  options.set("quiet", false);
+  options.set("verbose", false);
+
+  // Loader/Loading
+  options.set("animation-index", 0);
+  options.set("geometry-only", false);
+  options.set("progress", false);
+  options.set("camera-index", 0);
+  options.set("color", std::vector<double>({ 1., 1., 1. }));
```
Meakk, line 100: Would be awesome to support `std::initializer_list` to be able to call `options.set("color", { 1., 1., 1. });`

src/library/F3DOptions.cxx
mwestphal: Maybe someone do not want to initialize ? currently F3D do not require using it.

src/library/f3d_options.cxx
mwestphal: indeed, outside of the scope though

src/library/F3DOptions.cxx
Meakk: Then how to determine if a user specifies an option but it's not valid? The map should be initialized in the constructor imo, then any call to `set` should verify the key is already there.

mwestphal: I think you proposed behavior is indeed what we want. If we need more flexibility we can improve that in the future

mwestphal:
@Meakk please review

PR #236, opened 2022-02-26
Cleanup F3DLog

mwestphal:
- Separate F3DLog into vtkF3DLog and f3d_log
- vtkF3DLog is a simple vtkObject static methods only class to print a std::string with a severity
- f3d_log is a template based public API
- vtkF3D* class should use vtkF3DLog directly, and should not have concept of "verbose"
- vtkF3D* class generally only uses Warning and Error level, do not use Info
- f3d:: classes should use f3d::log and consider verbosity before printing anything
- Quiet is still general to the libf3d
- vtkWarning/ErrorMacro are still disabled and hidden from the user in release build


src/library/vtkF3DLog.h
```diff
@@ -0,0 +1,50 @@
...
+ * @brief   Class used to show logs
+ *
+ */
+
+#ifndef vtkF3DLog_h
+#define vtkF3DLog_h
+
+#include <vtkObject.h>
+
+#include <sstream>
```
Meakk, line 12: this include isn't needed

src/library/vtkF3DLog.h
```diff
@@ -0,0 +1,50 @@
...
+#include <string>
+
+// TODO could this be a namespace instead of inheriting vtkObject ?
```
Meakk, line 15: Yes, I don't see the point of inheriting from vtkObject here, especially if we never instantiate the object.

mwestphal: I was woried that iy may not have worked with vtk modules for some reason, but it seems to work fine. It may not be python wrappable but we do not care about that.

Meakk (approved):

PR #239, opened 2022-03-05
Renderer rework

mwestphal:
- Rework of the importer
- All vtkF3DRenderer options are controllable with a setter now
- All setters take care of updating what is needed or position flags as needed
- All setters check if the value is actually being changed
- In WithColoring, a UpdateColoringActors method still exists
- In WithColoring, setter update can be disabled, which means the UpdateColoringActors then needs to be called manually
- f3d::options is not used at all in the vtk* stack anymore
- some members have been renamed for better consistency

### DIFF
--- a/src/library/vtkF3DRendererWithColoring.cxx
+++ b/src/library/vtkF3DRendererWithColoring.cxx
@@ -166,7 +166,7 @@ bool vtkF3DRendererWithColoring::UsingInverseOpacityFunction()
 }
 
 //----------------------------------------------------------------------------
-void vtkF3DRendererWithColoring::SetScalarBarRange(std::vector<double> range, bool update)
+void vtkF3DRendererWithColoring::SetScalarBarRange(const std::vector<double>& range, bool update)
 {
   if (this->UserScalarBarRange != range)
   {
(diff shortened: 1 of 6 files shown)


src/library/vtkF3DRendererWithColoring.cxx
```diff
@@ -158,6 +165,32 @@ bool vtkF3DRendererWithColoring::UsingInverseOpacityFunction()
...
+    this->UserScalarBarRange = range;
+    if (update)
+    {
+      this->UpdateColoringActors();
+    }
+  }
+}
+
+//----------------------------------------------------------------------------
+void vtkF3DRendererWithColoring::SetColormap(std::vector<double> colormap, bool update)
```
Meakk, line 182: can you use `const std::vector<double>&` here?

src/library/vtkF3DRendererWithColoring.cxx
```diff
@@ -158,6 +165,32 @@ bool vtkF3DRendererWithColoring::UsingInverseOpacityFunction()
   return this->UseInverseOpacityFunction;
 }
 
+//----------------------------------------------------------------------------
+void vtkF3DRendererWithColoring::SetScalarBarRange(std::vector<double> range, bool update)
```
Meakk, line 169: can you use `const std::vector<double>&` here?

Meakk (approved):
minor comments, and good to go

PR #244, opened 2022-03-20
Disable tls in sanitizer

mwestphal:


mwestphal:
@Meakk I just cant manage to make the sanitizer fail. I'll let you take the lead on this.

PR #246, opened 2022-03-25
Adding simple sdk test architecture

mwestphal:
- Fix an issue with camera-index default value
- Add a testing architecture for adding libf3d specific tests
- prepend all test by either f3d or libf3d
- With previous version of VTK, force the engine to allocate loader, window and interactor in the right order
- add f3d::interactor::stop()
- add two simple tests
- hide leaks for sdk tests


src/library/f3d_engine.cxx
```diff
@@ -61,6 +61,15 @@ window& engine::getWindow()
           this->getOptions(), this->WindowType == engine::WindowTypeEnum::WINDOW_OFFSCREEN);
         break;
     }
+
+#if VTK_VERSION_NUMBER <= VTK_VERSION_CHECK(9, 1, 20400912) // TODO Actual version check
```
mwestphal, line 65: VERSION_CHECK is incorrect, code is not yet in VTK.

src/library/testing/TestSDKCompareWithFile.cxx
```diff
@@ -0,0 +1,15 @@
+#include <f3d_engine.h>
+#include <f3d_interactor.h>
```
mwestphal, line 2: not needed

src/library/f3d_engine.cxx
mwestphal: and add a comment

src/library/testing/TestSDKRenderAndInteract.cxx
```diff
@@ -0,0 +1,25 @@
...
+  load.loadFile(f3d::loader::LoadFileEnum::LOAD_CURRENT);
+  f3d::window& win = eng.getWindow();
+
+  win.render();
+
+  f3d::interactor& inter = eng.getInteractor();
+  inter.createTimerCallBack(1000, [&inter]() { inter.stop(); });
+  inter.start();
+
+  return 0;
```
mwestphal, line 24: EXIT_SUCCESS

.lsan.supp
```diff
@@ -7,3 +7,8 @@ leak:ld-linux-x86-64
 # OCCT
 leak:TKernel
 leak:TKXSBase
+
+# Potential mesa/VTK leak with incomplete callstack
+# forces us to hide all leaks from the libf3dSDKTests
+# https://gitlab.kitware.com/vtk/vtk/-/issues/18504
+leak:libf3dSDKTests
```
mwestphal, line 14: @Meakk 

This is the big hammer solution to the problem. Alternatively I could
- Add a cmake flags to enable the SDK_ONLY_RENDER_WINDOW_TESTS
- Force the creation of an interactor in all our tests (basically, just add `getInteractor()` before the first render.

Wdyt ?

Meakk: I think filtering this is fine for now. Let's investigate later

michaelmigliore (approved):

PR #269, opened 2022-04-18
Configure engine using binary flags

mwestphal:


src/library/f3d_engine.h
```diff
@@ -19,14 +20,29 @@ class interactor;
...
   };
 
-  engine(WindowTypeEnum windowType);
+  //======== Engine Flags =============
+  // engine::NO_WINDOW: Do not create a window to render into.
+  // engine::NO_INTERACTOR: Do not create an interactor to interact with
+  // engine::WINDOW_OFFSCREEN: Create an offscreen window to render into, ignored with NO_WINDOW
+  using engine_flags_t = std::bitset<8>;
+  static constexpr engine_flags_t FLAGS_NONE = engine_flags_t(0x0000);
+  static constexpr engine_flags_t NO_WINDOW = engine_flags_t(0x0001);
```
Meakk, line 41: Having `NO_*` flags look weird to me. What about doing the opposite? `WINDOW_NATIVE` and `INTERACTOR_DEFAULT` flags?

mwestphal: done

src/library/f3d_engine.h
```diff
@@ -19,14 +20,29 @@ class interactor;
...
+
+  //======== Engine Flags =============
+  // engine::CREATE_WINDOW: Create a window to render into.
+  // engine::CREATE_INTERACTOR: Create an interactor to interact with.
+  // engine::WINDOW_OFFSCREEN: Create an offscreen window to render into, need CREATE_WINDOW.
+  using flags_t = int;
+  enum Flags : flags_t
+  {
+    FLAGS_NONE = 0,             // 0000
+    CREATE_WINDOW = 1 << 0,     // 0001
```
Meakk, line 40: `CREATE_WINDOW` and `WINDOW_OFFSCREEN` sound weird to me. What about `WINDOW_NATIVE` and `WINDOW_OFFSCREEN`? And maybe just `INTERACTOR` or `INTERACTOR_DEFAULT`?

src/library/f3d_engine.h
```diff
@@ -19,14 +20,29 @@ class interactor;
...
+  struct interactor_exception : public std::exception
+  {
+    const char* what() const throw() { return "Cannot create interactor with this engine"; }
+  };
+
+  //======== Engine Flags =============
+  // engine::CREATE_WINDOW: Create a window to render into.
+  // engine::CREATE_INTERACTOR: Create an interactor to interact with.
+  // engine::WINDOW_OFFSCREEN: Create an offscreen window to render into, need CREATE_WINDOW.
+  using flags_t = int;
```
Meakk, line 36: `unsigned char` maybe enough, or `unsigned short`?

Meakk (changes requested):

src/library/f3d_engine.h
mwestphal: OFFSCREEN is NATIVE, but OFFSCREEN. With your logic, what happens when you set: 

WINDOW_NATIVE | WINDOW_OFFSCREEN ?

CREATE_* is to signify that without these flags, you will get an exception in the getters.

Meakk: Right, I forgot about that, I thought it was a true offscreen.  
In the future, it would be great to have a real offscreen (using OSMesa or EGL), then how would you differentiate native with hidden window, and offscreen?  
Ideally, the hidden state should be set using the window pointer, wdyt? Or maybe calling explicitly `window->show()`?

mwestphal: offscreen shoud be set on creation, before the first render, so no.

Here is how I see it:

You either have a window or you dont (getWindow works or not), controlled by the CREATE_WINDOW flag.
You either have an interactor or you dont (getInteractor works or not), controlled by the CREATE_INTERACTOR flag

the window can be of specific type:
- default
 - offscreen
 - egl
 - osmesa
 - external (?)

Controlled  by WINDOW_* flags

The interactor can be of specific type:
- default
- ?

Controlled  by INTERACTOR_* flags

Maybe the flags should be rennamed but I believe that the flag to create a window or not must exist.

Meakk (approved):

PR #270, opened 2022-04-26
Explicit public API export

Meakk:


Meakk:
@mwestphal 

https://github.com/f3d-app/f3d/runs/6170597578?check_suite_focus=true#step:13:68

This is exactly the issue I was talking to you about. On Windows, you cannot use the STL in public headers for this reason.  
You will not like it but I think we need raw pointers to declare the pimpl.

mwestphal:
@Meakk : Ok, I'll let you take care of the change

src/library/f3d_engine.h
```diff
@@ -55,7 +57,7 @@ class engine
 
 private:
   class F3DInternals;
-  std::unique_ptr<F3DInternals> Internals;
+  F3DInternals* Internals;
```
mwestphal, line 60: :cry:

src/library/F3DStarter.h
```diff
@@ -15,7 +20,7 @@ class F3DStarter
   /**
    * Parse the options and configure a f3d::loader accordingly
    */
-  int Start(int argc, char** argv);
+  F3D_EXPORT int Start(int argc, char** argv);
```
mwestphal, line 23: why does the methods needs to be exported for this one ?

mwestphal (approved):

src/library/F3DStarter.h
Meakk: One solution to keep the unique_ptr in public API is to export each method instead of exporting the whole class.  
Exporting each method is too verbose in my opinion, but for this particular class it's temporary, so I didn't wanted to change the unique_ptr to a raw ptr and revert this change when the class is not in the public API anymore.

mwestphal: got it.

PR #273, opened 2022-05-01
Rework F3DOptions to be moved in application

mwestphal:


.github/workflows/ci.yml
```diff
@@ -311,6 +311,11 @@ jobs:
         fetch-depth: 0
         lfs: 'false'
 
+    - name: gcc
+      run: |
+        gcc --version
+      shell: bash
```
mwestphal, line 317: to remove

src/application/CMakeLists.txt
```diff
@@ -23,15 +24,19 @@ endif()
 # Define f3d target
 add_executable(f3d ${F3D_SOURCE_FILES})
 target_link_libraries(f3d PUBLIC libf3d)
+
+if (UNIX AND NOT APPLE)
```
Meakk, line 28: I think clang does not need it. Maybe you should test if the compiler is GCC instead

src/application/F3DOptionsParser.cxx
```diff
@@ -179,101 +174,101 @@ class ConfigurationOptions
...
-    this->DeclareOption(grp1, "animation-index", "", "Select animation to show", options.AnimationIndex, true, true, "<index>");
-    this->DeclareOption(grp1, "camera-index", "", "Select the camera to use", options.CameraIndex, true, true, "<index>");
-    this->DeclareOption(grp1, "geometry-only", "", "Do not read materials, cameras and lights from file", options.GeometryOnly);
-    this->DeclareOption(grp1, "dry-run", "", "Do not read the configuration file", options.DryRun, true, false);
-    this->DeclareOption(grp1, "config", "", "Read a provided configuration file instead of default one", options.UserConfigFile, false, false, "<file path>");
-    this->DeclareOption(grp1, "font-file", "", "Path to a FreeType compatible font file", options.FontFile, false, false, "<file_path>");
+    auto grp0 = cxxOptions.add_options("Applicative");
+    this->DeclareOption(grp0, "input", "", "Input file", inputs, false, false, "<files>");
+    this->DeclareOption(grp0, "output", "", "Render to file", appOptions.Output, false, false,"<png file>");
+    this->DeclareOption(grp0, "no-background", "", "No background when render to file", appOptions.NoBackground);
```
Meakk, line 190: is it applicative? Having an option in the lib makes sense to me.

src/application/F3DOptionsParser.cxx
```diff
@@ -179,101 +174,101 @@ class ConfigurationOptions
...
-      this->DeclareOption(grp5, "denoise", "d", "Denoise the image", options.Denoise);
-    }
+    this->DeclareOption(grpCamera, "camera-position", "", "Camera position", options.getAsDoubleVectorRef("camera-position"), false, true, "<X,Y,Z>");
+    this->DeclareOption(grpCamera, "camera-focal-point", "", "Camera focal point", options.getAsDoubleVectorRef("camera-focal-point"), false, true, "<X,Y,Z>");
+    this->DeclareOption(grpCamera, "camera-view-up", "", "Camera view up", options.getAsDoubleVectorRef("camera-view-up"), false, true, "<X,Y,Z>");
+    this->DeclareOption(grpCamera, "camera-view-angle", "", "Camera view angle (non-zero, in degrees)", options.getAsDoubleRef("camera-view-angle"), false, true, "<angle>");
+    this->DeclareOption(grpCamera, "camera-azimuth-angle", "", "Camera azimuth angle (in degrees)", options.getAsDoubleRef("camera-azimuth-angle"), true, true, "<angle>");
+    this->DeclareOption(grpCamera, "camera-elevation-angle", "", "Camera elevation angle (in degrees)", options.getAsDoubleRef("camera-elevation-angle"), true, true, "<angle>");
+
+    // TODO find the right way to deal with modular options
```
Meakk, line 255: what's wrong with `if (f3d::HasRaytracingModule)`?

Meakk: Maybe just a runtime check `if (f3d::engine::hasRaytracingSupport())` would work?  
I think this public API is needed anyway.

src/application/F3DOptionsParser.cxx
```diff
@@ -528,57 +533,65 @@ std::string ConfigurationOptions::GetSystemSettingsDirectory()
...
-    if (directoryPath[directoryPath.size() - 1] != separator[0])
-    {
-      directoryPath.append(separator);
-    }
-#else
-    separator = "/";
-    directoryPath += separator;
-#endif
-    directoryPath += "..";
+    // TODO may not work on windows ?
```
Meakk, line 542: to check and remove comment if it works

src/application/F3DOptionsParser.cxx
```diff
@@ -528,57 +533,65 @@ std::string ConfigurationOptions::GetSystemSettingsDirectory()
...
+      }
     }
 #endif
-    directoryPath = vtksys::SystemTools::CollapseFullPath(directoryPath);
-    directoryPath += separator;
   }
-  return directoryPath;
+  catch (const std::filesystem::filesystem_error&)
+  {
+    f3d::log::error("Cannot recover binary config file directory: ", dirPath.string());
```
Meakk, line 559: calling `e.what()` could be useful?

src/application/F3DOptionsParser.cxx
```diff
@@ -600,20 +613,24 @@ void F3DOptionsParser::InitializeDictionaryFromConfigFile(const std::string& use
...
-F3DOptions F3DOptionsParser::GetOptionsFromCommandLine(std::vector<std::string>& files)
+void F3DOptionsParser::GetOptionsFromCommandLine(
+  F3DAppOptions& appOptions, f3d::options& options, std::vector<std::string>& files)
 {
   this->ConfigOptions->SetFilePathForConfigBlock("");
-  return this->ConfigOptions->GetOptionsFromArgs(files);
+  return this->ConfigOptions->GetOptionsFromArgs(appOptions, options, files);
 }
 
+// TODO Remove
```
Meakk, line 632: adding this discussion so you don't forget: to remove

src/library/f3d_options.h
```diff
@@ -40,11 +41,17 @@ class F3D_EXPORT options
...
+  int& getAsIntRef(const std::string& name);
+  double& getAsDoubleRef(const std::string& name);
+  std::string& getAsStringRef(const std::string& name);
+  std::vector<int>& getAsIntVectorRef(const std::string& name);
+  std::vector<double>& getAsDoubleVectorRef(const std::string& name);
+
 private:
   class F3DInternals;
   F3DInternals* Internals;
   options(const options& opt) = delete;
```
Meakk, line 54: I don't think we should delete the copy constructor if we implement the copy assignment.

Meakk (changes requested):

src/library/f3d_options.h
mwestphal: ok, i'll add it back

src/application/F3DOptionsParser.cxx
mwestphal: HasRaytracingModule is not installed. I need to think of the right way to check for the presence of optional libf3d module from the applicative side. Using the engine is fine but not great imo.

Meakk: Raytracing is a special module, it's not an extension of supported file, so it's not a problem to expose the support in the engine I think. Otherwise, you can set a public CMake variable `F3D_HAS_RAYTRACING_SUPPORT`, exported when `find_package` is called.

mwestphal: >  it's not an extension of supported file, so it's not a problem to expose the support in the engine I think

It would need to be a static method, but yes. In any case, I think cmake vars would be better.

src/application/CMakeLists.txt
mwestphal: it is needed:

```
 [100%] Linking CXX executable ../../bin/f3d
CMakeFiles/f3d.dir/F3DOptionsParser.cxx.o: In function `std::filesystem::__cxx11::path::path<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::filesystem::__cxx11::path>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::filesystem::__cxx11::path::format)':
/usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8/bits/fs_path.h:185: undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
CMakeFiles/f3d.dir/F3DOptionsParser.cxx.o: In function `ConfigurationOptions::InitializeDictionaryFromConfigFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
```

src/application/F3DOptionsParser.cxx
mwestphal: works when running from the terminal but does not work when running from the .exe

src/application/F3DOptionsParser.cxx
mwestphal: will do in another MR. for now, lets keep the option.

src/application/F3DOptionsParser.cxx
mwestphal: actually works

src/application/F3DOptionsParser.cxx
mwestphal: not needed after all, this exception is excpected in that specific case.

src/application/F3DOptionsParser.cxx
```diff
@@ -179,101 +174,101 @@ class ConfigurationOptions
...
 
     auto grp3 = cxxOptions.add_options("Window");
-    this->DeclareOption(grp3, "bg-color", "", "Background color", options.BackgroundColor, true, true, "<R,G,B>");
-    this->DeclareOption(grp3, "resolution", "", "Window resolution", options.Resolution, true, true, "<width,height>");
-    this->DeclareOption(grp3, "fps", "z", "Display frame per second", options.FPS);
-    this->DeclareOption(grp3, "filename", "n", "Display filename", options.Filename);
-    this->DeclareOption(grp3, "metadata", "m", "Display file metadata", options.MetaData);
-    this->DeclareOption(grp3, "fullscreen", "f", "Full screen", options.FullScreen);
-    this->DeclareOption(grp3, "blur-background", "u", "Blur background", options.BlurBackground);
+    this->DeclareOption(grp3, "bg-color", "", "Background color", options.getAsDoubleVectorRef("background-color"), true, true, "<R,G,B>"); // TODO change name ?
```
Meakk, line 229: `TODO`: yes, please use `bg-color` in `f3d::options`

src/application/F3DStarter.cxx
```diff
@@ -108,37 +109,40 @@ int F3DStarter::Start(int argc, char** argv)
     this->Internals->Engine->getWindow().setIcon(F3DIcon, sizeof(F3DIcon));
   }
 
+  // Copy command line options into the actual engine options
+  f3d::options& options = this->Internals->Engine->getOptions();
+  options = commandLineOptions;
```
Meakk, line 114: Instead of copying the `f3d::options` class, why not using the reference directly and pass it to `GetOptionsFromCommandLine`?  
I don't like the idea of making this class copyable.

src/library/f3d_loader_impl.cxx
```diff
@@ -291,6 +291,7 @@ bool loader_impl::loadFile(loader::LoadFileEnum load)
   // Reset loadedFile
   this->Internals->LoadedFile = false;
 
+  // TODO is this enough ? Should we set this only application side ?
```
Meakk, line 294: todo

src/library/f3d_options.h
```diff
@@ -15,6 +17,17 @@ class F3D_EXPORT options
 public:
   options();
   ~options();
+  options(const options& opt);
+  options& operator=(const options& opt);
+
+  class options_exception : public std::runtime_error
```
Meakk, line 23: what about just `exception` since it's already in the `options` scope?  
I think `catch(f3d::options::exception)` is cleaner than `catch(f3d::options::options_exception)`

Meakk (changes requested):

src/application/F3DStarter.cxx
mwestphal: you need to read the options to be able to create the engine, and you need the engine to create the options.

src/application/F3DOptionsParser.cxx
mwestphal: tbh i dont like bg.

Meakk: Yes, but the command line option is `bg-color`, and for backward compatibility it would be better to not rename it.

mwestphal: not in this MR though.

mwestphal:
Looks there is some issue with the diff. Opening a new PR.

PR #277, opened 2022-05-08
Add a drag and drop test

mwestphal:


src/library/VTKExtensions/Rendering/vtkF3DInteractorEventRecorder.cxx
```diff
@@ -75,3 +78,113 @@ void vtkF3DInteractorEventRecorder::ProcessEvents(
...
+
+      this->Interactor->SetEventPosition(pos);
+      this->Interactor->SetControlKey(ctrlKey);
+      this->Interactor->SetShiftKey(shiftKey);
+      this->Interactor->SetAltKey(altKey);
+      this->Interactor->SetKeyCode(static_cast<char>(keyCode));
+      this->Interactor->SetRepeatCount(repeatCount);
+      this->Interactor->SetKeySym(keySym);
+
+      this->Interactor->InvokeEvent(ievent, strArr);
```
mwestphal, line 167: @Meakk this is a copied from VTK with the addition of the strings at the end. While this is fine this is not perfect has our logs are not respective the VTK interactive log standard, if there is such a things, which we may be impacted if it ever evolved in the future. I could add a more generic approach in VTK, what do you think ?

src/application/F3DStarter.cxx
```diff
@@ -115,7 +116,7 @@ int F3DStarter::Start(int argc, char** argv)
 
   // Add and load file
   this->Internals->Engine->getLoader().addFiles(files);
-  bool loaded = this->LoadFile();
+  this->LoadFile();
```
Meakk, line 119: why is this change required?

src/library/VTKExtensions/Rendering/vtkF3DInteractorEventRecorder.cxx
Meakk: At this point, I think we should have our own serialization based on json for example.

Meakk (changes requested):

src/application/F3DStarter.cxx
mwestphal: bug with the d&d logic in some cases.

src/library/VTKExtensions/Rendering/vtkF3DInteractorEventRecorder.cxx
mwestphal: I dont think so. Do you plan to rewrite the whole logic in json ?

I think that what we need can be put in VTK and improving VTK gives us free maintenance on the code we put in VTK.

michaelmigliore: If we improve the VTK version, it will still not work with previous VTK versions, right?

mwestphal: it will, thanks tonthe "StreamVersion" logic

mwestphal: I'll move forward with a VTK side implementation. I think it make sense.

Meakk (approved):

PR #278, opened 2022-05-09
Move F3DOptions to Applications and improve f3d::options

mwestphal:
- Move F3DOptions to Applications/F3DOptionsParser
- Rewrite F3DOption to not use VTK at all and use filesystem instead
- Improve f3d::options for easier usages
- Use gcc-8 for filesystem support
- Add and update tests


.github/docker/Dockerfile
```diff
@@ -17,10 +17,15 @@ RUN apt update && apt install -y \
     openssh-client \
     software-properties-common \
     wget \
-    xvfb 
+    xvfb \
+    gcc-8 \
```
Meakk, line 21: sort the new packages

src/application/F3DOptionsParser.cxx
```diff
@@ -179,101 +174,101 @@ class ConfigurationOptions
...
-F3DOptions ConfigurationOptions::GetOptionsFromArgs(std::vector<std::string>& inputs)
+void ConfigurationOptions::GetOptionsFromArgs(
+  F3DAppOptions& appOptions, f3d::options& options, std::vector<std::string>& inputs)
 {
-  F3DOptions options;
   try
   {
-    cxxopts::Options cxxOptions(f3d::AppName, f3d::AppTitle);
+    // TODO From where ?
+    cxxopts::Options cxxOptions("F3D", "F3D");
```
Meakk, line 183: Can you add back f3d::AppName and f3d::AppTitle in an application-side CMake generated file?

src/library/f3d_engine.cxx
```diff
@@ -59,6 +59,12 @@ engine::~engine()
   delete this->Internals;
 }
 
+//----------------------------------------------------------------------------
+void engine::setOptions(options& opt)
```
Meakk, line 63: Sorry I don't like a `setOptions` function making the input invalid. In this case, I would prefer to explicitly call `engine.getOptions() = std::move(myOptions);`.  
Alternatively, you can have these two functions:
```cpp
void engine::setOptions(const options& opt); // call copy
void engine::setOptions(options&& opt); // call move
```
Having both will allow the user to call `engine.setOptions(std::move(myOptions))` if he wants to move it.

Meakk (changes requested):

src/library/f3d_config.h
```diff
@@ -13,7 +13,5 @@ extern const std::string LibVersion;
 extern const std::string LibBuildSystem;
 extern const std::string LibBuildDate;
 extern const std::string LibCompiler;
-extern const std::string ReservedString; // TODO to remove once not necessary
```
Meakk, line 16: `AppName` and `AppTitle` can be removed

src/library/f3d_engine.h
```diff
@@ -46,6 +46,16 @@ class F3D_EXPORT engine
...
 
+  // Engine provide a default options
+  // use this setter to use other options
+  // copy options into engine
+  void setOptions(const options& opt);
+
+  // Engine provide a default options
+  // use this setter to use other options
+  // move options into engine
+  void setOptions(options&& opt); // call move
```
Meakk, line 57: remove `//call move`

Meakk (approved):
minor comments, lgtm

src/library/f3d_config.h
mwestphal: not yet, no. I'm working on that on another branch.

PR #281, opened 2022-05-13
Strict build rework

mwestphal:
- Rework strict build to share it between app and lib
- Add more flags to warn on
- Fix some warnings
- use f3d::engine::exception instead of dedicated exception

### DIFF
--- a/src/application/CMakeLists.txt
+++ b/src/application/CMakeLists.txt
@@ -43,12 +43,11 @@ target_include_directories(f3d
   $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
   )
 
+# F3D_STRICT_BUILD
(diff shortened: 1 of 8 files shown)


Meakk (approved):

src/application/CMakeLists.txt
```diff
@@ -43,12 +43,11 @@ target_include_directories(f3d
   $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
   )
 
+# F3D_STRICT_BUILD
 if(F3D_STRICT_BUILD)
+  list(APPEND libf3d_compile_options_private ${f3d_strict_build_compile_options})
   if(MSVC)
-    target_compile_options(f3d PRIVATE /W4 /WX)
```
mwestphal, line 48: that was not right at all

PR #287, opened 2022-05-16
Add install components

Meakk:


src/library/VTKExtensions/CMakeLists.txt
```diff
@@ -6,7 +6,12 @@ vtk_module_scan(
   WANT_BY_DEFAULT   ON
   ENABLE_TESTS      ${BUILD_TESTING})
 
-vtk_module_build(MODULES ${modules} INSTALL_HEADERS OFF)
+# We need HEADERS_COMPONENT because of a bug in vtk_module_build in VTK <= 9.1
```
mwestphal, line 9: Ideally, add a version check once fixed in VTK, and a link to the issue

src/library/VTKExtensions/Rendering/vtk.module
```diff
@@ -7,6 +7,7 @@ DEPENDS
   VTK::RenderingAnnotation
   VTK::RenderingOpenGL2
   VTK::RenderingVolumeOpenGL2
+  VTK::opengl
```
mwestphal, line 10: why is it meeded ?

mwestphal (approved):

src/library/VTKExtensions/Rendering/vtk.module
michaelmigliore: Because `vtkF3DOpenGLGridMapper` uses an opengl function (`glDrawArrays`)

PR #289, opened 2022-05-17
Reorganize files and namespaces

mwestphal:


library/src/f3d_interactor.cxx
```diff
@@ -2,8 +2,8 @@
 
 namespace f3d
 {
-namespace details
-{
+
+// Is this ok to declare like this ? TODO
```
mwestphal, line 6: the only things I'm not sure of in this MR @Meakk . This is declared in the cxx only.

library/src/f3d_interactor_impl.cxx
```diff
@@ -460,7 +460,7 @@ bool interactor_impl::recordInteraction(const std::string& file)
   if (file.empty())
   {
     // TODO improve VTK to check file opening
-    f3d::log::error("Interaction record file is empty");
+    log::error("Interaction record file is empty");
```
mwestphal, line 463: shouldnt this line be covered by testing already ?

Meakk: I don't think so

library/src/f3d_loader_impl.cxx
```diff
@@ -17,134 +17,132 @@
...
-  F3DInternals(const options& options, window_impl& window)
-    : Options(options)
-    , Window(window)
+  static void DisplayImporterDescription(vtkImporter* importer)
   {
+#if VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 0, 20210228)
+    vtkIdType availCameras = importer->GetNumberOfCameras();
+    if (availCameras <= 0)
+    {
+      f3d::log::info("No camera available in this file");
```
Meakk, line 132: you don't need `f3d::` here and below

library/src/f3d_loader_impl.cxx
```diff
@@ -320,13 +318,13 @@ bool loader_impl::loadFile(loader::LoadFileEnum load)
   }
 
   // Recover the importer
-  this->Internals->Importer =
-    ::GetImporter(filePath, this->Internals->Options.getAsBool("geometry-only"));
+  this->Internals->Importer = loader_impl::F3DInternals::GetImporter(
```
Meakk, line 321: this `F3D` prefix looks weird now. Can we use just `internals` instead of `F3DInternals` for pimpls?

Meakk (changes requested):

PR #295, opened 2022-05-20
Doc and tests improvements

mwestphal:


library/src/loader_impl.cxx
```diff
@@ -191,24 +191,25 @@ void loader_impl::addFile(const std::string& path, bool recursive)
...
+    {
+      std::string currentFile = dir.GetFile(i);
+      if (currentFile != "." && currentFile != "..")
       {
-        std::string currentFile = dir.GetFile(i);
-        if (currentFile != "." && currentFile != "..")
-        {
-          sortedFiles.insert(currentFile);
-        }
+        sortedFiles.insert(currentFile);
```
Meakk, line 204: To check, but I'm pretty sure that preallocating a `std::vector` then sort it after the loop would be more efficient in time and space.

library/testing/TestSDKLoader.cxx
```diff
@@ -0,0 +1,114 @@
...
+#include <window.h>
+
+#include <iostream>
+
+int TestSDKLoader(int argc, char* argv[])
+{
+  f3d::engine eng(f3d::engine::FLAGS_NONE);
+  f3d::loader& load = eng.getLoader();
+
+  std::string cowFilename = "cow.vtp";
```
Meakk, line 13: `constexpr std::string_view` is better when the string is immutable and known at build time.

Meakk (approved):
minor comments, feel free to ignore them

library/src/loader_impl.cxx
mwestphal: It would be, but for such small size it does not matter imo.

library/testing/TestSDKLoader.cxx
mwestphal: Indeed, I need to learn to use these.

library/src/loader_impl.cxx
mwestphal: https://stackoverflow.com/questions/15638024/stdsort-vs-inserting-into-an-stdset

PR #296, opened 2022-05-21
Add / for a replacement of ? hotkey

mwestphal:


Meakk (approved):
Maybe it would be better to use a keyboard layout independent key, what about `tab`?

PR #298, opened 2022-05-24
Introduce f3d::image class

Meakk:


library/public/image.h
```diff
@@ -0,0 +1,52 @@
...
+
+  int getChannelCount() const;
+  image& setChannelCount(int dim);
+
+  image& setData(unsigned char* buffer);
+  unsigned char* getData() const;
+
+  bool compare(const image& reference, image& result, double threshold) const;
+
+  image& save(const std::string& path);
```
mwestphal, line 44: why do many of these methods return a ref to image ?

library/src/image.cxx
```diff
@@ -0,0 +1,194 @@
...
+  return error <= threshold;
+}
+
+//----------------------------------------------------------------------------
+image& image::save(const std::string& path)
+{
+  auto importer = this->Internals->GetVTKImporter();
+
+  vtkNew<vtkPNGWriter> writer;
+  writer->SetFileName(path.c_str());
```
mwestphal, line 188: check path is valid and can be written on ?

library/src/image.cxx
```diff
@@ -0,0 +1,194 @@
...
+  int Height = 0;
+  int Channels = 0;
+
+  void UpdateBufferSize() { this->Buffer.resize(this->Width * this->Height * this->Channels); }
+
+  vtkSmartPointer<vtkImageImport> GetVTKImporter() const
+  {
+    vtkNew<vtkImageImport> importer;
+    importer->CopyImportVoidPointer(
+      const_cast<unsigned char*>(this->Buffer.data()), this->Buffer.size());
```
mwestphal, line 29: sanity check of buffer size vs width/hieght/channels ?

library/src/image.cxx
```diff
@@ -0,0 +1,194 @@
...
+  : Internals(new image::internals())
+{
+}
+
+//----------------------------------------------------------------------------
+image::image(const std::string& path)
+  : Internals(new image::internals())
+{
+  auto reader = vtkSmartPointer<vtkImageReader2>::Take(
+    vtkImageReader2Factory::CreateImageReader2(path.c_str()));
```
mwestphal, line 65: check file exists ?

library/public/window.h
```diff
@@ -13,9 +14,7 @@ class F3D_EXPORT window
 public:
   virtual bool update() = 0;
   virtual bool render() = 0;
-  virtual bool renderToFile(const std::string& file, bool noBackground = false) = 0;
-  virtual bool renderAndCompareWithFile(const std::string& file, double threshold,
-    bool noBackground, const std::string& outputFile = "") = 0;
+  virtual image renderToImage(bool noBackground = false) = 0;
```
mwestphal, line 17: add a renderToImage(image&) ?

library/testing/TestSDKInteractionDirectory.cxx
```diff
@@ -46,8 +46,16 @@ int TestSDKInteractionDirectory(int argc, char* argv[])
...
+  f3d::image result = win.renderToImage();
+  f3d::image diff;
+  bool ret =
+    result.compare(f3d::image(std::string(argv[1]) + "/baselines/" + filename + ".png"), diff, 50);
+  if (!ret)
+  {
+    result.save(std::string(argv[2]) + filename + ".png");
+    diff.save(std::string(argv[2]) + filename + ".diff.png");
+    return EXIT_FAILURE;
+  }
```
mwestphal, line 58: we may want to put above logic in a testing method somewhere.

library/public/image.h
Meakk: I like this design, and I'd like to push more toward it. It's better in my opinion to return the object to chain calls like: 
 
```cpp
// nice
img.setResolution(300,300)
      .setChannelCount(3)
      .setData(buffer)
      .save("/tmp/img.png");

// not nice :(
img.setResolution(300,300)
img.setChannelCount(3)
img.setData(buffer)
img.save("/tmp/img.png");
```

library/public/window.h
Meakk: I think it's useless, this would be as efficient:
```cpp
image existingImage;
existingImage = std::move(window.renderToImage());
```

library/public/image.h
mwestphal: it makes sense with class that "are things" like image here, much less for class that "do things" like loader. My initial design was probably VTK-influenced hence the current API. I'm not against having a bit of both but I hope this will not be confusing for users. wdyt ?

library/public/window.h
mwestphal: In your case there is still a copy, is there not ?

Meakk: nop, it's calling `image::oparator=(image&&)` which does a swap internally.

library/public/image.h
Meakk: Yes you're right, it makes sense here, but much less for the loader. For the `options` class, it could be useful though to chain `set` calls.

library/public/window.h
mwestphal: ok !

library/public/image.h
mwestphal: ok, feel free to improve !

library/src/image.cxx
Meakk: This one is quite hard. How to determine if a path is writable?

library/src/image.cxx
```diff
@@ -0,0 +1,207 @@
...
+  unsigned int Height = 0;
+  unsigned int Channels = 0;
+
+  void UpdateBufferSize() { this->Buffer.resize(this->Width * this->Height * this->Channels); }
+
+  vtkSmartPointer<vtkImageImport> GetVTKImporter() const
+  {
+    if (this->Buffer.size() != this->Width * this->Height * this->Channels)
+    {
+      throw engine::exception("Image size mitmatch");
```
mwestphal, line 32: image::exception ?

library/src/image.cxx
```diff
@@ -0,0 +1,207 @@
...
+{
+}
+
+//----------------------------------------------------------------------------
+image::image(const std::string& path)
+  : Internals(new image::internals())
+{
+  if (!std::filesystem::exists(path))
+  {
+    throw engine::exception("Cannot open image " + path);
```
mwestphal, line 74: image exception ?

library/src/image.cxx
mwestphal: right

library/src/image.cxx
```diff
@@ -0,0 +1,207 @@
...
+
+//----------------------------------------------------------------------------
+image& image::save(const std::string& path)
+{
+  auto importer = this->Internals->GetVTKImporter();
+
+  vtkNew<vtkPNGWriter> writer;
+  writer->SetFileName(path.c_str());
+  writer->SetInputConnection(importer->GetOutputPort());
+  writer->Write();
```
mwestphal, line 203: check ErrorCode ? https://vtk.org/doc/nightly/html/classvtkAlgorithm.html#a31563282603f0b7df0fad9817f56b0be

library/testing/TestSDKHelpers.h
```diff
@@ -0,0 +1,22 @@
+#ifndef TestSDKHelpers_h
+#define TestSDKHelpers_h
+
+class TestSDKHelpers
+{
+public:
+  static bool RenderTest(f3d::window& win, const std::string& baseline, const std::string& output,
+    const std::string& diff)
```
mwestphal, line 8: I think RenderTest(argv, win, name) could be a more specific and simpler approach

library/src/image.cxx
```diff
@@ -0,0 +1,205 @@
...
+  : Internals(new image::internals())
+{
+  *this->Internals = *img.Internals;
+}
+
+//----------------------------------------------------------------------------
+image& image::operator=(const image& img)
+{
+  *this->Internals = *img.Internals;
+  return *this;
```
mwestphal, line 102: please add a TestSDKImage, covering 100% should be trivial.

library/public/options.h
```diff
@@ -30,15 +30,15 @@ class F3D_EXPORT options
...
-  void set(const std::string& name, std::initializer_list<double> values);
+  options& set(const std::string& name, bool value);
+  options& set(const std::string& name, int value);
+  options& set(const std::string& name, double value);
+  options& set(const std::string& name, const std::string& value);
+  options& set(const std::string& name, const char* value);
+  options& set(const std::string& name, const std::vector<int>& values);
+  options& set(const std::string& name, const std::vector<double>& values);
+  options& set(const std::string& name, std::initializer_list<int> values);
+  options& set(const std::string& name, std::initializer_list<double> values);
```
mwestphal, line 41: improve f3d::options testing to test the new return

library/src/image.cxx
mwestphal: not yet 100% coverage.

library/testing/TestSDKImage.cxx
```diff
@@ -0,0 +1,75 @@
...
+    nbCatch++;
+  }
+
+  if (nbCatch != 2)
+  {
+    std::cerr << "An exception has not been thrown" << std::endl;
+    return EXIT_FAILURE;
+  }
+
+  f3d::image baseline(std::string(argv[1]) + "/baselines/TestSDKImage.png");
```
mwestphal, line 53: you should read an image from the data, not from the baselines.

library/testing/TestSDKImage.cxx
```diff
@@ -0,0 +1,75 @@
...
+  {
+    std::cerr << "Baseline has wrong number of channels" << std::endl;
+    return EXIT_FAILURE;
+  }
+
+  if (baseline.getData() == nullptr)
+  {
+    std::cerr << "Baseline has no data" << std::endl;
+    return EXIT_FAILURE;
+  }
```
mwestphal, line 72: I would expect the baseline to only be used in the context of image::compare. The rest of the testing should either be performed on a generated data or an (identical to baseline) image from data.

library/testing/TestSDKImage.cxx
```diff
@@ -52,24 +52,29 @@ int TestSDKImage(int argc, char* argv[])
...
-    std::cerr << "Baseline has no data" << std::endl;
+    std::cerr << "Image has no data" << std::endl;
     return EXIT_FAILURE;
   }
 
+  // test operators
+  f3d::image imgCopy = generated; // copy constructor
+  imgCopy = baseline; // copy assignment
+  f3d::image imgMove = std::move(imgCopy); // move constructor
+  imgCopy = std::move(imgMove); // move assignment
```
mwestphal: check they are all actually equal ?

library/src/image.cxx
```diff
@@ -0,0 +1,224 @@
...
+  unsigned int Channels = 0;
+
+  void UpdateBufferSize() { this->Buffer.resize(this->Width * this->Height * this->Channels); }
+
+  vtkSmartPointer<vtkImageImport> GetVTKImporter() const
+  {
+    if (this->Buffer.size() != this->Width * this->Height * this->Channels)
+    {
+      throw exception("Image size mitmatch");
+    }
```
mwestphal, line 31: this should be coverable I think.

Meakk: I don't see how tbh

mwestphal: indeed.

mwestphal (approved):

PR #299, opened 2022-05-24
Improve public API and few cppcheck fixes

Meakk:


library/src/animationManager.cxx
```diff
@@ -216,7 +216,7 @@ void animationManager::Tick()
     this->Importer->UpdateTimeStep(*this->CurrentTimeStep);
     this->Window->render();
 
-    this->CurrentTimeStep++;
+    ++this->CurrentTimeStep;
```
mwestphal, line 219: what does this change ?

mwestphal (approved):

library/src/animationManager.cxx
Meakk: cppcheck recommends using preincrement for non-trivial types. I recommend this video on the subject btw: https://www.youtube.com/watch?v=ObVRSNvGitE

PR #301, opened 2022-05-28
Improve loader and engine testing

mwestphal:
- Improve loader and engine testing
- Improve loader and engine doc
- Fix a small issue in loader file list

### DIFF
--- a/library/public/engine.h
+++ b/library/public/engine.h
@@ -2,7 +2,20 @@
  * @class   engine
  * @brief   Class used to create instance of other classes
  *
- * TODO improve doc
+ * The main class of the libf3d to create all other instances
+ * Configured on creation using binary flags, then all objects
+ * can be accessed through their getter.
+ *
+ * Example usage:
+ *
+ * \code{.cpp}
+ *  f3d::engine eng(f3d::engine::CREATE_WINDOW | f3d::engine::CREATE_INTERACTOR);
+ *  f3d::loader& load = eng.getLoader();
+ *  load.addFile("path/to/file");
+ *  load.loadFile(f3d::loader::LoadFileEnum::LOAD_CURRENT);
+ *  f3d::interactor& inter = eng.getInteractor();
+ *  inter.start();
+ * \endcode
  */
 
 #ifndef f3d_engine_h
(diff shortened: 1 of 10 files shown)


library/public/engine.h
```diff
@@ -2,7 +2,18 @@
...
  * @brief   Class used to create instance of other classes
  *
- * TODO improve doc
+ * The main class of the libf3d to create all other instances
+ * Configured on creation using binary flags, then all objects
+ * can be accessed through their getter.
+ *
+ * Example usage:
+ *
+ *  f3d::engine eng(f3d::engine::CREATE_WINDOW | f3d::engine::CREATE_INTERACTOR);
```
Meakk, line 11: I think there are special tags for code in doxygen

library/src/window_impl_noRender.cxx
```diff
@@ -2,6 +2,7 @@
 
 #include <vtkNew.h>
 #include <vtkRenderWindow.h>
+#include <vtkVersion.h>
```
Meakk: is it needed?

library/src/window_impl_noRender.cxx
```diff
@@ -16,6 +17,7 @@ window_impl_noRender::window_impl_noRender(const options& options)
   : window_impl(options)
   , Internals(new window_impl_noRender::internals)
 {
+  this->Internals->RenWin->SetOffScreenRendering(true);
```
Meakk, line 19: I don't like the fact we have an OpenGL render window instance if we have a noRender f3d window. Maybe we should implement a `vtkNoRenderWindow` inheriting from `vtkRenderWindow` and doing nothing.

library/testing/TestSDKEngineExceptions.cxx
```diff
@@ -24,9 +28,14 @@ int TestSDKEngineExceptions(int argc, char* argv[])
...
   catch (const f3d::engine::exception& ex)
   {
-    std::cerr << ex.what() << std::endl;
-    exit(EXIT_SUCCESS);
+    std::cout << ex.what() << std::endl;
+    nbCatch++;
   }
 
-  return EXIT_FAILURE;
+  if (nbCatch != 2)
```
Meakk, line 35: If you look at my last changes for f3d::image test, I did something better. Just display a message and return in the `try` block.

Meakk (changes requested):

library/src/window_impl_noRender.cxx
mwestphal: Out of scope of this PR, but that would be better indeed if possible.

library/public/engine.h
mwestphal: I wil add them buyt am still unsure of the way we will handle API documentation. I hate doxygen with a passion.

library/testing/TestSDKEngineExceptions.cxx
```diff
@@ -9,24 +9,31 @@ int TestSDKEngineExceptions(int argc, char* argv[])
 {
   f3d::engine eng(f3d::engine::FLAGS_NONE);
 
+  // test exceptions
+  int nbCatch = 0;
```
Meakk, line 13: nbCatch is unused

Issue #52, opened 2021-11-12
Brainstorm libf3d API


mwestphal:
libf3d was introduced in !259 (closed), let's brainstorm a stable API for it.

Tasklist:

libf3d
- [x] Introduce f3d::options: #191 
- [x] Refactor F3DLoader and F3D to use f3d::options cleanly
- [x] Introduce f3d::loader and use it
- [x] Introduce f3d::window and use it
- [x] modularize VTK code
- [x] Add unit test for VTK modules : https://github.com/f3d-app/f3d/issues/265
- [x] sanity check call order with window and interactor
- [x] transform F3DLog into f3d::log
- [x] fix #32  by using categories for options
- [x] improve f3d_options by providing initialization methods
- [x] store init value in f3d::options for easier reset ?
- [x] support std::initializer_list to be able to call options.set("color", { 1., 1., 1. });
- [x] add a way to inform user when an options does not exist
- [x] introduce a factory and use it to create classes to hide the private API
- [x] Add SDK test infrastructure
- [x] Adress as many TODOs as possible
- [x] Rework engine for bit flag and init early
- [x] Add SDK tests
- [x] (optional) expand f3d_interactor to support user set behavior for many more context
- [x] add more window types
- [x] improve all classes from f3d namespaces to ensure all santiy checks are done correctly
- [x] unit testing of the API and general coverage improvements
- [x] Add a vtkNoRenderWindow
- [x] modernization of API
- [x] constness of API
- [x] Public Review of the API to anyone willing to help: #351 

F3D
- [x] Rework F3DOptions and F3DOptionsParser to interface with f3d::options
- [x] (Optional) introduce f3d::config in libf3d ?
- [x] rewrite F3DStarter from scratch to not depends on VTK

Doc
- [x] In header doc
- [x] example doc
- [x] website doc

mwestphal:
Ignoring completelly how libf3d and f3d are currently implemented, let's consider the fonctionnality we want to provide as a lib.

Features:
- Set a list of options
- Read a file
- Display resuting scene using file and options in a OpenGL context, potentially provided by user

Restrictions:
- User **must** not need to find_package VTK to use libf3d or use any vtk* class
- User does not need to know much about rendering to use libf3d

inputs @Meakk ?

Meakk:
I agree, VTK should be a private dependency of libf3d.  
Ideally, f3d executable should only provide the command-line parsing, and config file management.

mwestphal:
[...]

PR #304, opened 2022-05-30
Improve window doc and test

mwestphal:
- Add window testing
- Add window doc
- Add error ref param to image compare
- Improve TestSDKHelpers


library/public/window.h
```diff
@@ -6,16 +14,45 @@
...
 {
 class F3D_EXPORT window
 {
 public:
+  /**
+   * Use all the rendering related options to update the configuration of the window
+   * and the rendering stack below. This also initialize the rendering stack if needed.
+   * This will be called automatically when calling loader::loadFile but can also be called manually
+   * when needed. This must be called, either manually or automatically, before any render call.
+   * Return true on success, false otherwise.
```
Meakk, line 27: I think we should be consistent and more modern on our API. We should throw in case of failure, and we should return `window&` to chain calls, what do you think?  
Maybe not for `update()`, but doing `window.setIcon().setWindowName()` make sense to me.

Meakk (approved):
Minor comment, feel free to ignore, we can do it later

library/public/window.h
mwestphal: Lets keep as it is now but discuss the API in details in #52

PR #305, opened 2022-05-31
Add interactor test and doc

mwestphal:


Meakk (approved):
LGTM once the VTK version check is fixed.

PR #306, opened 2022-05-31
Improve test_sdk

Meakk:


.github/test_sdk/main.cxx
```diff
@@ -1,10 +1,16 @@
...
+#include <f3d/options.h>
+#include <f3d/window.h>
 
 int main()
 {
-  f3d::engine eng(f3d::engine::CREATE_WINDOW | f3d::engine::CREATE_INTERACTOR);
+  f3d::engine eng(f3d::engine::FLAGS_NONE);
 
-  // TODO Actual test
+  f3d::log::info("F3D engine is loaded")
```
mwestphal, line 13: what about more than just creating an empty engine ? Loading a file and rendering it would be better and will check that the installation contains more than just headers and a few engine symbols.

mwestphal: I still think current test is a bit lacking. What is wrong by adding a simple load and render test instead ?

Meakk: In practice, we don't have a file available to open here, we do not know the path to the source dir. I can create an obj file manually but it seems complicated for something that is tested in other tests.

mwestphal: Indeed, I did not consider that. That is fine then.

mwestphal (approved):

PR #307, opened 2022-05-31
Improve doc

mwestphal:
- Improve all headers doc
 - Change image::compare API

### DIFF
--- a/library/private/window_impl.h
+++ b/library/private/window_impl.h
@@ -1,9 +1,16 @@
+/**
+ * @class   window_impl
+ * @brief   A private abstract implementation of window
+ *
+ * A private abstract implementation of window that defines the private API
+ * See window.h for the class documentation
+ */
+
 #ifndef f3d_window_impl_h
 #define f3d_window_impl_h
 
 #include "window.h"
 
-// TODO Doc
 class vtkRenderWindow;
 class vtkF3DGenericImporter;
 namespace f3d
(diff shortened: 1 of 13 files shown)


library/public/image.h
```diff
@@ -27,33 +28,80 @@ class F3D_EXPORT image
...
-
+  //@}
+
+  /**
+   * Compare current image to a reference using the provided threshold.
+   * If the comparison fails, output the resulting diff and error and return false,
+   * return true otherwise.
+   * Can throw an image::exception if images are not sane.
+   */
+  bool compare(const image& reference, image& diff, double threshold, double& error) const;
```
mwestphal, line 99: @Meakk should this be ` compare(const image& reference, double threshold, image& diff, double& error) const;` instead ?

library/private/window_impl.h
```diff
@@ -15,17 +22,41 @@ namespace detail
 class window_impl : public window
 {
 public:
+
+  /**
+   * Default const
```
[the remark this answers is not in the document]
Meakk: Is it necessary btw?

library/public/image.h
```diff
@@ -27,33 +28,72 @@ class F3D_EXPORT image
...
   image(const image& img);
   image& operator=(const image& img);
-
   image(image&& img);
   image& operator=(image&& img);
+  //@}
+
+  //@{
+  /**
+   * Comparison operators, uses image::compare
```
Meakk, line 52: Add that it will return true if the images exactly match, otherwise we can use `compare` with a threshold

library/public/image.h
```diff
@@ -27,33 +28,72 @@ class F3D_EXPORT image
...
-
-  bool compare(const image& reference, image& result, double threshold, double& error) const;
-
-  bool operator==(const image& reference) const;
-  bool operator!=(const image& reference) const;
-
+  //@}
+
+  /**
+   * Compare current image to a reference using the provided threshold.
```
Meakk, line 86: Can we give the user a hint of what exactly is the threshold? Personally, I have no idea.

library/public/image.h
```diff
@@ -27,33 +28,72 @@ class F3D_EXPORT image
...
+   * If the comparison fails, output the resulting diff and error and return false,
+   * return true otherwise.
+   * Can throw an image::exception if images are not sane.
+   */
+  bool compare(const image& reference, image& diff, double threshold, double& error) const;
+
+  /**
+   * Save an image to a file in .png format
+   * Can throw an image::exception if images is not sane.
```
Meakk, line 95: images -> image

library/public/options.h
```diff
@@ -1,3 +1,11 @@
+/**
+ * @class   options
+ * @brief   Class used to control the different options
+ *
+ * A class to control the different options of f3d.
+ * See the TODO for the full listing of options
```
Meakk, line 6: the TODO?

Meakk (changes requested):

library/public/options.h
mwestphal: Yup, we need to discuss the options :)

library/public/image.h
mwestphal: The VTK doc is not super clear. This is basically a pixel to pixel comparison with averaging and shifting. The value is 256 based but accumulate over multiple pixels.

library/private/window_impl.h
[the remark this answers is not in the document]
mwestphal: What I mean by that is that the implementation is `=default` and nothing more is done.

Meakk (approved):

PR #314, opened 2022-06-04
Set verbose on f3d::log

mwestphal:
Introduce f3d::log::setVerboseLevel and remove quiet and verbose options

### DIFF
--- a/library/VTKExtensions/Core/F3DLog.h
+++ b/library/VTKExtensions/Core/F3DLog.h
@@ -17,10 +17,17 @@ namespace F3DLog
 {
 enum class Severity : unsigned char
 {
+  Debug = 0,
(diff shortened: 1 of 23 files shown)


library/VTKExtensions/Core/F3DLog.h
```diff
@@ -17,11 +17,18 @@ namespace F3DLog
...
   Info,
   Warning,
   Error
 };
 
+/**
+ * Set this global variable to control the verbose level
+ * that actually display something in Print
+ */
+extern Severity VerboseLevel;
```
Meakk, line 30: Why is it `extern` instead of just `static`?

Meakk (approved):

PR #317, opened 2022-06-05
Stricter cmake build

mwestphal:


library/VTKExtensions/Rendering/Testing/CMakeLists.txt
```diff
@@ -1,4 +1,7 @@
-# TODO add actual tests
+if(VTK_VERSION VERSION_LESS_EQUAL 9.1.0)
+  cmake_policy(SET CMP0115 OLD)
```
Meakk, line 2: Why is it needed?

mwestphal: VTK must be doing something strange internally with the file extensions. Without this policy with VTK 9.0.X, there is a CMake warning.

library/testing/TestSDKExternalWindowQT.cxx
```diff
@@ -67,9 +67,11 @@ int TestSDKExternalWindowQT(int argc, char* argv[])
     std::string(argv[1]) + "/data/cow.vtp", std::string(argv[1]) + "/baselines/", argv[2]);
   w.setTitle("F3D QT External Window");
   w.resize(300, 300);
-  w.startTimer(1000);
+  int timerId = w.startTimer(1000);
```
Meakk, line 70: why is the leak appears in this PR but not on master?

mwestphal: this PR fix sanitizer. It was incorrectly set since some PR ago.

mwestphal: What is the right way to silence it ?

Meakk (approved):

PR #329, opened 2022-07-02
clang-tidy pass

Meakk:


library/public/image.h
```diff
@@ -43,8 +43,8 @@ class F3D_EXPORT image
   ~image();
   image(const image& img);
   image& operator=(const image& img);
-  image(image&& img);
-  image& operator=(image&& img);
+  image(image&& img) noexcept;
+  image& operator=(image&& img) noexcept;
```
mwestphal, line 47: why adding it only here ?

mwestphal:
Nice, but adding a tidy job (see #203 ) seems necessary to prevent this.

library/public/image.h
Meakk: We should add it to other functions too, but this one is important because it has an impact on the performance, see https://clang.llvm.org/extra/clang-tidy/checks/performance/noexcept-move-constructor.html

.clang-tidy
```diff
@@ -0,0 +1,3 @@
+Checks: '-*,performance-*,modernize-*,-modernize-use-trailing-return-type,-modernize-use-nodiscard,-modernize-concat-nested-namespaces,-modernize-avoid-c-arrays,-modernize-use-auto'
```
Meakk, line 1: We may want to add more checks in the future

Meakk:
I decided to try SonarCloud instead.

.clang-tidy
```diff
@@ -0,0 +1,4 @@
+Checks: '-*,performance-*,modernize-*,-modernize-use-trailing-return-type,-modernize-use-nodiscard,-modernize-avoid-c-arrays,-modernize-use-auto'
```
mwestphal, line 1: does this line enable or disable checks ?

.cppcheck.supp
```diff
@@ -0,0 +1,16 @@
+// remove some checks
+missingInclude
+unusedFunction
+
+// external libraries
+*:external/json.hpp
+*:external/cxxopts.hpp
+
+// opencascade
+*:*.hxx
```
mwestphal, line 10: this looks a bit too wide for this to be opencascade specific

.github/workflows/style-checks.yml
```diff
@@ -7,11 +7,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - uses: DoozyX/clang-format-lint-action@v0.13
+    - uses: DoozyX/clang-format-lint-action@v0.16.1
       with:
         source: 'library application python'
         extensions: 'h,cxx'
-        clangFormatVersion: 13
+        clangFormatVersion: 14 # Last Ubuntu LTS version (22.04)
```
mwestphal, line 14: last ubuntu lts is 23.04

application/F3DOptionsParser.h
```diff
@@ -88,10 +88,10 @@ class F3DOptionsParser
   F3DOptionsParser();
   ~F3DOptionsParser();
 
-private:
   F3DOptionsParser(F3DOptionsParser const&) = delete;
   void operator=(F3DOptionsParser const&) = delete;
 
+private:
```
mwestphal, line 94: not sure why though

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
```diff
@@ -439,7 +436,7 @@ void vtkF3DRendererWithColoring::SetColoring(
       // Not coloring
       this->ArrayIndexForColoring = -1;
     }
-    else if (arrayName != F3D_RESERVED_STRING && nIndexes == 0)
+    else if (nIndexes == 0)
```
mwestphal, line 439: this doesnt look right

library/src/animationManager.cxx
```diff
@@ -26,7 +26,7 @@ void animationManager::Initialize(
   this->Importer = importer;
 
   // This can be -1 if animation support is not implemented in the importer
-  int availAnimations = this->Importer->GetNumberOfAnimations();
+  vtkIdType availAnimations = this->Importer->GetNumberOfAnimations();
 
   if (availAnimations > 0)
```
mwestphal, line 31: would != be better ?

.clang-tidy
Meakk: When it starts with `-` it disables the checks. So `-*` means disable everything, then enable all `performance` and `modernize` checks, then disable specific checks in the two enabled groups.

.cppcheck.supp
Meakk: I know... But we are not using `hxx` extension. Should I list all the opencascade headers having an issue? (there are MANY of them)

.github/workflows/style-checks.yml
Meakk: Nop. 23.04 is not LTS.

application/F3DOptionsParser.h
Meakk: https://stackoverflow.com/questions/18931133/must-a-deleted-constructor-be-private

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
Meakk: clang tidy reports that `arrayName != F3D_RESERVED_STRING` is always true.

library/src/animationManager.cxx
Meakk: unrelated to this PR

.cppcheck.supp
mwestphal: not possible to do something like "*:opencascase/*.hxx" ?

.github/workflows/style-checks.yml
mwestphal: indeed

mwestphal: maybe we should use 22.04 in the CI if EGL testing works with it. Looks like I got confused at some point.

mwestphal: unrelated anyway

.clang-tidy
mwestphal: thanks for the explanation

.cppcheck.supp
Meakk: `*:*opencascade/*.hxx` seems to work

.github/workflows/ci.yml
```diff
@@ -484,3 +484,93 @@ jobs:
...
+        -DCMAKE_PREFIX_PATH:PATH=$(pwd)/../dependencies/install/
+        -DF3D_MODULE_EXTERNAL_RENDERING=ON
+        -DF3D_PLUGIN_BUILD_ALEMBIC=ON
+        -DF3D_PLUGIN_BUILD_ASSIMP=ON
+        -DF3D_PLUGIN_BUILD_DRACO=ON
+        -DF3D_PLUGIN_BUILD_OCCT=ON
+        -DF3D_STRICT_BUILD=ON
+        -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
+
+    # It's necessary to run the build step to generate F3DIcon.h
```
mwestphal, line 560: we could add a target for this if needed to avoid building everything.

mwestphal (approved):

.github/workflows/ci.yml
Meakk: Yes I thought about it, but building F3DIcon.h target only builds pretty much everything, so there might be an issue in the dependency graph. We can take a look at this later though.

PR #331, opened 2022-07-09
Rework description logic

mwestphal:
Previously, rendering information was outputted on each hitkey stroke. This rework the logic to output only when needed, which means:

Coloring information when changing colouring (debug)
Scene and coloring when pressing ? (Info)
Scene and coloring and loading file (debug)


Meakk (approved):
lgtm

Issue #332, opened 2022-07-10
--texture-base-color with opacity and depth peeling is not correct


mwestphal:
**Describe the bug**
Using a texture base color with opacity and depth peeling, without setting --opacity, cause a wrong rendering

**To Reproduce**
 1. Open the file using `f3d --dry-run WaterBottle.glb --texture-base-color=/home/glow/f3d/src/testing/data/albedo_mod.png --depth-peeling --geometry-only` : 
![albedo_mod](https://user-images.githubusercontent.com/3129530/178157806-9a09988b-6a79-42d8-a65c-d80e3c20492e.png)

 2. wrong rendering:
 
![wrong](https://user-images.githubusercontent.com/3129530/178157876-66e97abd-58e6-423d-86c9-446e2db9128e.png)

 4. Using the following provide a correct rendering: `f3d --dry-run WaterBottle.glb --texture-base-color=/home/glow/f3d/src/testing/data/albedo_mod.png --depth-peeling --geometry-only --opacity=0`

**Expected behavior**
Correct translucent rendering, like this:
![right](https://user-images.githubusercontent.com/3129530/178157900-c082db41-1682-4a47-908a-c2d66ddce67d.png)

5. The --opacity parameter is not taken into account at all, apart from enabling translucentsy, it should multiply.

Meakk:
Indeed. This texture is RGBA, with an alpha channel containing translucent pixels. But VTK isn't aware the object is translucent. My suggestion is to force the translucent flag on the actor if the color texture has 4 channels.

mwestphal:
Sounds like a good approach

mwestphal:
Looks like this should be set inside VTK, not in F3D.

mwestphal:
Added a note a about the fact this is not multiplied.

mwestphal:
See #333 for more info

PR #333, opened 2022-07-10
Better texture tests

mwestphal:
- Improve testing of texture options


cmake/testing.cmake
```diff
@@ -142,6 +142,10 @@ f3d_test(NAME TestVolumeComp DATA vase_4comp.vti ARGS -vb --comp=3 LONG_TIMEOUT)
 f3d_test(NAME TestVolumeDirect DATA vase_4comp.vti ARGS -vb --comp=-2 LONG_TIMEOUT)
 f3d_test(NAME TestVolumeCells DATA waveletArrays.vti ARGS -vb --cells LONG_TIMEOUT)
 f3d_test(NAME TestVolumeNonScalars DATA waveletArrays.vti ARGS -vb --scalars=RandomPointScalars LONG_TIMEOUT)
+f3d_test(NAME TestTextureNormal DATA WaterBottle.glb ARGS --geometry-only --texture-normal=${CMAKE_SOURCE_DIR}/testing/data/normal.png --normal-scale=0.1 DEFAULT_LIGHTS)
+f3d_test(NAME TestTextureMaterial DATA WaterBottle.glb ARGS --geometry-only --texture-material=${CMAKE_SOURCE_DIR}/testing/data/red.jpg --roughness=1 --metallic=1 DEFAULT_LIGHTS) # roughness and metallic have no effect ?
```
mwestphal, line 146: @Meakk PTAL

cmake/testing.cmake
```diff
@@ -142,6 +142,10 @@ f3d_test(NAME TestVolumeComp DATA vase_4comp.vti ARGS -vb --comp=3 LONG_TIMEOUT)
...
+f3d_test(NAME TestTextureColor DATA WaterBottle.glb ARGS --geometry-only --texture-base-color=${CMAKE_SOURCE_DIR}/testing/data/albedo_mod.png --color=1,1,0 --opacity=0.1 --depth-peeling DEFAULT_LIGHTS) #may be impacted by https://github.com/f3d-app/f3d/issues/332
```
mwestphal, line 147: @Meakk PTAL

cmake/testing.cmake
Meakk: You are setting the material texture with a texture having only a red pixel. So you have occlusion=red=1, roughness=green=0, metallic=blue=0.  
The values are multiplied by the constant value given by the user, but since the value of roughness and metallic is 0, you can give whateever you want in the constant values, the result is the same.

cmake/testing.cmake
Meakk: Ok, it seems like the `opacity` argument isn't taken into account when using a texture.

cmake/testing.cmake
mwestphal: Indeed, this was the cause of my confusion.

However, are you sure that:
>The values are multiplied by the constant value given by the user, 
?

Here is another test with the following texture:

```
R=1
G=0.5
B=0.5
```
red_mod

`f3d  --geometry-only --texture-material=red_mod.jpg`:
a

`f3d  --geometry-only --texture-material=red_mod.jpg --roughness=1`:
a


If that is indeed a multiplication, shouldn't these images be the same ?

cmake/testing.cmake
mwestphal: It does have an effect though. What should be the right behavior here in your opinion ?

cmake/testing.cmake
Meakk: In the first case, you have:
- default roughness (0.3) * G channel (0.5) = 0.15
- custom roughness (1) * G channel (0.5) = 0.5

In both cases, the metallic value is 0 (because the default value is 0 so the texture has no effect)

cmake/testing.cmake
Meakk: The effect is coming from the texture. The correct behavior should be multiplication, just like metallic and roughness.

mwestphal: no, with ou without opacity is not the same result

cmake/testing.cmake
mwestphal: Ok. I understand now.

Do we want this behavior ? Someone using --texture-material would not expect the default roughness and default metallic to multiply its own texture material ?

We could either:
- Keep this behavior but with better documentation
- Change the default to zero but set it in the config file
- Change the default to -1 and implement a smart default

Let me know what you think

Meakk: IMHO, what we should do is change the default values to roughess=1, metallic=1, when the material texture is set, so the multiplier does nothing, but the user can still modify it if he wants to.

mwestphal: This is the third solution.

cmake/testing.cmake
mwestphal: you are right, it is not multiplying

mwestphal: added a note, outside of the scope of this MR.

cmake/testing.cmake
mwestphal: 1, will be implemented, but not in this PR, see #307

cmake/testing.cmake
```diff
@@ -142,6 +142,11 @@ f3d_test(NAME TestVolumeComp DATA vase_4comp.vti ARGS -vb --comp=3 LONG_TIMEOUT)
 f3d_test(NAME TestVolumeDirect DATA vase_4comp.vti ARGS -vb --comp=-2 LONG_TIMEOUT)
 f3d_test(NAME TestVolumeCells DATA waveletArrays.vti ARGS -vb --cells LONG_TIMEOUT)
 f3d_test(NAME TestVolumeNonScalars DATA waveletArrays.vti ARGS -vb --scalars=RandomPointScalars LONG_TIMEOUT)
+f3d_test(NAME TestTextureNormal DATA WaterBottle.glb ARGS --geometry-only --texture-normal=${CMAKE_SOURCE_DIR}/testing/data/normal.png --normal-scale=0.1 DEFAULT_LIGHTS)
+f3d_test(NAME TestTextureMaterial DATA WaterBottle.glb ARGS --geometry-only --texture-material=${CMAKE_SOURCE_DIR}/testing/data/red_mod.jpg --roughness=1 --metallic=1 DEFAULT_LIGHTS)
+f3d_test(NAME TestTextureMaterialWithOptions DATA WaterBottle.glb ARGS --geometry-only --texture-material=${CMAKE_SOURCE_DIR}/testing/data/red_mod.jpg --roughness=0.5 --metallic=0.5 DEFAULT_LIGHTS)
+f3d_test(NAME TestTextureColorWithOptions DATA WaterBottle.glb ARGS --geometry-only --texture-base-color=${CMAKE_SOURCE_DIR}/testing/data/albedo_mod.png --color=1,1,0 --opacity=0 --depth-peeling DEFAULT_LIGHTS) # TODO add a texture-base-color test without --opacity and improve current test once https://github.com/f3d-app/f3d/issues/332 is fixed
```
Meakk, line 148: `opacity=0` means the object is not visible. Is it what you want?

mwestphal: the object is visible until #332 is fixed. By using this value, I ensure that we do not forget to update this test when fixing, hence the comment after.

Meakk: ok

Meakk (approved):

PR #339, opened 2022-07-16
Add a f3d::camera

mwestphal:
- Remove camera options and add a f3d::camera
- Add and improve tests
- Move camera setup from VTKExtensions to libf3d
- Slight change in behavior in F3D because of the way OrthogonalizeViewUp is used


library/public/camera.h
```diff
@@ -0,0 +1,97 @@
...
+  virtual double getViewAngle() = 0;
+  //@}
+ 
+ //@{
+  /**
+   * TODO
+   * Call render on the window before calling any of these methods, or they will throw an exception.
+   */
+  virtual void setViewMatrix(const matrix_t& matrix) = 0;
+  virtual matrix_t getViewMatrix() = 0;
```
mwestphal, line 55: is this what you need @Meakk ?

library/src/camera_impl.cxx
```diff
@@ -0,0 +1,241 @@
...
+  std::move(std::begin(to), std::end(to), arr.begin() + 8);
+  arr[11] = 0;
+  std::move(std::begin(pos), std::end(pos), arr.begin() + 12);
+  arr[15] = 1;
+#else
+  // TODO  this implementation seems better
+  vtkMatrix4x4* mat = cam->GetModelViewTransformMatrix();
+  mat->Transpose();
+  double* data = mat->GetData();
+  std::move(data, data + 16, arr.begin());
```
mwestphal, line 140: does that looks correct to you @Meakk ?

library/src/camera_impl.cxx
```diff
@@ -0,0 +1,241 @@
...
+
+//----------------------------------------------------------------------------
+void camera_impl::setViewMatrix(const matrix_t& matrix)
+{
+  vtkCamera* cam = this->GetVTKCamera();
+  cam->SetPosition(matrix[12], matrix[13], matrix[14]);
+  cam->SetFocalPoint(matrix[8] + matrix[12], matrix[9] + matrix[13], matrix[10] + matrix[14]);
+  cam->SetViewUp(matrix[4], matrix[5], matrix[6]);
+//   cam->OrthogonalizeViewUp(); TODO NOT NEEDED ?
+  this->Internals->VTKRenderer->ResetCameraClippingRange();
```
mwestphal, line 101: does that looks correct to you @Meakk ?

library/public/camera.h
```diff
@@ -0,0 +1,97 @@
...
+  class exception : public std::runtime_error
+  {
+  public:
+    exception(const std::string& what = "")
+      : std::runtime_error(what)
+    {
+    }
+  };
+
+  typedef std::array<double, 3> vector_t;
```
Meakk, line 31: `vector3_t` would be better

library/public/camera.h
```diff
@@ -0,0 +1,97 @@
...
+  typedef std::array<double, 16> matrix_t;
```
Meakk, line 32: `matrix4_t`

library/public/camera.h
Meakk: Yes, I need this one, but probably other functions too.

library/src/camera_impl.cxx
Meakk: Yes, you need to reset the clipping range everytime the camera position/orientation changes

library/src/camera_impl.cxx
Meakk: Hmmm. Not sure why it's called `ModelViewTransform`, we do not care about the model transform. Anyway, yes it looks correct.  
However, make sure the call to `Transpose()` does not modify the matrix inside VTK.

library/src/camera_impl.cxx
mwestphal: my question is about the whole method

library/src/camera_impl.cxx
mwestphal: see the doc here:

https://vtk.org/doc/nightly/html/classvtkCamera.html#a40f9e41a86e5b7c94fe171e7b9193673

Ok I will check

library/public/camera.h
mwestphal: Could you precise what you need so I can implement it ?

Meakk: Maybe not related to the camera, but I'll need way to get the world position from the position in screen space.  
But I'm not sure if I need it, I will investigate first.

library/src/camera_impl.cxx
Meakk: Yes, should be good. And indeed `cam->OrthogonalizeViewUp()` is not needed.

application/F3DOptionsParser.h
```diff
@@ -31,6 +31,12 @@ struct F3DAppOptions
   std::vector<int> Resolution{ 1000, 600 };
   bool Quiet = false;
   bool Verbose = false;
+  double CameraAzimuthAngle = 0.0;
+  double CameraElevationAngle = 0.0;
+  std::vector<double> CameraFocalPoint;
```
Meakk, line 36: no default value like other options?

library/private/camera_impl.h
```diff
@@ -0,0 +1,86 @@
...
+public:
+  /**
+   * Default constructor
+   */
+  camera_impl();
+
+  /**
+   * Default destructor
+   */
+  ~camera_impl() override;
```
Meakk, line 32: Rule of three, add a copy constructor, and a copy assignment operator. Ideally, it should even be rule of five and add move constructor/assignment.

library/public/camera.h
```diff
@@ -0,0 +1,90 @@
...
+{
+class F3D_EXPORT camera
+{
+public:
+  typedef std::array<double, 3> vector3_t;
+  typedef std::array<double, 16> matrix4_t;
+
+  //@{
+  /**
+   * Set/Get the canera parameters.
```
Meakk, line 29: camera

library/private/camera_impl.h
Meakk: nevermind, the copy is deleted in the base class.

library/src/camera_impl.cxx
```diff
@@ -0,0 +1,256 @@
...
+class camera_impl::internals
+{
+public:
+  vtkRenderer* VTKRenderer = nullptr;
+  vtkNew<vtkCamera> DefaultVTKCamera;
+};
+
+//----------------------------------------------------------------------------
+camera_impl::camera_impl()
+  : Internals(new camera_impl::internals())
```
Meakk, line 18: `Internals(std::make_unique<camera_impl::internals>())`?

library/public/camera.h
```diff
@@ -0,0 +1,90 @@
+/**
+ * @class   camera
+ * @brief   Abstract class to control a camera in a window
+ *
+ * A class to control a camera in a window.
+ * There is in essence three API.
+ * The Position/FocalPoint/ViewUp API, the ViewMatrix API and the camera movements API.
+ * They can be mixed by using ViewMatrix API with the other APIs can result to unexpected results.
```
Meakk, line 8:
```suggestion
 * They can be mixed by using ViewMatrix API with the other APIs but it can result in unexpected results.
```

library/public/camera.h
```diff
@@ -0,0 +1,90 @@
+/**
+ * @class   camera
+ * @brief   Abstract class to control a camera in a window
+ *
+ * A class to control a camera in a window.
+ * There is in essence three API.
+ * The Position/FocalPoint/ViewUp API, the ViewMatrix API and the camera movements API.
```
Meakk, line 7: Maybe add a comment that `SetViewUp` orthogonalize the input, if it's not orthogonal to the view direction.

library/testing/TestSDKCamera.cxx
```diff
@@ -0,0 +1,230 @@
...
+#include <cmath>
+#include <iomanip>
+#include <iostream>
+
+bool compareDouble(double a, double b)
+{
+  return std::fabs(a - b) < 128 * std::numeric_limits<double>::epsilon();
+}
+
+bool compareVec(f3d::camera::vector3_t vec1, f3d::camera::vector3_t vec2)
```
Meakk, line 15: I'm nitpicking here so feel free to ignore. A better way to compare two vectors is to check the length is the same and that the angle between them is zero.

library/testing/TestSDKCamera.cxx
```diff
@@ -0,0 +1,230 @@
...
+}
+
+bool compareMat(f3d::camera::matrix4_t mat1, f3d::camera::matrix4_t mat2)
+{
+  bool ret = true;
+  for (size_t i = 0; i < mat1.size(); i++)
+  {
+    ret &= compareDouble(mat1[i], mat2[i]);
+  }
+  return ret;
```
Meakk, line 28: I don't like `&=`, and you can actually early exit in case of failure.

```suggestion
  for (size_t i = 0; i < mat1.size(); i++)
  {
    if (!compareDouble(mat1[i], mat2[i]))
    {
      return false;
    }
  }
  return true;
```

Meakk (changes requested):

application/F3DOptionsParser.h
mwestphal: no, if not set, it is computed automatically. that was already the case before.

library/public/camera.h
mwestphal: almost all methods OrthogonalizeViewUp. Should I mention which ones ?

library/testing/TestSDKCamera.cxx
mwestphal: it is indeed a better geometrical approach, but does not seem necessary in this context.

library/src/camera_impl.cxx
mwestphal: done in another pr: #340

application/F3DOptionsParser.h
mwestphal: (it is intialized to an empty vector, which we rely on to know that it was not set)

application/F3DStarter.cxx
```diff
@@ -221,6 +251,7 @@ int F3DStarter::Start(int argc, char** argv)
     // Start interaction
     else
     {
+      window.render();
```
mwestphal, line 254: to check if still needed

library/public/camera.h
Meakk: Ah ok. Then, maybe just add a note that the up direction is orthogonalized automatically if it's not orthogonal to the view direction defined by the position and the focal point.

Meakk (approved):
Please add the comment about the orthogonalized up direction, and it's good to merge

PR #351, opened 2022-08-10
libf3d API Public Review

mwestphal:
# *libf3d API Public Review*

Hello and welcome to the public review of the libf3d

> Wait, what is the libf3d ?

The libf3d is a simple BSD licensed C++ library to render meshes, currently in developpement.

> Ok, but what is a public review ?

The idea of a public review is to let anyone interested give us their remarks and ideas about the
API of the libf3d.

While we have some understanding on how the libf3d could be used, there is many usecases
that we may have missed. If you have such usecase in mind, now is the time to share.

Secondly, this is the first time we create an complete, modern, C++11 API from scratch.
We tried to do it right, but maybe you have some insight, which would be very welcome !

> Sounds good, can I try the libf3d already ?

Yes, you can ! take a look here: https://github.com/f3d-app/f3d, the F3D project contains the libf3d as well as F3D,
a executable using it. But let's focus on the review for now

> Ok, where do I start ?

Take a look at the *Files Changed* tab in this pull request:
https://github.com/f3d-app/f3d/pull/351/files

For starters, you can review README_libf3d.md, it will give you some examples about how to use the libf3d,
the different classes and options.
Click here for easier reading: https://github.com/f3d-app/f3d/blob/30afbd07ecaa7240efcae6ecb7b5ef5743a73792/README_libf3d.md

Then you may want to reader README_F3D.md. While not everything there is related to the libf3d, it will give you some hindsight
about how you could use the libf3d and what we do with it for now.
Click here for easier reading: https://github.com/f3d-app/f3d/blob/30afbd07ecaa7240efcae6ecb7b5ef5743a73792/README_F3D.md

Finally, you can take a look in the library/public directory where each public class headers can be found.
The in-header doc is quite extensive, although it may not be enough. Do not hesitate to ask questions, we will clarify and improve the doc in the process.

So you are welcome to ask questions, make suggestions either here in the PR body or on specific part of the diff.

We will make sure to try to adress all of them.

> I'm new to github and dont know much about it.

Don't worry, you can't break anything. Just click on the parts that interest you and leave comments, we will figure it out :)

> Ok, for how long will this run ?

We plan to close this review by the end of august.

> This is all very abstract, how does a libf3ed render looks like ?

Take a look at our gallery here: https://f3d-app.github.io/f3d/gallery/

Here is a quick peek :)

image


library/public/options.h
```diff
@@ -0,0 +1,107 @@
...
+  options& set(const std::string& name, const std::vector<double>& values);
+  options& set(const std::string& name, std::initializer_list<int> values);
+  options& set(const std::string& name, std::initializer_list<double> values);
+  //@}
+
+  //@{
+  /**
+   * Reference getters for all supported types
+   */
+  void get(const std::string& name, bool& value) const;
```
Connor-Bowley, line 63: Is my understanding of these interfaces correct?

```
void get(const std::string& name, bool& value) const;
```
This  "Reference getter" will copy the value into an existing bool. Because this is a copy, any further changes to the passed in parameter after this function will not have any effect.

```
bool& getAsBoolRef(const std::string& name);
```
This "Explicit reference getter" will return a reference to the actual bool underneath, such that doing the following is equivalent to the `toggle` method.

```
auto& b = option.getAsBoolRef();
b = !b;
```

If my understanding is correct, I think the use of the phrase "Reference getter" in both methods is misleading, as it indicates they are doing the same thing, which they are not.

library/public/window.h
```diff
@@ -0,0 +1,98 @@
...
+   * Set the size of the window
+   */
+  virtual window& setSize(int width, int heigth) = 0;
+
+  /**
+   * Set the icon to be shown by a window manager.
+   * icon should be an unsigned char array.
+   * iconSize should be the sizeof(icon).
+   */
+  virtual window& setIcon(const void* icon, size_t iconSize) = 0;
```
Connor-Bowley, line 83: If the icon should be an unsigned char array, why not actually require a `const unsigned char*`?

Connor-Bowley: I don't know what your policy on requiring other libraries is, but consider something like `gsl::span` from the guidelines support library here.

https://github.com/gsl-lite/gsl-lite
https://github.com/gsl-lite/gsl-lite/blob/4720a2980a30da085b4ddb4a0ea2a71af7351a48/include/gsl/gsl-lite.hpp#L3649

library/public/camera.h
```diff
@@ -0,0 +1,93 @@
...
+  virtual matrix4_t getViewMatrix() = 0;
+  virtual void getViewMatrix(matrix4_t& matrix) = 0;
+  //@}
+
+  //@{
+  /**
+   * Standard camera manipulation methods.
+   */
+  virtual camera& dolly(double val) = 0;
+  virtual camera& roll(double angle) = 0;
```
Connor-Bowley, line 63: You should consider doing some kind of [strong typing](https://www.fluentcpp.com/2016/12/08/strong-types-for-strong-interfaces/) here and explicitly calling out Radians or Degrees as part of the function signature.

As it stands I do not know if this takes degrees or radians, and even if the documentation told me, there is still a lot more chance for errors taking a raw double than taking a class where you get type safety.

library/public/camera.h
```diff
@@ -0,0 +1,93 @@
...
+
+#include <array>
+#include <string>
+
+namespace f3d
+{
+class F3D_EXPORT camera
+{
+public:
+  typedef std::array<double, 3> vector3_t;
```
Connor-Bowley, line 27: You should consider making a distinction between points and vectors. They both are essentially a 3 value tuple, but they have different semantic meaning.

Also, if you implement Point3d and Vector3d classes, you can give them reasonable math operators like

```
Point + Vector => Point
Vector + Vector => Vector
Vector * Scalar => Vector
etc
```

which will make the library easier to work with.

library/public/engine.h
```diff
@@ -0,0 +1,142 @@
...
+   * - "OpenCASCADE module"
+   * - "Assimp module"
+   * - "Alembic module"
+   * - "VTK version"
+   * - "Copyright_previous"
+   * - "Copyright"
+   * - "License"
+   * - "Authors"
+   */
+  static std::map<std::string, std::string> getLibInfo();
```
Connor-Bowley, line 119: Any particular reason a string to string map was chosen over a struct? An advantage of a struct is the compiler will tell you when you accidentally type "VKTVersion", and compile time errors are _much_ easier to debug.

library/public/engine.h
```diff
@@ -0,0 +1,142 @@
...
+
+namespace f3d
+{
+class options;
+class loader;
+class interactor;
+class F3D_EXPORT engine
+{
+public:
+  class exception : public std::runtime_error
```
Connor-Bowley, line 40: If there are multiple reasons for failure, consider making separate exceptions for each of them, so more granular `catch` statements are supported.

library/public/camera.h
```diff
@@ -0,0 +1,93 @@
...
+  virtual camera& resetToDefault() = 0;
+
+  /**
+   * Reset the camera using the bounds of actors in the scene.
+   */
+  virtual camera& resetToBounds() = 0;
+
+protected:
+  camera() = default;
+  virtual ~camera() = default;
```
Connor-Bowley, line 87: Are you sure you want the `camera` destructor to be protected? I believe this would prevent me from doing something like `std::unique_ptr<camera> c = std::make_unique<derived_camera>(args);`, which seems like a reasonable thing to do.

library/public/interactor.h
```diff
@@ -0,0 +1,111 @@
...
+ */
+
+#ifndef f3d_interactor_h
+#define f3d_interactor_h
+
+#include "export.h"
+
+#include <functional>
+#include <string>
+#include <vector>
```
Connor-Bowley, line 16: You probably need to add `<utility>` for standard pair.

library/public/engine.h
```diff
@@ -0,0 +1,142 @@
...
+#include "window.h"
+
+#include <map>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+namespace f3d
+{
+class options;
```
Connor-Bowley, line 34: You should consider including these classes instead of forward declaring them. There is some debate on forward declarations vs includes, but it increases usability of headers when they are self contained.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf11-header-files-should-be-self-contained

Now, something like `engine::internals` Pimpl declaration should definitely be forward declared, so I'm not saying don't ever forward declare, but consider preferring including thing that a user needs to use the class.

library/public/camera.h
Connor-Bowley: Note this comment applies to more classes than this one.

library/public/engine.h
```diff
@@ -0,0 +1,142 @@
...
+
+  /**
+   * Get a vector containing info about the supported readers
+   */
+  static std::vector<readerInformation> getReadersInfo();
+
+private:
+  class internals;
+  internals* Internals;
+  engine(const engine& opt) = delete;
```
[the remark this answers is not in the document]
mwestphal: good point, will be taken into account.

library/public/camera.h
mwestphal: since the constructor is protected to and users are not supposed to create camera themselves, i'd say that this is not supported.

library/public/options.h
```diff
@@ -0,0 +1,107 @@
...
+      : std::runtime_error(what)
+    {
+    }
+  };
+
+  //@{
+  /**
+   * Setters for all supported types
+   */
+  options& set(const std::string& name, bool value);
```
jcfr, line 48: Explicitly listing the options may be sensible. It would allow to maintain associated documentation & deprecation scheme as well as writing python code where such option would auto-complete.

jcfr: Say differently, available options available for each type of objects (model, ui, scene, ...) should probably be query-able. The associated description and type retrievable as well.

README_libf3d.md
```diff
@@ -0,0 +1,224 @@
...
+
+```python
+import f3d
+
+eng = f3d.engine()
+eng.getOptions()
+  .set("model.scivis.array-name", "Normals")
+  .set("model.scivis.component", 0)
+  .set("ui.bar", True)
+  .set("scene.grid", True)
```
jcfr, line 209: Updating existing options should likely support a dict:

```python
eng.options.update({
  "model.scivis.array-name": "Normals",
  "model.scivis.component": 0,
  "ui.bar": True,
  "scene.grid": True,
})
```

as well as explicit setting doing:

```python
eng.options.model.scivis.component = 0
```

library/public/loader.h
```diff
@@ -0,0 +1,94 @@
+/**
+ * @class   loader
```
jcfr, line 2: Should this be renamed `renderer` ? 

It seems like:
* addFile read from disk (or memory) and represent is an "f3d object"
`loadFile()` is in-fact `render(objectIdx)`

library/public/options.h
mwestphal: This is exactly what we do not want. By having a string based API we allow ouselves to add/deprecate options with minimum effort, from both lib dev and lib users.

However, having a way to query available options and generate automatically python code could be considered.

README_libf3d.md
mwestphal: The first API could be interesting.

The second one require more work to stay in the logic of the string based API.

library/public/loader.h
mwestphal: Im afraid it is not. loadFile do not render anything. window::render() does, as well as interactor::start()

mwestphal:
FYI @whiiiskyy @nyue @caioaao @schuhumi @kevinsmia1939 @SamuelTallet

library/public/options.h
Meakk: You are right. You get it from de function signature but the comment is misleading

library/public/window.h
Meakk: I agree about the type of the pointer.  
For gsl, we should avoid it, especially in the public headers.

library/public/camera.h
Meakk: I agree. It's also convenient when transforming with a 4x4 matrix to know if the translation should be applied.

library/public/options.h
jcfr: Having a specialized domain for specifying options would reduce chance of error and would help integrate and use the library in application.

Using any of these would streamline usage, reduce confusion and be less error prone:
* enum classes
* constant to reference the option key
* strongly type object allowing to easily know if an option is a boolean, integer, double, ...

library/public/loader.h
jcfr: Since the loader seems to a convenience class for selecting files, I would suggest to keep it specific to the application and not expose it in the public API.

Application integrating this library will likely have their own UI library providing a tailored way of organizing the files that is addressing their domain specific use case.

mwestphal: The loader load the file, I think this is easy to understand for users.

library/public/engine.h
mwestphal: The idea was to be as flexible as possible. Wdyt @Meakk ?

library/public/engine.h
mwestphal: good point

library/public/interactor.h
mwestphal: indeed !

library/public/engine.h
mwestphal: @Meakk https://stackoverflow.com/questions/9906402/should-one-use-forward-declarations-instead-of-includes-wherever-possible. I think this way of fwd declaration is idomatic from VTK, but do not make sens in the context of the public header of the libf3d.

library/public/loader.h
```diff
@@ -0,0 +1,94 @@
...
+    LOAD_CURRENT,
+    LOAD_NEXT,
+    LOAD_LAST
+  };
+
+  /**
+   * Add a list of files or directories to be loaded.
+   * Duplicate will not be added.
+   */
+  virtual loader& addFiles(const std::vector<std::string>& files) = 0;
```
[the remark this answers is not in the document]
mwestphal: indeed.

library/public/options.h
mwestphal: i will update the doc

library/public/window.h
mwestphal: indeed

library/public/camera.h
mwestphal: that would be interesting to add, but not only, in many more locations, wdyt @Meakk ?

In the meantime, Ill just improve the documentation.

library/public/options.h
mwestphal: Wdyt @Meakk ?

README_libf3d.md
mwestphal: Wdyt @Meakk ?

library/public/camera.h
mwestphal: I will add the types, not the operation, for now.

library/public/camera.h
mwestphal: I added a typedef but defining the templates as proposed in your link will not be in the alpha. Will we discuss if we want to do it,

library/public/engine.h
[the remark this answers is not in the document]
mwestphal: Adressed in https://github.com/f3d-app/f3d/pull/355

library/public/loader.h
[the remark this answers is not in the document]
mwestphal: Adressed in https://github.com/f3d-app/f3d/pull/355

library/public/options.h
mwestphal: Adressed in https://github.com/f3d-app/f3d/pull/355

library/public/camera.h
mwestphal: Partially adressed in https://github.com/f3d-app/f3d/pull/355, keeping this open for now.

library/public/camera.h
mwestphal: Adressed in https://github.com/f3d-app/f3d/pull/355

library/public/engine.h
mwestphal: Adressed in https://github.com/f3d-app/f3d/pull/355

library/public/camera.h
mwestphal: Will not be adressed.

library/public/engine.h
mwestphal: Adressed in https://github.com/f3d-app/f3d/pull/355

library/public/interactor.h
mwestphal: Adressed in https://github.com/f3d-app/f3d/pull/355

library/public/window.h
mwestphal: Adressed in https://github.com/f3d-app/f3d/pull/355

library/public/engine.h
Meakk: Yes the idea was to be flexible like the options class. However, for this particular map, we could create a new class or an internal struct, it's unlikely to change regularly and most users don't care about it anyway.

README_libf3d.md
Meakk: I like the first idea. For the second one, it may be difficult to implement and can also result in a more complicated api

library/public/options.h
Meakk: We already had this discussion, it's all about strong typing versus flexibility. With the current implementation, the api is much simpler, but yes, the user has to know about the options and will see errors at runtime only. We can discuss it offline again @mwestphal

library/public/engine.h
mwestphal: Thinking back on it, this was just laziness on my part. Adressed here: https://github.com/f3d-app/f3d/pull/360

README_libf3d.md
mwestphal: In any case, this would be python specific API. Lets keep this out for now.

library/public/camera.h
mwestphal: https://github.com/f3d-app/f3d/issues/361

README_libf3d.md
mwestphal: https://github.com/f3d-app/f3d/issues/362

mwestphal:
Thanks a lot @jcfr @Connor-Bowley , we have taken your comments into account.

We will keep this open for a week but if anyone else have some feedback, it would be now or never.

mwestphal:
Closing this as we are moving into release.

PR #355, opened 2022-08-14
Adressing libf3d API Public Review

mwestphal:
- adding recursive to addFIles
- improving options doc
- fix setIcon type
- better camera doc and typing
- rework exceptions
- missing include
- using includes instead of fwd decl when it makes sense
- rule of 5

see https://github.com/f3d-app/f3d/pull/351

### DIFF
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -68,12 +68,12 @@ configure_file(
 
 # Define libf3d target
 set(F3D_SOURCE_FILES
-  ${CMAKE_CURRENT_SOURCE_DIR}/src/camera_impl.cxx
   ${CMAKE_CURRENT_BINARY_DIR}/src/config.cxx
   ${CMAKE_CURRENT_SOURCE_DIR}/src/animationManager.cxx
-  ${CMAKE_CURRENT_SOURCE_DIR}/src/init.cxx
-  ${CMAKE_CURRENT_SOURCE_DIR}/src/image.cxx
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/camera_impl.cxx
   ${CMAKE_CURRENT_SOURCE_DIR}/src/engine.cxx
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/image.cxx
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/init.cxx
   ${CMAKE_CURRENT_SOURCE_DIR}/src/interactor.cxx
   ${CMAKE_CURRENT_SOURCE_DIR}/src/interactor_impl.cxx
   ${CMAKE_CURRENT_SOURCE_DIR}/src/loader_impl.cxx
@@ -84,8 +84,9 @@ set(F3D_SOURCE_FILES
 
 # List of headers that will be installed
 set(F3D_PUBLIC_HEADERS
-  ${CMAKE_CURRENT_SOURCE_DIR}/public/camera.h
+  ${CMAKE_CURRENT_BINARY_DIR}/public/exception.h
   ${CMAKE_CURRENT_BINARY_DIR}/public/export.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/public/camera.h
   ${CMAKE_CURRENT_SOURCE_DIR}/public/engine.h
   ${CMAKE_CURRENT_SOURCE_DIR}/public/image.h
   ${CMAKE_CURRENT_SOURCE_DIR}/public/interactor.h
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -93,6 +93,7 @@ set(F3D_PUBLIC_HEADERS
   ${CMAKE_CURRENT_SOURCE_DIR}/public/loader.h
   ${CMAKE_CURRENT_SOURCE_DIR}/public/log.h
   ${CMAKE_CURRENT_SOURCE_DIR}/public/options.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/public/types.h
   ${CMAKE_CURRENT_SOURCE_DIR}/public/window.h
   )
 
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -84,7 +84,7 @@ set(F3D_SOURCE_FILES
 
 # List of headers that will be installed
 set(F3D_PUBLIC_HEADERS
-  ${CMAKE_CURRENT_BINARY_DIR}/public/exception.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/public/exception.h
   ${CMAKE_CURRENT_BINARY_DIR}/public/export.h
   ${CMAKE_CURRENT_SOURCE_DIR}/public/camera.h
   ${CMAKE_CURRENT_SOURCE_DIR}/public/engine.h
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -68,9 +68,9 @@ configure_file(
 
 # Define libf3d target
 set(F3D_SOURCE_FILES
-  ${CMAKE_CURRENT_BINARY_DIR}/src/config.cxx
   ${CMAKE_CURRENT_SOURCE_DIR}/src/animationManager.cxx
   ${CMAKE_CURRENT_SOURCE_DIR}/src/camera_impl.cxx
+  ${CMAKE_CURRENT_BINARY_DIR}/src/config.cxx
   ${CMAKE_CURRENT_SOURCE_DIR}/src/engine.cxx
   ${CMAKE_CURRENT_SOURCE_DIR}/src/image.cxx
   ${CMAKE_CURRENT_SOURCE_DIR}/src/init.cxx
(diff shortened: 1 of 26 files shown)


mwestphal:
@Connor-Bowley ptal :)

library/public/engine.h
```diff
@@ -21,38 +21,31 @@
 #ifndef f3d_engine_h
 #define f3d_engine_h
 
+#include "exception.h"
 #include "export.h"
```
Meakk, line 25: I'm honestly not sure about this change. There are some drawbacks doing this, like increasing the compilation time for the user. I still think forward decl is better.

Meakk: If we want to simplify usability, we can provide a f3d.h file which include all other files. This way, users who want to include only required headers can still do it.

Meakk (changes requested):

library/public/camera.h
```diff
@@ -24,25 +25,39 @@ namespace f3d
 class F3D_EXPORT camera
 {
 public:
-  typedef std::array<double, 3> vector3_t;
-  typedef std::array<double, 16> matrix4_t;
+  //@{
+  /**
+   * Camera specific types.
+   */
+  struct point3_t : std::array<double, 3>
```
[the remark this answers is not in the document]
Connor-Bowley: This is more along the lines of strong typing, which will enable the compiler to prevent doing things like accidentally using a vector when meaning to use a point. There is more type safety.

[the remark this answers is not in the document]
mwestphal: Yup, that would be useless without the struct/class

library/public/engine.h
mwestphal: With less than 10 headers, I think there is no increase. Did you read the link provided by @Connor-Bowley ?

library/public/camera.h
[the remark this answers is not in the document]
Meakk: Ok, then it should probably goes in its own file, or at least move it outside of the camera class scope, so we can reuse these types in other places later.

library/public/engine.h
Meakk: Fair enough. I'm not used to that but I guess it's an old habit

library/public/camera.h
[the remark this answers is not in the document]
mwestphal: good point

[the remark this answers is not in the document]
Connor-Bowley: Agree with its own file outside of camera. I would think you would also want to add some math operators and things like `operator<<`.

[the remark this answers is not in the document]
mwestphal: it may be added later, this feels a bit outside of the scope of the first version imo.

library/CMakeLists.txt
```diff
@@ -68,12 +68,12 @@ configure_file(
 
 # Define libf3d target
 set(F3D_SOURCE_FILES
-  ${CMAKE_CURRENT_SOURCE_DIR}/src/camera_impl.cxx
   ${CMAKE_CURRENT_BINARY_DIR}/src/config.cxx
   ${CMAKE_CURRENT_SOURCE_DIR}/src/animationManager.cxx
-  ${CMAKE_CURRENT_SOURCE_DIR}/src/init.cxx
-  ${CMAKE_CURRENT_SOURCE_DIR}/src/image.cxx
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/camera_impl.cxx
```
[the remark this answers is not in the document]
mwestphal: they are sorted, but some are using binary dir. ill fix them manually

library/public/types.h
```diff
@@ -0,0 +1,40 @@
...
+ * Describe a 4x4 matrix
+ */
+struct matrix4_t : std::array<double, 16>
+{
+};
+
+/**
+ * Describe an angle in degrees
+ */
+typedef double angleDeg_t;
```
Meakk, line 37:
```suggestion
using angleDeg_t = double;
```

library/src/camera_impl.cxx
```diff
@@ -33,22 +33,22 @@ camera& camera_impl::setPosition(const vector3_t& pos)
 }
 
 //----------------------------------------------------------------------------
-camera::vector3_t camera_impl::getPosition()
+f3d::point3_t camera_impl::getPosition()
```
Meakk, line 36:
```suggestion
point3_t camera_impl::getPosition()
```

library/src/camera_impl.cxx
```diff
@@ -58,15 +58,15 @@ camera& camera_impl::setFocalPoint(const vector3_t& foc)
 }
 
 //----------------------------------------------------------------------------
-camera::vector3_t camera_impl::getFocalPoint()
+f3d::point3_t camera_impl::getFocalPoint()
```
Meakk, line 61:
```suggestion
point3_t camera_impl::getFocalPoint()
```

library/src/camera_impl.cxx
```diff
@@ -83,7 +83,7 @@ camera& camera_impl::setViewUp(const vector3_t& up)
 }
 
 //----------------------------------------------------------------------------
-camera::vector3_t camera_impl::getViewUp()
+f3d::vector3_t camera_impl::getViewUp()
```
Meakk, line 86:
```suggestion
vector3_t camera_impl::getViewUp()
```

library/src/camera_impl.cxx
```diff
@@ -108,15 +108,15 @@ camera& camera_impl::setViewAngle(const double& angle)
 }
 
 //----------------------------------------------------------------------------
-double camera_impl::getViewAngle()
+f3d::angleDeg_t camera_impl::getViewAngle()
```
Meakk, line 111:
```suggestion
angleDeg_t camera_impl::getViewAngle()
```

library/src/camera_impl.cxx
```diff
@@ -134,7 +134,7 @@ camera& camera_impl::setViewMatrix(const matrix4_t& matrix)
 }
 
 //----------------------------------------------------------------------------
-camera::matrix4_t camera_impl::getViewMatrix()
+f3d::matrix4_t camera_impl::getViewMatrix()
```
Meakk, line 137:
```suggestion
matrix4_t camera_impl::getViewMatrix()
```

library/src/image.cxx
```diff
@@ -25,19 +25,13 @@ class image::internals
 
   vtkSmartPointer<vtkImageImport> GetVTKImporter() const
   {
-    if (this->Buffer.size() != this->Width * this->Height * this->Channels)
```
Meakk, line 28: Why did you remove this block?

python/F3DPythonBindings.cxx
```diff
@@ -92,7 +92,7 @@ PYBIND11_MODULE(f3d, module)
     .export_values();
 
   loader
-    .def("addFiles", &f3d::loader::addFiles, "Add files to the loader", py::arg("list of files"))
+    .def("addFiles", &f3d::loader::addFiles, "Add files to the loader", py::arg("list of files"), py::arg("recursive") = false)
```
Meakk, line 95: This file will need more changes, but it can be done later

Meakk (changes requested):

python/F3DPythonBindings.cxx
mwestphal: yes

library/src/image.cxx
mwestphal: Because this test just cannot be true. I could put an assert if you want.

Meakk: Yes please

library/private/camera_impl.h
```diff
@@ -35,28 +35,28 @@ class camera_impl : public camera
...
+  camera& setFocalPoint(const point3_t& foc) override;
+  point3_t getFocalPoint() override;
+  void getFocalPoint(point3_t& foc) override;
   camera& setViewUp(const vector3_t& up) override;
   vector3_t getViewUp() override;
   void getViewUp(vector3_t& up) override;
-  camera& setViewAngle(const double& angle) override;
-  double getViewAngle() override;
-  void getViewAngle(double& angle) override;
+  camera& setViewAngle(const angleDeg_t& angle) override;
```
Meakk, line 47: I think it should be snake case: `angle_deg_t`. See https://en.cppreference.com/w/cpp/types/max_align_t for example

library/src/camera_impl.cxx
```diff
@@ -33,22 +33,22 @@ camera& camera_impl::setPosition(const vector3_t& pos)
...
-void camera_impl::getPosition(vector3_t& pos)
+void camera_impl::getPosition(point3_t& pos)
 {
   vtkCamera* cam = this->GetVTKCamera();
   cam->GetPosition(pos.data());
 }
 
 //----------------------------------------------------------------------------
-camera& camera_impl::setFocalPoint(const vector3_t& foc)
+camera& camera_impl::setFocalPoint(const point3_t& foc)
```
Meakk, line 51: Can you use initializer list here with the current implementation? I'm expecting user wanting to call `setFocalPoint({ 0, 0, 0})`

Meakk (changes requested):

library/src/camera_impl.cxx
mwestphal: should I put that in types.h or in camera_impl::setFocalPoint ?

Meakk: Actually, Maybe just add setters with 3 doubles.

Connor-Bowley: I do think that making sure that `std::array<double, 3>`'s initializer list constructor is visible would be useful. The downside of setters with 3 doubles is the loss of typing.

mwestphal: > Actually, Maybe just add setters with 3 doubles.

I dont think this is the right way to go, why even defining a vector3_t in that case ?

> that std::array<double, 3>'s initializer list constructor is visible would be useful. 

I also think so

library/private/camera_impl.h
mwestphal: why should typedef style be different than local variables ? Afaik, cppref seems to use snake case everywhere.

Meakk: If you choose to name the types with the _t suffix, it's to mimic the STL types, so we should use snake case in this case. I don't like point_3_t though, I think point3_t looks fine.

Meakk (approved):

Issue #356, opened 2022-08-14
Camera missing from python bindings


mwestphal:
Camera API is missing from the python bindings

PR #363, opened 2022-08-22
Add camera python bindings

mwestphal:
Fix: https://github.com/f3d-app/f3d/issues/356

### DIFF
--- a/python/F3DPythonBindings.cxx
+++ b/python/F3DPythonBindings.cxx
@@ -18,8 +18,6 @@ PYBIND11_MODULE(f3d, module)
   module.doc() = "f3d library bindings";
 
   // types
-  py::class_<f3d::angle_deg_t>(module, "angle_deg_t")
-    .def(py::init<double>());
   py::class_<f3d::point3_t>(module, "point3_t")
     .def(py::init<double, double, double>());
   py::class_<f3d::vector3_t>(module, "vector3_t")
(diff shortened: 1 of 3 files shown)


python/F3DPythonBindings.cxx
```diff
@@ -103,6 +106,34 @@ PYBIND11_MODULE(f3d, module)
     .def(
       "getFileInfo", &f3d::loader::getFileInfo, "Get the file index, path, and information string");
 
+  // f3d::camera
+  py::class_<f3d::camera, std::unique_ptr<f3d::camera, py::nodelete> > camera(module, "camera");
+
+  camera.def("setPosition", &f3d::camera::setPosition)
+    .def("getPosition", py::overload_cast<>(&f3d::camera::getPosition))
+    .def("getPosition", py::overload_cast<f3d::point3_t&>(&f3d::camera::getPosition))
```
Meakk, line 114: Don't you have to bind the types too?

Meakk (changes requested):
Please add a test

mwestphal:
The python API is missing many tests. Should all bindings be tested ?

python/F3DPythonBindings.cxx
mwestphal: Not sure what you mean, not a pybind11 expert here.

mwestphal: ok, got it now.

python/testing/TestPythonCamera.py
```diff
@@ -0,0 +1,43 @@
+import f3d
+import sys
```
Meakk, line 2: sys is not required I think

Meakk (approved):
Minor change but looks good to me

PR #373, opened 2022-08-26
Fix config file

mwestphal:
- Fix config file doc
- Fix binary config file recovery
- Modify binary config file location on linux
- Add missing component in packaging
- Change cmake logic for config file location so the config file works with .deb package


cmake/cpack_project_config.cmake
```diff
@@ -0,0 +1,4 @@
+# Override deb installation location to /use/local/
```
mwestphal, line 1: usr

application/F3DConfigFileTools.cxx
```diff
@@ -0,0 +1,139 @@
...
+      f3d::log::error("Executable is too long to recover path to config file");
+      return std::string();
+    }
+    execPath = buffer;
+#else
+    execPath = std::filesystem::canonical("/proc/self/exe").string();
+#endif
+#endif
+
+    // transfrom path to exe to path to install
```
Meakk, line 82:
```suggestion
    // transform path to exe to path to install
```

Meakk (changes requested):

Meakk (approved):

PR #380, opened 2022-08-30
MacOS package targets Catalina (10.15)

Meakk:
New github runners are using macOS 11 (Big Sur).  
Since most of the market share is still macOS 10.15 (Catalina), let's target this platform as the minimum supported version.

### DIFF
--- a/library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
+++ b/library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
@@ -58,16 +58,20 @@ void vtkF3DRendererWithColoring::SetPointSize(double pointSize)
   if (this->PointGaussianMapper)
   {
     vtkDataSet* dataset = this->PolyDataMapper->GetInput();
-    double bounds[6];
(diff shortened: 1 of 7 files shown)


library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
```diff
@@ -58,16 +58,20 @@ void vtkF3DRendererWithColoring::SetPointSize(double pointSize)
   if (this->PointGaussianMapper)
   {
     vtkDataSet* dataset = this->PolyDataMapper->GetInput();
-    double bounds[6];
-    dataset->GetBounds(bounds);
-    vtkBoundingBox bbox(bounds);
 
-    double gaussianPointSize = 1.0;
-    if (bbox.IsValid())
+    if (dataset)
```
Meakk, line 62: @mwestphal this change is necessary to avoid a crash I was experiencing if the object failed to load for strange reason on my machine. It's not required, but it does not hurt to keep it

mwestphal (approved):

PR #385, opened 2022-08-31
Fix file association on Windows

jpouderoux:


jpouderoux:
@mwestphal

mwestphal:
That was trivial, I wonder if it ever worked at all, thanks a lot !

mwestphal (approved):

PR #390, opened 2022-09-03
Add Java bindings

Meakk:
Add Java bindings architecture with a working test.  
All the APIs are not bound yet, but it will be fixed with #426  
A few notes:
- We need JDK 8+ to run the test without compiling it first
- Windows works fine, but on a headless server, we need to load Mesa `opengl32.dll` manually first
- On MacOS, we need to run java with the option `-XstartOnFirstThread` otherwise an exception is raised
- VTK 9.1 is crashing on MacOS for some reason (it's fixed on master)

When `F3D_JAVA_BINDINGS` is enabled, a native shared library called `f3d-java` is built containing the JNI symbols. A `f3d.jar` package is also generated containing the Java classes.  
So, in order to use the Java classes, you need to import the JAR file, and put the native `f3d-java` library somewhere in your `java.library.path`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,12 @@ jobs:
         fetch-depth: 0
         lfs: 'false'
 
+    - name: Setup JDK
+      uses: actions/setup-java@v3
+      with:
+        distribution: 'temurin'
+        java-version: '17'
+
     - name: Set LFS env var
       working-directory: ${{github.workspace}}/source
       shell: bash
@@ -104,6 +110,7 @@ jobs:
         -DF3D_MODULE_ASSIMP=ON
         -DF3D_MODULE_ALEMBIC=ON
         -DF3D_PYTHON_BINDINGS=ON
+        -DF3D_JAVA_BINDINGS=ON
         -DCMAKE_BUILD_TYPE=Release
         -DVTK_DIR:PATH=$(pwd)/../dependencies/vtk_install/lib/cmake/vtk
         -Dassimp_DIR:PATH=$(pwd)/../dependencies/assimp_install/lib/cmake/assimp-5.1
--- a/README_libf3d.md
+++ b/README_libf3d.md
@@ -3,7 +3,7 @@
 By Michael Migliore and Mathieu Westphal.
 
 libf3d is a BSD-licensed C++ library to open and render 3D meshes. It is of course used by F3D.
-libf3d API is simple and easy to learn. Python bindings are provided through pybind11.
+libf3d API is simple and easy to learn. Python bindings are provided through pybind11. Java bindings are also available.
 
 # Getting Started
 
@@ -186,7 +186,9 @@ Options|Default|Type|Description|F3D option|Trigger
 ------|------|------|------|------|------
 window.fullscreen|false|bool|Display in fullscreen.|--fullscreen|render
 
-# Python Bindings
+# Bindings
+
+## Python Bindings
 
 If the python bindings are generated, the libf3d can be used directly from python.
 Make sure to set `PYTHONPATH` to path where the python module is built.
@@ -204,6 +206,31 @@ eng.getOptions()
 
 eng.getLoader().addFile("f3d/testing/data/dragon.vtu").loadFile()
 eng.getInteractor().start()
+```
+
+## Java Bindings
+
+If the Java bindings are generated, the libf3d can be used directly from Java.
+You can import the `f3d.jar` package and use the provided Java classes directly.
+Make sure to set `java.library.path` to path where the JNI library is built.
+Here is an example showing how to use libf3d Java bindings:
+
+```java
+import io.github.f3d_app.f3d.*;
+
+public class F3DExample {
+  public static void main(String[] args) {
+
+    // Always use try-with-resources idiom to ensure the native engine is released
+    try (Engine engine = new Engine(Window.Type.NATIVE)) {
+      Loader loader = engine.getLoader();
+      loader.addFile("f3d/testing/data/dragon.vtu");
+      loader.loadFile(Loader.LoadFileEnum.LOAD_FIRST);
+
+      engine.getWindow().render();
+    }
+  }
+}
 ```
 
 # Building against the libf3d
--- a/cmake/installing.cmake
+++ b/cmake/installing.cmake
@@ -19,6 +19,13 @@ if (BUILD_WINDOWS_SHELL_THUMBNAILS_EXTENSION)
     RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT shellext)
 endif()
 
+# Java bindings
+if (F3D_JAVA_BINDINGS)
+  install(TARGETS javaf3d
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT java)
+  install_jar(f3d-jar DESTINATION "share/java" COMPONENT java)
+endif()
+
 # Python module
 if (F3D_PYTHON_BINDINGS)
   if(WIN32)
--- a/java/F3DJavaBindings.cxx
+++ b/java/F3DJavaBindings.cxx
@@ -1,0 +1,183 @@
+// Automatically generated headers
+#include <io_github_f3d_app_f3d_Camera.h>
+#include <io_github_f3d_app_f3d_Engine.h>
+#include <io_github_f3d_app_f3d_Loader.h>
+#include <io_github_f3d_app_f3d_Options.h>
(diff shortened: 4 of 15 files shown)


mwestphal:
Can we add some kind of tests ?

.github/workflows/ci.yml
```diff
@@ -36,6 +36,12 @@ jobs:
         fetch-depth: 0
         lfs: 'false'
 
+    - name: Setup JDK
+      uses: actions/setup-java@v3
+      with:
+        distribution: 'temurin'
+        java-version: '17'
```
mwestphal, line 43: what does this do, since it is already added to the docker image ?

mwestphal: And why is it not needed in android workflow ?

java/Options.java
```diff
@@ -0,0 +1,12 @@
+package io.github.f3d_app.f3d;
+
+public class Options {
+
+    public Options(long nativeAddress) {
+        mNativeAddress = nativeAddress;
+    }
+
+    public native void toggle(String name);
```
mwestphal, line 9: looks like you did not expose the whole public API, is that intentionnal ?

.github/workflows/ci.yml
Meakk: Github runners on macOS and Windows uses JDK 8 without this step that configures JDK 17.  
I need JDK 11+ for the test to run Java files directly without a compilation step (it's compiled on the fly)

Meakk: On the Docker image, the package is already JDK 11, so the Android CI does not need to install another version.

Meakk: I could add a condition `if not linux` on the step to rely on JDK 11 on the regular CI, but I think it's fine like that

java/Options.java
Meakk: Yes, I just need to add all of them, but I will do it in another PR later. The interactor is not bound yet too. I just bound what I needed for Android

Meakk: I just wanted to create the foundation for the Java bindings in this PR.

mwestphal: fair enough, ideally keep an issue alive with what would be needed next.

.github/workflows/ci.yml
mwestphal: fine as it is.

java/Options.java
Meakk: https://github.com/f3d-app/f3d/issues/426

README_libf3d.md
```diff
@@ -206,6 +208,31 @@ eng.getLoader().addFile("f3d/testing/data/dragon.vtu").loadFile()
 eng.getInteractor().start()
 ```
 
+## Java Bindings
+
+If the Java bindings are generated, the libf3d can be used directly from Java.
+You can import the `f3d.jar` package and use the provided Java classes directly.
+Make sure to set `java.library.path` to path where the JNI library is built.
+Here is an example showing how to use libf3d Java bindings:
```
mwestphal, line 216: in the PR descr there is some info that could be added here

cmake/installing.cmake
```diff
@@ -19,6 +19,13 @@ if (BUILD_WINDOWS_SHELL_THUMBNAILS_EXTENSION)
     RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT shellext)
 endif()
 
+# Java bindings
+if (F3D_JAVA_BINDINGS)
+  install(TARGETS javaf3d
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT java)
+  install_jar(f3d-jar DESTINATION "share/java" COMPONENT java)
```
mwestphal, line 26: where is instal_jar coming from ?

java/F3DJavaBindings.cxx
```diff
@@ -0,0 +1,183 @@
+// Automatically generated headers
+#include <io_github_f3d_app_f3d_Camera.h>
+#include <io_github_f3d_app_f3d_Engine.h>
+#include <io_github_f3d_app_f3d_Loader.h>
+#include <io_github_f3d_app_f3d_Options.h>
+#include <io_github_f3d_app_f3d_Window.h>
```
mwestphal, line 6: why would io_github be in the name of the header ?

java/testing/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+  # On macOS, this argument is mandatory in order to create native window
+  list(APPEND java_test_args "-XstartOnFirstThread")
+endif()
+
+# On macOS, VTK 9.1 is crashing in modifyDPIForBackingScaleFactorOfWindow, see:
+# https://gitlab.kitware.com/vtk/vtk/-/merge_requests/6845 (introduces the crash)
+# https://gitlab.kitware.com/vtk/vtk/-/merge_requests/9372 (fixes the crash)
+# Disabling the tests if we are between these two commits
+set(java_test_disabled OFF)
+if(APPLE AND VTK_VERSION VERSION_GREATER 9.0.0 AND VTK_VERSION VERSION_LESS_EQUAL 9.1.20220617)
```
mwestphal, line 16: so vtk 9.0.1 has the issue ?

java/testing/TestJavaBindings.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class TestJavaBindings {
+
+  static {
+    if (System.getProperty("os.name").startsWith("Windows"))
+    {
+      // On Windows, preload the OpenGL library
+      // This ensures the OpenGL used is the one in the working directory if any.
+      // In practice, it is used in F3D CI to run the test using Mesa, it's not required for production.
+      System.loadLibrary("opengl32");
```
mwestphal, line 11: no LD_PRELOAD equivalent on windows ?

cmake/installing.cmake
Meakk: CMake `UseJava` module: https://cmake.org/cmake/help/latest/module/UseJava.html

java/F3DJavaBindings.cxx
Meakk: We have no choice, it's the JNI nomenclature

java/testing/CMakeLists.txt
Meakk: It's funny enough, the commit producing the crash is the first one after the `9.0.0` tag

java/testing/TestJavaBindings.java
Meakk: nop, the variable `PATH` is used. But when `f3d-java.dll` is loaded by JVM, it uses this order:

- The directory from which the application loaded.
- The system directory. Use the [GetSystemDirectory](https://learn.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getsystemdirectorya) function to get the path of this directory.
- The 16-bit system directory. There is no function that obtains the path of this directory, but it is searched.
- The Windows directory. Use the [GetWindowsDirectory](https://learn.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getwindowsdirectorya) function to get the path of this directory.
- The current directory.
- The directories that are listed in the PATH environment variable. Note that this does not include the per-application path specified by the App Paths registry key. The App Paths key is not used when computing the DLL search path.

Since `open32.dll` is found in the system directory, this one is used, except if the library is already loaded in the application (that's why I load it manually)

README_libf3d.md
Meakk: The other comments in my PR are related to our CI infrastructure, I don't think it's relevant to add it to the documentation.

java/F3DJavaBindings.cxx
Meakk: Basically, the command `javac -h` internally called by CMake when `GENERATE_NATIVE_HEADERS` is enabled in the `add_jar` command is generating these files based on the Java package and classes names.

cmake/installing.cmake
mwestphal: ok !

README_libf3d.md
mwestphal: ok

java/F3DJavaBindings.cxx
mwestphal: fair enough, just wondering where it was coming from

java/testing/TestJavaBindings.java
mwestphal: ok

java/testing/CMakeLists.txt
mwestphal: lol

mwestphal (approved):

PR #395, opened 2022-09-04
Fix orientation marker 

### LINKED ISSUE #394: Segfault when loading a file with axis widget after showing a volume without axis widget
**Describe the bug**
F3D (VTK) segfaults when loading a file that shows with axis widget thanks to a config file

**To Reproduce**
 1. Open the file using `f3d --config=src/resource/config.json src/testing/data/multi` -> volume show up with axis
 2. Press X to remove axis
 3. Press right to load next file -> segfault

mwestphal:
Fix #394

### DIFF
--- a/library/VTKExtensions/Rendering/CMakeLists.txt
+++ b/library/VTKExtensions/Rendering/CMakeLists.txt
@@ -3,6 +3,7 @@ set(classes
   vtkF3DInteractorStyle
   vtkF3DNoRenderWindow
   vtkF3DOpenGLGridMapper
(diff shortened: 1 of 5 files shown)


Meakk:
Can you explain why setting the flag after construction is not working?

mwestphal:
The setter triggers a render, which can segfault, see the issue this PR fixes.

I think this should be redesigned on VTK side but this is out of scope. I can open an issue though.

Meakk (approved):

PR #396, opened 2022-09-04
Rework config and options

mwestphal:
Rework of config and option management in F3DStarter in order to improve the behavior.

 - add a "global" category in the config file that as a different behavior as other category
 - add API to f3d::options to getNames of options and getType of option
 - rework the option logic

Previous behavior:

When loading a file, the command line and config file are parsed in order to figure out which options to use

New behavior:

F3D has a dynamic option that will changed over time when interacting with F3D.
When loading a file for the first time, F3D keep a hand on:
 - A dynamic option that will evolve overtime
 - Used options with the file config applied to it and copied to the engine
 - The option in use in the engine (hold by engine)

If no file was load previously, dynamic option contains the command line option + global section of config file.
file options is copied from dynamic option and config for the file being loaded is added on it, and command line option are applied on top of it.
file options is then copied to the engine to be used

If a file was loaded previously, before doing anything else, F3D will compare each option between the current engine options and the previous file options.
Each different option is stored in the dynamic option.
File options is copied from dynamic option and config for the file being loaded is added on it.
**Command line options is NOT added on top of it.**
File options is then copied to the engine to be used

Practical example with the default config (see how it changed a bit in this PR):

- `f3d foo.vtp bar.step --grid=0`
- foo.vtp appears, with axis but no grid because command line override it
- press G to show the grid
- press Up to reload the file, grid stays on because command line is not reread.
- press X to hide axis
- press right to change file
- bar.step appears without axis because it was disabled previously, and with SSAO because it is enabled in step config file
- press right to change file
- foo.vtp appears without axis because it was disabled previously, and without SSAO because it has not been changed
- press A to enable SSAO
- press right to change file (SSAO changed is added to dynamic options here)
- bar.step appears without axis because it was disabled previously, and with SSAO because it is enabled in step config file
- press right to change file
- foo.vtp appears without axis because it was disabled previously, and with SSAO because it has been changed

I think this behavior is the most natural for multi, while still keeping full retro-compatibility with previous config file when loading a single file.


mwestphal:
let me know what you think @Meakk ?

application/F3DStarter.cxx
```diff
@@ -246,6 +250,10 @@ int F3DStarter::Start(int argc, char** argv)
         }
         return EXIT_FAILURE;
       }
+      else
+      {
+        f3d::log::info("Image comparison success with an error differencce of: ", error);
```
Meakk, line 255:
```suggestion
        f3d::log::info("Image comparison success with an error difference of: ", error);
```

application/F3DStarter.cxx
```diff
@@ -273,29 +281,107 @@ int F3DStarter::Start(int argc, char** argv)
...
 void F3DStarter::LoadFile(f3d::loader::LoadFileEnum load)
 {
+  if (this->Internals->LoadedFile)
+  {
+    // When loading a file after another, store the changed options
+    // into the dynamic options and use these dynamic option as the default
+    // for loading the next file while still applying file specific options on top of it
+
+    // Recover previous options from the engine
+    f3d::options previousOptions = this->Internals->Engine->getOptions();
```
Meakk, line 291: const ref

application/F3DStarter.cxx
```diff
@@ -273,29 +281,107 @@ int F3DStarter::Start(int argc, char** argv)
...
+    std::vector<std::string> optionNames = this->Internals->DynamicOptions.getNames();
+    for (auto name : optionNames)
+    {
+      switch (this->Internals->DynamicOptions.getType(name))
+      {
+        case f3d::options::Type::BOOL:
+        {
+          bool val;
+          previousOptions.get(name, val);
+          if (this->Internals->FileOptions.getAsBool(name) != val)
```
Meakk, line 304: why not creating a `f3d::options::copyFrom(const f3d::options& opt, const std::string& name)` to copy the variant instead of this switch case block?

Meakk (changes requested):

application/F3DStarter.cxx
Meakk: If F3D needs it, libf3d users can need it, so yes I think it would be useful.

mwestphal: agreed.

README.md
```diff
@@ -338,23 +342,28 @@ formatted file to provide default values for these options.
...
+and not on subsequent loads.
+The regular expression blocks are always taken into account, even when loading further files.
+
 A typical config file may look like this:
 
 ```javascript
 {
-   ".*": {
-       "resolution": "1200,800",
+   "global": {
```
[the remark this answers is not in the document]
mwestphal: I've been thinking about it, I tend to this now:
```
{
  "static":
  {
    ".*ext":
    {
    "axis": true,
    }
  },  
  "dynamic":
  {
    ".*ext":
    {
    "axis": false,
    }
  }
}  
```

wdyt ?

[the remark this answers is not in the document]
Meakk: Ok, looks good

application/F3DStarter.cxx
mwestphal: actually, copyFrom is not what we want as we only want to copy if a condifition is true.

It boils down to:

if the option value in the engine is different from the option passed to the engine, copy this value to the dynamic options.

There is three options involved here, and this is what makes it work.

of course I could add a `f3d::options::copyToIfDifferentWIth(f3d::options& to, const f3d::options& with, const std::string& name)` but that would actually be overspecific.

maybe I can just expose the std::variant in the api ?

README.md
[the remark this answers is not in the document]
mwestphal: After thinking about it, this is not right, static option are read even before checking if we actually have a file provided.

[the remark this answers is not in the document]
mwestphal: 1. this PR current implementation:

```
{
  "global":
  {
    "axis": true,
  },  
  ".*ext":
  {
  "axis": false,
  }
}  
```

2. You suggest an implementation like:

```
{
  "static":
  {
    "axis": true,
  },
"dynamic":
  {
    ".*ext":
    {
    "axis": false,
    }
  }
}  
```

3. I Suggested this non  viable approahc:

```
{
  "static":
  {
    ".*ext":
    {
    "axis": true,
    }
  },  
  "dynamic":
  {
    ".*ext":
    {
    "axis": false,
    }
  }
}  
```



3. is not viable, so 1. and 2. remains. I agree that 2. is nicer however it means that:
- Parsing code is more complex
- We break retrocompat **or**
- We support retrocompat but parsing code becomes very complex


The main benefit of 2. is that it ensure there is no confusion between regexp and "group", which will very likely never happen, so I'm not sure this is worth it.

Wdyt ?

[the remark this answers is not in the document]
mwestphal: keeping current implementation, but I can change `global` for anything else.

python/testing/TestPythonOptions.py
```diff
@@ -22,3 +22,12 @@
 assert engine.getOptions().getAsInt("render.raytracing.samples") == 2
 assert engine.getOptions().getAsDoubleVector("model.color.rgb") == [ 0., 1., 1.]
 assert engine.getOptions().getAsString("scene.up-direction") == "-Z"
+
+assert len(options.getNames()) > 0
+
+assert options.getType("interactor.axis") == f3d.options.BOOL
```
Meakk, line 28: forgot to remove that

Meakk (changes requested):

library/src/options.cxx
```diff
@@ -396,6 +396,48 @@ options& options::toggle(const std::string& name)
   return *this;
 }
 
+//----------------------------------------------------------------------------
+bool options::isSame(const options& other, const std::string& name) const
+{
+  try
+  {
+    return this->Internals->Options.at(name) == other.Internals->Options.at(name);
```
Meakk, line 404: is it working with vectors?

Meakk (approved):
please check if it works with vectors, otherwise looks good to me

library/src/options.cxx
mwestphal: it doesnt

PR #399, opened 2022-09-06
Support config filename/filestem only

mwestphal:
--config now support filename/fileStem , which is great and lets us specify a config file for thumbnails.

- Add support for filename/fileStem in --config
- Update/Improve doc
- Fix an issue with options not read from config file, introduced in #396 
- Add a thumbnail.json and use it in thumbnailers
- Add tests


Meakk (approved):
Cool! Looks good.

mwestphal:
last push is only commit reorg.

PR #401, opened 2022-09-06
Max size option

mwestphal:
Adds a --max-size option, that will just not load a file if it is bigger than provided max size, -1 being infinite.

Very useful tor thumbnails.


Meakk (approved):

mwestphal:
Doc is missing

mwestphal:
Ill update and fixup doc in another PR.

Issue #403, opened 2022-09-06
New formats are missing from MacOS bundle


mwestphal:
All formats should be listed in resources/BundleInfo.plist.in, it looks like recent formats have not been added and are missing, including:

- step/iges
- abc

mwestphal:
@Meakk

mwestphal:
fixed by #405

PR #405, opened 2022-09-06
Fix macOS bundle config

### LINKED ISSUE #404: MacOS Bundle issues
After installing F3D 1.3.0 bundle on MacOS,there are the following issues:

- config file is not found
- open with / double click opens F3D but without the file

Meakk:
Fix #404 and #403

### DIFF
--- a/application/F3DConfigFileTools.cxx
+++ b/application/F3DConfigFileTools.cxx
@@ -1,5 +1,7 @@
 #include "F3DConfigFileTools.h"
 
+#include "F3DConfig.h"
+
 #include "log.h"
 
 #include <cstring>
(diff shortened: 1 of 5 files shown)


Meakk:
@mwestphal should be fine now, I'll test the package tomorrow morning again with a fresh brain

application/F3DConfigFileTools.cxx
```diff
@@ -1,5 +1,7 @@
 #include "F3DConfigFileTools.h"
 
+#include "F3DConfig.h"
```
mwestphal, line 3: While we cant fail to build, I think we can improve our CI to actually test the binary config flag.

it would be as simple as adding a reference file to the "Check Install" step of our CI. This image should be using the installed config file.

Ok to do in another PT though.

mwestphal (approved):

application/F3DConfigFileTools.cxx
Meakk: I have a better idea :)

PR #409, opened 2022-09-07
Better default position for macOS native window

Meakk:


mwestphal:
+1 but please open an issue to test this on all OSes

mwestphal (approved):

PR #419, opened 2022-09-09
Fix engine constructor python binding

### LINKED ISSUE #416: python binding init engine need  arguments
**Describe the bug**
I followed the [doc](https://github.com/f3d-app/f3d/blob/master/README_libf3d.md) to build python binding from source .But when I use the follow code:
`
engine = f3d.engine()
`
and console print
` 
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. f3d.engine(arg0: f3d.window.Type)
`

**Additional context**
Maybe the part about python binding of document need to be updated

Meakk:
Fix #416

### DIFF
--- a/README_libf3d.md
+++ b/README_libf3d.md
@@ -195,7 +195,7 @@ Here is an example showing how to use libf3d python bindings:
 ```python
 import f3d
 
-eng = f3d.engine()
+eng = f3d.engine(f3d.window.NATIVE)
 eng.getOptions()
   .set("model.scivis.array-name", "Normals")
   .set("model.scivis.component", 0)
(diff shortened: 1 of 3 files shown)


README_libf3d.md
```diff
@@ -195,7 +195,7 @@ Here is an example showing how to use libf3d python bindings:
 ```python
 import f3d
 
-eng = f3d.engine()
+eng = f3d.engine(f3d.window.NATIVE)
```
mwestphal, line 198: is this really needed since you added the default val ?

Meakk: No, but the users will read master documentation for 1.3.1, so let's keep the argument in the documentation until the fix is released

mwestphal: good point

mwestphal (approved):

PR #429, opened 2022-09-19
Fix DLL export warning on Windows

### LINKED ISSUE #428: Warnings on Windows
```
D:\a\f3d\f3d\source\library\public\exception.h(20,1): warning C4275: non dll-interface class 'std::runtime_error' used as base for dll-interface struct 'f3d::exception' [D:\a\f3d\f3d\build\library\testing\libf3dSDKTests.vcxproj]
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.33.31629\include\stdexcept(101): message : see declaration of 'std::runtime_error' [D:\a\f3d\f3d\build\library\testing\libf3dSDKTests.vcxproj]
```

It may be tricky to fix, not sure if we can use a pimpl idiom here.

Meakk:
`f3d::exception` derives from `std::runtime_error` resulting in a warning in Windows because the exported symbol rely on the STL.  
We don't need to export the class since it's header only.

Fix #428

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,7 +112,6 @@ mark_as_advanced(F3D_STRICT_BUILD)
 set(f3d_strict_build_compile_options "")
 if(F3D_STRICT_BUILD)
   if(MSVC)
-    # Warning C4275 is essentially noise, disabling it to silent an issue with jsoncpp library
     set(f3d_strict_build_compile_options /W4 /WX)
   else()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -113,7 +113,7 @@ set(f3d_strict_build_compile_options "")
 if(F3D_STRICT_BUILD)
   if(MSVC)
     # Warning C4275 is essentially noise, disabling it to silent an issue with jsoncpp library
-    set(f3d_strict_build_compile_options /W4 /WX /wd4275)
+    set(f3d_strict_build_compile_options /W4 /WX)
```
mwestphal, line 116: how can reenabling a warning fix a warning ?

Meakk: Wait ^^  
I'm just running the CI right now.

mwestphal (approved):

PR #438, opened 2022-09-26
Introduce 2D interactor

### LINKED ISSUE #49: Consider adding a 2D interactor

jpouderoux:
Fixes #49 

- [x] Add 2D interactor
- [x] Allow to switch between 2D and 3D interactors with `2` and `3` keys
- [x] Set the 2D interactor automatically for 2D data
- [ ] Tests

Replaces the deprecated MR #132


library/src/interactor_impl.cxx
```diff
@@ -476,4 +493,25 @@ void interactor_impl::InitializeAnimation(vtkImporter* importer)
...
+//----------------------------------------------------------------------------
+void interactor_impl::selectBestInteractor()
+{
+  vtkRenderWindow* renWin = this->Internals->Window.GetRenderWindow();
+  vtkF3DRenderer* renderer = vtkF3DRenderer::SafeDownCast(renWin->GetRenderers()->GetFirstRenderer());
+  double bounds[6], lengths[3];
+  renderer->ComputeVisiblePropBounds(bounds);
+  vtkBoundingBox box(bounds);
+  box.GetLengths(lengths);
+  bool is2D = (lengths[0] == 0 || lengths[1] == 0 || lengths[2] == 0);
```
mwestphal, line 506: Wdyt @Meakk ? I dont really like the idea of this in the libf3d directly, but it can make sense from a user POV.

library/src/interactor_impl.cxx
```diff
@@ -193,7 +198,16 @@ class interactor_impl::internals
...
       default:
-        if (keySym == "Left")
+
+        if (keyCode == '2' || keySym == "KP_Down")
+        {
+          self->VTKInteractor->SetInteractorStyle(self->Style2D);
+        }
+        else if (keyCode == '3' || keySym == "KP_Next")
+        {
+          self->VTKInteractor->SetInteractorStyle(self->Style3D);
```
mwestphal, line 208: this should be controlled by an option

library/src/interactor_impl.cxx
```diff
@@ -193,7 +198,16 @@ class interactor_impl::internals
         self->Window.PrintSceneDescription(log::VerboseLevel::INFO);
         break;
       default:
-        if (keySym == "Left")
+
+        if (keyCode == '2' || keySym == "KP_Down")
+        {
+          self->VTKInteractor->SetInteractorStyle(self->Style2D);
```
mwestphal, line 204: does that do anything with trackball on/off ?

jpouderoux: Not sure what you mean. When enabling the 2d style, the trackball (ie Style3D) is disabled.

mwestphal: ok, just checking.

library/src/interactor_impl.cxx
```diff
@@ -193,7 +198,16 @@ class interactor_impl::internals
         self->Window.PrintSceneDescription(log::VerboseLevel::INFO);
         break;
       default:
-        if (keySym == "Left")
+
+        if (keyCode == '2' || keySym == "KP_Down")
```
mwestphal, line 202: what is kp down ?

jpouderoux: that's the symbol my linux distribution gives me when pressing the key "2" on the numpad...

library/src/interactor_impl.cxx
jpouderoux: Yeah, I will do that but there is a misconception somewhere IMHO because:
 - InteractorStyles are own by the interactor:
 - Options are managed by the window. But window doesn't have access to either interactor or engine (which owns both window and interactor).
Am I missing a link?

Not directly related:
  - options class should be free of "f3d" internals. We should have a derived f3doptions class (or so) that performs the declaration/initialization of the f3d specific options.

library/src/interactor_impl.cxx
jpouderoux: It is a simple way to deduce the best interactor style according the data dimension. As you can see, it is put in a dedicated method and called by F3DStarter, so lib user has choice to use it or not.

mwestphal: yes, but instead of choosing, having a way to know if the data is 2d and react accordingly may be better.

library/src/interactor_impl.cxx
mwestphal: > Options are managed by the window. But window doesn't have access to either interactor or engine (which owns both window and interactor). Am I missing a link?

See how it is done for Trackball. That being said, adding a logic for interactor specific options could make sense in the long run. Right now there are not a lot of such options, and an interactor may not always be available, so it is a bit easier this way.


> options class should be free of "f3d" internals. We should have a derived f3doptions class (or so) that performs the declaration/initialization of the f3d specific options.

You mean options.h ? There is only a pointer to an internal class, but it could be done indeed. not sure this is worth it. Please open a dedicate issue to discuss this.

>    to be more aligned with the lib approach, the mapping between key code/sym and options should be performed at a higher level, like in the f3dstart class.

Definitely, this is something which was planned but I did not get around to do it. It would be nice to track it in a dedicated issue.

Thanks for your suggestions !

library/src/interactor_impl.cxx
mwestphal: hum, not sure we want to support this.

jpouderoux: Not sure how to detect when key 2 is pressed then...

library/src/interactor_impl.cxx
jpouderoux: > See how it is done for Trackball.
I saw that but it is different. For trackball, option is set to to the vtkF3DRenderer, and then the InteractorStyle3D check its value because it has access to the renderer.
Here, the interactor styles are managed by the interactor_impl internal class. 
I think window_impl::UpdateDynamicOptions should be able to access to the interactor(_impl) instance own by the engine (which also is not visible from window_impl). If you agree with this statement, I can make changes to allow this access.

mwestphal: Yes, that may be needed.

library/src/interactor_impl.cxx
jpouderoux: Why not? If shortcut is "2" you can expect to use the 2 key of the numpad. But anyway, if we consider the non numpad key "2", VTK does not provide a keycode for it on my keyboard, and the sym is "ecute" which is the french e with accent on which the 2 stands but then it is not cross country... weired!

library/src/interactor_impl.cxx
Meakk: Indeed, I would prefer having a `loader.getSceneBoundingBox()` and a `interactor.setMode()` and put the logic in the application instead.

mwestphal: I think an option to control it is better for that case, instead of a method.

library/src/interactor_impl.cxx
mwestphal: Yeah, we also need to avoid non-international keys. tbh this design is not great, we may need to add a layer of abstraction.

mwestphal:
@jpouderoux the F3D community may pick this one up, are you ok with that ?

mwestphal:
Closing this as @jpouderoux agreed above.

Issue #410, opened 2022-09-07
Native window position is not set


Meakk:
**Describe the bug**

The default position of native windows is defined in VTK, and seems to be:
- 50,50 for Cocoa
- 0,0 for Windows and Linux (to confirm)

Moreover, Y=0 corresponds to the bottom of the screen for Cocoa, so the Y position must be reworked in VTK to flip it using the height of the screen.

Also, maybe a `setPosition` public API could be useful.

Meakk:
Related: #409

mwestphal:
Linux:
 - Awesome (floating) : Window is **not** positioned correctly, half of it is below the screen
 - Xfce: Window is positioned perfectly, top left, without hiding the "Xfce bar"

mwestphal:
Fixed by #444

PR #444, opened 2022-09-28
Adding window position

mwestphal:
- Adding a window::setPosition
- Adding a --position option
- Adding tests
- Updating doc

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ Options|Default|Description
 ------|------|------
 \-\-bg-color=&lt;R,G,B&gt;|0.2, 0.2, 0.2|Set the window *background color*.<br>Ignored if *hdri* is set.
 \-\-resolution=&lt;width,height&gt;|1000, 600|Set the *window resolution*.
-\-\-position=&lt;x,y&gt;||Set the *window position*, in pixels, starting from bottom left.
+\-\-position=&lt;x,y&gt;||Set the *window position* (top left corner) , in pixels, starting from the top left of your screens.
 -z, \-\-fps||Display a *frame per second counter*.
 -n, \-\-filename||Display the *name of the file*.
 -m, \-\-metadata||Display the *metadata*.<br>This only makes sense when using the default scene.
@@ -394,7 +394,7 @@ The following command-line options <br> cannot <br> be set via config file:
 `help`, `version`, `readers-list`, `config`, `dry-run`.
 
 The following command-line options <br>can only</br> be set in the global block of the config file:
-`no-render`, `inputs`, `output`, `quiet`, `verbose`, `resolution`, position and all testing options.
+`no-render`, `inputs`, `output`, `quiet`, `verbose`, `resolution`, `position` and all testing options.
 
 Boolean options that have been turned on in the configuration file can be turned
 off on the command line if needed, eg: `--point-sprites=false`.
--- a/README.md
+++ b/README.md
@@ -200,6 +200,7 @@ Options|Default|Description
 ------|------|------
 \-\-bg-color=&lt;R,G,B&gt;|0.2, 0.2, 0.2|Set the window *background color*.<br>Ignored if *hdri* is set.
 \-\-resolution=&lt;width,height&gt;|1000, 600|Set the *window resolution*.
+\-\-position=&lt;x,y&gt;||Set the *window position*, in pixels, starting from bottom left.
 -z, \-\-fps||Display a *frame per second counter*.
 -n, \-\-filename||Display the *name of the file*.
 -m, \-\-metadata||Display the *metadata*.<br>This only makes sense when using the default scene.
@@ -393,7 +394,7 @@ The following command-line options <br> cannot <br> be set via config file:
 `help`, `version`, `readers-list`, `config`, `dry-run`.
 
 The following command-line options <br>can only</br> be set in the global block of the config file:
-`no-render`, `inputs`, `output`, `quiet`, `verbose`, `resolution` and all testing options.
+`no-render`, `inputs`, `output`, `quiet`, `verbose`, `resolution`, position and all testing options.
 
 Boolean options that have been turned on in the configuration file can be turned
 off on the command line if needed, eg: `--point-sprites=false`.
--- a/cmake/testing.cmake
+++ b/cmake/testing.cmake
@@ -336,6 +336,9 @@ f3d_test(NAME TestOutputNoBackground DATA cow.vtp ARGS --no-background NO_BASELI
 f3d_test(NAME TestInteractionRecord DATA cow.vtp ARGS --interaction-test-record=${CMAKE_BINARY_DIR}/Testing/Temporary/interaction.log NO_BASELINE)
 f3d_test(NAME TestInteractionPlay DATA cow.vtp ARGS --interaction-test-play=${CMAKE_BINARY_DIR}/Testing/Temporary/interaction.log DEPENDS TestInteractionRecord NO_BASELINE)
 
+# Window position test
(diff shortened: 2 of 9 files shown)


mwestphal:
Fix #410

mwestphal:
@Meakk I did not add a default config or a default val for it. The MacOS case is managed specifically.


About MacOS, I wonder if aVTK fix or at least issue is needed.

README.md
```diff
@@ -393,7 +394,7 @@ The following command-line options <br> cannot <br> be set via config file:
 `help`, `version`, `readers-list`, `config`, `dry-run`.
 
 The following command-line options <br>can only</br> be set in the global block of the config file:
-`no-render`, `inputs`, `output`, `quiet`, `verbose`, `resolution` and all testing options.
+`no-render`, `inputs`, `output`, `quiet`, `verbose`, `resolution`, position and all testing options.
```
Meakk, line 397:
```suggestion
`no-render`, `inputs`, `output`, `quiet`, `verbose`, `resolution`, `position` and all testing options.
```

cmake/testing.cmake
```diff
@@ -336,6 +336,9 @@ f3d_test(NAME TestOutputNoBackground DATA cow.vtp ARGS --no-background NO_BASELI
 f3d_test(NAME TestInteractionRecord DATA cow.vtp ARGS --interaction-test-record=${CMAKE_BINARY_DIR}/Testing/Temporary/interaction.log NO_BASELINE)
 f3d_test(NAME TestInteractionPlay DATA cow.vtp ARGS --interaction-test-play=${CMAKE_BINARY_DIR}/Testing/Temporary/interaction.log DEPENDS TestInteractionRecord NO_BASELINE)
 
+# Window position test
+f3d_test(NAME TestPosition DATA dragon.vtu ARGS --position=100,100 NO_BASELINE)
```
Meakk, line 340: Is it only for the coverage? It does not test the feature itself?

README.md
```diff
@@ -200,6 +200,7 @@ Options|Default|Description
 ------|------|------
 \-\-bg-color=&lt;R,G,B&gt;|0.2, 0.2, 0.2|Set the window *background color*.<br>Ignored if *hdri* is set.
 \-\-resolution=&lt;width,height&gt;|1000, 600|Set the *window resolution*.
+\-\-position=&lt;x,y&gt;||Set the *window position*, in pixels, starting from bottom left.
```
Meakk, line 203: bottom left of what? The monitor? Also, the position corresponds to which corner of the window?

Meakk (changes requested):

library/public/window.h
```diff
@@ -76,6 +76,11 @@ class F3D_EXPORT window
    */
   virtual int getHeight() const = 0;
 
+  /**
+   * Set the position of the window
+   */
+  virtual window& setPosition(int x, int y) = 0;
```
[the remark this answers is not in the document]
mwestphal: @Meakk

README.md
```diff
@@ -200,6 +200,7 @@ Options|Default|Description
 ------|------|------
 \-\-bg-color=&lt;R,G,B&gt;|0.2, 0.2, 0.2|Set the window *background color*.<br>Ignored if *hdri* is set.
 \-\-resolution=&lt;width,height&gt;|1000, 600|Set the *window resolution*.
+\-\-position=&lt;x,y&gt;||Set the *window position* (top left corner) , in pixels, starting from the top left of your screens.
```
mwestphal, line 203: Could you validate that on MacOS @Meakk ?

library/public/window.h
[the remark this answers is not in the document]
Meakk: Since we have `getWidth` and `getHeight`, I think it makes sense to have `getPositionX` and `getPositionY`, but I don't know if VTK implementations works.

README.md
Meakk: It's incorrect: bottom left corner of the window, bottom left of the screen.

library/public/window.h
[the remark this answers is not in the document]
mwestphal: I'm unable to make it work, either before render, after a render or even during interaction using a timer. I'd say we drop it for now.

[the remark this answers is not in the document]
mwestphal: Not tested in VTK at first glance: `grep -R  GetPosition | grep Window | grep Testing`

README.md
mwestphal: Should be fixed in VTK imo.

Meakk: Is there any issue already opened in VTK?

mwestphal: None that I can find, ill create it

mwestphal: Some resource:

Xorg: https://tronche.com/gui/x/xlib/introduction/overview.html : upper left
Windows: http://www.jasinskionline.com/windowsapi/ref/s/setwindowpos.html: Upper left
Cocoa: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CocoaDrawingGuide/Transforms/Transforms.html: Lower left

mwestphal: SDL: https://wiki.libsdl.org/SDL_SetWindowPosition upper left

mwestphal: https://gitlab.kitware.com/vtk/vtk/-/issues/18681

mwestphal: So this may take a while. What about we handle that application side for now ?

library/src/window_impl.cxx
```diff
@@ -122,6 +115,20 @@ window& window_impl::setSize(int width, int height)
...
 }
 
+//----------------------------------------------------------------------------
+window& window_impl::setPosition(int x, int y)
+{
+#if __APPLE__
+  // vtkCocoaRenderWindow has a different behavior than other render windows
+  // https://gitlab.kitware.com/vtk/vtk/-/issues/18681
+  int* screenSize = this->Internals->RenWin->GetScreenSize();
+  this->Internals->RenWin->SetPosition(x, screenSize[1] - y);
```
Meakk, line 125: You are still positioning the bottom left corner of the window, relative to the top left of the screen.  
Should be `screenSize[1] - winSize[1] - y`. However in this case, the height of the title bar is missing...

library/src/window_impl.cxx
```diff
@@ -122,6 +115,20 @@ window& window_impl::setSize(int width, int height)
   return *this;
 }
 
+//----------------------------------------------------------------------------
+window& window_impl::setPosition(int x, int y)
+{
+#if __APPLE__
```
Meakk, line 121: It's not related to Apple, but Cocoa. What about testing `RenWin->IsA("vtkCocoaRenderWindow")`?

Meakk (changes requested):

library/src/window_impl.cxx
mwestphal: Ok, better then nothing.

However, in that case I'm not sure to follow your comment about the window being outside of the screen in 50,50 though.

Meakk: Right, I'm confused too. My understanding right now is the (50,50) is set before the first render, so the size is still (0,0), and top left / bottom left is the same position. Then, when expanding, the window is expanded in the bottom direction.  
It's just a guess, it needs to be clarified.

library/src/window_impl.cxx
mwestphal: bad for coverage, but ok :p

library/src/window_impl.cxx
mwestphal: please test the new impl

Meakk (approved):

PR #448, opened 2022-10-03
Fix warning with new versions of grep

Meakk:
`grep "\-\-"` produces a warning with recent versions of grep: 

> grep: warning: stray \ before -

Using `grep "--"` does not work because grep thinks `--` is an option.  
Using `grep "[-]-"` as a workaround.


mwestphal (approved):
Approved, assuming you tested on fish and szh

Meakk:
I did. The fish completion was working but the warning was pretty annoying. The fix is the same for zsh.

PR #451, opened 2022-10-03
Remove CMAKE_SYSTEM_PROCESSOR to fix reproducibility.

mzf-guest:
Hello,

the `CMAKE_SYSTEM_PROCESSOR` variable triggered a reproducibility error when cross compiling. See for example [this diffoscope report](https://tests.reproducible-builds.org/debian/rb-pkg/bookworm/i386/diffoscope-results/f3d.html) which contains the string `Linux·x86_64` when cross-compiling from amd64 to i386, instead of `Linux i686`. The string origin is in `config.cxx.in` and easily visible with `f3d --help` command.

The current fix works for the Debian package, but I'm not sure that this is the correct one as it may be a cross-compiling tool-chain misconfiguration, hence not related to F3D itself.

Feel free to edit this PR.

Best,
François

### DIFF
--- a/library/src/config.cxx.in
+++ b/library/src/config.cxx.in
@@ -5,7 +5,7 @@
 namespace f3d::detail
 {
 const std::string LibVersion = "@PROJECT_VERSION@"; // Version is synchronized with F3D
-const std::string LibBuildSystem = "@CMAKE_SYSTEM_NAME@ @CMAKE_SYSTEM_PROCESSOR@";
+const std::string LibBuildSystem = "@CMAKE_SYSTEM_NAME@";
 const std::string LibBuildDate = "@F3D_BUILD_DATE@";
 const std::string LibCompiler = "@CMAKE_CXX_COMPILER_ID@ @CMAKE_CXX_COMPILER_VERSION@";
 }


mwestphal:
In the Debian world, how do you inform the users that a software has been built for an architecture or another?

Strictly speaking, to be really cross platform with this logic we should not even add Linux/Windows/MacOS.

I don't see the issue this is fixing, please advise.

Meakk:
`CMAKE_SYSTEM_PROCESSOR` should contain the architecture of the target when cross compiling, not the host architecture.  
See https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html  
Not sure what's wrong in your case.

mzf-guest:
I've dig a bit further and my explanation about cross-compilation was (partly) wrong. Sorry for the mess.

The reproducibility problem is real because `CMAKE_SYSTEM_PROCESSOR` captures the output of `uname -m` (via [`CMAKE_SYSTEM_PROCESSOR`](https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html#unix-platforms) on Unix platform) which could be changed easily with the `setarch` command without affecting the build tool-chain. Hence, to achieve reproducibility, a software should not capture this value and [reprotest changes it on purpose](https://salsa.debian.org/reproducible-builds/reprotest/-/blob/master/reprotest/build.py#L290) to emphasize the problem as the package builds identically except this string.
We can see this information like the machine hostname, it may change at any time without affecting the build toolchain.

> Strictly speaking, to be really cross platform with this logic we should not even add Linux/Windows/MacOS.

The "Linux" info is the kernel name from `uname -s` which cannot be changed without affecting the build toolchain as far as I know. If you find a way to change it, we should update reprotest :smiley: 

I'm not CMake expert, so I can't advice on how to get the platform information in a reproducible way with CMake. An alternative is to get this information at runtime instead of compile time.

I'm fine with this Debian patch for the moment, and the problem is reported upstream so I did my duty here. Feel free to update or close this PR!

Thanks for your work with F3D.

mwestphal:
I'm trying to wrap my head around what debian expect (and actually means) by reproducible builds.

If I understand well, the problem is that if you are under the same architecture but with different system configuration (eg, using `setarch`), then the resulting build is may change, which is not acceptable.

So, in order to be truly "reproducible", the architecture should be recovered in a way that is **not** configurable by any means ?

I feel like this is a very common problem, how do other software adress this ?

F3D also note the build date and time, wouldn't that break reproducible builds ?

Reading this : https://docs.kernel.org/kbuild/reproducible-builds.html, it looks like the timestamp issue is covered using specific env var, could you not do the same thing for the architecture ?

I'm still a bit confused about what is expected from a software, should a software : 

1. Provide reproducible build out of the box, by default (which would justify you removing the arch from cmake code)
2. Provide a way for packager to generate reproducible build easilly (we could add a CMake var for this if that helps)
3. Let packager uses standard tools to manipulate the environnement in order to achieve reproducible build (more or less what @Meakk is suggesting)

I'm trying to find good resource on this, butwhat I find is either too technical or too broad, do not hesitate to link any introduction on this.

mzf-guest:
> Have you tried to put -DCMAKE_SYSTEM_PROCESSOR=x86_64 manually in the configuration step?

This would likely fix the reproducible issue, but the output of `f3d --version` would be wrong. The problem is not the value of this variable, but that it's currently recorded in the binary.

> So, in order to be truly "reproducible", the architecture should be recovered in a way that is not configurable by any means ?

In a way that changing the value really changes the build-toolchain and hence the output. `CMAKE_SYSTEM_PROCESSOR` is just an info variable and should not be captured (i.e. written in the binary). An other classic example is the source path: it's mandatory to use it to find files, but it should not be captured because it may changes without changing the build output.

> F3D also note the build date and time, wouldn't that break reproducible builds ?

Date and time are managed through the [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/) environment variable. CMake automatically uses this variable in the [`string`](https://cmake.org/cmake/help/latest/command/string.html#generation) command, so you're safe for now :smile:.

You should look at the [reproducible build website](https://reproducible-builds.org) which try to answer most of your questions, especially [this page](https://reproducible-builds.org/docs/deterministic-build-systems/) which gives the following guidelines:
 - Ensure stable inputs.
 - Ensure stable outputs.
 - Capture as little as possible from the environment.

On the same website you'll find resources, talks and advises on the subject. For a code approach you may read  [i-probably-didnt-backdoor-this](https://github.com/kpcyrd/i-probably-didnt-backdoor-this) and [the unreproducible package](https://github.com/bmwiedemann/theunreproduciblepackage).

This is not a Debian specific problem. [Arch Linux](https://wiki.archlinux.org/title/Reproducible_builds), [Tails](https://tails.boum.org/contribute/build/reproducible/), [NixOS](https://r13y.com/), [Yocto](https://wiki.yoctoproject.org/wiki/Reproducible_Builds), [F-Droid](https://f-droid.org/docs/Reproducible_Builds/), [SuSE](https://en.opensuse.org/openSUSE:Reproducible_Builds) and probably others are also working on making their packages reproducible. 

By the way, thanks a lot for taking this issue seriously. It's greatly appreciated to have upstream authors aware of reproducible build problems!

Meakk:
> By the way, thanks a lot for taking this issue seriously. It's greatly appreciated to have upstream authors aware of reproducible build problems!

I don't like the idea of packagers having to patch f3d, it means we failed somehow :)  
What about using `CMAKE_SIZEOF_VOID_P` and write `32-bits` or `64-bits` instead of relying on `CMAKE_SYSTEM_PROCESSOR`?

mzf-guest:
> What about using CMAKE_SIZEOF_VOID_P and write 32-bits or 64-bits instead of relying on CMAKE_SYSTEM_PROCESSOR value?

Indeed, this variable seems to do actual check by compiling a C file, so it must not  affect reproducibility.

Meakk:
Superseded by #454

Issue #352, opened 2022-08-10
Scenes are too dark


rafcon-dev:
hi guys. For some reason I'm getting really dark scenes. For example, for this file:
https://downloadfree3d.com/3d-models/vehicles/ship/south-korean-destroyer-chung-mu/
This is what I get in the viewer:
![image](https://user-images.githubusercontent.com/85711398/183935071-aa1cf268-2a18-49a2-9c31-a48d4cc2029f.png)

and this is what I get in an online glTF viewer for example:
![image](https://user-images.githubusercontent.com/85711398/183935474-75dadbd9-d3fc-4af1-bdf2-ae5dec2780f2.png)

Similar thing happens with stl files for example, the image is just too dark and flat. Am I missing something?
I'm just running a simple commad with no options. I've tried --dry-run and a few more, nothing. Tried --geometry-only, and it renders without textures, but still too dark.

 - OS: Ubuntu 22.04.1 LTS
 - GPU and GPU driver: GTX 770 450.203.03
 - f3d --version`: 1.2.1

Meakk:
It's a gltf file? You'll get best results using a HDRI file. You can also try to change the background color.

rafcon-dev:
It's a .glb file.

Chaning background color doesn't seem to affect the lighting.
Adding a light grey hdr image indeed improved the lighting.
![image](https://user-images.githubusercontent.com/85711398/184186000-dbd1bc0d-c14d-4803-82de-543c4404879f.png)
Still, it's a bit too flat, a bit of directional lighting wouldn't hurt.
The thing is, for .obj files it becomes too washed out, so I have to use a different command for those than from glb/gltf. I'm  trying to generate previews for lf file manager.
I suppose this also happens with the thumbnails generated with f3d?
Still, for stls, the thumbnails are a bit flat, when compared with stl-thumbnailer, for example:
f3d result:
![image](https://user-images.githubusercontent.com/85711398/184188088-6b8f4486-2d5e-47f7-8f43-69dc19b00666.png)

stl-thumbnailer:
![image](https://user-images.githubusercontent.com/85711398/184187726-3da6ac56-daf4-4d9e-966b-854e0d1aed22.png)

So, 3 things:
* Probably glft and glb files could have a default light grey hdr applied for the thumbnailer
* Lighting in the default scene could be improved, a bit brighter and more directional
* default stl files orientation does not appear to be correct?

Meakk:
[...]

PR #461, opened 2022-10-27
add --light-intensity option

snoyer:
This PR adds a `--light-intensity` option to address problem described in #352.

It is implemented by simply adding a [`vtkLightKit`](https://vtk.org/doc/release/4.0/html/classvtkLightKit.html) to the renderer.

Here are some samples using the same model as in previous issue:

```bash
f3d 'ROKN Chung Mu.glb' --camera-position 37.3,20.4,61.6 --camera-focal-point -5.5,1.4,6.7 --camera-view-up -0.2,1.0,-0.2 --camera-view-angle 30 --ssao --fxaa --tone-mapping
```
light-intensity-no

```bash
f3d 'ROKN Chung Mu.glb' --light-intensity=1 --camera-position 37.3,20.4,61.6 --camera-focal-point -5.5,1.4,6.7 --camera-view-up -0.2,1.0,-0.2 --camera-view-angle 30 --ssao --fxaa --tone-mapping --light-intensity=1
```
light-intensity-1

```bash
f3d 'ROKN Chung Mu.glb' --light-intensity=2 --camera-position 37.3,20.4,61.6 --camera-focal-point -5.5,1.4,6.7 --camera-view-up -0.2,1.0,-0.2 --camera-view-angle 30 --ssao --fxaa --tone-mapping --light-intensity=2
```
light-intensity-2

```bash
f3d 'ROKN Chung Mu.glb' --light-intensity=4 --camera-position 37.3,20.4,61.6 --camera-focal-point -5.5,1.4,6.7 --camera-view-up -0.2,1.0,-0.2 --camera-view-angle 30 --ssao --fxaa --tone-mapping --light-intensity=4
```
light-intensity-4

### DIFF
--- a/application/F3DOptionsParser.cxx
+++ b/application/F3DOptionsParser.cxx
@@ -246,6 +246,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp3, "filename", "n", "Display filename", options.getAsBoolRef("ui.filename"), HasDefault::YES, MayHaveConfig::YES);
     this->DeclareOption(grp3, "metadata", "m", "Display file metadata", options.getAsBoolRef("ui.metadata"), HasDefault::YES, MayHaveConfig::YES);
     this->DeclareOption(grp3, "blur-background", "u", "Blur background", options.getAsBoolRef("render.background.blur"), HasDefault::YES, MayHaveConfig::YES);
+    this->DeclareOption(grp3, "light-intensity", "", "Light intensity", options.getAsDoubleRef("render.light.intensity"), HasDefault::YES, MayHaveConfig::YES, "<intensity>");
 
     auto grp4 = cxxOptions.add_options("Scientific visualization");
     this->DeclareOption(grp4, "scalars", "s", "Color by scalars", options.getAsStringRef("model.scivis.array-name"), HasDefault::YES, MayHaveConfig::YES, "<array_name>", HasImplicitValue::YES, "");
--- a/application/F3DOptionsParser.h
+++ b/application/F3DOptionsParser.h
@@ -40,6 +40,7 @@ struct F3DAppOptions
   std::vector<double> CameraPosition = { 0 };
   std::vector<double> CameraViewUp = { 0 };
   double CameraViewAngle = 0.0;
+  double LightIntensity = 0.0;
 };
 
 class F3DOptionsParser
--- a/application/F3DOptionsParser.h
+++ b/application/F3DOptionsParser.h
@@ -40,7 +40,6 @@ struct F3DAppOptions
   std::vector<double> CameraPosition = { 0 };
   std::vector<double> CameraViewUp = { 0 };
   double CameraViewAngle = 0.0;
-  double LightIntensity = 0.0;
 };
 
 class F3DOptionsParser
--- a/library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
+++ b/library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
@@ -97,7 +97,7 @@ void vtkF3DRenderer::Initialize(const std::string& fileInfo, const std::string& 
 {
   this->RemoveAllViewProps();
   this->RemoveAllLights();
-  this->originalLightIntensities.clear();
+  this->OriginalLightIntensities.clear();
 
   this->AddActor(this->FilenameActor);
   this->AddActor(this->GridActor);
@@ -525,14 +525,15 @@ void vtkF3DRenderer::SetLightIntensity(const double intensityFactor)
   vtkCollectionSimpleIterator it;
   for (lc->InitTraversal(it); (light = lc->GetNextLight(it));)
   {
-    double originalIntensity = light->GetIntensity();
-    if (this->originalLightIntensities.count(light))
+    double originalIntensity;
+    if (this->OriginalLightIntensities.count(light))
     {
-      originalIntensity = this->originalLightIntensities[light];
+      originalIntensity = this->OriginalLightIntensities[light];
     }
     else
     {
-      this->originalLightIntensities[light] = originalIntensity;
+      originalIntensity = light->GetIntensity();
+      this->OriginalLightIntensities[light] = originalIntensity;
     }
 
     light->SetIntensity(originalIntensity * intensityFactor);
--- a/library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
+++ b/library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
@@ -515,6 +515,29 @@ void vtkF3DRenderer::SetBackground(const double* color)
 }
 
 //----------------------------------------------------------------------------
+void vtkF3DRenderer::SetLightIntensity(const double intensity)
+{
+  if (intensity > 0)
+  {
+    if (intensity != LightKit->GetKeyLightIntensity() || !this->LightKitAdded)
+    {
+      LightKit->SetKeyLightIntensity(intensity);
+      LightKit->SetKeyLightWarmth(.5);
+      if (!this->LightKitAdded)
+      {
+        LightKit->AddLightsToRenderer(this);
+        this->LightKitAdded = true;
+      }
+    }
+  }
+  else if (this->LightKitAdded)
+  {
+    LightKit->RemoveLightsFromRenderer(this);
+    this->LightKitAdded = false;
+  }
+}
+
+//----------------------------------------------------------------------------
 void vtkF3DRenderer::SetUseDepthPeelingPass(bool use)
 {
   if (this->UseDepthPeelingPass != use)
--- a/library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
+++ b/library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
@@ -5,6 +5,8 @@
 #include "vtkF3DOpenGLGridMapper.h"
 #include "vtkF3DRenderPass.h"
 
+#include "vtkLight.h"
+#include "vtkLightCollection.h"
 #include <vtkAxesActor.h>
 #include <vtkBoundingBox.h>
 #include <vtkCamera.h>
@@ -95,6 +97,7 @@ void vtkF3DRenderer::Initialize(const std::string& fileInfo, const std::string& 
 {
   this->RemoveAllViewProps();
   this->RemoveAllLights();
+  this->originalLightIntensities.clear();
 
   this->AddActor(this->FilenameActor);
   this->AddActor(this->GridActor);
@@ -515,25 +518,24 @@ void vtkF3DRenderer::SetBackground(const double* color)
 }
 
 //----------------------------------------------------------------------------
-void vtkF3DRenderer::SetLightIntensity(const double intensity)
+void vtkF3DRenderer::SetLightIntensity(const double intensityFactor)
 {
-  if (intensity > 0)
+  vtkLightCollection* lc = this->GetLights();
+  vtkLight* light;
+  vtkCollectionSimpleIterator it;
+  for (lc->InitTraversal(it); (light = lc->GetNextLight(it));)
   {
-    if (intensity != LightKit->GetKeyLightIntensity() || !this->LightKitAdded)
+    double originalIntensity = light->GetIntensity();
+    if (this->originalLightIntensities.count(light))
     {
-      LightKit->SetKeyLightIntensity(intensity);
-      LightKit->SetKeyLightWarmth(.5);
-      if (!this->LightKitAdded)
-      {
-        LightKit->AddLightsToRenderer(this);
-        this->LightKitAdded = true;
-      }
+      originalIntensity = this->originalLightIntensities[light];
     }
-  }
-  else if (this->LightKitAdded)
-  {
-    LightKit->RemoveLightsFromRenderer(this);
-    this->LightKitAdded = false;
+    else
+    {
+      this->originalLightIntensities[light] = originalIntensity;
+    }
+
+    light->SetIntensity(originalIntensity * intensityFactor);
   }
 }
 
--- a/library/VTKExtensions/Rendering/vtkF3DRenderer.h
+++ b/library/VTKExtensions/Rendering/vtkF3DRenderer.h
@@ -12,6 +12,7 @@
 #define vtkF3DRenderer_h
 
 #include <vtkOpenGLRenderer.h>
+#include <vtkLightKit.h>
 
 class vtkCornerAnnotation;
 class vtkOrientationMarkerWidget;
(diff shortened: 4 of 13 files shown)


library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -514,6 +514,29 @@ void vtkF3DRenderer::SetBackground(const double* color)
...
+      if (!this->LightKitAdded)
+      {
+        LightKit->AddLightsToRenderer(this);
+        this->LightKitAdded = true;
+      }
+    }
+  }
+  else if (this->LightKitAdded)
+  {
+    LightKit->RemoveLightsFromRenderer(this);
```
mwestphal, line 535: What happens if you set the lightIntensity to 0 instead ?

application/F3DOptionsParser.cxx
```diff
@@ -246,6 +246,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp3, "filename", "n", "Display filename", options.getAsBoolRef("ui.filename"), HasDefault::YES, MayHaveConfig::YES);
     this->DeclareOption(grp3, "metadata", "m", "Display file metadata", options.getAsBoolRef("ui.metadata"), HasDefault::YES, MayHaveConfig::YES);
     this->DeclareOption(grp3, "blur-background", "u", "Blur background", options.getAsBoolRef("render.background.blur"), HasDefault::YES, MayHaveConfig::YES);
+    this->DeclareOption(grp3, "light-intensity", "", "Light intensity", options.getAsDoubleRef("render.light.intensity"), HasDefault::YES, MayHaveConfig::YES, "<intensity>");
```
mwestphal, line 249: README.md update needed

library/VTKExtensions/Rendering/vtkF3DRenderer.h
```diff
@@ -175,6 +177,9 @@ class vtkF3DRenderer : public vtkOpenGLRenderer
   std::string HDRIFile;
   std::string FontFile;
 
+  vtkNew<vtkLightKit> LightKit;
```
mwestphal, line 180: Isn't there already a default light kit in all scenes ? What about Renderer->GetLights() ?

mwestphal:
Hi @snoyer 

Thanks a lot for your contribution. This is a needed feature, but we will need to make sure this is the right implementation :). Let's wait for @Meakk feedback on this.

In the meantime, you may want to add tests.

library/VTKExtensions/Rendering/vtkF3DRenderer.h
snoyer: According to the vtkLightKit doc

> To enforce the relationship between the three lights, the intensity of the fill and headlights are set as a ratio to the key light brightness. Thus, the brightness of all the lights in the scene can be changed by changing the key light intensity.

so having a handle on a full light kit rather than individual lights would make changing the intensity consistently easier I would guess? (disclaimer: again, I know nothing about f3d or VTK)

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
mwestphal: Ok, lets see what @Meakk thinks of that.

library/VTKExtensions/Rendering/vtkF3DRenderer.h
mwestphal: tbh @snoyer , for a F3D and VTK beginner, this is great code !

We will check that and get back to you.

Meakk (changes requested):
Thanks for this @snoyer!  
I think the current approach might cause issues in some cases.  
Let me explain quickly how lights are handled in F3D.  

There are two code paths depending if the file format support "full scene" mode (e.g. OBJ, GLTF), or "geometry only" (if file format does not support lights like STL for example, or if `--geometry-only` option is set).  
If the mode is "geometry only", then we are using a vtkLightKit (see `vtkF3DGenericImporter::ImportLights`)  
If the move is "full scene", then the VTK importer is responsible of adding the lights (most of the time a single headlight is added, but we can have more complicated light configuration)  
So adding a `vtkLightKit` if `--light-intensity` is set will not change the intensity of the lights added by the importer.  

My recommendation is to loop on existing lights (Use `this->GetLights()`) and change all the intensities using `vtkLight::SetIntensity` in your function `vtkF3DRenderer::SetLightIntensity`.  
Don't forget to call this function everytime by adding it in `window_impl::UpdateDynamicOptions` to make sure it's properly called when a new file is loaded.

Meakk:
For the loop, something like that should work:

```cpp
  vtkLight* light;
  vtkLightCollection* lc = this->GetLights();
  vtkCollectionSimpleIterator it;
  for (lc->InitTraversal(it); (light = lc->GetNextLight(it));)
  {
    light->SetIntensity(intensity);
  }
```

snoyer:
> Thanks for this @snoyer! I think the current approach might cause issues in some cases. Let me explain quickly how lights are handled in F3D.
> 
> There are two code paths depending if the file format support "full scene" mode (e.g. OBJ, GLTF), or "geometry only" (if file format does not support lights like STL for example, or if `--geometry-only` option is set). If the mode is "geometry only", then we are using a vtkLightKit (see `vtkF3DGenericImporter::ImportLights`) If the move is "full scene", then the VTK importer is responsible of adding the lights (most of the time a single headlight is added, but we can have more complicated light configuration) So adding a `vtkLightKit` if `--light-intensity` is set will not change the intensity of the lights added by the importer.

Thanks @Meakk, this is very helpful.

> My recommendation is to loop on existing lights (Use `this->GetLights()`) and change all the intensities using `vtkLight::SetIntensity` in your function `vtkF3DRenderer::SetLightIntensity`. Don't forget to call this function everytime by adding it in `window_impl::UpdateDynamicOptions` to make sure it's properly called when a new file is loaded.

Alright, @mwestphal also suggested working with existing lights and that all makes sense, I have adjusted the code accordingly.

snoyer:
Interestingly, this works in window mode but not always in mode.

It looks like `window_impl::UpdateDynamicOptions` is called both before and after the lights get added when in window mode, but when `--output` is set it is only called before and not after.

mwestphal:
@snoyer This sounds like an actual bug, I will take a look.

mwestphal:
I reproduce the issue, I'll take a look.

mwestphal:
@Meakk Turns out there is a **third** light code path

```
  ///@{
  /**  
   * Turn on/off a flag which disables the automatic light creation capability.
   * Normally in VTK if no lights are associated with the renderer, then a light
   * is automatically created. However, in special circumstances this feature is
   * undesirable, so the following boolean is provided to disable automatic
   * light creation. (Turn AutomaticLightCreation off if you do not want lights
   * to be created.)
   */
  vtkGetMacro(AutomaticLightCreation, vtkTypeBool);
  vtkSetMacro(AutomaticLightCreation, vtkTypeBool);
  vtkBooleanMacro(AutomaticLightCreation, vtkTypeBool);
  ///@}

```


`vtkOpenGLRenderer::UpdateLights()`
```
  // create alight if needed
  if (!lightingCount)
  {
    if (this->AutomaticLightCreation)
    {    
      vtkDebugMacro(<< "No lights are on, creating one.");
      this->CreateLight();
      lc->InitTraversal(sit);
```

When an importer do not provide any light, VTK kindly add a single point light. However it adds it only one the first render, and this is too late for @snoyer feature.

I will investigate how to fix that.

mwestphal:
@snoyer : here is the fix: https://github.com/f3d-app/f3d/pull/463

You can either rebase on my branch or just copy the lines, as you see fit.

mwestphal:
@snoyer : I would suggest a rebase instead of a merge, cleaner git history.

snoyer:
> @snoyer : I would suggest a rebase instead of a merge, cleaner git history.

Sorry, I suck at git and did the lazy thing of updating the branch form the website :/

mwestphal:
No worries, I'll check that before the merge :)

application/F3DOptionsParser.h
```diff
@@ -40,6 +40,7 @@ struct F3DAppOptions
   std::vector<double> CameraPosition = { 0 };
   std::vector<double> CameraViewUp = { 0 };
   double CameraViewAngle = 0.0;
+  double LightIntensity = 0.0;
```
snoyer, line 43: this is actually unused, should be removed

library/VTKExtensions/Rendering/vtkF3DRenderer.h
```diff
@@ -175,6 +178,8 @@ class vtkF3DRenderer : public vtkOpenGLRenderer
   std::string HDRIFile;
   std::string FontFile;
 
+  std::map<vtkLight*, double> originalLightIntensities;
```
mwestphal, line 181: member var start with capital letter: OriginalLightIntensities

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -514,6 +517,28 @@ void vtkF3DRenderer::SetBackground(const double* color)
...
+  for (lc->InitTraversal(it); (light = lc->GetNextLight(it));)
+  {
+    double originalIntensity = light->GetIntensity();
+    if (this->originalLightIntensities.count(light))
+    {
+      originalIntensity = this->originalLightIntensities[light];
+    }
+    else
+    {
+      this->originalLightIntensities[light] = originalIntensity;
```
mwestphal, line 535: I think this map should be cleaned on initialization

mwestphal: already done,

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -514,6 +517,28 @@ void vtkF3DRenderer::SetBackground(const double* color)
...
+    }
```
mwestphal, line 536: how about:

```
double originalIntensity;
if (this->originalLightIntensities.contains(light))
{
  originalIntensity = this->originalLightIntensities[light];
}
else
{
  originalIntensity = light->GetIntensity();
  this->originalLightIntensities[light] = originalIntensity;
}
```

library/VTKExtensions/Rendering/vtkF3DRenderer.h
```diff
@@ -11,6 +11,8 @@
 #ifndef vtkF3DRenderer_h
 #define vtkF3DRenderer_h
 
+#include <map>
+#include <vtkLightKit.h>
```
mwestphal, line 15: vtkLight

mwestphal (changes requested):
Still needs a test :)

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
snoyer: `.clear()` in `vtkF3DRenderer::Initialize` not good?

library/testing/TestSDKDynamicLightIntensity.cxx
```diff
@@ -0,0 +1,47 @@
+#include <engine.h>
+#include <loader.h>
+#include <options.h>
+#include <window.h>
+
+#include "TestSDKHelpers.h"
+
+int TestSDKDynamicLightIntensity(int argc, char* argv[])
```
Meakk, line 8: You could have added one line in `cmake/testing.cmake` it would have been easier, but it's fine the way you did it.  
@mwestphal we should add a section in the documentation explaining how to make test.

Meakk (commented):
Please add the option in the documentation and it's good to go in my opinion.  
Thanks again for your contribution.

library/testing/TestSDKDynamicLightIntensity.cxx
snoyer: Yeah I couldn't figure it out so I went with TestSDKDynamicBackgroundColor as a model.

In the end it worked out best this way because the test covers changing the intensity dynamically (ie. the test will fail if `this->OriginalLightIntensities[light] = originalIntensity;` is removed, which would not be a problem when `vtkF3DRenderer::SetLightIntensity` is only called once)

mwestphal: Very good point @Meakk !

It would be nice to also add a --light-intensity test in cmake/testing.cmake though.

library/VTKExtensions/Rendering/vtkF3DRenderer.h
```diff
@@ -11,6 +11,8 @@
 #ifndef vtkF3DRenderer_h
 #define vtkF3DRenderer_h
 
+#include <map>
```
mwestphal, line 14: std include should be put after vtk includes, with an empty lines in betweem

snoyer: Could the formatter enforce that? something like:
```yaml
IncludeBlocks: Regroup
IncludeCategories:
  - Regex:           '".*"'
    Priority:        1
  - Regex:           '^<(vtk.*)>'
    Priority:        2
    CaseSensitive:   true
  # ... anything else ?
```

library/testing/TestSDKDynamicLightIntensity.cxx
snoyer: I barely managed to write and run this one test by poking around and making educated guesses, I still don't know how to properly run the whole test suite.
You guys need to document how to 1. run the existing tests, and 2. write new ones if you don't want new contributors to give up before their feature is considered good enough

library/VTKExtensions/Rendering/vtkF3DRenderer.h
mwestphal: good point, wdyt @Meakk ?

library/testing/TestSDKDynamicLightIntensity.cxx
mwestphal: Ok, that's fair. We are currently rewriting the whole documentation but it is taking some time.

Thanks a lot for your patience and tenacity. I'll add another test once we merge this.

mwestphal (approved):

library/VTKExtensions/Rendering/vtkF3DRenderer.h
mwestphal: alternative is to write a guide style, but such a doc are hard to write and maintain

library/testing/TestSDKDynamicLightIntensity.cxx
mwestphal: Here is the doc, still a WIP, but should make things clearer:

https://github.com/mwestphal/f3d/blob/jekyll_doc/doc/dev/TESTING.md

Meakk (approved):

library/VTKExtensions/Rendering/vtkF3DRenderer.h
mwestphal: First try at a coding style doc: https://github.com/mwestphal/f3d/blob/jekyll_doc/doc/dev/CODING_STYLE.md

PR #462, opened 2022-10-28
Fix ZSH completions

allemangD:
Characters `[]` may not appear in the explanation text of ZSH completions, so change ` to not include those characters.

- Format choices as `{A, B, C}` instead of `[A|B|C]` [(the same syntax as Python `argparse`](https://docs.python.org/3/library/argparse.html#choices))

- Format implicit values as `<arg>(=value)` instead of `[=<arg>(=value)]` (don't know an existing alternate syntax, so just stripped `[]`).

The `completions.zsh` regex should also skip short-options that occur in option explanations (text `-y` occurs in the explanation for `--comp`).

Add a trailing `*:file:_files` optspec so that filenames may autocomplete.

### DIFF
--- a/resources/completion.zsh
+++ b/resources/completion.zsh
@@ -13,4 +13,4 @@ longopts=$(echo $f3dhelp | grep "[-]-" | sed 's/=.*>//g' | sed 's/-.,//g' | sed 
 arguments=("${(f)shortopts}")
 arguments+=("${(f)longopts}")
 
-_arguments "$arguments[@]"
+_arguments "$arguments[@]" "*:file:_files"
--- a/resources/completion.zsh
+++ b/resources/completion.zsh
@@ -5,9 +5,9 @@ local shortopts
 local arguments
 local f3dhelp
 
-f3dhelp=$(f3d --help 2>&1 | sed '1,/Examples/!d')
+f3dhelp=$(f3d --help 2>&1 | sed '1,/Examples/!d' | sed 's/\[.*\] *//')
 
-shortopts=$(echo $f3dhelp | grep "[-].," | sed "s/^ *\(-.\), *--[^ ]* *\(.*\)$/\1[\2]/")
+shortopts=$(echo $f3dhelp | grep "^\s*[-].," | sed "s/^ *\(-.\), *--[^ ]* *\(.*\)$/\1[\2]/")
 longopts=$(echo $f3dhelp | grep "[-]-" | sed 's/=.*>//g' | sed 's/-.,//g' | sed "s/^ *\([^ ]*\) *\(.*\)$/\1[\2]/g")
 
 arguments=("${(f)shortopts}")
--- a/application/F3DOptionsParser.cxx
+++ b/application/F3DOptionsParser.cxx
@@ -213,7 +213,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp1, "quiet", "", "Enable quiet mode, which superseed any verbose options and prevent any console output to be generated at all", appOptions.Quiet,  HasDefault::YES, MayHaveConfig::YES );
     this->DeclareOption(grp1, "progress", "", "Show progress bar", options.getAsBoolRef("ui.loader-progress"), HasDefault::YES, MayHaveConfig::YES);
     this->DeclareOption(grp1, "geometry-only", "", "Do not read materials, cameras and lights from file", options.getAsBoolRef("scene.geometry-only"), HasDefault::YES, MayHaveConfig::YES);
-    this->DeclareOption(grp1, "up", "", "Up direction", options.getAsStringRef("scene.up-direction"), HasDefault::YES, MayHaveConfig::YES, "[-X|+X|-Y|+Y|-Z|+Z]");
+    this->DeclareOption(grp1, "up", "", "Up direction", options.getAsStringRef("scene.up-direction"), HasDefault::YES, MayHaveConfig::YES, "{-X, +X, -Y, +Y, -Z, +Z}");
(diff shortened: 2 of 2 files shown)


application/cxxopts.hpp
```diff
@@ -1834,7 +1834,7 @@ namespace cxxopts
       {
         if (o.has_implicit)
         {
-          result += " [=" + arg + "(=" + toLocalString(o.implicit_value) + ")]";
+          result += " " + arg + "(=" + toLocalString(o.implicit_value) + ")";
```
Meakk, line 1837: The other changes look fine to me, but this one is inside cxxopts so your change may be overwritten if we update cxxopts at some point.  
What about removing it completely all the text in the brackets by piping `sed "s/\[.*\]//g"` just after `echo $f3dhelp`?

Meakk (changes requested):
Thanks @allemangD!  
Please take a look at my comment. If we can avoid modifying cxxopts, it would be great!

mwestphal:
Hello @allemangD , do you need any help to address the comments ?

allemangD:
> Hello @allemangD , do you need any help to address the comments ?

I don't think so, I've just been distracted by other projects. I should be able to get to it today. Thanks for checking in.

allemangD:
@Meakk Simply removing `[]` via sed substitution did work, but it meant the help text for those options was entirely empty. In general I noticed all the options with multiline helptext (`--comp`, `--range`, `--emissive-factor`, etc) only included the first line of the text.

I rewrote the `sed` substitutions in `completions.zsh` to join those multiline helptext, normalize spacing, remove invalid chars, etc. so that the ZSH suggestions are more useful:

```
$ f3d --c
--camera-azimuth-angle    -- <angle> Camera azimuth angle (in degrees) (default: 0)
--camera-elevation-angle  -- <angle> Camera elevation angle (in degrees) (default: 0)
--camera-focal-point      -- <X,Y,Z> Camera focal point (default: 0)
--camera-index            -- <index> Select the camera to use (default: -1)
--camera-position         -- <X,Y,Z> Camera position (default: 0)
--camera-view-angle       -- <angle> Camera view angle (non-zero, in degrees)
--camera-view-up          -- <X,Y,Z> Camera view up (default: 0)
--cells                   -- Use a scalar array from the cells
--colormap                -- <color_list> Specify a custom colormap (default: 0,0,0,0,0.4,0.9,0,0,0.8,0.9,0.9,0,1,1,1,1)
--color                   -- <R,G,B> Solid color (default: 1,1,1)
--comp                    -- Component from the scalar array to color with. -1 means magnitude, -2 or the short option, -y, means direct scalars (default: -1)
--config                  -- <filePath/filename/fileStem> Specify the configuration file to use. absolute/relative path or filename/filestem to search in configuration file locations.
```

I included a link in a comment, but the sed strategy for the multiline text is described here https://stackoverflow.com/a/37951863/4672189. I don't have a suitable MacOS environment to check that it works there, but I tried to use the Mac/BSD-compliant version. It might be worth checking that compatibility before merging.

Meakk:
This is some next level sed commands :)  
I'll test it this weekend on a Mac.

resources/completion.zsh
```diff
@@ -1,16 +1,32 @@
...
+#   - Illegal characters [].
+#   - Leading whitespace.
+# Join multiline helptext. See https://stackoverflow.com/a/37951863/4672189 for details.
+# Normalize whitespace.
+# Discard trailing whitespace.
+commandlist=$(
+  f3d --help 2>&1 |
+    tail -n +5 |
+    sed '1,/Keys:/!d' | sed '/.*:$/d' | sed 's/\[.*\] *//g' | sed 's/^\s*//' |
+    sed -e ':a' -e 'N' -e '/\n-[a-z\-]/!s/\n/ /' -e 'ta' -e 'P' -e 'D' |
```
Meakk, line 22: Unfortunately, this line does not seem to work on macOS and remove everything.

Meakk: More specifically, this part causes issues: `/\n-[a-z\-]/!s/\n/ /`

allemangD: :slightly_frowning_face:

It looks like MacOS doesn't respect the `\n` escape. I tried using the `$'\n'` technique mentioned in that StackOverflow answer, but that breaks functionality in Linux.

I've just pushed 25e617f where I move that complex regex into a separate variable with a multiline string. This works on Linux, and the StackOverflow post suggests it should also work on MacOS.

Meakk: I'll check again

Meakk: Still failing:

```
➜  build git:(fix-zsh-completion) f3d --
sed: 1: "/\
-[a-z\-]/!s/\
/ /;
": unterminated regular expression

_arguments:comparguments:327: invalid argument: 
sed: 1: "/\
-[a-z\-]/!s/\
/ /;
": unterminated regular expression
_arguments:comparguments:327: invalid argument: 
sed: 1: "/\
-[a-z\-]/!s/\
/ /;
": unterminated regular expression
_arguments:comparguments:327: invalid argument: 
```

Meakk: I agree. Please remove the last two commits and it's good to go.

allemangD: :+1: I've moved those two commits to https://github.com/allemangD/f3d/tree/multiline-helptext. This PR should be good now.

Meakk (approved):

Meakk:
Thanks @allemangD !

resources/completion.zsh
Meakk: There is no problem with MacOS on master, it's just that the last improvements made by @allemangD to the autocomplete didn't work so he reverted them.

mwestphal: ok, thats fair.

PR #471, opened 2022-11-04
fix image data python bindings

snoyer:
`pybind` doesn't automagically bind C-style arrays ([which makes sense](https://github.com/pybind/pybind11/issues/1460)) so there needs to be some explicit conversion to properly expose `char*` image data in python as `bytes`/`bytearray` type.


Meakk (approved):
Looks good to me, thanks!

mwestphal:
Nice, thanks @snoyer !

PR #477, opened 2022-11-09
Fixing some issues with F3D website

mwestphal:
- Sadly using `include` breaks all jekyll links inside the included part, reverting to duplicating meain readme.
- Use local logo in readme as it works perfectly
- Use a cloud icon for favicon as the path is relative aand breaks in the doc tree
- Fix a wrong link
- Fix license detection by using a raw license file
- Last point means the link in the "summary" and the link in the readme are not the same, the best I could do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ permalink: /
 # F3D - Fast and minimalist 3D viewer
 By Michael Migliore and Mathieu Westphal.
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -1,9 +1,17 @@
+---
```
Meakk, line 1: why did you restore the front matter in the readme file?

mwestphal: > Sadly using include breaks all jekyll links inside the included part, reverting to using front matter in readme

Meakk (approved):
That's fine for now

PR #491, opened 2022-11-19
Focal point picking

snoyer:
This is a feature I use all the time in Meshlab: double click on the model and the camera will target the picked point. Very useful to zoom in and inspect details.

The following video capture shows rotating the model about the default origin, then double clicking on a wheel and rotating about it, and finally double clicking and zooming on the hand of the figure to have a better look: 
[focal-point-picking-buggy.webm](https://user-images.githubusercontent.com/489715/202837046-fd695675-f7e5-4683-b372-09d335406425.webm)

Discussion points:
- we have a low level window so we need to implement double click detection ourselves?
- is there a good way to animate the transition? (dedicated animation thread/queue?)


mwestphal:
Thanks @snoyer for your features and ideas !

> we need to implement double click detection ourselves?

Indeed, so I wonder if a modifier + click mayu be better interactions for this. Do you know if this double click gesture is standard in other software ?

> is there a good way to animate the transition? (dedicated animation thread/queue?)

No such thing in VTK, though it should be implementable.
That being said do you think animating the transition provide anything apart from looking "nice" ?


BTW We are thinking of creating a discord channel for informal discussions about F3D, would you be interested in joining ?

Meakk:
Really nice!  
FYI CATIA uses the mouse wheel click to set the focus on release (if the mouse has not moved between press and release)  
Also, I'm afraid the point picker will not work 3d model with large triangles (like a box for example), I suspect the vtkCellPicker will give you more precise results (but will not work on point clouds...)

snoyer:
> > we need to implement double click detection ourselves?
> 
> Indeed, so I wonder if a modifier + click may be better interactions for this. Do you know if this double click gesture is standard in other software ?

Meshlab is the only other viewer that comes to mind so I can't be positive either way.

> > is there a good way to animate the transition? (dedicated animation thread/queue?)
> 
> No such thing in VTK, though it should be implementable. That being said do you think animating the transition provide anything apart from looking "nice" ?

I believe it does slightly more than looking nice. Being able to follow the subject as the camera moves (even if the transition is quite short) saves your brain the split second of "oh yeah, right." it needs to register the new position and confirm the move is what you expected it to be.

It would be a "nice to have" detail, but it's totally functional without the animation so it can be done without if it is problematic.

> BTW We are thinking of creating a discord channel for informal discussions about F3D, would you be interested in joining ?
Sure


> FYI CATIA uses the mouse wheel click to set the focus on release (if the mouse has not moved between press and release)

That would also make sense as middle click is what is used for dragging currently.

> Also, I'm afraid the point picker will not work well on 3d model with large triangles (like a box for example), I suspect the vtkCellPicker will give you more precise results (but will not work on point clouds...)

I'll admit I haven't investigated the different pickers (or tried to load a point cloud). Is it possible to decide which picker to used based on the model? Or just have 2 pickers, pick with the CellPicker first and fall back to the PointPicker if there's no hit?

library/VTKExtensions/Rendering/vtkF3DInteractorStyle.h
```diff
@@ -44,6 +44,16 @@ class vtkF3DInteractorStyle : public vtkInteractorStyleTrackballCamera
...
+  /**
+   * Handle middle mouse button
+   */
+  void OnMiddleButtonPress();
+
+  /**
+   * Handle middle mouse button
+   */
+  void OnMiddleButtonRelease();
+
```
Meakk, line 56: these functions are useless, you can remove them

library/src/interactor_impl.cxx
```diff
@@ -268,6 +282,88 @@ class interactor_impl::internals
...
+    self->Style->OnMiddleButtonDown();
+  }
+
+  static void OnMiddleButtonRelease(vtkObject*, unsigned long, void* clientData, void*)
+  {
+    internals* self = static_cast<internals*>(clientData);
+
+    const auto click_pos = self->VTKInteractor->GetEventPosition();
+
+    const auto x_delta = click_pos[0] - self->MiddleClickPosision[0];
```
Meakk, line 302: Can you get the last position using `self->VTKInteractor->GetLastEventPosition();`? If it works, it would simplify the code.

library/src/interactor_impl.cxx
```diff
@@ -268,6 +282,88 @@ class interactor_impl::internals
...
+    self->MiddleClickPosision[1] = click_pos[1];
+
+    self->Style->OnMiddleButtonDown();
+  }
+
+  static void OnMiddleButtonRelease(vtkObject*, unsigned long, void* clientData, void*)
+  {
+    internals* self = static_cast<internals*>(clientData);
+
+    const auto click_pos = self->VTKInteractor->GetEventPosition();
```
Meakk, line 300: Avoid `auto` for simple types. I'm not against it, but we try to be consistent with the existing code style.

library/src/interactor_impl.cxx
```diff
@@ -268,6 +282,88 @@ class interactor_impl::internals
...
+
+    const auto x_delta = click_pos[0] - self->MiddleClickPosision[0];
+    const auto y_delta = click_pos[1] - self->MiddleClickPosision[1];
+    const auto sq_pos_delta = x_delta * x_delta + y_delta * y_delta;
+    if (sq_pos_delta < self->DragDistanceTol * self->DragDistanceTol)
+    {
+      bool pick_successful = false;
+      double picked_pos[3];
+      self->CellPicker->Pick(self->MiddleClickPosision[0], self->MiddleClickPosision[1], 0,
```
Meakk, line 309: Typo: `MiddleClickPosision` -> `MiddleClickPosition`

library/src/interactor_impl.cxx
```diff
@@ -268,6 +282,88 @@ class interactor_impl::internals
...
+        const auto dz = picked_pos[2] - foc[2];
+
+        if (self->TransitionDuration > 0)
+        {
+          const auto start = std::chrono::high_resolution_clock::now();
+          const auto end = start + std::chrono::milliseconds(self->TransitionDuration);
+          auto now = start;
+          while (now < end)
+          {
+            const double linear_t =
```
Meakk, line 343: We are using camel case for variables. What about `linearTime` instead of `linear_t`.

library/src/interactor_impl.cxx
```diff
@@ -268,6 +282,88 @@ class interactor_impl::internals
...
+              std::chrono::duration_cast<std::chrono::milliseconds>(now - start).count() /
+              (double)self->TransitionDuration;
+            const double tween_t = (1 - std::cos(vtkMath::Pi() * linear_t)) / 2;
```
Meakk, line 346: Here too

Meakk (changes requested):
Here are some comments.  
I built and tested your branch, and it works very well!  
This feature is very nice, thank you!

library/src/interactor_impl.cxx
snoyer: It seems `GetLastEventPosition` gets updated while dragging. If the pointer has stopped moving before releasing the button the values are the same as `GetEventPosition`.

library/src/interactor_impl.cxx
snoyer: This is the hardest change of the bunch. They're interpolation parameters, can we just name then `t` and `u`?

Meakk: sure

library/src/interactor_impl.cxx
snoyer: I kept `auto` for the `std::chrono::` stuff that gets quite verbose, let me know if I should change that too.

library/src/interactor_impl.cxx
```diff
@@ -297,41 +295,44 @@ class interactor_impl::internals
   {
     internals* self = static_cast<internals*>(clientData);
 
-    const auto click_pos = self->VTKInteractor->GetEventPosition();
+    const int* MiddleButtonUpPosition = self->VTKInteractor->GetEventPosition();
```
Meakk, line 298: Local variables start with a lowercase

library/src/interactor_impl.cxx
```diff
@@ -340,15 +341,15 @@ class interactor_impl::internals
...
               (double)self->TransitionDuration;
-            const double tween_t = (1 - std::cos(vtkMath::Pi() * linear_t)) / 2;
+            const double u = (1 - std::cos(vtkMath::Pi() * t)) / 2;
 
             self->Window.getCamera().setFocalPoint(
-              { foc[0] + dx * tween_t, foc[1] + dy * tween_t, foc[2] + dz * tween_t });
+              { foc[0] + dx * u, foc[1] + dy * u, foc[2] + dz * u });
             self->Window.getCamera().setPosition(
-              { pos[0] + dx * tween_t, pos[1] + dy * tween_t, pos[2] + dz * tween_t });
+              { pos[0] + dx * u, pos[1] + dy * u, pos[2] + dz * u });
```
Meakk, line 352: If the picked point is closer or further, there is a side effect of producing a zoom.  
I think this is because you are applying the same vector to the focal point and the position.  
I think it can be fixed easily by projecting the new position on the camera plane.
In practice, it's just a dot product. Here's a sketch to illustrate what I'm suggesting, let me know if you need help implementing it.

image

Meakk (changes requested):
Besides these two comments, you will need to add a test to cover your code.  
Please add a line after this line: https://github.com/f3d-app/f3d/blob/185baa389446f509937b2393341fdc9ee0c66c92/cmake/testing.cmake#L322  
You will need to provide an interaction file, they are located in `testing/recordings` (you just have to name it `<test_name>.log`. It can be recorded using `--interaction-test-record` option. @mwestphal can help you if you have problems.  
Also, please add the documentation in `doc/user/INTERACTIONS.md`

mwestphal:
More info (but nothing about interactions sadly) in https://f3d.app/doc/dev/TESTING.html

library/src/interactor_impl.cxx
snoyer: If I understand correctly, on the following sketch, when focusing from `f0` to `f1`, the camera should go form `p0` to `p2` instead of `p1`, is that right?
focal-move

Meakk: Exactly. I think it makes more sense. What do you think?

snoyer: The camera will travel in the plane and therefore will not go back and forth (which sounds nice) but the distance to the focal point will not be constant anymore (which could backfire?). This new method might actually produce more of a "zoom" effect (getting closer/further to the object) while the old one has more of a "dolly" effect but maintains the same distance.

Meakk: The focal point is used only to compute the camera direction and is used as the rotation center, so I don't think it will affect the zoom factor.  
The "dolly" effect you are mentioning is what I'd like to avoid, but maybe it's just my preference, I'm not pushing for it but it would be great if we can compare the two behaviors and decide what makes more sense.

library/src/interactor_impl.cxx
```diff
@@ -308,31 +308,44 @@ class interactor_impl::internals
...
       }
 
       if (pickSuccessful)
       {
+        /*     pos.--------------------.foc
+         *       /|                   /
+         *      / |                  /
+         *     .--.-----------------.picked
+         * pos1     pos2
+         */
```
snoyer, line 335: Not sure how you guys feel about ascii art diagrams, coding style doesn't mention it ;)

Meakk: No problem on my side, it looks good :)

library/src/interactor_impl.cxx
Meakk: @snoyer I quickly tested and it behaves better now in my opinion. So I vote to keep this new version :)

snoyer: Agreed, I haven't found a situation where the varying distance would feel worse than flying along the model when picking far away. Good call.

cmake/testing.cmake
```diff
@@ -320,6 +320,9 @@ f3d_test(NAME TestInteractionDumpSceneState DATA dragon.vtu NO_BASELINE INTERACT
 f3d_test(NAME TestInteractionCycleVerbose DATA dragon.vtu ARGS --verbose -s NO_BASELINE INTERACTION REGEXP "Not coloring")#SSSSYC
 f3d_test(NAME TestInteractionEmptyDrop INTERACTION REGEXP "Drop event without any provided files.")#DropEvent Empty;
 f3d_test(NAME TestInteractionCameraUpdate DATA dragon.vtu INTERACTION DEFAULT_LIGHTS) #MouseWheel;MouseWheel;MouseWheel;S
+f3d_test(NAME TestInteractionFocalPointPicking DATA dragon.vtu INTERACTION DEFAULT_LIGHTS)
+f3d_test(NAME TestInteractionFocalPointPicking2 DATA dragon.vtu INTERACTION DEFAULT_LIGHTS)
+f3d_test(NAME TestInteractionFocalPointPicking3 DATA pointsCloud.vtp INTERACTION DEFAULT_LIGHTS)
```
mwestphal, line 325: Lets try avoiding numbering of test but instead choose a descriptive name

mwestphal: (the reason for this is that VTK support multiple baselines for a single test by numbering them. F3D does nt supports that and never will but still, lets make sure not to confuse anyone)

mwestphal (changes requested):
Great code ! I'll let @Meakk answer regharding the Shift usage since I did not try the feature.

I've added a single remark.

library/src/interactor_impl.cxx
```diff
@@ -49,6 +53,16 @@ class interactor_impl::internals
     dropFilesCallback->SetCallback(OnDropFiles);
     this->Style->AddObserver(vtkF3DInteractorStyle::DropFilesEvent, dropFilesCallback);
 
+    vtkNew<vtkCallbackCommand> MiddleButtonPressCallback;
```
Meakk, line 56:
```suggestion
    vtkNew<vtkCallbackCommand> middleButtonPressCallback;
```

library/src/interactor_impl.cxx
```diff
@@ -49,6 +53,16 @@ class interactor_impl::internals
...
+    MiddleButtonPressCallback->SetClientData(this);
+    MiddleButtonPressCallback->SetCallback(OnMiddleButtonPress);
+    this->Style->AddObserver(vtkCommand::MiddleButtonPressEvent, MiddleButtonPressCallback);
+
+    vtkNew<vtkCallbackCommand> MiddleButtonReleaseCallback;
```
Meakk, line 61:
```suggestion
    vtkNew<vtkCallbackCommand> middleButtonReleaseCallback;
```

library/src/interactor_impl.cxx
```diff
@@ -268,6 +282,103 @@ class interactor_impl::internals
...
+        if (self->TransitionDuration > 0)
+        {
+          const auto start = std::chrono::high_resolution_clock::now();
+          const auto end = start + std::chrono::milliseconds(self->TransitionDuration);
+          auto now = start;
+          while (now < end)
+          {
+            const double t =
+              std::chrono::duration_cast<std::chrono::milliseconds>(now - start).count() /
+              (double)self->TransitionDuration;
```
Meakk, line 362:
```suggestion
              static_cast<double>(self->TransitionDuration);
```

Meakk (changes requested):
Minor style changes, and I'm good on my side (after tests renaming)

Meakk (approved):

PR #500, opened 2022-11-26
Small doc fix

mwestphal:


library/doxygen/index.rst
```diff
@@ -23,5 +23,4 @@ Indices and tables
 ==================
 
 * :ref:`genindex`
-* :ref:`modindex`
```
Meakk, line 26: what was this line used for?

mwestphal: it generates the dead link "Module Index" in the doxygen doc: https://f3d.app/doc/libf3d/doxygen/

It is used for python documentation, see here: https://github.com/svenevs/exhale/issues/182

Meakk (approved):

Issue #490, opened 2022-11-17
Light intensity control


snoyer:
Control over the lighting of the models/scenes is currently limited to the `--light-intensity` option from the command line.
It would be nice to be able to have additional ways to change the intensity more _interactively_ and/or more _automatically_.

I have made a proof-of-concept exploring 2 ways of achieving that:

1. Manual increase and decrease intensity with `L` and `shift+L` keys. Pretty straight forward and self-explanatory.
2. Guesstimation of an "optimal" intensity for the given view when pressing `ctrl+L`. This is basically an auto-exposure feature that will run a few render passes at different intensities to try and optimize the result. This could be exposed as an option to be performed automatically after the model is loaded (could be good for thumbnail generation?).

This is implemented in the following branch: [snoyer:light-intensity-control](https://github.com/f3d-app/f3d/compare/master...snoyer:light-intensity-control) if anyone is willing to give it a try and discuss if/how it could be properly integrated.

mwestphal:
That is great ! I will take a look this weekend. Could you create a WIP PR ? That would make it easier.

BTW did you consider adding a --light-intensity-optimal ?

Meakk:
This would be a great feature indeed.  
However, I'd like to try modifying the exposure coefficient of the tone mapping pass instead of playing with the light intensity.  
Right now, the tone mapping pass is handled here: https://github.com/f3d-app/f3d/blob/52e51527e48d6d75d5e875f829368caf1f6316a4/library/VTKExtensions/Rendering/vtkF3DRenderer.cxx#L173

There is an exposure coefficient unmodified by f3d (default is 1).  
Can you try to create a new option `render.effect.tone-mapping.exposure`, use this option to call `SetExposure`, and modify this with your keybinds? I suspect you will get better image quality with this method.

I'm also concerned about the iterative method doing a render and traversing all the pixels every iteration, it might be too expensive.

snoyer:
Thanks for the feedback guys, definitely a lot of loose ends right now which is why I opened an issue rather than a PR.

> Adding --light-intensity-optimal definitelly make sense but I wonder if the results varies a lot with the default scene, what kind of results did you get ?

The results do vary depending on the model, there's a lot of margin for improvement on the actual metric use to evaluate what is good lighting or not. I'll try and get a few examples of what I'm able to get with the few models I played with.

[...]

PR #504, opened 2022-11-27
Simple light intensity controls

snoyer:
This PR adds `L` and `shift+L` as key bindings to increase and decrease light intensity as discussed in #490.


library/src/interactor_impl.cxx
```diff
@@ -198,6 +198,14 @@ class interactor_impl::internals
         self->Options.toggle("interactor.trackball");
         render = true;
         break;
+      case 'L':
+      {
+        const double intensity = self->Options.getAsDouble("render.light.intensity");
+        const double factor = rwi->GetShiftKey() ? 0.95 : 1.05;
```
mwestphal, line 204: I'm not sure the values generated by this are the best we could have, eg, pressing the key three times will set the value to .857375. 

increase/decrease of 0.05 mayb be better ?

snoyer: True, on the other hand, increasing by a fixed percentage point (eg. `+= 0.05` instead of `*= 1.05`) feel quite slow after the first few increments, and quite fast going down.
Can you try it with and see what you think?
```
const double offset = rwi->GetShiftKey() ? -0.05 : +0.05;
self->Options.set("render.light.intensity", intensity + offset);
```

Maybe there's some magic math to produce a nicer curve

mwestphal: I will try your branch

snoyer: How about having the values increasing/decreasing as follow?
- `0.00, 0.01, ... 0.48, 0.49,` (by 0.01 until 0.5)
- `0.50, 0.52, ... 0.96, 0.98,` (by 0.02 until 1)
- `1.00, 1.05, ... 4.90, 4.95,` (by 0.05 until 5)
- `5.00, 5.10, ... 9.80, 9.90,` (by 0.10 until 10)
- `10.00, 10.25, 10.50, ...` (by 0.25 above 10)

mwestphal: LGTM

mwestphal: @Meakk

Meakk: looks good to me too

snoyer:
Light intensity value now change in nicer increments (and with integer math to avoid floating point funniness).

However, the Windows CI tests are failing and I'm not competent enough to identify why.

mwestphal:
Yeah, this is breaking in master too since yestarday, no ideai waht is happening yet.

mwestphal:
@snoyer you can rebase on master to get clean CI

library/src/interactor_impl.cxx
```diff
@@ -24,6 +24,7 @@
 
 #include <chrono>
 #include <map>
+#include <math.h>
```
mwestphal, line 27: should be cmath

mwestphal (changes requested):

mwestphal (approved):

mwestphal:
@Meakk ptal

Meakk (approved):

mwestphal:
Thanks a lot @snoyer !

PR #518, opened 2022-12-04
Add missing dep example plugin

mwestphal:


examples/plugins/example-plugin/module/vtk.module
```diff
@@ -5,3 +5,4 @@ DESCRIPTION
 DEPENDS
   VTK::CommonCore
   VTK::CommonExecutionModel
+  VTK::IOImport
```
Meakk, line 8: I know this is needed because reader.h expose a vtkImporter, maybe we should use raw pointers and forward declare it. Not essential though so feel free to merge

Meakk (approved):

examples/plugins/example-plugin/module/vtk.module
mwestphal: Yes, currently needed, but of course can be improved if you feel like it.

PR #519, opened 2022-12-04
Add SDK plugin cmake options

mwestphal:
- Add a F3D_INSTALL_SDK_PLUGIN cmake option
- Add a f3d_SDK_PLUGIN config option
- Use options when needed


cmake/installing.cmake
```diff
@@ -180,10 +180,19 @@ if(F3D_INSTALL_SDK)
...
-      "${CMAKE_CURRENT_LIST_DIR}/plugin.cxx.in"
-      "${CMAKE_CURRENT_LIST_DIR}/readerBoilerPlate.h.in"
     DESTINATION
       "lib/cmake/${PROJECT_NAME}"
     COMPONENT cmake)
+
+  if(F3D_INSTALL_SDK_PLUGIN)
+    install(
+      FILES
+        "${CMAKE_CURRENT_LIST_DIR}/f3dPlugin.cmake"
```
Meakk, line 190: Should we fill a list of files and install this list instead of having 2 install calls?

examples/plugins/example-plugin/CMakeLists.txt
```diff
@@ -6,6 +6,12 @@ include(GNUInstallDirs)
 
 find_package(f3d REQUIRED)
 
+# Check Plugin SDK is available
+if(NOT (DEFINED f3d_SDK_PLUGIN AND ${f3d_SDK_PLUGIN}))
```
Meakk, line 10: Can we do something at the SDK level? I don't expect users to check that manually

mwestphal: if the user do not check, it will just fail on the next call.

What we could do is to have a component logic on f3dConfig, but I'm not sure what our component will be and how to implement it.

In any case, user will still need to do something like:

`find_package(f3d REQUIRED COMPONENTS sdk plugin)`

Meakk: What about raising a fatal error in `f3d_init_plugin` if the plugin SDK is not present?

mwestphal: f3d_init_plugin is F3DPlugin.cmake, which is not installed

Meakk: ok then

PR #521, opened 2022-12-05
Fix compile options for plugins

Meakk:
Fix coverage, sanitizers and strict build with plugins


mwestphal (approved):
Does this have an impact on externally built plugin ?

Meakk:
@mwestphal please review again, I fixed a lot of things.

CMakeLists.txt
```diff
@@ -72,7 +72,7 @@ if(F3D_STRICT_BUILD)
   if(MSVC)
     set(f3d_strict_build_compile_options /W4 /WX)
   else()
-    set(f3d_strict_build_compile_options -Wall -Wextra -Wshadow -Woverloaded-virtual -Wno-deprecated -Wno-strict-overflow -Wno-array-bounds -Wunreachable-code -Wno-missing-field-initializers -Wno-unused-parameter -Wmissing-declarations -Wredundant-decls -Wpointer-arith -Werror)
```
mwestphal, line 75: why removing it ?

plugins/occt/module/vtkF3DOCCTReader.cxx
```diff
@@ -1,5 +1,10 @@
 #include "vtkF3DOCCTReader.h"
 
+#if defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Woverloaded-virtual"
```
mwestphal, line 5: not fixable ?

mwestphal:
reader.h coverage is lower than expected. Do you know why these methods are not hit ?

```
   * The reader having the highest score (from 0 to 100) is used to read the file.
   * Default is 50.
   */
  virtual int getScore() const { return 50; }

  /**
   * Create the geometry reader (VTK reader) for the given filename
   */
  virtual vtkSmartPointer<vtkAlgorithm> createGeometryReader(const std::string&) const
  {
    return nullptr;
  }
```

plugins/occt/module/vtkF3DOCCTReader.cxx
Meakk: No, it's in the OCCT headers

Meakk:
> reader.h coverage is lower than expected. Do you know why these methods are not hit ?
> 
> ```
>    * The reader having the highest score (from 0 to 100) is used to read the file.
>    * Default is 50.
>    */
>   virtual int getScore() const { return 50; }
> 
>   /**
>    * Create the geometry reader (VTK reader) for the given filename
>    */
>   virtual vtkSmartPointer<vtkAlgorithm> createGeometryReader(const std::string&) const
>   {
>     return nullptr;
>   }
> ```

I fixed the `getScore` by overriding it only if the CMake argument is set.  
Regarding `getGeometryReader`, it's a bit more difficult. The base implementation would be called only if the plugin is defining no importer and no reader, which makes no sense.

mwestphal:
> Regarding getGeometryReader, it's a bit more difficult. The base implementation would be called only if the plugin is defining no importer and no reader, which makes no sense.

Then I'm not sure we should provide an implementation then.

mwestphal:
> Then the impl should be {} since it cant be reached.
 
 It is already. great.

PR #536, opened 2022-12-10
Use workarounds instead of patches for Draco

Meakk:


.github/actions/draco-install-dep/action.yml
```diff
@@ -50,7 +41,8 @@ runs:
         -DBUILD_SHARED_LIBS=OFF
         -DDRACO_JS_GLUE=OFF
         -DCMAKE_BUILD_TYPE=Release
-        -DCMAKE_INSTALL_PREFIX:PATH=../draco_install
+        -DCMAKE_POSITION_INDEPENDENT_CODE=ON
+        -DCMAKE_INSTALL_PREFIX=../draco_install
```
mwestphal, line 45: how does that fix the installation issue ?

mwestphal:
Please fix the superbuild too

.github/actions/draco-install-dep/action.yml
Meakk: No idea :)  
I read the CMake doc but I don't understand.  
Looks like adding `:PATH` causing issues.

mwestphal (approved):

PR #543, opened 2022-12-18
Improve plugin SDK

### LINKED ISSUE #537: libf3d plugin sdk header should not be in public directory
libf3d plugin sdk header should not be in public directory

Meakk:
- Move `plugin.h` and `reader.h` from `public` to `plugin` folder. Fix #537 
- Plugins have responsibility to install the mimetype XML files, so the xml files have been moved from `resources` to the plugins folder.

Note that the option to install the mimetypes files have been removed and we know rely on component-specific install using the built-in generated `cmake_install.cmake` script because xml files installation is `EXCLUDE_FROM_ALL`.


.github/workflows/ci.yml
```diff
@@ -164,10 +164,15 @@ jobs:
...
 
-    - name: Install
+    - name: Install (ALL)
       working-directory: ${{github.workspace}}/build
       run: cmake --install .
 
+    - name: Install (mimetypes on Linux)
+      if: matrix.os == 'ubuntu-latest'
+      working-directory: ${{github.workspace}}/build
+      run: cmake -DCOMPONENT=mimetypes -P cmake_install.cmake
```
mwestphal, line 174: I'm a bit confused by this. How will compiling user find that they need to do that ?

Meakk: We cannot really use an option I think because a plugin can be built externally.  
So I added `EXCLUDE_FROM_ALL` because we don't want the mime-type XML always installed, right?  
So I found this component-specific installation trick.  

However, looking at the comments from [this answer](https://stackoverflow.com/a/9192877/2609654), it seems like we can replace `cmake -DCOMPONENT=mimetypes -P cmake_install.cmake` by `cmake --install . --component mimetypes` which looks way better.  
I also think we can generalize this usage to more components, by installing optional component explicitly instead of having CMake variables that configure what's installed in `ALL`.

mwestphal: Got it. Indeed, that is much better looking an generally acceptable. Make sure to document it thoroughly in build.md. We will need to inform packagers.

doc/dev/BUILD.md
```diff
@@ -45,3 +45,21 @@ Some modules, plugins and bindings depending on external libraries can be option
...
+
+Optional components can be installed by specifying the component name:
+
+```
+cmake --build ${your_build_dir} --component ${component_name}
+```
+
+Here is the list of the optional components you can install:
+
+* `mimetypes`: Install plugins mimetype XML files for integration with Freedesktop (Linux only)
```
mwestphal, line 65: Do you plan to add other components in another MR ?

mwestphal: Would it be possible to have a all_components target ?

Meakk: I do plan to `exclude from all` every single component that is not required (e.g. windows shell extension, linux desktop integration, shell completion, python bindings, java bindings, ...)

> Would it be possible to have a all_components target ?

By default all the component are in the group `all_components` and are installed if you don't explicitly specify a component.

mwestphal: sounds good. In another PR then ?

Meakk: yup

doc/dev/BUILD.md
```diff
@@ -45,3 +45,21 @@ Some modules, plugins and bindings depending on external libraries can be option
 * `F3D_PLUGIN_BUILD_DRACO`: Support for DRC file format. Requires `Draco`. Disabled by default.
 * `F3D_BINDINGS_PYTHON`: Generate python bindings (requires `Python` and `pybind11`). Disabled by default.
 * `F3D_BINDINGS_JAVA`: Generate java bindings (requires `Java` and `JNI`). Disabled by default.
+
+## Installing
+
+Installation can be done through CMake, by running the following command:
+
+```
+cmake --build ${your_build_dir}
```
mwestphal, line 54: --install ?

doc/dev/BUILD.md
```diff
@@ -45,3 +45,21 @@ Some modules, plugins and bindings depending on external libraries can be option
...
+```
+
+Optional components can be installed by specifying the component name:
+
+```
+cmake --build ${your_build_dir} --component ${component_name}
+```
+
+Here is the list of the optional components you can install:
```
mwestphal, line 63: mention `all_components` somewhere ?

Meakk: Do you want to list of all the components installed by default?

mwestphal: No, I mean that mention that you can install all components easily by using all_components

Meakk: There is a misunderstanding. There is no `all_components` group. If `--component` is not specified, all the components not flagged as `EXCLUDE_FROM_ALL` are installed.

mwestphal: So, once this is merged, someone compiling F3D wanting to install all components of F3D has no way to know the list of component to potentially install apart reading the CMakeLists.txt ?

Meakk: Currently, if the user specify nothing (ie `cmake --install .`), everything is installed, except `mimetypes` component.  
If he wants to install it, then yes, the user has to know the name of the component (if there is a way to get the components name, I'm not aware of it).

mwestphal: Then all of these optional component should be at least documented and packagers will need to be informed. I'm afraid that some package will miss the mimetypes in the next release.

mwestphal: Not a big fan but I see the point of it.

Meakk: I find it much cleaner than a CMake configuration-time option (which is not possible in case anyway)  
I'd rather exclude all optional components from `ALL`, and let the packager install what he wants.

mwestphal: Yes I agree it is better, I'd just hope to have a way to make sure nobody miss out on a feature because of it.

mwestphal (approved):

PR #559, opened 2023-01-07
Build VTK shared on all linux and macos

mwestphal:
- VTK build shared libs on all linux
- VTK build shared libs on macos
- Small fixes for macos compat with build shared libs
- Set Windows PATH for upcoming shared libs (tested here: #562)
- Set LD_LIBRARY_PATH for shared libs on linux

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,11 +79,6 @@ jobs:
          mkdir build
          mkdir install
 
-    - name: Set LD_LIBRARY_PATH ubuntu
-      working-directory: ${{github.workspace}}/dependencies/
-      shell: bash
-      run: echo "LD_LIBRARY_PATH=$(pwd)/install/lib:$LD_LIBRARY_PATH" >> $GITHUB_ENV
-
     - name: Set PATH windows
       if: matrix.os == 'windows-latest'
       working-directory: ${{github.workspace}}/dependencies/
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -305,6 +312,11 @@ jobs:
       with:
         vtk_sha_file: ./source/.github/actions/vtk_commit_sha
 
+    - name: Set LD_LIBRARY_PATH ubuntu
```
Meakk, line 315: Missing `if os == ubuntu`?

Meakk: Nothing is needed for macOS?

mwestphal: Added more info to explain why it is needed.

This is a linux only and vtk < 9.2.0 only issue that requires this.

All the rest (macos and other VTK versions) rely fully on rpath to find the libs.

I have no choice because macos image have the security (SIP) enabled and prevent the use of DYLD_LIBRARY_PATH through multiple steps.

See this: 
 - https://briandfoy.github.io/macos-s-system-integrity-protection-sanitizes-your-environment/
 - https://github.com/conan-io/conan-center-index/issues/8443
 - https://github.com/pypa/cibuildwheel/issues/816
 
 An alternative would be to use DYLD_LIBRARY_PATH whenever needed, but its not as elegant and it pushes me to actually go and fix rpath issues in our deps.

Meakk (approved):

Issue #472, opened 2022-11-04
Improve reader description


Meakk:
**Describe the bug**

The reader description is shown in the column "type" of Windows explorer.  
We should put a better description.  

![image](https://user-images.githubusercontent.com/6069752/200028185-2a54da9a-3f9f-4762-b4f9-54a6c137af90.png)

In the case above, it should be "GL Transmission format" and not "glTF files reader".

PR #566, opened 2023-01-15
Improve reader description

Meakk:
Also split OCCT and Assimp readers in a single reader per file type.
Mimetypes for OCCT have been improved too.
Finally, `readers-list` option have been reworked to include a plugin name column.

Fix https://github.com/f3d-app/f3d/issues/472


cmake/f3dPlugin.cmake
```diff
@@ -46,7 +46,7 @@ The `NAME` argument is required. The arguments are as follows:
   * `NAME`: (Required) The name of the reader.
   * `VTK_IMPORTER`: The VTK importer class to use.
   * `VTK_READER`: The VTK reader class to use.
-  * `DESCRIPTION`: The description of the reader.
+  * `FORMAT_DESCRIPTION`: The description of the reader.
```
mwestphal, line 49: The description of the format read by the reader ?

plugins/native/CMakeLists.txt
```diff
@@ -119,7 +119,7 @@ f3d_plugin_declare_reader(
   EXTENSIONS vtp vtu vtr vti vts vtm
   MIMETYPES application/vnd.vtp application/vnd.vtu application/vnd.vtr application/vnd.vti application/vnd.vts application/vnd.vtm
```
mwestphal, line 120: should this be split ?

plugins/occt/CMakeLists.txt
```diff
@@ -23,12 +23,21 @@ endif()
 f3d_plugin_init()
 
 f3d_plugin_declare_reader(
-  NAME OCCT
-  EXTENSIONS stp step igs iges
-  MIMETYPES model/iges application/vnd.step
+  NAME STEP
+  EXTENSIONS stp step stpnc p21 210
```
mwestphal, line 27: is that the exact same format as step ?

plugins/occt/f3d-occt-formats.xml
```diff
@@ -1,8 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <mime-info xmlns="http://www.freedesktop.org/standards/shared-mime-info">
-  <mime-type type="application/vnd.step">
+  <mime-type type="model/step">
```
mwestphal, line 3: While step is part of iana: https://www.iana.org/assignments/media-types/model/step, it is not present in freedesktop yet https://gitlab.freedesktop.org/xdg/shared-mime-info

We should open an issue so it can be added. I did that successfully already for certain formats.

plugins/occt/f3d-occt-formats.xml
```diff
@@ -1,8 +1,16 @@
...
     <comment>CAD STEP exchange ISO file</comment>
     <glob pattern="*.step"/>
     <glob pattern="*.stp"/>
+    <glob pattern="*.stpnc"/>
+    <glob pattern="*.p21"/>
+    <glob pattern="*.210"/>
+  </mime-type>
+  <mime-type type="model/iges">
```
mwestphal, line 11: already present in shared-mime-info

plugins/occt/CMakeLists.txt
Meakk: According to the wikipedia and iana page, yes

plugins/native/CMakeLists.txt
Meakk: I have split them, but I use `vtkXMLGenericDataObjectReader` in each reader. Do you think it's fine or should I specialize it? (e.g. using `vtkXMLPolyDataReader` for vtp files)

Meakk: The goal of splitting them is to have a different format description

plugins/occt/CMakeLists.txt
```diff
@@ -23,12 +23,21 @@ endif()
...
+  MIMETYPES model/step
```
mwestphal, line 28: this mimetypes does not exists in freedesktop

mwestphal: should be vnd.step

mwestphal (approved):

PR #568, opened 2023-01-17
Add Examples

mwestphal:
- Fix an issue with PUBLIC link in public
- Fix a potential segfault with an invalid file
- Add examples
- Use examples instead of test_sdk to validate install


mwestphal:
data issue resolved, needs macos fix and rendering fix to go forward

.github/workflows/ci.yml
```diff
@@ -193,25 +194,36 @@ jobs:
...
+      if: matrix.os == 'windows-latest'
+      uses: ./source/.github/actions/mesa-install
+      with:
+        path: ${{github.workspace}}\build_examples\render-image\Release\
+
+    - name: Install Mesa Windows Examples2
+      if: matrix.os == 'windows-latest'
+      uses: ./source/.github/actions/mesa-install
+      with:
+        path: ${{github.workspace}}\build_examples\render-interact\Release\
```
mwestphal, line 213: I could not find a better way to fix this

.github/workflows/ci.yml
```diff
@@ -129,7 +130,7 @@ jobs:
         -DF3D_BINDINGS_JAVA=ON
         -DF3D_INSTALL_SDK=ON
         -DF3D_INSTALL_SDK_PLUGIN=ON
-        -DCMAKE_BUILD_TYPE=Release
+        -DCMAKE_BUILD_TYPE=Debug
```
Meakk, line 133: why this change?

.github/workflows/ci.yml
```diff
@@ -193,25 +194,38 @@ jobs:
...
+        path: ${{github.workspace}}\build_examples\render-interact\Release\
+
+    # older VTK version create different renderings so the are not tested
+    - name: Test libf3d examples
+      if: matrix.vtk_version != 'v9.0.0' && matrix.vtk_version != 'v9.1.0'
+      working-directory: ${{github.workspace}}
+      run: ctest --test-dir build_examples -C Release -VV
+
+    # Windows do not support shared VTK in CI yet
+    - name: Validate plugin exaples
```
Meakk, line 222:
```suggestion
    - name: Validate plugin examples
```

cmake/f3dPlugin.cmake
```diff
@@ -243,7 +243,7 @@ macro(f3d_plugin_build)
 
   list(TRANSFORM F3D_PLUGIN_VTK_MODULES PREPEND "VTK::")
 
-  target_link_libraries(f3d-plugin-${F3D_PLUGIN_NAME} PUBLIC
+  target_link_libraries(f3d-plugin-${F3D_PLUGIN_NAME} PRIVATE
```
Meakk, line 246: Why is it needed?

examples/README.md
```diff
@@ -1,3 +1,23 @@
...
+
+Examples plugins for the libf3d, which requires the `F3D_INSTALL_PLUGIN_SDK` cmake option when building F3D.
+
+## Building
+
+Each example is intended to be built as a separate CMake project, however top level `CMakeLists.txt` are provided if needed.
+
+```
+cmake -Df3d_DIR=/path/to/f3d/install/lib/cmake/f3d /path/to/example/dir
+cmake --build
```
Meakk, line 20:
```suggestion
cmake --build .
```

examples/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+project(examples)
+
+cmake_minimum_required(VERSION 3.2)
```
Meakk, line 3: Are we sure about this version?

examples/plugins/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+project(libf3d-examples)
+
+cmake_minimum_required(VERSION 3.2)
+
+find_package(f3d REQUIRED)
```
Meakk, line 5: Check for `f3d_SDK_PLUGIN`

Meakk (changes requested):

.github/workflows/ci.yml
mwestphal: should have been removed, nice find

cmake/f3dPlugin.cmake
mwestphal: actually not needed, but you mentionned that it was a mistake for it to be public

examples/CMakeLists.txt
mwestphal: Not at all, do you want to find out the min version ?

cmake/f3dPlugin.cmake
Meakk: It does not make any difference here I think since there is no linkage to plugin libraries.

examples/CMakeLists.txt
Meakk: Not necessary in this PR, but it would be great to find out what version we are really using.

cmake/f3dPlugin.cmake
mwestphal: since there is no reason for it to be pulic, lets keep it private ?

examples/CMakeLists.txt
mwestphal: done in this PR, libf3d is cmake 3.1 (or earlier, did not check cmake 2), plugins and F3D are 3.13. (` Unknown CMake command "target_link_options".`

Issue #456, opened 2022-10-18
Misplaced cmake lines related to the python binding


yurivict:
[This line](https://github.com/f3d-app/f3d/blob/master/CMakeLists.txt#L195) really belongs [here](https://github.com/f3d-app/f3d/blob/master/python/CMakeLists.txt#L1).

And [this line](https://github.com/f3d-app/f3d/blob/master/cmake/installing.cmake#L38) belongs [here](https://github.com/f3d-app/f3d/blob/master/python/CMakeLists.txt#L27).

This way one could build the project in ```python/``` separately. But now these relevant lines are in other files.

mwestphal:
What do you mean by building the project separately ?

Meakk:
I suppose the idea is to split f3d in different packages (libf3d, f3d, pyf3d...)  
We should think about it indeed.  
It's also something I have in mind with the plugin architecture I'm currently implementing (occt/assimp/alembic could be a dependency of the plugin packages)

yurivict:
> I suppose the idea is to split f3d in different packages (libf3d, f3d, pyf3d...)

Yes. I am going to add the FreeBSD port for f3d Python binding, and have encountered this problem with misplaced lines.

Meakk:
There is also a dependency on `libf3d` target though, so we also maybe need to add something like:
```cmake
if (NOT TARGET libf3d)
  find_package(f3d REQUIRED)
  add_library(libf3d ALIAS f3d::libf3d)
endif()
```

mwestphal:
> I suppose the idea is to split f3d in different packages (libf3d, f3d, pyf3d...)

That would be nice, but I'm not sure how @yurivict suggestion address this.

yurivict:
> That would be nice, but I'm not sure how @yurivict suggestion address this.


This is the movement in the same direction.
The inner CMakeLists.txt can be a complete, separately buildable cmake file.

For example, the  [Faiss](https://github.com/facebookresearch/faiss/blob/main/CMakeLists.txt#L37) project does it this way.

mwestphal:
Great, do you think you could provide a PR for this change ? @yurivict

PR #569, opened 2023-01-17
Allow external python bindings

Meakk:
Fix https://github.com/f3d-app/f3d/issues/456

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -1,9 +1,20 @@
+cmake_minimum_required(VERSION 3.12)
+
+project(pyf3d)
+
+find_package(Python 3.6 COMPONENTS Interpreter Development)
 find_package(pybind11 2.2 REQUIRED)
-message(STATUS "Bindings: Python: pybind11 ${pybind11_VERSION} found")
 
 pybind11_add_module(pyf3d F3DPythonBindings.cxx)
 
-target_link_libraries(pyf3d PRIVATE libf3d)
+# In case pyf3d is built separately, we need to retrieve the existing libf3d
+if(TARGET libf3d)
+  target_link_libraries(pyf3d PRIVATE libf3d)
+else()
+  find_package(f3d REQUIRED)
+  target_link_libraries(pyf3d PRIVATE f3d::libf3d)
+  target_include_directories(pyf3d PRIVATE "${f3d_INCLUDE_DIR}/f3d")
+endif()
 
 set_target_properties(pyf3d PROPERTIES
   LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib"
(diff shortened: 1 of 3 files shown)


mwestphal:
Nice ! I guees we need the same for the libf3d, dont we ?

python/CMakeLists.txt
```diff
@@ -1,9 +1,29 @@
...
+project(pyf3d)
+
+find_package(Python 3.6 COMPONENTS Interpreter Development)
+message(STATUS "Bindings: Python: Python ${Python_VERSION} found")
+
 find_package(pybind11 2.2 REQUIRED)
 message(STATUS "Bindings: Python: pybind11 ${pybind11_VERSION} found")
 
+# In case pyf3d is built separately, we need to retrieve the existing libf3d
+if(NOT TARGET libf3d)
```
mwestphal, line 12: invert the testing logic ?

python/CMakeLists.txt
```diff
@@ -1,9 +1,29 @@
...
+  set(_f3d_include_path "${f3d_INCLUDE_DIR}/f3d")
+else()
+  add_library(f3d::libf3d ALIAS libf3d)
+  set(_f3d_include_path "${CMAKE_SOURCE_DIR}/library/public")
+endif()
+
 pybind11_add_module(pyf3d F3DPythonBindings.cxx)
 
-target_link_libraries(pyf3d PRIVATE libf3d)
+target_include_directories(pyf3d PRIVATE "${_f3d_include_path}" )
```
mwestphal, line 24: this was not needed before strangely

Meakk: It's not strictly needed for local build, because libf3d target has public includes to `library/public`, but for external build the target has public includes to `/usr/include` and the headers are in `/usr/include/f3d`.  
Alternatively, we could move all headers to `library/public/f3d` and use `#include <f3d/xxx.h>` everywhere.

mwestphal: as you see fit. ok as is.

Meakk:
> Nice ! I guees we need the same for the libf3d, dont we ?

What do you mean?

python/CMakeLists.txt
```diff
@@ -1,9 +1,20 @@
+cmake_minimum_required(VERSION 3.12)
+
+project(pyf3d)
```
mwestphal, line 3: >>    Nice ! I guees we need the same for the libf3d, dont we ?

> What do you mean?

Do we want the same for the libf3d ?

mwestphal: this is what I mean. Do we want to be able to build the libf3d independantly ?

mwestphal (approved):

python/CMakeLists.txt
Meakk: Let's discuss this offline

PR #577, opened 2023-01-26
rework cmake logic

mwestphal:
- Remove packaging
- Move installation to their respective correct locations
- Move testing to application


CMakeLists.txt
```diff
@@ -180,7 +175,15 @@ if(UNIX AND NOT APPLE)
...
 
-include(cmake/installing.cmake)
+# Installing licenses
+set(F3D_LIC_DIR ".")
+
+if (UNIX AND NOT APPLE AND NOT ANDROID)
+  set(F3D_LIC_DIR ${CMAKE_INSTALL_DOCDIR})
+endif()
+
+install(FILES LICENSE.md doc/THIRD_PARTY_LICENSES.md
```
Meakk, line 185: It would be better to put licenses files where they are used (e.g. assimp in assimp plugin) so the licenses are installed only if it is used (ie when assimp plugin is enabled)

library/CMakeLists.txt
```diff
@@ -173,7 +173,52 @@ if (APPLE OR UNIX)
...
 endif()
+
+# Installing
+if(F3D_INSTALL_SDK OR BUILD_SHARED_LIBS)
+  install(TARGETS libf3d
+    EXPORT f3dTargets
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT library
+    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT library
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT library
+    PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/${CMAKE_PROJECT_NAME} COMPONENT headers) #TODO check that
```
Meakk, line 188: Todo?

Meakk (commented):
Great! Much cleaner now.

CMakeLists.txt
mwestphal: assimp license is not in THIRD_PARTY_LICENSES.md anymore.
This project do not distribute assimp, the superbuild-does and assimp license is distributed in the package.

THIRD_PARTY_LICENSES.md contains only jsoncpp and cxxopts (check the diff), which could arguably be put in the application folder as only application rely on them. Do you think this is worth it ?

library/CMakeLists.txt
mwestphal: useless, removed.

mwestphal: turns out it is needed for the SDK, and puttin in SDK specific code is more complex.

CMakeLists.txt
```diff
@@ -182,8 +182,13 @@ if (UNIX AND NOT APPLE AND NOT ANDROID)
   set(F3D_LIC_DIR ${CMAKE_INSTALL_DOCDIR})
 endif()
 
-install(FILES LICENSE.md doc/THIRD_PARTY_LICENSES.md
+install(FILES LICENSE.md
   DESTINATION ${F3D_LIC_DIR} COMPONENT licenses)
 
+if (F3D_BUILD_APPLICATION)
+  install(FILES doc/THIRD_PARTY_LICENSES.md
```
Meakk, line 189: We are using third party libraries even if the application is not built

mwestphal: discussed offline, not needed.

Meakk (approved):

PR #581, opened 2023-01-28
Enable raytracing windows in CI

mwestphal:


.github/workflows/ci.yml
```diff
@@ -67,7 +67,7 @@ jobs:
 
     - name: Install Raytracing Dependencies
       if: matrix.enable_raytracing == 'true'
-      uses: ./source/.github/actions/ospray-sb-install-dep
+      uses: ./source/.github/actions/ospray-superbuild-install-dep
```
Meakk, line 70: Looks like the file is still named ospray-sb-install-dep

mwestphal: it is not, it si been renamed

Meakk: Indeed. The diff on mobile doesn't show the file name change

Meakk (approved):

PR #587, opened 2023-01-31
Emscripten bindings

Meakk:
Here's a proof of concept of libf3d running in a browser.  

2023-03-16-215338_2595x1466

This PR fixes the compilation issues and set-up a CI to make sure it does not break in the future.  
The complete bindings will be done later when the public API and options are stabilized.  
I tried to create a unit test, but I don't know if we can create a canvas without running a browser, I'll take care of that later.


.github/workflows/webassembly.yml
```diff
@@ -0,0 +1,51 @@
...
+         mkdir install
+
+    - name: Configure
+      working-directory: ${{github.workspace}}/build
+      run: >
+        cmake ../source
+        -DBUILD_SHARED_LIBS=OFF
+        -DCMAKE_INSTALL_PREFIX:PATH=../install
+        -DF3D_STRICT_BUILD=ON
+        -DF3D_PLUGIN_BUILD_EXODUS=OFF
```
mwestphal, line 45: do not work with emscriptem ?

mwestphal (approved):

mwestphal:
Approved but CI is not green yet.

.github/workflows/webassembly.yml
Meakk: nop, HDF5 is not building with emscripten

Issue #605, opened 2023-02-05
Assimp lights corrupts vtkPolyDataMapper


mwestphal:
**Describe the bug**
Whenever a FBX or DAE file (assimp plugin) is read, if it contains a light of any kind, other full scene files not containing such light cannot be rendered at all and make shader fails

**To Reproduce**
 0. download these three files  in [segfault.zip](https://github.com/f3d-app/f3d/files/10611189/segfault.zip) , cubeLight.fbx contains a cube with a sun light,  cube.fbx and cube.glb same without any light
 1. f3d cube*
 2. press left two times
 3. fails with:
```
ERROR: In vtkShaderProgram.cxx, line 438
vtkShaderProgram (0x5646278bb870): 1: #version 150
2: #ifndef GL_ES
3: #define highp
4: #define mediump
5: #define lowp
6: #endif // GL_ES
7: #define attribute in
8: #define varying out
9: 
10: 
11: /*=========================================================================
12: 
13:   Program:   Visualization Toolkit
14:   Module:    vtkPolyDataVS.glsl
15: 
16:   Copyright (c) Ken Martin, Will Schroeder, Bill Lorensen
17:   All rights reserved.
18:   See Copyright.txt or http://www.kitware.com/Copyright.htm for details.
19: 
20:      This software is distributed WITHOUT ANY WARRANTY; without even
21:      the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
22:      PURPOSE.  See the above copyright notice for more information.
23: 
24: =========================================================================*/
25: 
26: in vec4 vertexMC;
27: 
28: 
29: 
30: 
31: // frag position in VC
32: //VTK::PositionVC::Dec
33: 
34: // optional normal declaration
35: //VTK::Normal::Dec
36: 
37: // extra lighting parameters
38: //VTK::Light::Dec
39: 
40: // Texture coordinates
41: //VTK::TCoord::Dec
42: 
43: // material property values
44: //VTK::Color::Dec
45: 
46: // clipping plane vars
47: //VTK::Clip::Dec
48: 
49: // camera and actor matrix values
50: uniform mat4 MCDCMatrix;
51: 
52: // Apple Bug
53: //VTK::PrimID::Dec
54: 
55: // Value raster
56: //VTK::ValuePass::Dec
57: 
58: // picking support
59: //VTK::Picking::Dec
60: 
61: void main()
62: {
63:   
64: 
65:   //VTK::Color::Impl
66: 
[...]

PR #625, opened 2023-02-11
Remove a wrong call that disable lights on F3D renderer

mwestphal:
Fix https://github.com/f3d-app/f3d/issues/605

### DIFF
--- a/plugins/assimp/module/vtkF3DAssimpImporter.cxx
+++ b/plugins/assimp/module/vtkF3DAssimpImporter.cxx
@@ -88,8 +88,6 @@ public:
 
       if (nbLights > 0)
       {
-        renderer->AutomaticLightCreationOff();
-
         for (int i = 0; i < nbLights; i++)
         {
           aiLight* aLight = this->Scene->mLights[i];


plugins/assimp/module/vtkF3DAssimpImporter.cxx
```diff
@@ -88,8 +88,6 @@ class vtkF3DAssimpImporter::vtkInternals
 
       if (nbLights > 0)
       {
-        renderer->AutomaticLightCreationOff();
```
Meakk, line 91: I remembered doing that on purpose. The idea is to remove VTK lights when there is at least one light provided in the file.

mwestphal: not needed:

https://github.com/Kitware/VTK/blob/master/Rendering/OpenGL2/vtkOpenGLRenderer.cxx#L141

mwestphal: (and light test would fail if that were not true.)

mwestphal: (and the renderer parameters should not be changed by the importer anyway)

Meakk (approved):

PR #632, opened 2023-02-16
Update logo.ico

### LINKED ISSUE #631: [Windows] Desktop shortcut icon is not hidpi
**Describe the bug**
Looks like modern applications have bigger and cleaner icons (see icons on the left vs icons on the right)

image

Meakk:
Includes size 16,24,32,48 and 256.
Generated by:

```bash
convert -background transparent -define 'icon:auto-resize=16,24,32,48,256' logo.svg logo.ico
```

Fix #631 

See https://learn.microsoft.com/en-us/windows/win32/uxguide/vis-icons?redirectedfrom=MSDN#size-requirements


mwestphal (commented):
THe new ico has a smaller size, is that expected ?

mwestphal:
To test you can build the superbuild with your branch of F3D.

Meakk:
> THe new ico has a smaller size, is that expected ?

Not sure, but there is no 64 and 128 size, at the cost of an additional 256 size. That might be the reason.

mwestphal (approved):

Issue #597, opened 2023-02-01
"multi-file" file loading API could be moved to F3D application


mwestphal:
The multi-file part of the loader API is quite specific to F3D application so it could make sens to move it there and provide a simpler API in libf3d. eg:

The loading API could be reduced to:

- loadModel(filename, options) clear scene, load a model from disk
- loadModel(buffer, options) clear scene, load a model from memory

mwestphal:
I've been thinking about this and starting to like the idea, especially in regards to https://github.com/f3d-app/f3d/issues/127.

Here is what we could do:

- drop `geometry-only` f3d::options
- modify `f3d::loader` API like this:

```
loader::loadGeometry(std::string file, bool resetScene)
loader::loadScene(std::string file)
```

- implement the `resetScene == false` case in the `vtkF3DGenericImporter` by using multiple reader when needed in the default scene

- Move multi files management to F3D Application and use `loadGeometry/loadScene` accordingly to `--geometry-only` option

- Add a `--multi-files-mode` command line option (global options only) with following possible parameters:

1. switch: current behavior, when loading multiples files, one can switch between files using the arrow keys. default.
2. load: load all provided files as multiple models in the same default scene.
3. folder: show all files in a folder view (cf https://user-images.githubusercontent.com/3129530/145460406-bc9b190a-ad7b-4b45-80d3-95cd01f49bcc.png) will not be implemented right away.

wdyt @snoyer @Meakk @nermolov @SamuelTallet ?

mwestphal:
FYI @GitCodeBoy

SamuelTallet:
"Multi files mode" option is a good idea! 🙂

snoyer:
>     * Add a `--multi-files-mode` command line option (global options only) with following possible parameters:
> 
> 
>     1. switch: current behavior, when loading multiples files, one can switch between files using the arrow keys. default.
> 
>     2. load: load all provided files as multiple models in the same default scene.
> 
>     3. folder: show all files in a folder view (cf https://user-images.githubusercontent.com/3129530/145460406-bc9b190a-ad7b-4b45-80d3-95cd01f49bcc.png) will not be implemented right away.

These points all feel orthogonal to one another.

The "folder view" sounds like an alternative way to render/navigate, independently of the multi-file stuff (ie. the folder view could be implemented with the current file loading as an alternative to left/right arrows if I'm undertanding correctly).

I think the general multi-file case should be somehow allowing to group filenames instead of treating them individually.
[...]

Issue #417, opened 2022-09-08
make Loader class able to remove file


GitCodeBoy:
**Is your feature request related to a problem? Please describe.**
I want to edit the file list of loader when I use f3d python binding，but I can only find "addFile" and "loadFile" in f3d.loader class,I cant find  a way to edit or remove a file from loader

mwestphal:
Good point, should be easy to add a method to remove the current file.

mwestphal:
to put in context of #597

mwestphal:
Will be fixed by https://github.com/f3d-app/f3d/pull/634

PR #634, opened 2023-02-16
Rework Loader API

mwestphal:
- [x] Move multifile logic to application
- [x] Rework multifile logic under F3D application constraints
- [x] Rework loader API to support multiple geometry in default scene
- [x] Implement multigeometry support
- [x] Update doc
- [x] Add --group-geometries options
- [x] Add testing
- [x] Add examples

Fix: https://github.com/f3d-app/f3d/issues/597
Fix: https://github.com/f3d-app/f3d/issues/417

New Loader API looks like this:

```
  /** 
   * Load a geometry from a provided file to the scene.
   * Reset the scene before loading if a full scene was loaded previously or if reset is set to false,
   * do not reset if only loaded geometries previously.
   * Geometries loader using this method will be available in a default scene and use all default
   * scene related options.
   * Throw a load_failure_exception on failure.
   */
  virtual loader& loadGeometry(const std::string& filePath, bool reset = false) = 0;

  /** 
   * Reset scene and load a full scene from provided file.
   * Please note default scene related options are not taken into account when loading a full scene.
   * Throw a load_failure_exception on failure.
   */
  virtual loader& loadScene(const std::string& filePath) = 0;

  /** 
   * Return true if the loader has a scene reader for the providen file, false otherwise.
   */
  virtual bool hasSceneReader(const std::string& filePath) = 0;

  /** 
   * Return true if the loader has a geometry reader for the providen file, false otherwise.
   */
  virtual bool hasGeometryReader(const std::string& filePath) = 0;
```

A new  command line option has been added to F3D:

`"group-geometries", "", "When opening multiple files, show them all in the same scene. Force geometry-only. The configuration file for the first file will be loaded."`


application/F3DStarter.cxx
```diff
@@ -330,10 +332,33 @@ void F3DStarter::LoadFile(f3d::loader::LoadFileEnum load)
...
+  int size = static_cast<int>(this->Internals->FilesList.size());
+  if (size > 0)
+  {
+    if (load == LoadFileEnum::LOAD_LAST)
+    {
+      this->Internals->CurrentFileIndex = size - 1;
+    }
+    else
+    {
+      this->Internals->CurrentFileIndex += load;
```
snoyer, line 346: Relying on the underlying value of the enum feels (is?) sketchy.

application/F3DStarter.cxx
```diff
@@ -330,10 +332,33 @@ void F3DStarter::LoadFile(f3d::loader::LoadFileEnum load)
...
   std::string filePath, fileName, fileInfo;
-  this->Internals->Engine->getLoader().getFileInfo(load, index, filePath, fileName, fileInfo);
+  int size = static_cast<int>(this->Internals->FilesList.size());
+  if (size > 0)
+  {
+    if (load == LoadFileEnum::LOAD_LAST)
+    {
+      this->Internals->CurrentFileIndex = size - 1;
+    }
+    else
```
snoyer, line 344: no test for `LOAD_FIRST`?

application/F3DStarter.cxx
```diff
@@ -330,10 +332,33 @@ void F3DStarter::LoadFile(f3d::loader::LoadFileEnum load)
...
+    {
+      this->Internals->CurrentFileIndex += load;
+      this->Internals->CurrentFileIndex %= size;
+      this->Internals->CurrentFileIndex = this->Internals->CurrentFileIndex < 0
+        ? this->Internals->CurrentFileIndex + size
+        : this->Internals->CurrentFileIndex;
```
snoyer, line 350: Would it be worth defining a proper modulo function? to do something like:
```cpp
switch(load){
  case LOAD_FIRST: i = 0; break
  case LOAD_LAST: i = size-1; break
  case LOAD_NEXT: i = modulo(i+1, size); break
  case LOAD_PREV: i = modulo(i-1, size); break
}```

**Review by snoyer — COMMENTED** [review]
A few comments/questions on the `LoadFileEnum` usage.

Alternatively would it be simpler to drop the enum altogether and have something like `F3DStarter::LoadFile(int index, bool relative_index=false)` that would do `new_index = modulo(relative_index ? old_index+index : index, size)` instead? That would allow to index over the whole range from either end by default (`f(0)`: first, `f(1)`: second, ..., `f(-2)`: second to last, `f(-1)`: last), and do previous/current/next using the second parameter (`f(-1, true)`/`f(0, true)`/`f(1, true)` respectively)

File: application/F3DStarter.cxx (hunk 3 above)
Comment (mwestphal) [reply application/F3DStarter.cxx:350]: Definitely, for now I'm just moving methods around before the actual redesign

File: application/F3DStarter.cxx (hunk 2 above)
Comment (mwestphal) [reply application/F3DStarter.cxx:344]: Load first us gone

**Comment by mwestphal** [comment]
> Alternatively would it be simpler to drop the enum altogether ...

TBH it doesnt realy matter is this is part of the F3D application, so not exposed to the libf3d users. Of course I can still improve the API but we should implement just what we need.

File: application/F3DStarter.cxx (hunk 3 above)
Comment (mwestphal) [reply application/F3DStarter.cxx:350]: reworked the whole thing

File: examples/libf3d/render-interact/main.cxx [hunk 4]
```diff
@@ -23,7 +23,7 @@ int main(int argc, char** argv)
 
   // Load a model
   f3d::loader& load = eng.getLoader();
-  load.addFile(std::string(argv[1])).loadFile();
+  load.loadFile(std::string(argv[1]));
```
mwestphal, line 26: incorrect

examples/libf3d/render-image/main.cxx
```diff
@@ -23,7 +23,7 @@ int main(int argc, char** argv)
 
     // Load a model
     f3d::loader& load = eng.getLoader();
-    load.addFile(std::string(argv[1])).loadFile();
+    load.loadFile(std::string(argv[1]));
```
mwestphal, line 26: incorrect

mwestphal:
@Meakk @snoyer : this is starting to take actual form. Most features have been implemented. Please review, especially the new loader API.

doc/user/OPTIONS.md
```diff
@@ -69,7 +69,7 @@ Options|Default|Description
 
 Options|Default|Description
 ------|------|------
--s, \-\-scalars=\<array_name\>||Specify an array to *Color* with if present in the file. If no array_name is provided, one will be picked if any are available. <br>Requires a default scene.<br>Use \-\-verbose to recover the usable array names.
+-s, \-\-scalars=\<array_name\>||Specify an array to *Color* with if present in the file. If no array_name is provided, the first in alphabital order will be picked if any are available. <br>Requires a default scene.<br>Use \-\-verbose to recover the usable array names.
```
mwestphal, line 72: typo

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.h
```diff
@@ -166,7 +100,7 @@ class vtkF3DRendererWithColoring : public vtkF3DRenderer
 
   ///@{
   /**
-   * Get current coloring infomatiuons,
+   * Get current coloring infomations,
```
mwestphal, line 103: typo

library/public/engine.h
```diff
@@ -27,8 +27,7 @@ namespace f3d
  * \code{.cpp}
  *  f3d::engine eng(f3d::engine::CREATE_WINDOW | f3d::engine::CREATE_INTERACTOR);
  *  f3d::loader& load = eng.getLoader();
- *  load.addFile("path/to/file");
- *  load.loadFile(f3d::loader::LoadFileEnum::LOAD_CURRENT);
+ *  load.loadFile("path/to/file");
```
mwestphal, line 30: to update

library/public/engine.h
```diff
@@ -153,6 +152,8 @@ class F3D_EXPORT engine
     std::vector<std::string> Extensions;
     std::vector<std::string> MimeTypes;
     std::string PluginName;
+    bool HasSceneReader;
+    bool HasGeometryReader;
```
mwestphal, line 156: not set ?

library/public/loader.h
```diff
@@ -3,6 +3,7 @@
 
 #include "export.h"
 
+#include <functional>
```
mwestphal, line 6: not needed ?

library/public/loader.h
```diff
@@ -13,75 +14,25 @@ namespace f3d
...
-   * Get information about the next file to load according to the load param.
-   * Set the load argument to LOAD_FIRST, LOAD_PREVIOUS, LOAD_NEXT or LOAD_LAST as needed.
-   * nextFileIndex will provide the index of the file or -1 if not available.
-   * filePath the path to the file and fileInfo a more complete information about the file,
-   * including index in list.
-   */
-  virtual void getFileInfo(LoadFileEnum load, int& nextFileIndex, std::string& filePath,
-    std::string& fileName, std::string& fileInfo) const = 0;
+  virtual bool canReadScene(const std::string& filePath) = 0;
+  virtual bool canReadGeometry(const std::string& filePath) = 0;
```
mwestphal, line 35: see the new API above @snoyer @Meakk

library/src/interactor_impl.cxx
```diff
@@ -232,28 +232,28 @@ class interactor_impl::internals
...
+                  self->Loader.loadFile(load);
+                  render = true;
+                }
+                else if (keySym == "Up")
+                {
+                  self->AnimationManager.StopAnimation();
+                  loader::LoadFileEnum load = loader::LoadFileEnum::LOAD_CURRENT;
+                  self->Loader.loadFile(load);
+                  render = true;
+                }*/
```
mwestphal, line 255: todo

library/VTKExtensions/Readers/vtkF3DGenericImporter.cxx
```diff
@@ -116,138 +222,110 @@ bool vtkF3DGenericImporter::GetTemporalInformation(
...
-  this->GeometryActor->GetProperty()->SetEmissiveTexture(
-    this->GetTexture(this->TextureEmissive, true));
-  this->GeometryActor->GetProperty()->SetEmissiveFactor(this->EmissiveFactor);
-  this->GeometryActor->GetProperty()->SetNormalTexture(this->GetTexture(this->TextureNormal));
-  this->GeometryActor->GetProperty()->SetNormalScale(this->NormalScale);
-
-  // If the input texture is RGBA, flag the actor as translucent
-  if (colorTex && colorTex->GetImageDataInput(0)->GetNumberOfScalarComponents() == 4)
+  // Update each reader
+  for (size_t readerIndex = 0; readerIndex < this->Internals->Readers.size(); readerIndex++)
```
Meakk, line 228: Is it streamed or the actors are visible only when everything is loaded?

library/VTKExtensions/Readers/vtkF3DGenericImporter.cxx
```diff
@@ -331,20 +433,215 @@ std::string vtkF3DGenericImporter::GetDataObjectDescription(vtkDataObject* objec
...
+      }
+    }
+  }
+
+  auto& data = useCellData ? this->Internals->CellDataArrayVectorForColoring
+                           : this->Internals->PointDataArrayVectorForColoring;
+  data.clear();
+
+  // Create a vector of arrays by name
+  for (std::string arrayName : arrayNames)
```
Meakk, line 542:
```suggestion
  for (const std::string& arrayName : arrayNames)
```

Meakk (commented):
Here's a first review. I have not finished yet, I'll continue tomorrow, but you can start addressing my comments.

application/F3DStarter.cxx
```diff
@@ -11,7 +11,21 @@
...
+
+namespace fs = std::filesystem;
+
+namespace
+{
+long long fileModulo(long long index, long long size)
+{
+  index %= size;
+  index = index < 0 ? index + size : index;
+  return index;
```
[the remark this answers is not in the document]
snoyer: @Meakk, `(index + size) % size` would most probably work in the app as we're unlikely to ever decrement to lower than `-size` but technically it could still return negative values. Wouldn't it be worth it to keep the true modulo for correctness/safety?

application/F3DStarter.cxx
```diff
@@ -53,6 +67,8 @@ class F3DStarter::F3DInternals
   f3d::options DynamicOptions;
   f3d::options FileOptions;
   std::unique_ptr<f3d::engine> Engine;
+  std::vector<fs::path> FilesList;
+  long long CurrentFileIndex = -1;
```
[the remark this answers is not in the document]
snoyer: Obviously anything other than `std::optional<std::vector<fs::path>::size_type>` would be wrong and unacceptable ;)

what about `size_t` and have a separate variable to flag whether a file is loaded/active instead of using `-1`? (or keep using `-1`, it's good enough for `std::string::` with `npos`)

library/VTKExtensions/Readers/vtkF3DGenericImporter.cxx
mwestphal: Not streamed. Loading all is needed however if a reader has already read the file, it will not read again.

application/F3DStarter.cxx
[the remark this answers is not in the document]
mwestphal: Well, my code has the same issue anyway.

application/F3DStarter.cxx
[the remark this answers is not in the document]
mwestphal: just using an `int`  is fine I guess.

application/F3DStarter.cxx
```diff
@@ -11,7 +11,21 @@
 #include "options.h"
 #include "window.h"
 
+#include <cassert>
 #include <filesystem>
+#include <set>
+
+namespace fs = std::filesystem;
```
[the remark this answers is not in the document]
mwestphal: is there an actual difference ?

library/VTKExtensions/Readers/vtkF3DGenericImporter.cxx
```diff
@@ -37,29 +38,130 @@
 
 #include <sstream>
 
+namespace
+{
+//----------------------------------------------------------------------------
+// TODO : add this function in a utils file for rendering in VTK directly
+vtkSmartPointer<vtkTexture> GetTexture(const std::string& filePath, bool isSRGB = false)
```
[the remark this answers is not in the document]
mwestphal: outside of the scope of this PR though

library/VTKExtensions/Readers/vtkF3DGenericImporter.cxx
```diff
@@ -37,29 +38,130 @@
...
+
+  bool AnimationEnabled = false;
+  std::set<double> TimeSteps;
+};
+
 vtkStandardNewMacro(vtkF3DGenericImporter);
 
+//----------------------------------------------------------------------------
+vtkF3DGenericImporter::vtkF3DGenericImporter()
+  : Internals(new vtkF3DGenericImporter::vtkInternals())
```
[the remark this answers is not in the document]
mwestphal: indeed we should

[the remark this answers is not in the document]
mwestphal: the member and the name of the struct should be different though. Suggestions ?

application/F3DStarter.cxx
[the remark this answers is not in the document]
mwestphal: well, there is. Using do not work like that.
https://en.cppreference.com/w/cpp/language/namespace_alias

library/VTKExtensions/Readers/vtkF3DGenericImporter.cxx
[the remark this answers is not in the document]
Meakk: ok

library/VTKExtensions/Readers/vtkF3DGenericImporter.cxx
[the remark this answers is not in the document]
Meakk: `Internals* Pimpl` or `Internals* Private`?  
Obviously `Internals* mInternals` would be better but against the coding rules.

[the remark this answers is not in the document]
Meakk: You can keep the prefix if your prefer though

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
```diff
@@ -114,17 +108,22 @@ void vtkF3DRendererWithColoring::SetUseInverseOpacityFunction(bool use)
   if (this->UseInverseOpacityFunction != use)
   {
     this->UseInverseOpacityFunction = use;
-    if (this->VolumeProp)
+
+    auto volPropsAndMappers = this->Importer->GetVolumePropsAndMappers();
```
Meakk, line 112: you probably want a reference here: `auto&`. Const-ness is automatically deduced so you don't have to specify it.

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
```diff
@@ -114,17 +108,22 @@ void vtkF3DRendererWithColoring::SetUseInverseOpacityFunction(bool use)
...
+    for (auto volPropAndMapper : volPropsAndMappers)
```
Meakk, line 113: here too

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.h
```diff
@@ -84,78 +73,23 @@ class vtkF3DRendererWithColoring : public vtkF3DRenderer
    */
   void SetColormap(const std::vector<double>& colormap);
 
-  enum CycleTypeEnum
+  enum class CycleType
   {
-    F3D_FIELD_CYCLE = 0,
-    F3D_ARRAY_CYCLE,
-    F3D_COMPONENT_CYCLE
+    F3D_CYCLE_NONE,
```
Meakk, line 78: `F3D_CYCLE_` prefix not needed and too verbose.

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
```diff
@@ -259,114 +315,169 @@ int vtkF3DRendererWithColoring::GetColoringComponent()
...
+
+  if (hasColoring && !this->ColorTransferFunctionConfigured)
+  {
+    this->ConfigureRangeAndCTFForColoring(info);
+    this->ColorTransferFunctionConfigured = true;
+  }
+
+  // Handle surface geometry
+  bool geometriesVisible = this->UseRaytracing || (!this->UseVolume && !this->UsePointSprites);
+  auto actorsAndMappers = this->Importer->GetGeometryActorsAndMappers();
```
Meakk, line 364: reference

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
```diff
@@ -259,114 +315,169 @@ int vtkF3DRendererWithColoring::GetColoringComponent()
...
-      this->ConfigureRangeAndCTFForColoring(this->ArrayForColoring, this->ComponentForColoring);
-      this->ColorTransferFunctionConfigured = true;
+      actorAndMapper.second->ScalarVisibilityOff();
     }
   }
 
-  if (this->GeometryActor)
+  // Handle point sprites
+  bool pointSpritesVisible = !this->UseRaytracing && !this->UseVolume && this->UsePointSprites;
+  auto psActorsAndMappers = this->Importer->GetPointSpritesActorsAndMappers();
```
Meakk, line 393: reference

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
```diff
@@ -259,114 +315,169 @@ int vtkF3DRendererWithColoring::GetColoringComponent()
...
-    if (volumeVisibility && (!mapper || !mapper->GetInput() || !this->ArrayForColoring))
+    auto propAndMapper = volPropsAndMappers[i];
+    vtkDataArray* coloringArray = nullptr;
+    if (hasColoring && info.Arrays.size() > i)
+    {
+      coloringArray = info.Arrays[i];
+    }
+    if (!volumeVisible)
+    {
+      propAndMapper.first->SetVisibility(false);
```
Meakk, line 432: There is no `VisibilityOff()` function?

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
```diff
@@ -463,15 +574,10 @@ void vtkF3DRendererWithColoring::ConfigureVolumeForColoring(vtkSmartVolumeMapper
...
+void vtkF3DRendererWithColoring::ConfigureScalarBarActorForColoring(vtkScalarBarActor* scalarBar,
+  const std::string& arrayName, int component, vtkColorTransferFunction* ctf)
 {
-  if (!array)
-  {
-    return;
-  }
-
-  std::string title = array->GetName();
+  std::string title = arrayName;
```
Meakk, line 580: you can change the function to take a string copy and avoid this local variable.

library/public/engine.h
```diff
@@ -22,16 +22,26 @@ namespace f3d
...
  *
- * Example usage:
+ * Example usage for a default scene file:
  *
  * \code{.cpp}
  *  f3d::engine eng(f3d::engine::CREATE_WINDOW | f3d::engine::CREATE_INTERACTOR);
  *  f3d::loader& load = eng.getLoader();
- *  load.addFile("path/to/file");
- *  load.loadFile(f3d::loader::LoadFileEnum::LOAD_CURRENT);
+ *  load.resetToDefaultScene().addGeometry("path/to/file");
```
Meakk, line 30: `resetToDefaultScene()` is needed even if nothing is added yet?

library/public/loader.h
Meakk: Are these `canReadXXX` useful? Just call `addGeometry` and check the result.  
Alternatively you can keep these function but `addGeometry` and `loadFullScene` should return `loader&`

library/public/engine.h
mwestphal: yes, see the addGeometry comment.

mwestphal:
> Are these canReadXXX useful? Just call addGeometry and check the result.

What if someone just want to know if a file has a geom reader or a full scene reader without actually reading ?

> Alternatively you can keep these function but addGeometry and loadFullScene should return loader&

How do you know if the file was actually read then ?

library/public/loader.h
mwestphal: > Are these canReadXXX useful? Just call addGeometry and check the result.

What if someone just want to know if a file has a geom reader or a full scene reader without actually reading ?

> Alternatively you can keep these function but addGeometry and loadFullScene should return loader&

How do you know if the file was actually read then ?

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
mwestphal: const is needed though.

library/public/loader.h
Meakk: > What if someone just want to know if a file has a geom reader or a full scene reader without actually reading ?

Then why give the full path? The extension is enough?

> How do you know if the file was actually read then ?

Throw if it fails

application/F3DStarter.cxx
```diff
@@ -330,47 +332,171 @@ void F3DStarter::LoadFile(f3d::loader::LoadFileEnum load)
...
+                nGeom++;
+              }
+              else
+              {
+                f3d::log::warn(geomPath, " is not a geometry of a supported file format\n");
+              }
+            }
+            if (nGeom > 0)
+            {
+              filenameInfo = std::to_string(nGeom) + " Geometries loaded";
```
Meakk, line 423:
```suggestion
              filenameInfo = std::to_string(nGeom) + " geometries loaded";
```

application/F3DStarter.cxx
```diff
@@ -330,47 +332,171 @@ void F3DStarter::LoadFile(f3d::loader::LoadFileEnum load)
...
+        else
+        {
+          f3d::log::warn(filePath, " is not a file of a supported file format\n");
+          filenameInfo += " [UNSUPPORTED]";
+        }
+      }
+      catch (const f3d::loader::load_failure_exception& ex)
+      {
+        f3d::log::error("Could not load file:");
+        f3d::log::error(ex.what());
```
Meakk, line 447: can you combine these two lines? Maybe just `f3d::log::error("Could not load file: ", e.what());`

library/testing/TestSDKInteractorDropFullScene.cxx
```diff
@@ -0,0 +1,33 @@
...
+  f3d::window& win = eng.getWindow();
+  f3d::interactor& inter = eng.getInteractor();
+  win.setSize(300, 300);
+
+  std::string filename = "TestSDKInteractorDropFullScene";
+#if defined(_WIN32)
+  std::string interactionFilePath = std::string(argv[2]) + "..\\..\\" + filename + ".log";
+#else
+  std::string interactionFilePath = std::string(argv[2]) + "../../" + filename + ".log";
+#endif
```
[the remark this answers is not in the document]
mwestphal: we could, but this is outside of the scope of this PR. Should I open an issue ?

Meakk (commented):
It's a bit painful to review, but it seems ok to me.

library/testing/TestSDKInteractorDropFullScene.cxx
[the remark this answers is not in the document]
mwestphal: Btw im not sure we want that. On windows, path are windows path arent they ?

application/F3DStarter.cxx
```diff
@@ -330,47 +332,168 @@ void F3DStarter::LoadFile(f3d::loader::LoadFileEnum load)
...
+      // Recursively add all files
+      this->AddFile(entryPath);
+    }
+  }
+  else
+  {
+    auto it =
+      std::find(this->Internals->FilesList.begin(), this->Internals->FilesList.end(), tmpPath);
+
+    if (it == this->Internals->FilesList.end())
```
Meakk, line 494: should we warn the user the file is already added?

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
```diff
@@ -259,114 +318,172 @@ int vtkF3DRendererWithColoring::GetColoringComponent()
...
+  if (hasColoring && !this->ColorTransferFunctionConfigured)
+  {
+    this->ConfigureRangeAndCTFForColoring(info);
+    this->ColorTransferFunctionConfigured = true;
+  }
+
+  // Handle surface geometry
+  bool geometriesVisible = this->UseRaytracing || (!this->UseVolume && !this->UsePointSprites);
+  const auto& actorsAndMappers = this->Importer->GetGeometryActorsAndMappers();
+  for (size_t i = 0; i < actorsAndMappers.size(); i++)
```
Meakk, line 364: Let's try to use structured bindings here:
```cpp
for (auto& [actor, mapper] : actorsAndMappers)
```

Meakk: you'll lose access to `i` though

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
```diff
@@ -486,29 +597,36 @@ void vtkF3DRendererWithColoring::ConfigureScalarBarActorForColoring(
 }
 
 //----------------------------------------------------------------------------
-void vtkF3DRendererWithColoring::ConfigureRangeAndCTFForColoring(vtkDataArray* array, int component)
+void vtkF3DRendererWithColoring::ConfigureRangeAndCTFForColoring(
+  const vtkF3DGenericImporter::ColoringInfo& info)
 {
-  if (!array || component == -2)
+  if (this->ComponentForColoring == -2) // TODO use options
```
Meakk, line 602: todo?

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
```diff
@@ -534,258 +652,112 @@ void vtkF3DRendererWithColoring::ConfigureRangeAndCTFForColoring(vtkDataArray* a
...
-    {
-      this->ComponentForColoring = this->ArrayForColoring->GetNumberOfComponents() - 1;
-    }
+    return;
   }
+
+  // -2 -1 0 1 2 3 4
+  this->ComponentForColoring =
+    (this->ComponentForColoring + 3) % (info.MaximumNumberOfComponents + 2) -
+    2; // TODO separate direct scalars and magnitude from component number
```
Meakk, line 735: todo?

library/public/loader.h
```diff
@@ -12,76 +13,64 @@ namespace f3d
...
+ *  f3d::loader& load = eng.getLoader();
+ *
+ *  if (load.hasSceneReader(path)
+ *  {
+ *    load.loadFullScene(path);
+ *  }
+ *  else if (load.hasFullSceneReader(path)
+ *  {
+ *    load.loadGeometry(path);
+ *  }
```
Meakk, line 33:
```suggestion
 *  if (load.hasFullSceneReader(path)
 *  {
 *    load.loadFullScene(path);
 *  }
 *  else if (load.hasSceneReader(path)
 *  {
 *    load.loadGeometry(path);
 *  }
```

library/public/loader.h
```diff
@@ -12,76 +13,64 @@ namespace f3d
...
+  virtual loader& loadGeometry(const std::string& filePath, bool reset = false) = 0;
 
   /**
-   * Get the current file index.
+   * Reset scene and load a full scene from provided file.
+   * Please note default scene related options are not taken into account when loading a full scene.
+   * Throw a load_failure_exception on failure.
    */
-  virtual int getCurrentFileIndex() const = 0;
+  virtual loader& loadFullScene(const std::string& filePath) = 0;
```
Meakk, line 63: We have `hasGeometryReader` and `loadGeometry`, but `hasSceneReader` and `loadFullScene`.  
Let's rename the API `loadScene`, it's more consistent and shorter.

library/testing/TestSDKInteractorDropFullScene.cxx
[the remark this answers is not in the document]
Meakk: I don't want users to have this kind of code to workaround non-portability issues in libf3d.  
You don't use backslashes for CMake on Windows, right?  
Cross-platform tools usually use normal slashes and conversion is done internally is needed.

Meakk (changes requested):

application/F3DStarter.cxx
mwestphal: Yes, or at least we should switch to it anyway, even if already added. I will make the change.

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
mwestphal: and i need access to i.

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
mwestphal: yes, not in this PR.

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
mwestphal: yes, not in this PR

library/public/loader.h
mwestphal: both version, mine and yours, are both incorrect. I will fix it.

library/testing/TestSDKInteractorDropFullScene.cxx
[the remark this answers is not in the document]
mwestphal: Not related to this PR. Please open an issue or create a PR. Such code already exists in other tests.
I'm not even sure if loadGeometry works with unix backslashes on windows..

> Cross-platform tools usually use normal slashes and conversion is done internally is needed.

No, they use std::filesystem that handles that for them, but I cant use std::filesystem here.

library/VTKExtensions/Rendering/vtkF3DRendererWithColoring.cxx
Meakk: can't want for C++20 :P

library/public/loader.h
Meakk: I'd like to have the same `has[...]Reader` and `load[...]` pattern. In the first case you have `Geometry` and it's fine. For the second case you have `hasSceneReader` an `loadFullScene` so `Scene`/`FullScene` which is inconsistent.  
It should be either `hasSceneReader`/`loadScene` or `hasFullSceneReader`/`loadFullScene` (I prefer the first proposition)

mwestphal: got it, I agree.

library/testing/TestSDKInteractorDropFullScene.cxx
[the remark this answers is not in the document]
mwestphal: @Meakk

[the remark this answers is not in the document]
Meakk: https://github.com/f3d-app/f3d/pull/661

application/F3DStarter.cxx
mwestphal: actually that makes it way too complex with the need of index tracking. I will just emit a warning.

webassembly/F3DEmscriptenBindings.cxx
```diff
@@ -28,8 +28,8 @@ f3d::options* getOptionsPtr(f3d::engine& e) { return &e.getOptions(); }
 f3d::options* toggle(f3d::options& o, const std::string& name) { return &o.toggle(name); }
 
 f3d::loader* getLoaderPtr(f3d::engine& e) { return &e.getLoader(); }
-f3d::loader* addFile(f3d::loader& l, const std::string& p) { return &l.addFile(p); }
-void loadFile(f3d::loader& l) { l.loadFile(); }
+f3d::loader* loadGeometry(f3d::loader& l, const std::string& p) { return &l.loadGeometry(p); }
```
Meakk, line 31: You have to change the HTML example file too. Go ahead blindly it should be straigthforward, I'll double check it still work later.

mwestphal: on it

library/testing/TestSDKInteractorDropFullScene.cxx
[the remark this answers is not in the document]
mwestphal: fixed here too.

webassembly/F3DEmscriptenBindings.cxx
Meakk: I don't see the change to the HTML file

mwestphal: done this time

Meakk (approved):

PR #635, opened 2023-02-19
[CI] Use new version of wipe caches

mwestphal:


mwestphal:
I will go ahead and merge this @Meakk

PR #641, opened 2023-02-28
Fix grid shader on OpenGL ES

Meakk:
When I rebased #587 I noticed the grid shader wasn't working anymore.  
Default uniform values are not allowed, and implicit cast are not allowed too.  
This PR fix the shader compilation when using a GLES backend.


snoyer:
> @snoyer FYI
> Default uniform values are not allowed, and implicit cast are not allowed too.

Thanks! I didn't know about these differences. Is there a way to compile/run F3D with GLES (apart from the Emscripten bindings)?

Meakk:
I think you can build VTK using OpenGL ES backend.  
We'll have to add tests at some point to ensure non-regression but it's not straightforward.

PR #642, opened 2023-03-04
Grid configuration

snoyer:
Add options to control the size of the unit square and the number of subdivisions of the grid.

This would be useful when rendering physical objects (3d printing comes to mind) as the grid could give a better indication of actual scale if these parameters can be set to a given value (eg. unit-square of 10 and 10 subdivisions for a model in millimeters, or unit square of 25.4 and 4 or 8 subdivision if the model is in mm but the target audience is used to imperial units).


Meakk (approved):
Awesome. Looks good to me.  
@mwestphal PTAL

library/src/options.cxx
```diff
@@ -116,7 +116,9 @@ options::options()
   this->Internals->init("render.show-edges", false);
   this->Internals->init("render.line-width", 1.0);
   this->Internals->init("render.point-size", 10.0);
-  this->Internals->init("render.grid", false);
+  this->Internals->init("render.grid.enable", false);
+  this->Internals->init("render.grid.unit", 0.0);
```
mwestphal, line 120: shouldnt the default be negative ?

library/VTKExtensions/Rendering/vtkF3DOpenGLGridMapper.h
```diff
@@ -58,7 +58,7 @@ class vtkF3DOpenGLGridMapper : public vtkOpenGLPolyDataMapper
 
   double FadeDistance = 10.0;
   double UnitSquare = 1.0;
```
mwestphal, line 60: shoulnt the default be negative ?

mwestphal (changes requested):

library/src/options.cxx
snoyer: It gets recomputed if `unitSquare <= 0` so zero does work as a "not set" value (while also being a valid distance value but that's pretty deep nitpicking).

It can be set to `-1` here if it makes it easier to read/maintain, let me know.

library/VTKExtensions/Rendering/vtkF3DOpenGLGridMapper.h
snoyer: Technically it could be initialized to anything because it gets explicitly set later but making it non-positive here in `vtkF3DOpenGLGridMapper` would mean we have an invalid value by default which relies on an external piece of code to be made right (the automatic adjustment is outside the class in `vtkF3DRenderer.cxx`).

mwestphal: good point.

library/src/options.cxx
mwestphal: Right, non-positive.

mwestphal (approved):

PR #645, opened 2023-03-11
Rework light creation for a better uniformity of lights

mwestphal:
Previously, vtkF3DRenderer relied on vtkOpenGLRenderer to create automatic lights when the importer did not provide one.

vtkF3DGenericImporter did provided a vtkLightKit but also removed any previously added light, which is not standard.

This uniformise the behavior.

 - vtkF3DGenericImporter new do not importer any lights, as there is none by definition 
 - vtkF3DRenderer now creates its own vtkLightKit whenever needed, without reliying on the vtkOpenGLRenderer automatic light creation mechanism.

Of course, this means that all full scene file that did not provide any lights will have a slightly different rendering, hence the baseline changes. 

I personally think the new lighting is much better than a single headlight.


Meakk:
I agree it looks better

Meakk (approved):

PR #646, opened 2023-03-11
Fix codespell

mwestphal:


.codespellrc
```diff
@@ -1,3 +1,3 @@
 [codespell]
 skip = ./testing/**/*,./.git/**/*,./application/cxxopts.hpp,./cmake/NSIS.template.in
-ignore-words-list=nnumber,unknwn,dota
+ignore-words-list=nnumber,unknwn,dota,superseed
```
Meakk, line 3: Why do you want to whitelist it instead of fixing errors?

mwestphal: superseed is a word

Meakk: It's not ^^

mwestphal: turns out its not

Meakk (approved):

PR #647, opened 2023-03-11
Fix a leak with HDRI caching

mwestphal:
- Fix a leak with HDRI caching and improve testing so that such leak will be catched in the future
- Add an explicit  TestHDRICache

### DIFF
--- a/library/testing/TestSDKDynamicHDRI.cxx
+++ b/library/testing/TestSDKDynamicHDRI.cxx
@@ -6,10 +6,18 @@
 
 #include "TestSDKHelpers.h"
 
+#include <random>
+
 int TestSDKDynamicHDRI(int argc, char* argv[])
 {
+  // Generate a random number to avoid reusing any existing cache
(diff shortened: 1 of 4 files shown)


library/testing/TestSDKDynamicHDRI.cxx
```diff
@@ -40,7 +48,7 @@ int TestSDKDynamicHDRI(int argc, char* argv[])
   }
 
   // Check caching is working
-  std::ifstream lutFile(std::string(argv[2]) + "/cache/lut.vti");
+  std::ifstream lutFile(cachePath+ "/lut.vti");
```
Meakk, line 51:
```suggestion
  std::ifstream lutFile(cachePath + "/lut.vti");
```

Meakk:
Is it possible to set up a symbolic link in git for the test baseline in order to make sure it's exactly the same than the initial test?

Meakk (approved):

PR #658, opened 2023-03-26
Use a BMP3 150x57 logotype for NSIS packaging

mwestphal:
For NSIS packaging, the logotype needs to be of BMP3 format and should be 150x57: https://stackoverflow.com/questions/28768417/how-to-set-an-icon-in-nsis-install-cmake

Alternatively, this file (and logo.ico) could be moved to f3d-superbuild as NSIS packagin is their only usage


Meakk:
Can you explain what is the change here?

mwestphal:
For NSIS packaging, the logotype needs to be of BMP3 format and should be 150x57: https://stackoverflow.com/questions/28768417/how-to-set-an-icon-in-nsis-install-cmake

Meakk (approved):

PR #660, opened 2023-03-28
Allow external plugin builds

Meakk:


plugins/alembic/CMakeLists.txt
```diff
@@ -1,4 +1,15 @@
-include(${CMAKE_SOURCE_DIR}/cmake/f3dPlugin.cmake)
+cmake_minimum_required(VERSION 3.19)
+
+project(f3d-plugin-alembic)
```
mwestphal, line 3: Add a CI step building one internal plugin externally ?

mwestphal (approved):

plugins/alembic/CMakeLists.txt
Meakk: I'll do it later.

PR #661, opened 2023-03-29
Remove uneeded uses of antislash in testing

Meakk:


library/src/interactor_impl.cxx
```diff
@@ -551,7 +553,7 @@ bool interactor_impl::playInteraction(const std::string& file)
     this->Internals->Recorder->SetInteractor(this->Internals->VTKInteractor);
 #endif
 
-    std::string cleanFile = vtksys::SystemTools::CollapseFullPath(file);
+    std::string cleanFile = vtksys::SystemTools::CollapseFullPath(outPath);
```
mwestphal, line 556: we should rely on std::filesystem, not vtksys imo

library/testing/TestSDKInteractorCallBack.cxx
```diff
@@ -31,11 +31,8 @@ int TestSDKInteractorCallBack(int argc, char* argv[])
 
   // Test callbacks
   std::string filename = "TestSDKInteractorCallBack";
-#if defined(_WIN32)
-  std::string interactionFilePath = std::string(argv[2]) + "..\\..\\" + filename + ".log";
-#else
+
   std::string interactionFilePath = std::string(argv[2]) + "../../" + filename + ".log";
```
mwestphal, line 35: Is this actually needed ? Maybe it works out of the box ?

library/testing/TestSDKInteractorCallBack.cxx
```diff
@@ -31,11 +31,8 @@ int TestSDKInteractorCallBack(int argc, char* argv[])
...
-#endif
```
mwestphal, line 38: If it doesn't work, can you check that it works for --input, --ouput and --config ?

library/testing/TestSDKInteractorCallBack.cxx
Meakk: Actually yes, it works out of the box.

mwestphal (approved):

Issue #640, opened 2023-02-27
Options name incorrect in the documentation


Meakk:
**Describe the bug**
In `doc/libf3d/OPTIONS.md`, the options names are incorrect.  
Example: `scene.background.blur` -> `render.background.blur`

PR #667, opened 2023-03-30
Fix libf3d options in documentation

Meakk:
Fix https://github.com/f3d-app/f3d/issues/640


mwestphal:
We definitely should be more attentive to option changes update in the doc. Thanks.

mwestphal (approved):

PR #676, opened 2023-04-07
F3DRenderer: Update light intensity only once

mwestphal:


mwestphal:
@snoyer please take a look :)

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -1074,8 +1060,32 @@ int vtkF3DRenderer::UpdateLights()
   {
     vtkNew<vtkLightKit> lightKit;
     lightKit->AddLightsToRenderer(this);
-    return this->GetLights()->GetNumberOfItems();
+    lightCount = this->GetLights()->GetNumberOfItems();
```
snoyer, line 1063: needs `this->LightIntensitiesConfigured = false` here maybe?

snoyer: also the doc for `vtkRenderer.UpdateLights()` says "This method will return the actual number of lights that were on", does that mean that there could be some "not on" lights that don't get counted? in which case the `lightCount == 0` test for this block would be better checked on `this->GetLights()->GetNumberOfItems()` directly?
sidenote:  no `isEmpty()` method in `vtkCollection`? :(

mwestphal: > needs this->LightIntensitiesConfigured = false here maybe?

Indeed needed.

mwestphal: > does that mean that there could be some "not on" lights that don't get counted?

Indeed, that is correct. However the only case where it matters would be if somehow a user was able to import a file with turned off lights.

In that case, I'd rather add lights to the scene instead of having a unlighted scene. I'll add a comment.

mwestphal: > no isEmpty() method in vtkCollection? :(

tbf, vtkCollection should be reimplemented fully to be just a wrapper of std::vector. This class is coming from the time were VTK did not support C++11 at all.

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -1069,13 +1055,41 @@ void vtkF3DRenderer::Render()
...
+  // Recover the number of light that are on
   int lightCount = this->Superclass::UpdateLights();
+
+  // If no lights are turned on, add a light kit
   if (lightCount == 0 && !this->UseImageBasedLighting)
   {
     vtkNew<vtkLightKit> lightKit;
     lightKit->AddLightsToRenderer(this);
-    return this->GetLights()->GetNumberOfItems();
+    lightCount = this->GetLights()->GetNumberOfItems();
```
snoyer, line 1066: extra nitpicky at that point but isn't that breaking the spec of the parent method now (returning the total light count instead of the "on" light count)? should be `lightCount += number_of_lights_in_the_lightkit` to be fully correct, even though I don't know if it's worth the trouble of calling `this->GetLights()->GetNumberOfItems()` before and after adding the lightkit to figure out the difference.

mwestphal: Definitely correct although I dont see a case where it would matter. I will make the changes though.

Meakk (approved):

PR #683, opened 2023-04-13
Down interaction add current file directory to the list of files

### LINKED ISSUE #129: Add a hotkey to load all file in current containing folder
Would it be possible to add a key that allows to add all files from a folder after opening one file out of it? E.g. I drop a file from a folder, then - while looking at the file - consider to look at all. Also (it's a shame to say that, but) as I am forced to use win at work I got used to its picture viewer that by default allows to browse all files in the same folder.

from @drtrigon

mwestphal:
- Add a Down interaction that adds current file parent directory
- Add testion with/withou --group-geometries
- Add doc
- Fix #129

### DIFF
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -159,6 +159,17 @@ int F3DStarter::Start(int argc, char** argv)
           this->Render();
           return true;
         }
+        else if (keySym == "Down")
+        {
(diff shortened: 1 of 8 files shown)


Meakk (approved):
Good idea 👍

drtrigon:
Very cool! Thanks for implementing!

Issue #443, opened 2022-09-28
Key Interaction management in libf3d is not great and should be improved


mwestphal:
**Describe the bug**
Currently the management of interactions is fully handled inside of the libf3d, without the possibility of customization.
Moreover, somekey are not accessible as the key input from VTK is "raw", making some key unreachable by just using the keyboard on some keyboard layout.

**To Reproduce**
 1. Open the file using `f3d --dry-run example.glb`
 2. press ? to show the camera information in the terminal while using a US keyboard
 3. nothing happens

**Expected behavior**
Camera information appears.

**Additional context**
This may require some changes in VTK, unless we want to take care of converting raw input from VTK and converting them to actual user input.

Moreover, if we tackled customization, we need to make sure the cheatsheet stay synchronized.

Wdyt @jpouderoux @Meakk ?

mwestphal:
related to #438

Meakk:
I agree it should be improved, and customizable.

mwestphal:
how about an API like this:

`setKeyPressCallBack(mod, key, fn)` ?

What I'm not sure about is how do we convert the intention of the user : `I want to do X so I press Y Key` to receiving actually `Y` systematically and reacting upon it.

I'm not even sure about the `mod` part. eg, on a qwerty keyboard, pressing  `?` is actually pressing `Shift + /`.

`showkeys -a` display it correctly:

```
/        47 0057 0x2f
?        63 0077 0x3f
/        47 0057 0x2f
?        63 0077 0x3f

```

but VTK do not provide the right keysym or keycode:

```
/
Slash

Shift_L
/
Slash
```

If we want to provide better interaction, I think we definitely needs to improve VTK to give us the correct key code. We should not have to handle keyboard layouts in the libf3d.

Wdyt @snoyer @Meakk ?

Meakk:
The current API is `virtual interactor& setKeyPressCallBack(std::function<bool(int, std::string)> callBack)` and you want to replace it with `virtual interactor& setKeyPressCallBack(int mod, int keycode, std::function<bool()> callBack)`, right?

I think it makes sense. We can save a map in libf3d (initialized on the application side?) so the user can provide his own bindings.  
Regarding the `mod`, what about `ctrl` and `alt`? It's a bit different than `shift` since the result is not a different character, so I think `mod` is required.

snoyer:
[...]

PR #705, opened 2023-04-16
Keep camera on reload

snoyer:
Felt bad opening an issue without at least looking into it a bit so here's an attempt at reloading the current file without resetting the camera.

The use case is:
1. loading a procedural or otherwise generated model, navigating around
2. externally regenerating the model
3. reloading the model in F3D keeping the current view.

This implementation is based on what's been discussed on Discord:
1. adding a mechanism to save and restore camera states by key. I feel like this would be better using opaque state objects but I could not get it to work :(
2. saving the camera state before reloading, restoring afterwards before the next render

I don't know if this behaviour should be the default or not?
If not, `shift+up` could be used, but do we have access to modifiers in interactor callbacks?
The type of model/scene could also play a role maybe (eg. only reset the camera if the file defines cameras)?

### DIFF
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -144,7 +144,8 @@ int F3DStarter::Start(int argc, char** argv)
         f3d::camera& cam = this->Internals->Engine->getWindow().getCamera();
         const auto camState = cam.saveState();
         this->LoadFile(index, true);
-        cam.restoreState(camState, true);
+        cam.restoreState(camState);
+        cam.deleteState(camState);
       }
       else
       {
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -135,29 +135,40 @@ int F3DStarter::Start(int argc, char** argv)
     f3d::window& window = this->Internals->Engine->getWindow();
     f3d::interactor& interactor = this->Internals->Engine->getInteractor();
 
+    const auto loadFile = [this](int index, bool restoreCamera = false) -> bool
+    {
+      this->Internals->Engine->getInteractor().stopAnimation();
+
+      if (restoreCamera)
+      {
+        f3d::camera& cam = this->Internals->Engine->getWindow().getCamera();
+        const auto camState = cam.saveState();
+        this->LoadFile(index, true);
+        cam.restoreState(camState, true);
+      }
+      else
+      {
+        this->LoadFile(index, true);
+      }
+
+      this->Render();
+      return true;
+    };
+
     interactor.setKeyPressCallBack(
-      [this](int, std::string keySym) -> bool
+      [this, loadFile, &interactor](int, std::string keySym) -> bool
       {
         if (keySym == "Left")
         {
-          this->Internals->Engine->getInteractor().stopAnimation();
-          this->LoadFile(-1, true);
-          this->Render();
-          return true;
+          return loadFile(-1);
         }
         else if (keySym == "Right")
         {
-          this->Internals->Engine->getInteractor().stopAnimation();
-          this->LoadFile(1, true);
-          this->Render();
-          return true;
+          return loadFile(+1);
         }
         else if (keySym == "Up")
         {
-          this->Internals->Engine->getInteractor().stopAnimation();
-          this->LoadFile(0, true);
-          this->Render();
-          return true;
+          return loadFile(0, true); // TODO decide how to control whether we restore the camera
         }
         else if (keySym == "Down")
         {
@@ -168,12 +179,14 @@ int F3DStarter::Start(int argc, char** argv)
               this->Internals->FilesList[static_cast<size_t>(this->Internals->CurrentFileIndex)]
                 .parent_path(),
               true);
-            this->LoadFile(0, true);
-            this->Render();
+            return loadFile(0);
           }
           return true;
         }
-        return false;
+        else
+        {
+          return false;
+        }
       });
 
     interactor.setDropFilesCallBack(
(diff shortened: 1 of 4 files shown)


mwestphal:
> but do we have access to modifiers in interactor callbacks?

Yes we do. But I think behavior could be default :)

snoyer:
I added `deleteState` to make the state management complete (could not delete without restoring before), so the API is now:
- `CameraStateKey saveState()`
- `bool restoreState(const CameraStateKey&)`
- `bool deleteState(const CameraStateKey&)`
- with an internal `map<CameraStateKey, ...>` storing the states (as `vtkCamera` deep copies)

Using opaque state objects would allow to delegate the storing of the states (which would then be optional, and not needed for this PR for example) to the app and simplify the API to:
- `CameraState saveState()`
- `bool restoreState(const CameraState&)`

However my lack of c++ skills is standing in the way of me implementing it that way. Options are then:
1. we want opaque state object because it's nicer and:
   a. I keep banging my head against it until it works
   b. someone more competent implements it instead
2. we leave it as is (using keys and internal management)

Thoughts?

mwestphal:
I do think storing the states in the App would be better, but I'm not sure I will be able to help shortly. Could you push a branch showing what did not work for you ?

snoyer:
> Could you push a branch showing what did not work for you ?

Basically I have a `class cameraState_impl : public cameraState` to hold and hide a `vtkCamera`, so far so good.
The problem was I was trying to do:
- `cameraState saveState(){ return new cameraState_impl(...); }`
- `restoreState(cameraState&){ ... dynamic_cast<cameraState_impl&>(cameraState) ... }`
Which, according to the internet, wasn't going to work because polymorphism without pointers just isn't a thing (for reason I can't be bothered to dig into really). Compiler was fine building it all only for it to `bad_cast` at runtime though, fun times!

Using pointers (`CameraState* saveState()`, `void restoreState(CameraState*)`) works. Maybe using smart pointers would be better but I'm not fluent enough to tell.

mwestphal:
Let me know if you need help with this @snoyer , the release rush is running out.

I'm not available on discord for a few days though.

mwestphal:
Could you give a go to the implementation with an opaque object and I can try to make it work ?

snoyer:
> Could you give a go to the implementation with an opaque object and I can try to make it work ?

That's where's it at now, `cameraState_impl` holds a `vtkCamera` to be used internally, and the exposed `cameraState` just gets passed around exposing nothing:
https://github.com/f3d-app/f3d/blob/5416af1204f832f5987bcbf2c5edacafb2e49f06/library/private/camera_impl.h#L23-L28
https://github.com/f3d-app/f3d/blob/5416af1204f832f5987bcbf2c5edacafb2e49f06/library/public/camera.h#L12-L16

mwestphal:
ok ! Could you rebase your branch on master ? I will then take a look this week I hope

mwestphal:
Just tested and it seems to work fine. Am I missing anything ?

mwestphal:
BTW I wonder if we should add a modifier to keep cameraState when switching between files, it seems like a usefull feature.

snoyer:
> Just tested and it seems to work fine. Am I missing anything ?

It is functional.
But `CameraState* saveState()`, `void restoreState(CameraState*)` feels... meh? I believe (maybe wrongly) that `CameraState saveState()`, `void restoreState(CameraState&)` (no pointers) would be better. Thoughts?


> BTW I wonder if we should add a modifier to keep cameraState when switching between files, it seems like a usefull feature.

1. See first post: need to pass modifiers to callbacks then
 https://github.com/f3d-app/f3d/blob/e035ec361deeecba9fa71e1004f4f80dc2e04109/library/private/interactor_impl.h#L37
https://github.com/f3d-app/f3d/blob/e035ec361deeecba9fa71e1004f4f80dc2e04109/library/src/interactor_impl.cxx#L86

2. Would need to specify the expected behavior for edge cases. I did the easy thing of copying the `vtkCamera` which works just fine for a single model by restoring the exact same absolute camera position. However when going across models that are centered/oriented differently we may need to save/restore the F3D camera _parameters_ instead of the current VTK camera.

mwestphal:
Ok, got it, Ill take a look.

mwestphal:
> However when going across models that are centered/oriented differently we may need to save/restore the F3D camera parameters instead of the current VTK camera.

I think I'll go this way, and we will not need any opaque object :)

mwestphal:
Here it is @snoyer : https://github.com/f3d-app/f3d/pull/788.

Can I let you finish it ?

Regarding the modifier, you are absolutely right, I will look into adding an API for modifiers in the callbacks.

snoyer:
> Here it is @snoyer : #788.
> 
> Can I let you finish it ?

I'm going to need to learn myself some more git first to rebranch from yours but I'll see what I can do :)

> Regarding the modifier, you are absolutely right, I will look into adding an API for modifiers in the callbacks.

Careful you don't run into #443 when you go there, there's no telling how deep it could drag you ;p

mwestphal:
Yeah, I'll avoid #443 for now.

> some more git first to rebranch from yours but I'll see what I can do 

Here is how I would do it:

```
git remote add mwestphal https://github.com/mwestphal/f3d.git
git fetch mwestphal
git cherry-pick 64c38377cbff0c3d502bcef41cad48e18346ccf8
```

hth,

snoyer:
Are we looking at restoring the camera on previous/next model in this PR or just on reloading the current model?

If we want to do prev/next we need to first assess and decide whether the camera state should be:
1. absolute in space (cam position, focal point, ...), or
2. relative to the models (azimuth, elevation, ...)

mwestphal:
imo a camera state should be absolute, so I'd focus on that first. If we get fancy we camera positions in the future, we may want to revisit it at some point, but for now seems good enough.

So lets only focus on keeping camera location on the UP interaction for now with an absolute position :)

mwestphal:
Need any help with this one @snoyer ?

mwestphal:
closing in favor of: https://github.com/f3d-app/f3d/pull/705

PR #726, opened 2023-04-18
Macos bundle static

mwestphal:


CMakeLists.txt
```diff
@@ -154,6 +154,9 @@ endif()
 # plugins
 option(F3D_PLUGINS_STATIC_BUILD "Make all plugins static (embedded into libf3d) and automatically loaded by F3D" ON)
 mark_as_advanced(F3D_PLUGINS_STATIC_BUILD)
+if (F3D_MACOS_BUNDLE AND NOT F3D_PLUGINS_STATIC_BUILD)
```
Meakk, line 157: shouldn't we define `F3D_PLUGINS_STATIC_BUILD` as a dependent option to enforce it to OFF if `F3D_MACOS_BUNDLE` is ON?

Meakk: fair enough

Meakk (approved):

PR #747, opened 2023-04-21
Dynaymic plugin with sanitizer

mwestphal:


mwestphal:
FYI @Meakk , sanitizer is broken with dynamic plugins

PR #756, opened 2023-04-22
make `AddFile` return index

snoyer:
When loading files, F3D checks for duplicate to avoid loading them twice but always displays the last file of the unique list regardless.

Making `F3DStarter::AddFile` return the index of the last added file allows to show the last dropped file even it was already present in the list.

### DIFF
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -180,11 +180,15 @@ int F3DStarter::Start(int argc, char** argv)
       [this](std::vector<std::string> filesVec) -> bool
       {
         this->Internals->Engine->getInteractor().stopAnimation();
+        int index = -1;
         for (std::string file : filesVec)
         {
-          this->AddFile(fs::path(file));
+          index = this->AddFile(fs::path(file));
         }
-        this->LoadFile(static_cast<int>(this->Internals->FilesList.size()) - 1);
+        if (index > -1)
+        {
+          this->LoadFile(index);
+        }
         this->Render();
         return true;
       });
@@ -495,7 +499,7 @@ void F3DStarter::Render()
 }
 
 //----------------------------------------------------------------------------
-void F3DStarter::AddFile(const fs::path& path, bool quiet)
+int F3DStarter::AddFile(const fs::path& path, bool quiet)
 {
   auto tmpPath = fs::absolute(path);
   if (!fs::exists(tmpPath))
@@ -504,10 +508,9 @@ void F3DStarter::AddFile(const fs::path& path, bool quiet)
     {
       f3d::log::error("File ", tmpPath, " does not exist");
     }
-    return;
+    return -1;
   }
-
-  if (fs::is_directory(tmpPath))
+  else if (fs::is_directory(tmpPath))
   {
     std::set<fs::path> sortedPaths;
     for (const auto& entry : fs::directory_iterator(tmpPath))
@@ -519,6 +522,7 @@ void F3DStarter::AddFile(const fs::path& path, bool quiet)
       // Recursively add all files
       this->AddFile(entryPath, quiet);
     }
+    return this->Internals->FilesList.size() - 1;
   }
   else
   {
@@ -533,5 +537,6 @@ void F3DStarter::AddFile(const fs::path& path, bool quiet)
     {
       f3d::log::warn("File ", tmpPath, " has already been added");
     }
+    return std::distance(this->Internals->FilesList.begin(), it);
   }
 }
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -522,7 +522,7 @@ int F3DStarter::AddFile(const fs::path& path, bool quiet)
       // Recursively add all files
       this->AddFile(entryPath, quiet);
     }
-    return this->Internals->FilesList.size() - 1;
+    return static_cast<int>(this->Internals->FilesList.size()) - 1;
   }
   else
   {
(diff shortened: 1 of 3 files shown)


mwestphal:
> to show the last dropped file even it was already present in the list.

I think current behavior was voluntary on my part and related to the "Down" interaction. With your change, pressing "Down" will switch to the last file of the current folder, which is not expected imo.

snoyer:
> I think current behavior was voluntary on my part and related to the "Down" interaction. With your change, pressing "Down" will switch to the last file of the current folder, which is not expected imo.

I see, that makes sense. However when dropping an already loaded file I would be more intuitive if that file showed up, instead of jumping to the end of the list. Should be handled differently in each case maybe?

I thought that was going to be a quick fix but apparently not, who'd have guessed...

mwestphal:
> Should be handled differently in each case maybe?

I think you are right.

> I thought that was going to be a quick fix but apparently not, who'd have guessed...

indeed :p

snoyer:
> > Should be handled differently in each case maybe?
> 
> I think you are right.

Turns out it's already 2 different code paths. I only use the newly returned index from `F3DStarter::AddFile` in `interactor.setDropFilesCallBack` so the "Down" interactions shouldn't be affected.

application/F3DStarter.cxx
```diff
@@ -180,11 +180,15 @@ int F3DStarter::Start(int argc, char** argv)
...
         this->Internals->Engine->getInteractor().stopAnimation();
+        int index = -1;
         for (std::string file : filesVec)
         {
-          this->AddFile(fs::path(file));
+          index = this->AddFile(fs::path(file));
+        }
+        if (index > -1)
+        {
+          this->LoadFile(index);
```
mwestphal, line 190: what happends if you drop the folder containing the currently loaded file ?

If I understand well it loads the last file of the folder. I think that it is fine, just checking.

snoyer: It would display whatever was "added" last, according to whatever the traversing/recursing order is, I assume. I haven't really looked into the recursive cases tbh.

mwestphal: as long as it is not broken and file are actually added, fine by me.

snoyer: It would need to be looked into deeper to be done properly right. Not sure I'm up to dealing with what I expect to find down there though (eg. it will shove a bunch of possibly random files in the list on add, and we'd need to handle the unreadable files to skip them and remove them from the list on load).

mwestphal: I just tested `master` it with this:

`./bin/f3d ../src/testing/data/multi/suzanne.obj`

Then dropped multi into it, zsuzanne.stl was loaded.
So current behavior is as described, and you did not change that as far as I can see, so this is perfectly fine :)


> we'd need to handle the unreadable files to skip them and remove them from the list on load).

Not needed, you can drop an unreadable file, it will just show up as an empty "UNSUPPORTED" file. There is already tests for that.

application/F3DStarter.h
```diff
@@ -23,7 +23,7 @@ class F3DStarter
   /**
    * Add a file or directory to the list of paths
    */
-  void AddFile(const std::filesystem::path& path, bool quiet = false);
+  int AddFile(const std::filesystem::path& path, bool quiet = false);
```
mwestphal, line 26: I think this returned index could be checked and potentially used whenever AddFile is used.

snoyer: the other uses I can see are the "Down" action:
https://github.com/f3d-app/f3d/blob/82aedc9d31603fa52dac5ef3ae50d286ecba9465/application/F3DStarter.cxx#L162-L175

and the initial loading from the CLI
https://github.com/f3d-app/f3d/blob/82aedc9d31603fa52dac5ef3ae50d286ecba9465/application/F3DStarter.cxx#L217-L223

It would make sense to leave the initial case as is, ie. display the first file and have the other ones queued up.

For the down action I don't know, I'm unclear on the use-cases/semantics.

application/F3DStarter.cxx
snoyer: > Not needed, you can drop an unreadable file, it will just show up as an empty "UNSUPPORTED" file. There is already tests for that.

Good to know it's a feature and not a bug that needs fixing. Scenarios that may be worth considering however:
1. I open an `.stl` I saved in my `~/Download` folder full of random stuff, I press down to see what other 3d files I have in there, I am now scrolling through a possibly high percentage of "[UNSUPPORTED]" memes, cat pictures, movies...
2. I open a model in my home directory I press down by accident with my fat fingers, F3D it now off rummaging through the depth of my `~/`, hidden folders and all...

application/F3DStarter.h
mwestphal: I think you are right :)

application/F3DStarter.cxx
mwestphal: Correct, this feature may need some improvements, but this is completely outside of the scope of this PR. Thanks for checking.

mwestphal (approved):

mwestphal:
@snoyer let me know if this is ready to be merged :)

snoyer:
> @snoyer let me know if this is ready to be merged :)

From what I can tell it does fix the simple case that I ran into using the app (drag and drop repeated files one by one), for which there so happened to already have been an existing test (`TestInteractionDropSameFiles` drops Cow, then Dragon, then Cow; used to end up showing Dragon, now shows Cow) so it has that going for it.

However I'm going to leave that final call of whether or not it's prod ready to you and @Meakk because I don't feel comfortable enough with the fully recursive behavior of `AddFile` to have a solid opinion.

mwestphal:
> However I'm going to leave that final call of whether or not it's prod ready to you and @Meakk because I don't feel comfortable enough with the fully recursive behavior of AddFile to have a solid opinion.

You did not change anything in regards to that, so that's fine.

PR #760, opened 2023-04-24
add absolute grid positioning

### LINKED ISSUE #755: Respect model origin when displaying scene
When F3D opens an object, it puts the body on the ground plane, then centres it on two remaining axes, losing the model's origin in the process

I'd like an option (I'd like it to be enabled by default) to keep the model's origin and place the model in the viewport scene according to the model creator's intent.

**Describe alternatives you've considered**

- An option: **Model placement: Origin / Centred**
- A hotkey to switch it
- A default in the config
- Model's origin displayed in the viewer

**Additional context**

The way F3d centres model mimics the behaviour of 3D slicer software like PrusaSlicer.   
It's desirable for a slicer (that can't print under the ground plane, can't print in air and have limited build plate dimensions), but that's completely irrelevant to a 3D viewer.

It might be desirable in some cases (Like a scene with an object that's moved way off the origin due to human error).

In other cases, it isn't:
- You're dealing with multi-component model split into several files in the same origin (ex.: multicoloured prints)
- The origin has a special meaning (an attachment point, or swivel point) 
- probably more

--

**I've made a test part** that contains 2 cubes, starting at world's origin: [origin-cubes.obj.zip](https://github.com/f3d-app/f3d/files/11300324/origin-cubes.obj.zip) (human-readable, you can see everything starts at 0), [FreeCAD source origin-cubes.FCStd.zip](https://github.com/f3d-app/f3d/files/11300325/origin-cubes.FCStd.zip)

You can notice F3D completely ignores model's origin on every axis.

| FreeCAD source | OJB in Meshlab | OBJ in F3D |
| --- | --- | --- |
| freecad | meshlab | f3d |

**Contribution**
I probably can't contribute in this

snoyer:
Adding `render.grid.absolute` option and `--grid-absolute` CLI argument to avoid positioning the grid under the model and keep the (0,0,0) origin. Fixes #755.


application/F3DOptionsParser.cxx
```diff
@@ -288,6 +288,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp1, "up", "", "Up direction", options.getAsStringRef("scene.up-direction"), HasDefault::YES, MayHaveConfig::YES, "{-X, +X, -Y, +Y, -Z, +Z}");
     this->DeclareOption(grp1, "axis", "x", "Show axes", options.getAsBoolRef("interactor.axis"), HasDefault::YES, MayHaveConfig::YES);
     this->DeclareOption(grp1, "grid", "g", "Show grid", options.getAsBoolRef("render.grid.enable"), HasDefault::YES, MayHaveConfig::YES);
+    this->DeclareOption(grp1, "grid-absolute", "G", "Show grid at (0,0,0)", options.getAsBoolRef("render.grid.absolute"), HasDefault::YES, MayHaveConfig::YES);
```
mwestphal, line 291: I'm not a big fan of adding small/big cases for short option. I'd rather not have a short option at all. wdyt @Meakk ?

mwestphal: > even just --grid-absolute

Two option to do the same thing with slight difference is a bad idea and will cause headaches in the future.
`-g` turn on the grid, other options configure the grid. Simple (yet verbosy, I know, that is why config files can help :) )

Meakk: I agree, I prefer `-g --grid-absolute`.  
If we want to swap between relative/absolute easily, you could just do `-g --grid-absolute=true` and change `true` by `false`.  
Maybe `0` and `1` is also supported by cxxopts, I have never tried.

snoyer: the council has spoken

mwestphal: > you could just do -g --grid-absolute=true and change true by false.
> Maybe 0 and 1 is also supported by cxxopts, I have never tried.

correct, true/false and 0/1 are both supported

application/F3DOptionsParser.cxx
```diff
@@ -288,6 +288,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp1, "up", "", "Up direction", options.getAsStringRef("scene.up-direction"), HasDefault::YES, MayHaveConfig::YES, "{-X, +X, -Y, +Y, -Z, +Z}");
     this->DeclareOption(grp1, "axis", "x", "Show axes", options.getAsBoolRef("interactor.axis"), HasDefault::YES, MayHaveConfig::YES);
     this->DeclareOption(grp1, "grid", "g", "Show grid", options.getAsBoolRef("render.grid.enable"), HasDefault::YES, MayHaveConfig::YES);
+    this->DeclareOption(grp1, "grid-absolute", "", "Show grid at (0,0,0)", options.getAsBoolRef("render.grid.absolute"), HasDefault::YES, MayHaveConfig::YES);
```
mwestphal, line 291: Show -> Position ?

mwestphal: instead of below the model.

doc/libf3d/OPTIONS.md
```diff
@@ -63,6 +63,7 @@ render.line-width|double<br>1.0<br>render|Set the *width* of lines when showing
 render.show-edges|bool<br>false<br>render|Show the *cell edges*|\-\-edges
 render.point-size|double<br>10.0<br>render|Set the *size* of points when showing vertices and point sprites.|\-\-point-size
 render.grid.enable|bool<br>false<br>render|Show *a grid* aligned with the horizontal (orthogonal to the Up direction) plane.|\-\-grid
+render.grid.absolute|bool<br>false<br>render|Show the grid at the *absolute origin* of the model's coordinate system.|\-\-grid
```
mwestphal, line 66: Position the grid at the *absolute origin* of the model's coordinate system instead of below the model.

mwestphal (approved):

application/F3DOptionsParser.cxx
snoyer: "Position grid at the absolute origin instead of below the model" to match your other suggestion below?

mwestphal: sounds good to me

Meakk (approved):

PR #764, opened 2023-04-25
validate up direction string

snoyer:
Running `f3d foo.obj --up=X`, without the sign for the axis, we can observe that:

1. omitting sign does not default to positive
2. F3D does not warn/complain about `up` strings it does not expect/understand

This PR addresses both points.

### DIFF
--- a/library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
+++ b/library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
@@ -58,6 +58,7 @@
 
 #include <cctype>
 #include <chrono>
+#include <regex>
 #include <sstream>
 
 vtkStandardNewMacro(vtkF3DRenderer);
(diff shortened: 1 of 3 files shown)


mwestphal:
NIce ! FYI i'm hoping to add enum based option at some point that will only accept specific strings, but this is a perfectable solution for  now.

mwestphal (approved):

snoyer:
newly added test failed on all `v9.0.0` runners, not sure what to do with that information...

mwestphal:
> newly added test failed on all `v9.0.0` runners, not sure what to do with that information...

Add DEFAULT_LIGHTS in the test. This disable the test on vtk9.0 as lights changed between vtk9.0 and 9.1

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -187,15 +188,18 @@ void vtkF3DRenderer::Initialize(const std::string& up)
...
+  const std::regex re("([-+]?)([XYZ])", std::regex_constants::icase);
+  std::smatch match;
+  if (std::regex_match(up, match, re))
   {
-    char sign = up[0];
-    char axis = std::toupper(up[1]);
-    if ((sign == '-' || sign == '+') && (axis >= 'X' && axis <= 'Z'))
+    const float sign = match[1].str() == "-" ? -1.0 : +1.0;
+    const int index = std::toupper(match[2].str()[0]) - 'X';
+    if (index >= 0 && index < 3)
```
Meakk, line 197: Can this condition fail? If not, it should be an assert.

snoyer: Good point. Straight up `assert(index >= 0 && index < 3)` or is there a more c++ way?

mwestphal: looks good to me

PR #772, opened 2023-04-25
Test OCCT coloring

mwestphal:


mwestphal:
I need to add freetype to the ci

.github/actions/occt-install-dep/action.yml
```diff
@@ -62,22 +63,6 @@ runs:
       shell: bash
       run: cmake --build . --parallel 2 --target install --config Release
 
-    - name: Patch OCCT install macos
```
Meakk, line 65: This has been patched upstream? Nice!

Meakk (approved):

.github/actions/occt-install-dep/action.yml
mwestphal: no, but the patch does not seem needed when building more modules

PR #782, opened 2023-04-27
Avoid hardcoded path to conform to current CMake pratice

topazus:
It maybe be better to use the `${CMAKE_INSTALL_LIBDIR}`, `${CMAKE_INSTALL_INCLUDEDIR}`, instead of hardcoding the directory. `${CMAKE_INSTALL_LIBDIR}`, `${CMAKE_INSTALL_INCLUDEDIR}`, et al. are defined by the `GNUInstallDirs` that represents the platform-specific installation directory and provides install directory variables defined by the [GNU Coding Standards](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html).

Ref:
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8111
https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html
https://github.com/libsdl-org/SDL/commit/6956f4aa1982b66b234026b46f7bb2dd44c67894

### DIFF
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -174,20 +174,20 @@ if(BUILD_SHARED_LIBS)
     LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT library)
 
   install(FILES ${F3D_PUBLIC_HEADERS}
-    DESTINATION "include/f3d"
+    DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/f3d"
     COMPONENT sdk
     EXCLUDE_FROM_ALL)
 
   install(EXPORT f3dTargets
     NAMESPACE f3d::
-    DESTINATION "lib/cmake/f3d"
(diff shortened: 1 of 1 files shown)


mwestphal:
Definitely a good idea !

mwestphal:
strange unrelated CI issues, investigating.

mwestphal:
I've rerun, looks like an intermitent docker repository failure

topazus:
Thanks for your quick looking at this.

mwestphal (approved):

topazus:
> I see a few other usages of `lib` in application/CMakeLists.txt, I think they should be fixed too.

You mean the usage of `lib` here? I don't know apple platform.
https://github.com/f3d-app/f3d/blob/e035ec361deeecba9fa71e1004f4f80dc2e04109/application/CMakeLists.txt#L46-L51

mwestphal:
No worries, I will take case of it.

mwestphal:
Perfect, thanks @topazus !

PR #784, opened 2023-04-27
Update docker images and add a EGL CI

mwestphal:
- Update for new docker images
- Add a EGL CI


mwestphal:
Looks like the way of counting coverage slightly changed between gcov/lcov versions.

.github/workflows/ci.yml
```diff
@@ -130,7 +137,7 @@ jobs:
         -DF3D_LINUX_GENERATE_MAN=ON
         -DF3D_LINUX_INSTALL_DEFAULT_CONFIGURATION_FILE_IN_PREFIX=ON
         -DF3D_MACOS_BUNDLE=${{ matrix.bundle_label == 'bundle' && 'ON' || 'OFF' }}
-        -DF3D_MODULE_EXTERNAL_RENDERING=ON
+        -DF3D_MODULE_EXTERNAL_RENDERING=${{ matrix.egl_label == 'egl' && 'OFF' || 'ON' }}
```
Meakk, line 140: Why would you disable external context with egl?

mwestphal: the external rendering testing does not work  at all with EGL, and this is expected.

Meakk: Why is it expected? The external feature means the context is external so there is no use of GLX or EGL directly. It's agnostic of the method used by the caller.
If the tests are not working because QT and GLFW cannot be used, I got it, but the tests themselves should be disabled while the module should be built.  
What happens if the user creates his own EGL context and wants to use libf3d with an external context?

mwestphal: Doable but there is no control for egl en f3d code yet. That was the easiest way. I will add egl check and specific code path for egl in f3d later.

mwestphal: The external feature of course builds fine with egl, only the test with glfw et qt are not working

Meakk (approved):

PR #786, opened 2023-04-27
Vtk occt

mwestphal:


mwestphal:
@Meakk I see a few issues with this, but I also see how it could be beneficial for F3D. What are your thougts ?

mwestphal:
This is not ready and the fork are diverging, closing.

PR #788, opened 2023-04-27
Keep camera on reload

mwestphal:


mwestphal:
@snoyer ptal

snoyer:
> @snoyer ptal

Inconsequential nitpick: not a big fan of the hardcoded default values in `camera_state_t`, but it looks like they are hardcoded in the `vtkCamera` constructor too so I guess there's no way to avoid the duplication short of instantiating a `vtkCamera` to `get...()` the values at runtime :(

mwestphal:
> Inconsequential nitpick: not a big fan of the hardcoded default values in `camera_state_t`, but it looks like they are hardcoded in the `vtkCamera` constructor too so I guess there's no way to avoid the duplication short of instantiating a `vtkCamera` to `get...()` the values at runtime :(

Not a big fan either tbh, but better to have something than uninitialized values.

mwestphal:
resetToDefault now do not rely on a vtk camera

library/private/camera_impl.h
```diff
@@ -12,6 +12,8 @@
 
 #include "camera.h"
 
+#include <vtkCamera.h>
```
mwestphal, line 15: to delete

application/F3DStarter.cxx
```diff
@@ -171,12 +182,14 @@ int F3DStarter::Start(int argc, char** argv)
...
                 .parent_path(),
               true);
-            this->LoadFile(0, true);
-            this->Render();
+            return loadFile(0);
           }
           return true;
         }
-        return false;
+        else
```
Meakk, line 189: no reason to add a `else` here

application/F3DStarter.cxx
```diff
@@ -138,29 +138,40 @@ int F3DStarter::Start(int argc, char** argv)
     f3d::window& window = this->Internals->Engine->getWindow();
     f3d::interactor& interactor = this->Internals->Engine->getInteractor();
 
+    const auto loadFile = [this](int index, bool restoreCamera = false) -> bool
```
Meakk, line 141: since the lambda is used only in the other lambda, you can move it there to reduce the scope.

library/public/camera.h
```diff
@@ -9,6 +9,14 @@
 
 namespace f3d
 {
+struct F3D_EXPORT camera_state_t
```
Meakk, line 12: You do not need to export header-complete struct/class because no symbol is generated

library/public/types.h
```diff
@@ -1,14 +1,16 @@
 #ifndef f3d_types_h
 #define f3d_types_h
 
+#include "export.h"
```
Meakk, line 4: revert changes in this file

Meakk (changes requested):

library/public/types.h
mwestphal: it is needed, fails on windows without it

mwestphal: https://github.com/f3d-app/f3d/actions/runs/5141210018/jobs/9253451184

library/public/camera.h
mwestphal: since it is needed for types.h, I assume it may be needed here.

library/public/camera.h
```diff
@@ -9,6 +9,14 @@
...
+{
+  point3_t pos = { 0, 0, 1 };
+  point3_t foc = { 0, 0, 0 };
+  vector3_t up = { 0, 1, 0 };
+  angle_deg_t angle = 30;
```
mwestphal, line 17: @snoyer @Meakk so let me know what yhou think of this init. imo it is fine.

mwestphal:
@snoyer please review

library/public/camera.h
Meakk: it's a double, so `30.0`? And similar for the values of the vectors.

Meakk (approved):

PR #789, opened 2023-04-29
Fix dynamic plugin loading with multi-config generators

mwestphal:


application/F3DOptionsParser.cxx
```diff
@@ -200,8 +200,10 @@ std::vector<std::string> ConfigurationOptions::GetPluginSearchPaths() const
   // Add a executable related path
   auto tmpPath = F3DSystemTools::GetApplicationPath();
   tmpPath = tmpPath.parent_path().parent_path();
-  tmpPath /= "lib";
-  searchPaths.push_back(tmpPath.string());
+  for (auto dir : F3D::PluginDirectories)
```
mwestphal, line 203: should be const auto&

mwestphal:
Will be handled by @Meakk

PR #800, opened 2023-05-05
Add VCPKG manifest file and documentation

Meakk:


vcpkg.json
```diff
@@ -0,0 +1,15 @@
...
+        "assimp",
+        "draco",
+        "opencascade",
+        {
+            "name": "vtk",
+            "default-features": false,
+            "features": [ "opengl" ]
+        }
+    ],
+    "builtin-baseline": "60e2c07d20aeb4bf534a8deabf92a70cbcb07617"
```
mwestphal, line 14: no raytracing in vcpkg ?

mwestphal (approved):

vcpkg.json
Meakk: Looking at the port file of vtk, it does not seem so.

PR #833, opened 2023-05-29
Camera: Reset using screenspace

mwestphal:


mwestphal:
The only change that looks strange is : testing/baselines/TestInteractionFocalPointPickingPoints.png , but I checked it and it is fine, focal point picking happens on another point.

library/src/camera_impl.cxx
```diff
@@ -202,7 +209,23 @@ camera& camera_impl::resetToDefault()
...
 {
+
+#if VTK_VERSION_NUMBER < VTK_VERSION_CHECK(9, 0, 20210331)
   this->Internals->VTKRenderer->ResetCamera();
+#else
+#if F3D_MODULE_EXTERNAL_RENDERING
+  if (vtkExternalOpenGLRenderWindow::SafeDownCast(
+        this->Internals->VTKRenderer->GetRenderWindow()) != nullptr)
+  {
+    // External render window does not support ResetCameraScreenSpace correctly
```
Meakk, line 220: why?

library/src/camera_impl.cxx
```diff
@@ -202,7 +209,23 @@ camera& camera_impl::resetToDefault()
 //----------------------------------------------------------------------------
 camera& camera_impl::resetToBounds()
 {
+
+#if VTK_VERSION_NUMBER < VTK_VERSION_CHECK(9, 0, 20210331)
   this->Internals->VTKRenderer->ResetCamera();
+#else
+#if F3D_MODULE_EXTERNAL_RENDERING
+  if (vtkExternalOpenGLRenderWindow::SafeDownCast(
```
Meakk, line 217: Calling `IsA` might be better, what do you think?

mwestphal: definitely

library/src/camera_impl.cxx
mwestphal: no idea, I tried tracking it down but no luck and did not want to spend too much time on it. Using ResetCameraScreenSpace results on an empty image even though the computed bounds looks fine.

Meakk (approved):

PR #836, opened 2023-06-01
Rework animation suport

mwestphal:
Rework animation suport

- Rely on real elapsed time instead of timestep to animate
- Store animationManager in loader instead of interactor
- Add an --animation-speed-factor option to slow, speed and invert time\
- Add an --animation-time option to set an animation time to load
- Add an --animation-frame-rate option to control animation frame rate
- Fix multiple issues related to time and improve internal doc 
- update tests accordingly
- Add doc for options and animations


plugins/assimp/module/vtkF3DAssimpImporter.cxx
```diff
@@ -895,6 +895,7 @@ void vtkF3DAssimpImporter::UpdateTimeStep(double timestep)
 
   // get the animation tick
   timestep *= this->Internals->Scene->mAnimations[0]->mTicksPerSecond;
+  // TODO figure out how to make this work with a timevalue
```
mwestphal, line 898: @Meakk this code worked with a timestep for some reasons. I need to figure out how to make it work with a time value in the time range. I'm sure it is simple. Feel free to take care of it and push on this branch.

library/src/loader_impl.cxx
```diff
@@ -199,10 +195,14 @@ loader& loader_impl::loadGeometry(const std::string& filePath, bool reset)
   this->Internals->GenericImporter->RemoveObservers(vtkCommand::ProgressEvent);
   progressWidget->Off();
 
-  if (this->Internals->Interactor)
+  // Initialize the animation using temporal information from the importer
+  this->Internals->AnimationManager.Initialize(&this->Internals->Options, &this->Internals->Window,
+    this->Internals->Interactor, this->Internals->GenericImporter);
+
+  double animationTime = this->Internals->Options.getAsDouble("scene.animation.time");
+  if (animationTime != 0)
```
mwestphal, line 203: Since we can only know if an option was changed or if it is default by comparing the default value, this has a very small side effect that if an animation has a time range that include 0 inside, we wont be able to load the time value zero, it will just be as if no time value was set, which can be different with the generic importer and certain specificaly crafted exodus file.

I dont think it matters and I have no fix anyway, at least until we have a concept of "isDefault" in the options.

mwestphal:
@snoyer please review

library/src/animationManager.cxx
```diff
@@ -98,39 +98,46 @@ void animationManager::Initialize(
...
+      // Accumulate time ranges
+      if (timeRange[0] < this->TimeRange[0])
       {
-        this->TimeSteps.insert(timeSteps->GetValue(i));
+        this->TimeRange[0] = timeRange[0];
+      }
+      if (timeRange[1] > this->TimeRange[1])
+      {
+        this->TimeRange[1] = timeRange[1];
       }
```
snoyer, line 132:
```cpp
timeRange[0] = std::min(timeRange[0], this->TimeRange[0]);
timeRange[1] = std::max(timeRange[1], this->TimeRange[1]);
```
more explicit?

library/src/animationManager.cxx
```diff
@@ -184,24 +201,62 @@ void animationManager::ToggleAnimation()
...
+  // elapsedTime can be negative
+  this->CurrentTime += elapsedTime;
+  if (this->CurrentTime < this->TimeRange[0])
+  {
+    this->CurrentTime = this->TimeRange[1];
+  }
+  else if (this->CurrentTime > this->TimeRange[1])
+  {
+    this->CurrentTime = this->TimeRange[0];
+  }
```
snoyer, line 226: Shouldn't the delta past the bounds be added when wrapping around?
If so can the 2 `if` be replaced by a modulo operation?

doc/libf3d/OPTIONS.md
```diff
@@ -17,7 +17,9 @@ See the exhaustive list below, but note that this may change in the future.
 
 Option|Type<br>Default<br>Trigger|Description|F3D option
 :---:|:---:|:---|:---:
-scene.animation.index|int<br>0<br>load|Select the animation to load.<br>Any negative value means all animations.<br>The default scene always has at most one animation.|\-\-animation-index
+scene.animation.index|int<br>0<br>load|Select the animation to load.<br>Any negative value means all animations (glTF only).<br>The default scene always has at most one animation.|\-\-animation-index
```
snoyer, line 20: If multiple values are supported why not actually have this argument accept multiple values rather that rely on a magic value?

snoyer (commented):
I have no idea how any of this is expected to work so I did a pass just looking at the diff

doc/libf3d/OPTIONS.md
mwestphal: Two reasons:
1. Only gltf supports multiples animations, not other importers sadly
2. Even if we added support of multiples arguments, being able to show them all in a simple command is practical as this is a standard usecase for many users.

Meakk: I think what @snoyer is suggesting could be useful if there are many animations and you want to enable only a subset.  
Out of the scope of this PR though but we could think about it.

plugins/assimp/module/vtkF3DAssimpImporter.cxx
```diff
@@ -894,9 +894,15 @@ void vtkF3DAssimpImporter::UpdateTimeStep(double timestep)
   }
 
   // get the animation tick
-  timestep *= this->Internals->Scene->mAnimations[0]->mTicksPerSecond;
+  double fps =
+    this->Internals->Scene->mAnimations[this->Internals->ActiveAnimation]->mTicksPerSecond;
+  if (fps == 0)
+  {
+    fps = 1;
```
Meakk, line 901:
```suggestion
  if (fps == 0.)
  {
    fps = 1.;
```

plugins/assimp/module/vtkF3DAssimpImporter.cxx
```diff
@@ -1031,11 +1037,16 @@ bool vtkF3DAssimpImporter::IsAnimationEnabled(vtkIdType animationIndex)
...
-  int& nbTimeSteps, double timeRange[2], vtkDoubleArray* timeSteps)
+bool vtkF3DAssimpImporter::GetTemporalInformation(vtkIdType animationIndex,
+  double vtkNotUsed(frameRate), int& vtkNotUsed(nbTimeSteps), double timeRange[2],
+  vtkDoubleArray* vtkNotUsed(timeSteps))
 {
   double duration = this->Internals->Scene->mAnimations[animationIndex]->mDuration;
   double fps = this->Internals->Scene->mAnimations[animationIndex]->mTicksPerSecond;
+  if (fps == 0)
+  {
+    fps = 1;
```
Meakk, line 1048:
```suggestion
  if (fps == 0.)
  {
    fps = 1.;
```

doc/libf3d/OPTIONS.md
mwestphal: first fixing #841  is needed imo.

library/src/animationManager.cxx
mwestphal: and just better.

library/src/animationManager.cxx
```diff
@@ -212,18 +205,14 @@ void animationManager::Tick()
...
-  }
-  else if (this->CurrentTime > this->TimeRange[1])
-  {
-    this->CurrentTime = this->TimeRange[0];
-  }
+  elapsedTime *= animationSpeedFactor;
+
+  // Modulo computation, compute CurrentTime + elapsedTime in the time range.
+  double delta = this->TimeRange[1] - this->TimeRange[0];
+  this->CurrentTime = fmod(this->CurrentTime - this->TimeRange[0] + fmod(elapsedTime, delta) + delta, delta) + this->TimeRange[0];
```
snoyer:
```suggestion
  this->CurrentTime = fmod(fmod(this->CurrentTime - this->TimeRange[0], delta) + delta, delta) + this->TimeRange[0];
```

We could really do with a proper modulo function defined somewhere (it came up before):
```cpp
double modulo(double a, double b)
{
    const double remainder = fmod(a, b);
    return remainder < 0 ? remainder + b : remainder;
}
```
because the whole `fmod(fmod(a, b)+b, b)` thing is clever but ugly, then:
```suggestion
  this->CurrentTime = this->TimeRange[0] + modulo(this->CurrentTime - this->TimeRange[0], this->TimeRange[1] -  this->TimeRange[0]);
```

Also I think that may not work when exactly at the bounds so it'd need an `if(this->CurrentTime < this->TimeRange[0] || this->CurrentTime > this->TimeRange[1])` first

mwestphal: good point, especially I hurt my brain to write it. Ill take care of that.

mwestphal: > (it came up before)

Do you remember where ? Cant find it.

I will add it locally for now.

snoyer: https://github.com/f3d-app/f3d/blob/master/application/F3DStarter.cxx#L412-L413

mwestphal: > Also I think that may not work when exactly at the bounds

It behave as expected on this->TimeRange[0]

On this->TimeRange[1], it wraps to this->TimeRange[0], which I guess we may not want.

snoyer: > > Also I think that may not work when exactly at the bounds
> 
> It behave as expected on this->TimeRange[0]
> 
> On this->TimeRange[1], it wraps to this->TimeRange[0], which I guess we may not want.

ah yes, because `fmod(x,x) == 0` so the lower bound would be included, but the upper bound would be excluded.

So technically only `if(this->CurrentTime > this->TimeRange[1])` is necessary, but `if(this->CurrentTime < this->TimeRange[0] || this->CurrentTime > this->TimeRange[1])` wouldn't hurt either, up to you

mwestphal: that's what I did :)

mwestphal: > https://github.com/f3d-app/f3d/blob/master/application/F3DStarter.cxx#L412-L413

This one is int based and way less complex, not sure that's worth adding a (private) shared namespace for that yet ?

Meakk (approved):

PR #853, opened 2023-06-13
Initialize lib in instance instead of global

mwestphal:
- Move initialization logic from a global logic to an instance logic
 - Remove a few useless `f3d::`

### DIFF
--- a/library/src/engine.cxx
+++ b/library/src/engine.cxx
@@ -1,7 +1,7 @@
 #include "engine.h"
 
 #include "config.h"
-#include "init.h" // This includes is needed on mac so that the global variable is created
+#include "init.h"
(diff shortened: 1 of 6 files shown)


library/src/engine.cxx
```diff
@@ -36,6 +36,9 @@ class engine::internals
 engine::engine(window::Type windowType)
   : Internals(new engine::internals)
 {
+  // Ensure all lib initialization is done (once)
+  f3d::detail::init::initialize();
```
Meakk, line 40: You don't have to specify `f3d::` since we are already in the namespace

Meakk (approved):
Good to go after minor change

PR #857, opened 2023-06-16
`--camera-zoom-factor` metavar

snoyer:
Found typo when looking up zoom option for recent question on Discord


mwestphal:
Thanks @snoyer !

PR #858, opened 2023-06-16
Support for wider channel size in f3d::image

### LINKED ISSUE #854: Image API do not support hdr/exr
Image API do not support hdr/exr

```
f3d::image("path/to/file.hdr")
```

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff3d6cd37 in ?? () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff3d6cd37 in ?? () from /usr/lib/libc.so.6
#1  0x00007ffff46c4e9a in vtkImageExport::Export (this=0x5555555d0ec0, output=0x7ffff347c010) at /home/glow/dev/vtk/vtk1/src/IO/Image/vtkImageExport.cxx:147
#2  0x00007ffff46948bf in f3d::image::internals::SetFromVTK (this=0x5555555da750, alg=0x5555555d0290) at /home/glow/dev/f3d/f3d/src/library/src/image.cxx:55
#3  0x00007ffff4693ba7 in f3d::image::image (this=0x7fffffffd1a0, path="/home/glow/dev/f3d/f3d/src/testing//data/palermo_park_1k.hdr") at /home/glow/dev/f3d/f3d/src/library/src/image.cxx:84
#4  0x0000555555567dca in TestSDKImage (argc=4, argv=0x7fffffffe7a0) at /home/glow/dev/f3d/f3d/src/library/testing/TestSDKImage.cxx:49
#5  0x000055555555af07 in main (ac=4, av=0x7fffffffe7a0) at /home/glow/dev/f3d/f3d/build/library/testing/libf3dSDKTests.cxx:282
```

Meakk:
Fix #854 

Also, add the architecture required to deprecate libf3d functions.


snoyer:
What does that entail for the data buffer exposed through `unsigned char* image::getData()`? 

For example, in the case of a `float`-based image would `getData()` return the raw encoded floating point values as bytes?

library/src/image.cxx
```diff
@@ -21,25 +25,44 @@ class image::internals
...
+    switch (this->ChannelSize)
+    {
+      case 1:
+        importer->SetDataScalarType(VTK_UNSIGNED_CHAR);
+        break;
+      case 2:
+        importer->SetDataScalarType(VTK_UNSIGNED_SHORT);
+        break;
+      case 4:
+        importer->SetDataScalarType(VTK_FLOAT);
```
mwestphal, line 52: naive question here. Is there any image format of channel size 4 that is not VTK_FLOAT, eg VTK_UNSIGNED_INT ?

library/src/image.cxx
```diff
@@ -21,25 +25,44 @@ class image::internals
...
+        break;
+    }
     importer->SetWholeExtent(0, this->Width - 1, 0, this->Height - 1, 0, 0);
     importer->SetDataExtentToWholeExtent();
     return importer;
   }
 
   void SetFromVTK(vtkAlgorithm* alg)
   {
+    alg->Update();
```
mwestphal, line 64: strange that this was not needed before

library/public/image.h
```diff
@@ -18,6 +18,22 @@ namespace f3d
...
+  {
+    PNG,
+    JPG,
+    TIF,
+    BMP
+  };
+
   /**
    * Create an image from file, the following formats are supported:
    * PNG, PNM, TIFF, BMP, HDR, JPEG, GESigna, MetaImage, TGA.
```
mwestphal, line 39: I dont this that list is up to date ?

library/src/image.cxx
```diff
@@ -208,11 +246,30 @@ bool image::operator!=(const image& reference) const
 }
 
 //----------------------------------------------------------------------------
-void image::save(const std::string& path) const
+void image::save(const std::string& path, Format format) const
 {
   auto importer = this->Internals->GetVTKImporter();
 
-  vtkNew<vtkPNGWriter> writer;
+  vtkSmartPointer<vtkImageWriter> writer;
```
mwestphal, line 253: no vtkImageWriterFactory I suppose ?

mwestphal: no way to use the image exporter here ?

library/src/window_impl.cxx
```diff
@@ -418,10 +418,12 @@ image window_impl::renderToImage(bool noBackground)
 
   int* dims = exporter->GetDataDimensions();
   int cmp = exporter->GetDataNumberOfScalarComponents();
+  vtkIdType size = exporter->GetInput()->GetScalarSize();
```
mwestphal, line 421: what format does this code supports by the way ?

library/testing/TestSDKImage.cxx
```diff
@@ -46,6 +54,16 @@ int TestSDKImage(int argc, char* argv[])
   {
   }
 
+  // check reading a 32-bits image
+  f3d::image hdrImg(std::string(argv[1]) + "/data/palermo_park_1k.hdr");
```
mwestphal, line 58: missing a exr test

library/src/image.cxx
Meakk: There's nothing in `vtkExporter` to get the scalar type without updating the input algorithm and getting the input data, unfortunately.

library/public/image.h
Meakk: plait-il?

library/src/image.cxx
Meakk: Nop, no factory. The `vtkExporter` is used to get the data from a `vtkImageData`, I don't see how it can be useful.

Meakk:
Nevermind, I have another idea to refactor this class.

library/public/image.h
mwestphal: I dont **think** that list is up to date ?

snoyer:
> > What does that entail for the data buffer exposed through `unsigned char* image::getData()`?
> > For example, in the case of a `float`-based image would `getData()` return the raw encoded floating point values as bytes?
> 
> Just to answer the question. Usually, when you have a buffer of an unknown type, you can just use a buffer of bytes and reinterpret_cast the buffer later. You have to be careful about the endianness though.

Alright so if the bindings expose this raw data too, multiplying the length by the new channel width [here](https://github.com/f3d-app/f3d/blob/master/python/F3DPythonBindings.cxx#L41) and [here](
https://github.com/f3d-app/f3d/blob/master/python/F3DPythonBindings.cxx#L51) should do.

Meakk:
You're right! Nice catch :)

Meakk:
@mwestphal I've modified the public API without taking care of backward compatibility. Do you think it's fine?

mwestphal:
Ideally they should be deprecated, if possible by keeping the feature with a warning, if not, by keeping the method but erroring out.

Not saying this will impact a lot of people but we need to think of a process that works for the future of F3D. I know very well the system in VTK, maybe we need something similar ? I can create an issue about it if you want.

library/CMakeLists.txt
```diff
@@ -122,9 +122,14 @@ set_target_properties(libf3d PROPERTIES
 
 # Generate export macros for exported public APIs
 include(GenerateExportHeader)
+option(F3D_EXCLUDE_DEPRECATED "Exclude deprecated functions of the library" OFF)
+if(F3D_EXCLUDE_DEPRECATED)
+  set(libf3d_no_deprecated DEFINE_NO_DEPRECATED)
```
mwestphal: how does that DEFINE_ logic works ?

mwestphal: Awesome

library/testing/TestSDKImage.cxx
mwestphal: still missing

library/testing/TestSDKImage.cxx
```diff
@@ -19,9 +19,18 @@ int TestSDKImage(int argc, char* argv[])
...
-  f3d::image generated;
-  generated.setResolution(width, height).setChannelCount(channels).setData(pixels.data());
+  f3d::image generated(width, height, channels);
+  generated.setData(pixels.data());
+
+  // test save in different formats
   generated.save(std::string(argv[2]) + "TestSDKImage.png");
+  generated.save(std::string(argv[2]) + "TestSDKImage.jpg", f3d::image::SaveFormat::JPG);
+  generated.save(std::string(argv[2]) + "TestSDKImage.tif", f3d::image::SaveFormat::TIF);
+  generated.save(std::string(argv[2]) + "TestSDKImage.bmp", f3d::image::SaveFormat::BMP);
```
mwestphal, line 29: cannot save to exr or hdr ?

library/public/image.h
mwestphal: what about EXR ?

library/testing/TestSDKImage.cxx
Meakk: There's no EXR or HDR writer.

.github/workflows/ci.yml
```diff
@@ -380,13 +380,15 @@ jobs:
         -DCMAKE_PREFIX_PATH:PATH=$(pwd)/../dependencies/install/
         -DF3D_COVERAGE=ON
         -DF3D_MODULE_EXTERNAL_RENDERING=ON
+        -DF3D_MODULE_EXR=ON
```
mwestphal, line 383: rebase on master ? this should not be here

library/src/image.cxx
```diff
@@ -81,7 +71,13 @@ image::image(const std::string& path)
...
   {
     reader->SetFileName(fullPath.c_str());
-    this->Internals->SetFromVTK(reader);
+    reader->Update();
+    this->Internals->Image = reader->GetOutput();
+  }
+
+  if (!this->Internals->Image)
+  {
+    throw read_exception("Cannot read image " + path);
```
mwestphal, line 80: testing this should be trivial

library/src/image.cxx
```diff
@@ -19,48 +22,35 @@ namespace f3d
...
-
-    int* dims = exporter->GetDataDimensions();
-    this->Width = dims[0];
-    this->Height = dims[1];
-    this->Channels = exporter->GetDataNumberOfScalarComponents();
-    this->UpdateBufferSize();
-
-    exporter->Export(this->Buffer.data());
-  }
+  vtkSmartPointer<vtkImageData> Image;
```
mwestphal, line 25: could you explain in a few words why changing the implementation logic is needed ?

library/src/image.cxx
```diff
@@ -121,63 +119,85 @@ image& image::operator=(image&& img) noexcept
...
 
 //----------------------------------------------------------------------------
 image& image::setData(unsigned char* buffer)
 {
-  std::copy(buffer, buffer + this->Internals->Buffer.size(), this->Internals->Buffer.begin());
+  int scalarSize = this->Internals->Image->GetScalarSize();
+  int totalSize = this->getWidth() * this->getHeight() * this->getChannelCount() * scalarSize;
+  unsigned char* internalBuffer =
+    reinterpret_cast<unsigned char*>(this->Internals->Image->GetScalarPointer());
+  std::copy(buffer, buffer + totalSize, internalBuffer);
```
mwestphal, line 182: check for ChannelType ?

library/src/image.cxx
```diff
@@ -121,63 +119,85 @@ image& image::operator=(image&& img) noexcept
...
+    case VTK_UNSIGNED_SHORT:
+      return ChannelType::SHORT;
+    case VTK_FLOAT:
+      return ChannelType::FLOAT;
+  }
+  throw read_exception("Unknown channel type");
+}
 
 //----------------------------------------------------------------------------
 image& image::setData(unsigned char* buffer)
```
mwestphal, line 176: Templated version for different channel types ?

library/src/image.cxx
```diff
@@ -121,63 +119,85 @@ image& image::operator=(image&& img) noexcept
...
+    reinterpret_cast<unsigned char*>(this->Internals->Image->GetScalarPointer());
+  std::copy(buffer, buffer + totalSize, internalBuffer);
   return *this;
 }
 
 //----------------------------------------------------------------------------
 unsigned char* image::getData() const
 {
-  return this->Internals->Buffer.data();
+  return reinterpret_cast<unsigned char*>(this->Internals->Image->GetScalarPointer());
```
mwestphal, line 189: check for channel type ?

library/src/image.cxx
```diff
@@ -187,15 +207,14 @@ bool image::compare(const image& reference, double threshold, image& diff, doubl
...
 
-  if (error <= threshold)
+  if (error > threshold)
   {
-    return true;
-  }
-  else
-  {
-    diff.Internals->SetFromVTK(imDiff);
+    imDiff->Update();
```
mwestphal, line 212: should not be necessary

python/F3DPythonBindings.cxx
```diff
@@ -68,22 +87,37 @@ PYBIND11_MODULE(f3d, module)
...
+    .value("JPG", f3d::image::SaveFormat::JPG)
+    .value("TIF", f3d::image::SaveFormat::TIF)
+    .value("BMP", f3d::image::SaveFormat::BMP)
+    .export_values();
+
+  py::enum_<f3d::image::ChannelType>(image, "ChannelType")
+    .value("BYTE", f3d::image::ChannelType::BYTE)
+    .value("SHORT", f3d::image::ChannelType::SHORT)
+    .value("FLOAT", f3d::image::ChannelType::FLOAT)
+    .export_values();
```
mwestphal, line 103: python bindings tests needs update

.github/workflows/ci.yml
Meakk: Already rebased on master, it was missing so I think nothing about EXR was actually tested for coverage.

mwestphal: it is: https://github.com/f3d-app/f3d/blob/e3d0dd398c9f4f5bec1f935915248e8e17170e7b/.github/workflows/ci.yml#L170

Meakk: line 383 vs line 170 🤨

Meakk: The variable is enabled for the regular CI, but it wasn't enabled for the coverage.

library/src/image.cxx
Meakk: there's no way to know what's inside the user buffer

library/src/image.cxx
Meakk: I think it is

library/src/image.cxx
Meakk: I changed it to a generic `void*`. No need to have templated versions, it accepts any pointer now.

library/src/image.cxx
Meakk: it returns a void* now.

library/src/image.cxx
Meakk: I was managing the buffer, resolution, and number of channels. Adding the scalar type was too much work and I realized that I could just rely on `vtkImageData` which already supports everything we need.

mwestphal: right, VTK rules after all :)

library/src/image.cxx
mwestphal: right, in some cases it can be

library/src/image.cxx
```diff
@@ -121,63 +122,89 @@ image& image::operator=(image&& img) noexcept
...
+    case VTK_UNSIGNED_CHAR:
+      return ChannelType::BYTE;
+    case VTK_UNSIGNED_SHORT:
+      return ChannelType::SHORT;
+    case VTK_FLOAT:
+      return ChannelType::FLOAT;
+    default:
+      break;
+  }
+  throw read_exception("Unknown channel type");
```
mwestphal, line 178: I suppose this is not testable ? Would require some VTK trickery to reach.

library/src/image.cxx
```diff
@@ -121,63 +122,89 @@ image& image::operator=(image&& img) noexcept
...
+}
 
 //----------------------------------------------------------------------------
-image& image::setData(unsigned char* buffer)
+image& image::setData(void* buffer)
```
mwestphal, line 182: why changing to void* then casting to unsigned char* later ?

.github/workflows/ci.yml
mwestphal: right!

library/public/image.h
```diff
@@ -49,27 +86,41 @@ class F3D_EXPORT image
...
   ///@{ @name Buffer Data
   /**
    * Set/Get image buffer data.
-   * Its size is expected to be `width * height * channelCount`.
+   * Its size is expected to be `width * height * channelCount * typeSize`.
    */
-  image& setData(unsigned char* buffer);
-  unsigned char* getData() const;
+  image& setData(void* buffer);
+  void* getData() const;
```
mwestphal, line 123: is this breaking retrocompat ? I dont think so apart from a potential cast warning ?

library/src/image.cxx
Meakk: Yeah, too complicated

library/src/image.cxx
Meakk: Because std::copy needs a type.

library/public/image.h
Meakk: Ah yes indeed. People using `getData` will have to cast the returned pointer to the expected pointer type

library/public/image.h
```diff
@@ -49,27 +86,41 @@ class F3D_EXPORT image
...
+   * Get image channel type.
+   * throw an image::read_exception if the type is unknown.
+   */
+  ChannelType getChannelType() const;
+
   ///@{ @name Buffer Data
   /**
    * Set/Get image buffer data.
-   * Its size is expected to be `width * height * channelCount`.
+   * Its size is expected to be `width * height * channelCount * typeSize`.
```
mwestphal, line 120: What is typeSize ? If I understand well it is related to ChannelType but I think we need to provide a API for it, like `getTypeSize` or `getChannelSize`.

library/src/image.cxx
```diff
@@ -121,63 +122,89 @@ image& image::operator=(image&& img) noexcept
...
 {
-  std::copy(buffer, buffer + this->Internals->Buffer.size(), this->Internals->Buffer.begin());
+  unsigned int scalarSize = this->Internals->Image->GetScalarSize();
+  unsigned int totalSize =
+    this->getWidth() * this->getHeight() * this->getChannelCount() * scalarSize;
+  unsigned char* internalBuffer =
+    static_cast<unsigned char*>(this->Internals->Image->GetScalarPointer());
+  std::copy_n(static_cast<unsigned char*>(buffer), totalSize, internalBuffer);
```
mwestphal, line 189: so you are casting to unsigned char so that copy_n as a size to work with but it can be any actual type, is that correct ? In that case using int8_t is maybe more clear.

library/testing/CMakeLists.txt
```diff
@@ -82,6 +82,8 @@ set_target_properties(libf3dSDKTests PROPERTIES
   CXX_VISIBILITY_PRESET hidden
   )
 
+target_compile_definitions(libf3dSDKTests PRIVATE F3D_MODULE_EXR=$<BOOL:${F3D_MODULE_EXR}>)
```
mwestphal, line 85: so do you need to remove it from the config file then ?

library/src/image.cxx
mwestphal: see other discussions

library/testing/CMakeLists.txt
Meakk: I'll do it in another PR

library/src/image.cxx
Meakk: fair enough

library/public/image.h
Meakk: I'll keep `unsigned char*` functions but flag them as deprecated

library/testing/CMakeLists.txt
mwestphal: could you use config in this one and cleanup everything in another ? I'd rather not have a partial use of this.

Meakk: No sorry it's not possible, I don't have access to the config in the test. That's one of the reason I had to change the method.

Meakk: I'd rather do the migration everywhere if you insist, but I think this PR is already big enough

library/public/image.h
```diff
@@ -49,27 +86,50 @@ class F3D_EXPORT image
...
+   * Get image channel type size in bytes.
+   */
+  unsigned int getChannelTypeSize() const;
+
   ///@{ @name Buffer Data
   /**
    * Set/Get image buffer data.
-   * Its size is expected to be `width * height * channelCount`.
+   * Its size is expected to be `width * height * channelCount * typeSize`.
    */
```
mwestphal, line 126: add a note in the doc about deprecated method

library/testing/CMakeLists.txt
mwestphal: lets not do it but add a TODO to not forget about it

python/F3DPythonBindings.cxx
```diff
@@ -237,4 +257,36 @@ PYBIND11_MODULE(f3d, module)
     .def_static("getLibInfo", &f3d::engine::getLibInfo)
     .def_static("getReadersInfo", &f3d::engine::getReadersInfo)
     .def_static("getPluginsList", &f3d::engine::getPluginsList);
+
+// deprecated functions, will be removed in the next major release
```
mwestphal, line 261: in the next major release, F3D v3.0.0

mwestphal (changes requested):
some changes needed

.github/workflows/ci.yml
```diff
@@ -380,13 +380,15 @@ jobs:
...
         -DF3D_MODULE_RAYTRACING=ON
         -DF3D_PLUGINS_STATIC_BUILD=OFF
         -DF3D_PLUGIN_BUILD_ALEMBIC=ON
         -DF3D_PLUGIN_BUILD_ASSIMP=ON
         -DF3D_PLUGIN_BUILD_DRACO=ON
         -DF3D_PLUGIN_BUILD_OCCT=ON
         -DF3D_STRICT_BUILD=ON
+        -DF3D_EXCLUDE_DEPRECATED=ON
```
mwestphal, line 391: should be off then ?

mwestphal (changes requested):

.github/workflows/ci.yml
Meakk: It's only the coverage CI. The main CI has the default value (OFF)

Issue #14, opened 2021-11-12
Add shortcut to rotate view-up by 90degrees


mwestphal:
**Context**
F3D support rotating camera by 90 degrees by pressing dedicated bindings.
Hoever, this does not rotate the view-up. It would be nice to be able to do that.

**Suggested solution**
- Add new bindings in F3D on Ctrl+4 and Ctrl+6 to rotate the view-up +/- 90 degrees.
- Implement the bindings using a command, reloading the scene will be needed
- Update/Add tests 

a bit more info here: https://gitlab.kitware.com/f3d/f3d/-/issues/144

lknknm:
Hello! As stated on Discord, I'm currently tackling this issue.

mwestphal:
@Meakk @lknknm : I'll improve this issue, i just want to make sure I get it right. 

The idea is to have a bindings to change the `--up` direction ? is that correct ?
Being able to switch to +X or -X by pressing a combination of keys.

lknknm:
> @Meakk @lknknm : I'll improve this issue, i just want to make sure I get it right.
> 
> The idea is to have a bindings to change the `--up` direction ? is that correct ? Being able to switch to +X or -X by pressing a combination of keys.

I don't know if it was the intended direction of the original issue, but in my view, we could implement the 90° rotation shortcut as described in another issue I've opened here: https://github.com/f3d-app/f3d/issues/886#issuecomment-1629534293

mwestphal:
@dk1242 any news on this ?

mwestphal:
no feedback, opening to contribution

mwestphal:
This issue should be updated to reflect the up-direction now being not axis-aligned since #2023

wil92:
Is this issue still needed?

mwestphal:
yes, it is.

mwestphal:
Fixed by #2727 , but not as rotation, simply as changing the view up, which is enough.

PR #862, opened 2023-06-20
refactor camera transition code

snoyer:
Refactor camera transition used in the focal point picking feature, separating the camera interpolation and update from the timing and rendering.

This should be functionally identical by itself but would come in handy for future features modifying the camera position such as #14.

### DIFF
--- a/library/private/interactor_impl.h
+++ b/library/private/interactor_impl.h
@@ -84,8 +84,6 @@ public:
   void UpdateRendererAfterInteraction();
 
 private:
-  void AnimateCameraTransition(const std::function<void(camera&, double)>& update_camera);
-
   class internals;
   std::unique_ptr<internals> Internals;
 };
--- a/library/private/interactor_impl.h
+++ b/library/private/interactor_impl.h
@@ -84,6 +84,8 @@ public:
   void UpdateRendererAfterInteraction();
 
 private:
+  void AnimateCameraTransition(std::function<void(camera& cam, double t)> update_camera);
+
   class internals;
   std::unique_ptr<internals> Internals;
 };
--- a/library/private/interactor_impl.h
+++ b/library/private/interactor_impl.h
@@ -84,7 +84,7 @@ public:
   void UpdateRendererAfterInteraction();
 
 private:
-  void AnimateCameraTransition(std::function<void(camera&, double)> update_camera);
+  void AnimateCameraTransition(const std::function<void(camera&, double)>& update_camera);
 
   class internals;
   std::unique_ptr<internals> Internals;
--- a/library/private/interactor_impl.h
+++ b/library/private/interactor_impl.h
@@ -84,7 +84,7 @@ public:
   void UpdateRendererAfterInteraction();
 
 private:
-  void AnimateCameraTransition(std::function<void(camera& cam, double t)> update_camera);
+  void AnimateCameraTransition(std::function<void(camera&, double)> update_camera);
 
   class internals;
   std::unique_ptr<internals> Internals;
--- a/library/src/interactor_impl.cxx
+++ b/library/src/interactor_impl.cxx
@@ -346,35 +346,35 @@ public:
          *     .--.-----------------.picked
          * pos1    pos2
          */
-        const camera_state_t state0 = self->Window.getCamera().getState();
+        const camera_state_t state = self->Window.getCamera().getState();
 
         double focV[3];
-        vtkMath::Subtract(picked, state0.foc.data(), focV); /* foc -> picked */
+        vtkMath::Subtract(picked, state.foc.data(), focV); /* foc -> picked */
 
         double posV[3];
-        vtkMath::Subtract(picked, state0.foc.data(), posV); /* pos -> pos1, parallel to focV */
+        vtkMath::Subtract(picked, state.foc.data(), posV); /* pos -> pos1, parallel to focV */
         if (!self->Style->GetInteractor()->GetShiftKey())
         {
           double v[3];
-          vtkMath::Subtract(state0.foc.data(), state0.pos.data(), v); /* pos -> foc */
-          vtkMath::ProjectVector(focV, v, v);                         /* pos2 -> pos1 */
+          vtkMath::Subtract(state.foc.data(), state.pos.data(), v); /* pos -> foc */
+          vtkMath::ProjectVector(focV, v, v);                       /* pos2 -> pos1 */
           vtkMath::Subtract(posV, v, posV); /* pos -> pos2, keeps on camera plane */
         }
 
-        const auto iterpolateCameraState = [&state0, &focV, &posV](double ratio) -> camera_state_t
+        const auto iterpolateCameraState = [&state, &focV, &posV](double ratio) -> camera_state_t
         {
           return { //
             {
-              state0.pos[0] + posV[0] * ratio,
-              state0.pos[1] + posV[1] * ratio,
-              state0.pos[2] + posV[2] * ratio,
+              state.pos[0] + posV[0] * ratio,
+              state.pos[1] + posV[1] * ratio,
+              state.pos[2] + posV[2] * ratio,
             },
             {
-              state0.foc[0] + focV[0] * ratio,
-              state0.foc[1] + focV[1] * ratio,
-              state0.foc[2] + focV[2] * ratio,
+              state.foc[0] + focV[0] * ratio,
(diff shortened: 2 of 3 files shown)


mwestphal:
Def a good idea

snoyer:
my auto format (called from `vscode`) disagrees with the CI
 ```diff
-  void StartInteractor() { this->VTKInteractor->Start(); }
+  void StartInteractor()
+  {
+    this->VTKInteractor->Start();
+  }
```
Did I break something in my setup or can someone else confirm?

**Comment by snoyer** [comment]
Not sure what's up with untouched [application/F3DOptionsParser.cxx](https://app.codecov.io/gh/f3d-app/f3d/pull/862?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=f3d-app#diff-YXBwbGljYXRpb24vRjNET3B0aW9uc1BhcnNlci5jeHg=) decreasing in coverage... my fault or gremlins in the CI?

**Comment by mwestphal** [comment]
I have the same issue, clang format version différence probably

**Comment by mwestphal** [comment]
Yeah, that one showed up on other pr already, you can ignore

**Comment by Meakk** [comment]
I noticed that issue too. You can ignore it.

**Comment by snoyer** [comment]
> I have the same issue, clang format version différence probably

@mwestphal, looks like `AllowShortFunctionsOnASingleLine: false` in `.clang-format` could solve it but it affects other files too so it should be a separate PR.

File: library/src/interactor_impl.cxx [hunk 1]
```diff
@@ -363,29 +363,13 @@ class interactor_impl::internals
           vtkMath::Subtract(posV, v, posV);             /* pos -> pos2, keeps on camera plane */
         }
 
-        if (self->TransitionDuration > 0)
+        const auto update_camera = [&foc, &focV, &pos, &posV](camera& c, double t)
```
mwestphal, line 366: avoid single letter var

library/src/interactor_impl.cxx
```diff
@@ -408,6 +392,33 @@ class interactor_impl::internals
...
 
+  void AnimateCameraTransition(const std::function<void(camera&, double)>& update_camera)
+  {
+    window& win = this->Window;
+    camera& cam = win.getCamera();
+    const int duration = this->TransitionDuration;
+
+    if (duration > 0)
+    {
+      // TODO can we disable inputs so we don't queue key presses while the animation is running?
```
mwestphal, line 403: there is a flag to disable mouse movement in the interactor, maybe look at this and see it should be expanded/copied into something that fits what you needd ?

snoyer: Would it be fine to leave it as is for now and deal with it when it actually comes up?

Right now the animation is only triggered from mouse clicks which seems to be fine, the problematic case is with key presses: when the key is held down events get queued faster than the animation plays out and they keep going after the key has been released until all the queued key presses are processes.

I believe that will eventually show up when implementing animation for #14

mwestphal: > Would it be fine to leave it as is for now and deal with it when it actually comes up?

Yes, no worries

library/src/interactor_impl.cxx
```diff
@@ -363,29 +363,13 @@ class interactor_impl::internals
           vtkMath::Subtract(posV, v, posV);             /* pos -> pos2, keeps on camera plane */
         }
 
-        if (self->TransitionDuration > 0)
+        const auto update_camera = [&foc, &focV, &pos, &posV](camera& cam, double t)
```
mwestphal, line 366: avoid single letter var :)

library/src/interactor_impl.cxx
mwestphal: please rework the comment not into a question but into a potential action to take in the future

mwestphal (changes requested):
reviewed

library/src/interactor_impl.cxx
snoyer: [precedent](https://github.com/f3d-app/f3d/pull/491#discussion_r1029550176) :)

kinda stuck (again!) on the variable name here, `interpolant` would be the most descriptive but is kind of a drag. Could be `time` but is that confusing as it goes from `0` through `1` without relation to physical time? Any suggestions?

mwestphal: `interpolant` is fine by me, `time` is too. maybe `timeRatio` ?

library/private/interactor_impl.h
```diff
@@ -84,6 +84,8 @@ class interactor_impl : public interactor
   void UpdateRendererAfterInteraction();
 
 private:
+  void AnimateCameraTransition(const std::function<void(camera&, double)>& update_camera);
```
mwestphal, line 87: Add some doc for your fellow F3D developers :)

mwestphal: wait, this method is in `interactor_impl::internals`, so the declaration is not needed anyway

library/src/interactor_impl.cxx
```diff
@@ -408,6 +400,34 @@ class interactor_impl::internals
     this->VTKInteractor->ExitCallback();
   }
 
+  void AnimateCameraTransition(const std::function<void(camera&, double)>& update_camera)
```
mwestphal, line 403: I'm not completely convinced by the API, could you explain the rationale ?

snoyer: The idea is to provide a generic way to transition the camera from one position to the next without having to know too much about the animation itself as that would be handled by the app. 

So here the calling code provides a function that modifies a given camera according to an interpolant ratio in `[0,1]` where `0` is the original position and `1` is the final position.
Then the app will call that function at the appropriate time(s) with the correct ratio value. Right now we call it as fast as we can render but we could include rate limiting, callbacks on each frames, change the duration, easing, ... later without the calling code needing to know.

If the issue is the code itself we could go with `void AnimateCameraTransition(const std::function<camera_state_t(double interpolant)>& interpolate_state)` for a more Functional Programming (pretty! no side effects!) style at the cost of having to build the state object every frame instead of mutating the camera directly.

Not sure that answers your concern, let me know

snoyer: > Do you think it will happen often that one will want to redefined how the interpolation should be done ?

I can see 3 cases that would realistically come up short term:
- the (only) case we have currently is moving the focal point by picking, that's all xyz linear interpolation.
- there's been talk about snapping to nearest axis (or three-quarter view) and that could be done with spherical coords (you'd lerp the radius and angles before going back to xyz)
- if you wanted to animate the current "reset default view" from an arbitrary position you'd probably want quaternions to do it right

mwestphal: thats fair

library/src/interactor_impl.cxx
```diff
@@ -346,46 +346,39 @@ class interactor_impl::internals
          *     .--.-----------------.picked
          * pos1    pos2
          */
-        camera& cam = self->Window.getCamera();
-        const point3_t pos = cam.getPosition();
-        const point3_t foc = cam.getFocalPoint();
+        const camera_state_t state0 = self->Window.getCamera().getState();
```
mwestphal, line 349: state0 -> state ?

mwestphal (approved):

mwestphal:
@Meakk ?

library/src/interactor_impl.cxx
```diff
@@ -408,6 +401,40 @@ class interactor_impl::internals
     this->VTKInteractor->ExitCallback();
   }
 
+  /**
+   * Run a camera transition animation based on a camera state interpolation function.
+   * The provided function will be called with an interpolation parameter
+   * varying from `0.` for the initial to `1.` for the final state;
+   * it shall return an appropriate linearly interpolated `camera_state_t` for any value in between.
+   */
+  void AnimateCameraTransition(const std::function<camera_state_t(double)>& interpolateCameraState)
```
Meakk, line 410: It's usually preferred to use a template parameter instead of `std::function` when it's a private "inlined" function.
See https://rules.sonarsource.com/cpp/tag/bad-practice/RSPEC-5213/

```suggestion
  template<typename Func>
  void AnimateCameraTransition(Func interpolateCameraState)
```

Meakk (approved):
Here's a minor comment. It's not really important, so feel free to ignore if you prefer `std::function`

Meakk (approved):

PR #863, opened 2023-06-20
Add shortcut to rotate camera

lknknm:
Feature to add shortcuts to rotate the camera view (Front, Right, Top). 
This will implement the following set of shortcuts to manipulate camera views:
- `1`, `3` and `7`: Axis movement (Front, Right, Top View);
- `9`: Front isometric view;


library/src/interactor_impl.cxx
```diff
@@ -615,4 +718,49 @@ void interactor_impl::UpdateRendererAfterInteraction()
 {
   this->Internals->Style->UpdateRendererAfterInteraction();
 }
+
+//----------------------------------------------------------------------------
+void interactor_impl::setViewOrbit(const std::string& view)
+{
+  const auto toZup = zUpTransform(self->VTKInteractor);
+  const auto fromZup = toZup->GetInverse();
```
mwestphal, line 726: not sure transfrom to  and from zUp is needed. wdyt @Meakk ?

library/src/interactor_impl.cxx
```diff
@@ -21,16 +21,45 @@
...
+      right[0], right[1], right[2], 0, //
+      fwd[0], fwd[1], fwd[2], 0,       //
+      up[0], up[1], up[2], 0,          //
+      0, 0, 0, 1,                      //
+    };
+    vtkNew<vtkTransform> toZup;
+    toZup->SetMatrix(m);
+    return toZup;
+  }
+  vtkNew<vtkTransform> zUpTransform(vtkRenderWindowInteractor* interactor)
```
mwestphal, line 55: these two methods API is not expected. A single method like this would do the trick:

`void zUpTransform(vtkRenderer* renderer, vtkTransform* transform)`

Then use it like that:

```
vtkNew<vtkTransform> transform;
zUpTransform(randerer, transform);
```

mwestphal: and make it a static method of interactor_impl::internals for simplicity, I think.

library/src/interactor_impl.cxx
```diff
@@ -615,4 +718,49 @@ void interactor_impl::UpdateRendererAfterInteraction()
 {
   this->Internals->Style->UpdateRendererAfterInteraction();
 }
+
+//----------------------------------------------------------------------------
+void interactor_impl::setViewOrbit(const std::string& view)
```
mwestphal, line 723: use an enum class instead of a string param

library/src/interactor_impl.cxx
snoyer: That one was my suggestion, it's a change of basis from the renderer's `right,forward,up` (which is configurable) to `x,y,z` (and then back) so we can just internally work with `+z` as up regardless of the configured up direction.

The current implementation of this PR could do without this transforms approach (using  `renderer->GetEnvironmentUp()` and `renderer->GetEnvironmentRight()` directly for the direction vectors).

But it may be nice to have the more generic change of basis for some more involved math, eg. if we want to manipulate or animate the camera based on azimuth/elevation (we'd want a reliable z up to compute spherical coordinates so we don't have to do the whole `{vector[axes[0]], vector[axes[1]], vector[axes[2]]}` dance to reorder the axes all the time)

library/src/interactor_impl.cxx
```diff
@@ -71,11 +75,85 @@ class interactor_impl::internals
...
+    const double m[16] = {
+      right[0], right[1], right[2], 0, //
+      fwd[0],   fwd[1],   fwd[2],   0, //
+      up[0],    up[1],    up[2],    0, //
+      0,        0,        0,        1, //
+    };
+    vtkNew<vtkTransform> toZup;
+    toZup->SetMatrix(m);
+    return toZup;
+  }
```
snoyer, line 97: what @mwestphal meant was to use an out parameter instead of a return
```suggestion
  static void zUpTransform(vtkRenderer* renderer, vtkTransform* transform)
  {
    const double* up = renderer->GetEnvironmentUp();
    const double* right = renderer->GetEnvironmentRight();
    double fwd[3];
    vtkMath::Cross(right, up, fwd);

    const double m[16] = {
      right[0], right[1], right[2], 0, //
      fwd[0],   fwd[1],   fwd[2],   0, //
      up[0],    up[1],    up[2],    0, //
      0,        0,        0,        1, //
    };
    transform->SetMatrix(m);
  }
```

---

But if we're going that route we may want to go all out and do
```suggestion
  static void zUpTransforms(vtkRenderer* renderer, vtkTransform* to, vtkTransform* from)
  {
    const double* up = renderer->GetEnvironmentUp();
    const double* right = renderer->GetEnvironmentRight();
    double fwd[3];
    vtkMath::Cross(right, up, fwd);

    const double m[16] = {
      right[0], right[1], right[2], 0, //
      fwd[0],   fwd[1],   fwd[2],   0, //
      up[0],    up[1],    up[2],    0, //
      0,        0,        0,        1, //
    };
    to->SetMatrix(m);
    from->SetMatrix(m);
    from->Inverse();
  }
```
calling as:
```cpp
        vtkRenderer* renderer = self->VTKInteractor->GetRenderWindow()->GetRenderers()->GetFirstRenderer();
        vtkNew<vtkTransform> toZup, fromZup;
        zUpTransform(renderer, toZup, fromZup);
```

---

but wait! there's more! if the function is in the class, why not make it a member instead of static and grab the renderer from inside?
```suggestion
  void zUpTransforms(vtkTransform* to, vtkTransform* from)
  {
    vtkRenderer* renderer = this->VTKInteractor->GetRenderWindow()->GetRenderers()->GetFirstRenderer();
    const double* up = renderer->GetEnvironmentUp();
    const double* right = renderer->GetEnvironmentRight();
    double fwd[3];
    vtkMath::Cross(right, up, fwd);

    const double m[16] = {
      right[0], right[1], right[2], 0, //
      fwd[0],   fwd[1],   fwd[2],   0, //
      up[0],    up[1],    up[2],    0, //
      0,        0,        0,        1, //
    };
    to->SetMatrix(m);
    from->SetMatrix(m);
    from->Inverse();
  }
```
```cpp
        vtkNew<vtkTransform> toZup, fromZup;
        self->zUpTransforms(toZup, fromZup);
```

mwestphal: make sense to me :)

doc/user/INTERACTIONS.md
```diff
@@ -48,12 +48,18 @@ Other options can be toggled directly by pressing the following hotkeys:
 
 Note that some hotkeys can be available or not depending on the file being loaded and the F3D configuration.
 
+Camera Hotkeys:
+* `1`: front view camera.
+* `3`: right view camera.
+* `7`: top view camera.
+* `9`: isometric view camera.
```
mwestphal, line 55: do you expect other number to be used for camera in the future ? What behavior will it be ?

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -964,12 +964,16 @@ void vtkF3DRenderer::ConfigureCheatSheet()
...
     cheatSheetText << " SPACE: Play animation if any\n";
     cheatSheetText << " LEFT : Previous file \n";
     cheatSheetText << " RIGHT: Next file \n";
     cheatSheetText << "  UP  : Reload current file \n";
     cheatSheetText << " DOWN : Add files from dir of current file\n";
+    cheatSheetText << "\n 1: Front View camera\n";
+    cheatSheetText << " 3: Right View camera\n";
+    cheatSheetText << " 7: Top View camera\n";
+    cheatSheetText << " 9: Isometric View camera\n";
+    cheatSheetText << " ENTER: Reset camera to initial parameters\n";
```
mwestphal, line 976: you probably needs to update cheatsheet baselines too, the CI will tell you :)

library/private/interactor_impl.h
```diff
@@ -76,6 +76,11 @@ class interactor_impl : public interactor
    */
   void UpdateRendererAfterInteraction();
 
+  /**
+   * Set the view orbit position on the viewport.
+   */
+  virtual void setViewOrbit(const std::string& file) = 0;
```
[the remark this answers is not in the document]
mwestphal: however I dont think this is needed in the header, you can define setViewOrbit in the cxx only

library/src/interactor.cxx
```diff
@@ -31,6 +31,10 @@ const std::vector<std::pair<std::string, std::string> >& interactor::getDefaultI
     { "L", "Increase (+Shift: decrease) lights intensity" },
     { "H", "Toggle cheat sheet display" },
     { "?", "Dump camera state to the terminal" },
+    { "1", "Front View camera" },
+    { "3", "Left View camera" },
+    { "7", "Top View camera" },
+    { "9", "Isometric View camera" },
     { "Escape", "Quit" },
     { "Enter", "Reset camera to initial parameters" },
     { "Space", "Play animation if any" },
```
mwestphal, line 40: reorg as in other location

library/src/interactor_impl.cxx
```diff
@@ -21,16 +21,20 @@
...
+#include "camera.h"
+
 namespace f3d::detail
 {
-class interactor_impl::internals
-{
-public:
+  class interactor_impl::internals
+  {
+  public:
```
mwestphal, line 37: not needed

library/private/interactor_impl.h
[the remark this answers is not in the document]
mwestphal: actually you can just remove this, the method is in internal already,

library/src/interactor_impl.cxx
```diff
@@ -71,6 +75,79 @@ class interactor_impl::internals
...
+    vtkMath::Cross(right, up, fwd);
+    const double m[16] = {
+      right[0], right[1], right[2], 0, //
+      fwd[0],   fwd[1],   fwd[2],   0, //
+      up[0],    up[1],    up[2],    0, //
+      0,        0,        0,        1, //
+    };
+    to->SetMatrix(m);
+    from->SetMatrix(m);
+    from->Inverse();
```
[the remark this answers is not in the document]
mwestphal: I defer to @Meakk though

library/src/interactor_impl.cxx
```diff
@@ -71,6 +75,79 @@ class interactor_impl::internals
...
+      case ViewType::VT_ISOMETRIC:
+        A = { -1, 1, 1 }; fromZup->TransformPoint(up.data(), up.data()); break;
+    }
+
+    fromZup->TransformPoint(A.data(), A.data());
+    newPos[0] = foc[0] + radius * A[0];
+    newPos[1] = foc[1] + radius * A[1];
+    newPos[2] = foc[2] + radius * A[2];
+
+    /* convert coordinatess back to whatever up is according to model/options */
```
mwestphal, line 141: coordinates

library/src/interactor_impl.cxx
```diff
@@ -615,4 +708,5 @@ void interactor_impl::UpdateRendererAfterInteraction()
 {
   this->Internals->Style->UpdateRendererAfterInteraction();
 }
+
```
mwestphal, line 711: not needed

mwestphal:
@lknknm you need to rebase on master and fix potential conflicts, do not hesitate if you need help with that

doc/user/INTERACTIONS.md
lknknm: We could make it similar to Blender so it's familiar with Blender users, but also with some minor changes:
- `1, 3 and 7`: Axis movement (Front, Right, Top View);
- `9`: Front isometric view; 
- `Shift`, `Alt` or `Ctrl` modifiers: Invert views (Rear, Left, Bottom, Rear Isometric);
- `5`: Orthographic toggle.
- `4` and `6`: rotate camera left and right;
- `2` and `8` rotate camera top or bottom;

mwestphal: thanks for sharing, good start in this PR anyway :)

lknknm: I also forgot to mention that we could also be able to snap to axis movement using `Ctrl + mouse camera pan`.

mwestphal: another feature entirely, please open an issue :)

library/src/interactor_impl.cxx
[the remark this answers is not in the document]
lknknm: Waiting @Meakk 's response on this so I can resolve the conversation. So far I've been using @snoyer 's solution.

doc/user/INTERACTIONS.md
lknknm: Should I open 2 issues? One new issue addressing the updated camera shortcuts feature and another one addressing the `Ctrl + mouse camera rotate` feature?

library/src/interactor_impl.cxx
```diff
@@ -71,8 +75,102 @@ class interactor_impl::internals
 #endif
   }
 
+  //----------------------------------------------------------------------------
+  // Method defined to normalize the Z axis so all models are treated temporarily
+  // as Z-up axis models.
+  void zUpTransforms(vtkTransform* to, vtkTransform* from)
```
Meakk, line 81: Replace all types vtkTransform by vtkMatrix3x3

library/src/interactor_impl.cxx
```diff
@@ -71,8 +75,102 @@ class interactor_impl::internals
...
+    const double* right = renderer->GetEnvironmentRight();
+    double fwd[3];
+    vtkMath::Cross(right, up, fwd);
+    const double m[16] = {
+      right[0], right[1], right[2], 0, //
+      fwd[0], fwd[1], fwd[2], 0,       //
+      up[0], up[1], up[2], 0,          //
+      0, 0, 0, 1,                      //
+    };
+    to->SetMatrix(m);
```
Meakk, line 95: Set all elements using `to->SetElement(0, 0, right[0])` (9 calls in total)

library/src/interactor_impl.cxx
```diff
@@ -71,8 +75,102 @@ class interactor_impl::internals
...
+    from->SetMatrix(m);
+    from->Inverse();
```
Meakk, line 97: Copy the matrix and transpose:
```cpp
from->DeepCopy(to);
from->Transpose();
```

library/src/interactor_impl.cxx
```diff
@@ -71,8 +75,102 @@ class interactor_impl::internals
...
+    vtkNew<vtkTransform> toZup, fromZup;
+    self->zUpTransforms(toZup, fromZup);
+    camera& cam = self->Window.getCamera();
+    vector3_t up = { 0, 0, 1 };
+    point3_t pos = cam.getPosition();
+    point3_t foc = cam.getFocalPoint();
+    point3_t Axis, newPos;
+
+    /* convert coords to +Z up */
+    toZup->TransformPoint(pos.data(), pos.data());
```
Meakk, line 122: You will have a matrix3x3 instead of a transform here. Just replace `->TransformPoint` by `->MultiplyPoint`

Meakk (commented):
You can find the doc here: https://vtk.org/doc/nightly/html/classvtkMatrix3x3.html

library/src/interactor_impl.cxx
snoyer:
```cpp
    const double m[9] = {
      right[0], right[1], right[2], //
      fwd[0], fwd[1], fwd[2],       //
      up[0], up[1], up[2],          //
    };
    to->DeepCopy(m);
```

library/src/interactor.cxx
lknknm: I didn't understand what location are you refering to. Since `library/src/interactor.cxx` and `library/VTKExtensions/Rendering/vtkF3DRenderer.cxx` already follow different orders from each other, I decided to group every the new `1, 3, 7, 9` hotkeys with the other character hotkeys in the `interactor.cxx` file. Then `Escape`, `Enter`, etc would follow.

mwestphal (approved):

library/src/interactor_impl.cxx
```diff
@@ -71,8 +75,101 @@ class interactor_impl::internals
...
+    const double dy = foc[1] - pos[1];
+    const double dz = foc[2] - pos[2];
+    double radius = sqrt(dx * dx + dy * dy + dz * dz);
+    switch (view)
+    {
+      case ViewType::VT_FRONT:
+        Axis = { 0, 0, 1 };
+        fromZup->MultiplyPoint(up.data(), up.data());
+        break;
+      case ViewType::VT_BACK:
```
Meakk, line 134: Looks like `VT_BACK` and `VT_LEFT` are not covered by the unit test. Is there a way to include these cases?

library/src/interactor_impl.cxx
```diff
@@ -71,8 +75,101 @@ class interactor_impl::internals
...
+  };
+  static void SetViewOrbit(ViewType view, internals* self)
+  {
+    vtkNew<vtkMatrix3x3> toZup, fromZup;
+    self->zUpTransforms(toZup, fromZup);
+    camera& cam = self->Window.getCamera();
+    vector3_t up = { 0, 0, 1 };
+    point3_t pos = cam.getPosition();
+    point3_t foc = cam.getFocalPoint();
+    point3_t Axis, newPos;
```
Meakk, line 118:
```suggestion
    point3_t axis, newPos;
```

Local variables start with a lowercase char.

library/src/interactor_impl.cxx
```diff
@@ -231,6 +328,22 @@ class interactor_impl::internals
         self->Window.PrintColoringDescription(log::VerboseLevel::INFO);
         self->Window.PrintSceneDescription(log::VerboseLevel::INFO);
         break;
+      case '1':
```
Meakk, line 331: Ok I understand why it's not covered now. Should we keep the enum?  
Maybe just doing:
```cpp
// set front view
self->SetViewOrbit({ 0, 0, 1 }, self);
```
instead of
```cpp
self->SetViewOrbit(ViewType::VT_FRONT, self);
```
would simplify the code and improve the coverage.

Meakk: See discussion on Discord, we can also keep the enum but rmove `VT_BACK` and `VT_LEFT`

library/src/interactor_impl.cxx
```diff
@@ -71,8 +75,101 @@ class interactor_impl::internals
...
+      case ViewType::VT_TOP:
+        Axis = { 0, 1, 0 };
+        break;
+      case ViewType::VT_ISOMETRIC:
+        Axis = { -1, 1, 1 };
+        fromZup->MultiplyPoint(up.data(), up.data());
+        break;
+    }
+
+    fromZup->MultiplyPoint(Axis.data(), Axis.data());
```
snoyer, line 155: The point of the zUp idea was to be able to hardcode the internal math assuming z points up, so the axis values here look off, for example `ViewType::VT_FRONT: Axis = { 0, 0, 1 }` should not have a non-zero z, and `fromZup->MultiplyPoint(Axis.data(), Axis.data())` feels wrong too, because you want to transform your hardcoded values _back to_ the renderer's orientation, not _from_ it.
Looks like that code could work because somehow 2 wrongs do make a right in this case :)

I think it should go something like this instead (untested but hopefully you get the idea enough to fix it):
```suggestion
    switch (view)
    {
      case ViewType::VT_FRONT:
        Axis = { 0, +1, 0 };
        break;
      case ViewType::VT_BACK:
        Axis = { 0, -1, 0 };
        break;
      case ViewType::VT_RIGHT:
        Axis = { +1, 0, 0 };
        break;
      case ViewType::VT_LEFT:
        Axis = { -1, 0, 0 };
        break;
      case ViewType::VT_TOP:
        Axis = { 0, 0, -1 };
        up = { -1, 0, 0 };
        break;
      case ViewType::VT_ISOMETRIC:
        Axis = { -1, -1, -1 };
        break;
    }
    fromZup->MultiplyPoint(up.data(), up.data());
```

All of that needs to be double checked, I meant to build and run locally to do it more thoroughly but I haven't had time yet so here's my educated guess / gut feeling from just looking at the PR instead.

library/src/interactor_impl.cxx
lknknm: The idea was that `self->SetViewOrbit(ViewType::VT_FRONT, self);` could state clearly that `VT_FRONT` is being passed as a parameter when the function is called for example.
In my opinion that would be a bit less cryptic than using `{ 0, 0, 1 }` as a parameter when the function is called inside the big switch case.

Meakk: Alright, sounds good. Then, please just remove the unused enum values.

library/src/interactor_impl.cxx
lknknm: After a few tweaks, it worked. I'll now update this on my new commit to this PR. Thanks for the suggestion!

Meakk (approved):

library/src/interactor_impl.cxx
```diff
@@ -71,8 +75,90 @@ class interactor_impl::internals
...
+        axis = { +1, 0, 0 };
+        break;
+      case ViewType::VT_TOP:
+        axis = { 0, +1, 0 };
+        up = { 0, -1, 0 };
+        break;
+      case ViewType::VT_ISOMETRIC:
+        axis = { -1, +1, +1 };
+        break;
+    }
```
snoyer, line 141: That still looks like your hardcoded values are assuming Y is up not Z

library/src/interactor_impl.cxx
```diff
@@ -71,8 +75,90 @@ class interactor_impl::internals
...
+    fromZup->MultiplyPoint(up.data(), up.data());
+    fromZup->MultiplyPoint(axis.data(), axis.data());
+
+    newPos[0] = foc[0] + radius * axis[0];
+    newPos[1] = foc[1] + radius * axis[1];
+    newPos[2] = foc[2] + radius * axis[2];
+
+    /* convert coordinates back to whatever up is according to model/options */
+    fromZup->MultiplyPoint(newPos.data(), newPos.data());
```
snoyer, line 150: shouldn't need to transform `newPos` because you've just built it above by combining `foc` and `axis` that are both already transformed by `fromZup`

snoyer (commented):
Maybe my suggestion of using a transform was a bit over-complicated...

library/src/interactor_impl.cxx
lknknm: Fixed after @Meakk's PR: https://github.com/f3d-app/f3d/pull/863/commits/58f6f99b8acc4e003345d13825d2fbe6c85ec649

library/src/interactor_impl.cxx
lknknm: Fixed after @Meakk's PR: https://github.com/f3d-app/f3d/pull/863/commits/58f6f99b8acc4e003345d13825d2fbe6c85ec649

Meakk (approved):

mwestphal (approved):
still looks good

PR #867, opened 2023-06-22
Allow python camera's `setPosition`, `setFocalPoint`, `setViewUp` to accept tuples

snoyer:
Also allow `vector3_t` and `point3_t` to be subscriptable  (eg. `point[1]` to get y coord) and iterable (eg. `xyz = tuple(point)` or `x,y,z = point`)

### DIFF
--- a/python/F3DPythonBindings.cxx
+++ b/python/F3DPythonBindings.cxx
@@ -14,13 +14,38 @@
 
 namespace py = pybind11;
 
+template<typename T, size_t S>
+void pyseq_to_array(const py::sequence& seq, std::array<T, S>& arr)
+{
+  const size_t len = py::len(seq);
+  if (len != arr.size())
+    throw py::value_error("wrong length");
+  for (size_t i = 0; i < len; ++i)
+    arr[i] = py::cast<T>(seq[i]);
+}
+
 PYBIND11_MODULE(f3d, module)
 {
   module.doc() = "f3d library bindings";
 
   // types
-  py::class_<f3d::point3_t>(module, "point3_t").def(py::init<double, double, double>());
-  py::class_<f3d::vector3_t>(module, "vector3_t").def(py::init<double, double, double>());
+  py::class_<f3d::point3_t>(module, "point3_t")
+    .def(py::init<double, double, double>())
+    .def("__len__", [](const f3d::point3_t& p) { return p.size(); })
+    .def(
+      "__iter__", [](const f3d::point3_t& p) { return py::make_iterator(p); },
+      py::keep_alive<0, 1>()) /* keep alive while iterator is used */
+    .def("__getitem__", [](const f3d::point3_t& p, int i) { return p[i]; })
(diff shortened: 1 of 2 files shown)


python/F3DPythonBindings.cxx
```diff
@@ -14,13 +14,38 @@
...
-  py::class_<f3d::point3_t>(module, "point3_t").def(py::init<double, double, double>());
-  py::class_<f3d::vector3_t>(module, "vector3_t").def(py::init<double, double, double>());
+  py::class_<f3d::point3_t>(module, "point3_t")
+    .def(py::init<double, double, double>())
+    .def("__len__", [](const f3d::point3_t& p) { return p.size(); })
+    .def(
+      "__iter__", [](const f3d::point3_t& p) { return py::make_iterator(p); },
+      py::keep_alive<0, 1>()) /* keep alive while iterator is used */
+    .def("__getitem__", [](const f3d::point3_t& p, int i) { return p[i]; })
+    .def("__setitem__", [](f3d::point3_t& p, int i, double value) { p[i] = value; });
```
Meakk, line 39: I have no idea what it does but I'll suppose you know what you are doing :)

Meakk (approved):

python/F3DPythonBindings.cxx
snoyer: I think I do but I'm open to be proven wrong :)

- `__getitem__` and `__setitem__` are for subscripting (`x = point[0]`, `point[1] = y`)
- `__iter__` makes the object iterable, the basic usage would be to do `for value in my_collection: ...` which isn't really useful for point/vectors, but it's also used internally when you do `tuple(point)` so in this case you can convert a `point3_t` to a raw tuple in a pythonic way (without needing a `point.to_tuple()` method for example). The language also lets you explode iterables so here you could do `x,y,z = point` to extract the coords in 3 separate variables
- `__len__` is for the built-in `len(obj)` function, here `len(point)` should always return 3. I'm not sure where (or if at all) it is used internally but I left it in anyway, doesn't seem to hurt

snoyer: oh and `py::keep_alive<0, 1>()` is just your regular deep magic incantation you do because the internet said so, as per tradition. Couldn't tell you how that one works or if we _really_ need it.

Meakk: ok thanks!

mwestphal:
Thats great but I'd have hoped pybind11 to be more helping with such cases. thanks a lot @snoyer

mwestphal (approved):

snoyer:
> Thats great but I'd have hoped pybind11 to be more helping with such cases. thanks a lot @snoyer

It does have ["custom type casters"](https://pybind11.readthedocs.io/en/stable/advanced/cast/custom.html) that look like they would be a better alternative at scale but I don't have a good enough handle on the concept right now to implement it that way. I could look into it later.

snoyer:
replaced by #869

PR #869, opened 2023-06-23
Alternative python `point3_t` and `vector3_t` refactor

snoyer:
In the previous episode (#867) we added support for built-in method to the `point3_t` and `vector3_t` classes so they could behave like tuples, and we added overloads to methods so they could accept tuples in place of these classes.

In this alternative installment we drop the bindings for `point3_t` and `vector3_t` altogether and use `pybind`'s type casters to transparently use tuples instead.
We also drop the getter methods with out parameters that won't work anymore (or haven't ever worked like `camera.getViewAngle(angle_out)`).

This approach seems simpler and more straight-forward, `point3_t` and `vector3_t` being simply `array<double, 3>` in c++ it makes sense they'd be simple `(float, float, float)` tuples in python.

The only thing that might be an issue for some is that python tuples are immutable so this wouldn't work anymore:
```py
p = camera.getPosition()  # point3_t auto casts to tuple
p[1] += 10  # TypeError: 'tuple' object does not support item assignment
camera.setPosition(p)  # point3_t auto casts from tuple, won't run cause we crashed above
```
but one could easilly enough do:
```py
p = list(camera.getPosition())  # manually re-convert to list
p[1] += 10  # p is a list now so all good
camera.setPosition(p)  # point3_t auto casts from any Sequence type so a list works
```

Thoughts?

### DIFF
--- a/python/F3DPythonBindings.cxx
+++ b/python/F3DPythonBindings.cxx
@@ -14,25 +14,27 @@
 
 namespace py = pybind11;
 
-namespace PYBIND11_NAMESPACE
-{
-namespace detail
-{
 template<>
-class type_caster<f3d::point3_t>
+class py::detail::type_caster<f3d::point3_t>
 {
 public:
   bool load(handle src, bool convert)
   {
     if (!isinstance<sequence>(src))
+    {
       return false;
+    }
     py::sequence l = reinterpret_borrow<sequence>(src);
     if (l.size() != 3)
+    {
       return false;
+    }
 
     size_t i = 0;
     for (auto it : l)
+    {
       value[i++] = py::cast<double>(it);
+    }
     return true;
   }
 
@@ -45,20 +47,26 @@ public:
 };
 
 template<>
-class type_caster<f3d::vector3_t>
+class py::detail::type_caster<f3d::vector3_t>
 {
 public:
   bool load(handle src, bool convert)
   {
     if (!isinstance<sequence>(src))
+    {
       return false;
+    }
     py::sequence l = reinterpret_borrow<sequence>(src);
     if (l.size() != 3)
+    {
       return false;
+    }
 
     size_t i = 0;
     for (auto it : l)
+    {
       value[i++] = py::cast<double>(it);
+    }
     return true;
   }
 
@@ -69,9 +77,6 @@ public:
 
   PYBIND11_TYPE_CASTER(f3d::vector3_t, const_name("f3d.vector3_t"));
 };
-
-}
-}
 
 PYBIND11_MODULE(f3d, module)
 {
--- a/python/F3DPythonBindings.cxx
+++ b/python/F3DPythonBindings.cxx
@@ -43,7 +43,7 @@ public:
 
   static handle cast(const f3d::point3_t& src, return_value_policy, handle /* parent */)
   {
-    return Py_BuildValue("fff", src[0], src[1], src[2]);
+    return Py_BuildValue("ddd", src[0], src[1], src[2]);
   }
 
   PYBIND11_TYPE_CASTER(f3d::point3_t, const_name("f3d.point3_t"));
@@ -57,7 +57,7 @@ public:
 
   static handle cast(const f3d::vector3_t& src, return_value_policy, handle /* parent */)
   {
-    return Py_BuildValue("fff", src[0], src[1], src[2]);
+    return Py_BuildValue("ddd", src[0], src[1], src[2]);
   }
 
   PYBIND11_TYPE_CASTER(f3d::vector3_t, const_name("f3d.vector3_t"));
--- a/python/F3DPythonBindings.cxx
+++ b/python/F3DPythonBindings.cxx
@@ -14,28 +14,34 @@
 
 namespace py = pybind11;
 
+template<typename T, size_t S>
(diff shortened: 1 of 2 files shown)


python/F3DPythonBindings.cxx
```diff
@@ -14,14 +14,69 @@
 
 namespace py = pybind11;
 
+namespace PYBIND11_NAMESPACE
```
mwestphal, line 17: possible collision ? should be part of f3d namespace ?

python/F3DPythonBindings.cxx
```diff
@@ -14,14 +14,69 @@
...
+    size_t i = 0;
+    for (auto it : l)
+      value[i++] = py::cast<double>(it);
+    return true;
+  }
+
+  static handle cast(const f3d::vector3_t& src, return_value_policy, handle /* parent */)
+  {
+    return Py_BuildValue("fff", src[0], src[1], src[2]);
+  }
```
mwestphal, line 68: add brackets everywhere ?

mwestphal:
this definitely looks good, they only issue is that it breaks retro compat. I think it is fine unless there is a way to keep retro compat for one version with a deprecation system. wdyt @Meakk ?

python/F3DPythonBindings.cxx
snoyer: Has to be part of `pybind11::detail` for the magic to work I think.

In the docs they do
```cpp
namespace PYBIND11_NAMESPACE { namespace detail {
    template <> struct type_caster<MyClass> {
    // ...
    };
}}
```
It may be possible to inline the namespace rather than enclose
```cpp
template<>
class py::detail::type_caster<f3d::point3_t> { /* ... */ };
```
I'll check if it still works

python/F3DPythonBindings.cxx
snoyer: ah yes, I skipped them on the single line blocks, was in python mode I guess :)
Will fix to conform to prescribed non-compact style

python/F3DPythonBindings.cxx
mwestphal: fait enough

python/F3DPythonBindings.cxx
snoyer: `InsertBraces: true` another one for `.clang-format`

Meakk:
Is there any reason we are using tuples instead of lists? The fact we cannot modify tuple values seems a bit annoying.  
Regarding retrocompat, I think that's not too bad, people early adopting python api should be able to do the migration easily.

snoyer:
> Is there any reason we are using tuples instead of lists?

Tuples are fixed length so `tuple[float, float, float]` is a better python type to match `std::array<double, 3>` than `list[float]` (in the same way you wouldn't use `std::vector<double>` for 3d points).

> The fact we cannot modify tuple values seems a bit annoying

The example I completely made up above was to show the minimal change needed to keep an hypothetical legacy C-style `p[1] += 10` working, but if you actually needed to implement "move the camera by 10 along Y" with your python hat on you'd most likely do:
```py
x, y, z = camera.getPosition()  # unpack tuple
camera.setPosition((x, y+10, z))  # pass new tuple
camera.setPosition([x, y+10, z])  # or pass new list if you like square brackets better
```
so unless there's some exciting use-cases I'm not seeing I'm not convinced the immutability is an issue (I'd personally even argue immutable is better regardless, but that's getting subjective).

Meakk:
I wasn't aware of the unpack tuple syntax. I think you're right and that makes sense.

snoyer:
hmmmm...

https://github.com/f3d-app/f3d/actions/runs/5373754632/jobs/9748445792:
```diff
-  bool load(handle src, bool convert)
-  {
-    return load_array(src, convert, value);
-  }
+  bool load(handle src, bool convert) { return load_array(src, convert, value); }
```

https://github.com/f3d-app/f3d/actions/runs/5324148812/jobs/9643034996
```diff
-  void StartInteractor() { this->VTKInteractor->Start(); }
+  void StartInteractor()
+  {
+    this->VTKInteractor->Start();
+  }
```

snoyer:
> does this need a review ?

have at it :)

python/F3DPythonBindings.cxx
```diff
@@ -14,14 +14,59 @@
...
+
+template<>
+class py::detail::type_caster<f3d::point3_t>
+{
+public:
+  bool load(handle src, bool convert) { return load_array(src, convert, value); }
+
+  static handle cast(const f3d::point3_t& src, return_value_policy, handle /* parent */)
+  {
+    return Py_BuildValue("fff", src[0], src[1], src[2]);
```
mwestphal, line 46: fff ? what does this do ?

mwestphal (approved):

python/F3DPythonBindings.cxx
snoyer: It's the format string to make a `(float, float, float)` tuple, see https://docs.python.org/3/c-api/arg.html#building-values
Actually reading the page after looking up the link it looks like `"ddd"` could be more correct. Nice catch!

michaelmigliore (approved):

Meakk (approved):

PR #319, opened 2022-06-20
WIP: Add python packaging

sitic:
Initial implementation of python packaging, looking for feedback if's something you want / would accept.

Goal would be to make f3d easily installable with `pip`. The python packages generated here include both the libf3d (#52) bindings, as well as the `f3d` executable. I advocate for including the `f3d` binary in the python distribution wheel, to me `pip` the fastest & easiest way to install software (especially if I only need it for a specific project).

`pip install .` installs the python package, `pip wheel .` builds a python distribution wheel. I've updated the python test files so that `pytest` is able to run.

I'll note that because pybind11 does [not support](https://github.com/pybind/pybind11/issues/1755) the stable cpython ABI we unfortunately would have to build a `.whl`for each python minor version & OS combination.

- [x] Make `pip install .` compile & install a python package for libf3d
- [x] Add `f3d` executable as a `console_script` (using a python wrapper)
- [ ] Don't use `BUILD_SHARED_LIBS=OFF`
- [ ] Add CI to build platform independent wheels for Linux / MacOS / Windows
- [ ] Complete metadata information in `setup.py`
- [ ] Add CI job to publish on PyPi on release
- [ ] Maintainers: Register https://pypi.org/project/f3d/

PR #871, opened 2023-06-23
Add python packaging

Meakk:


python/packaging/__main__.py
```diff
@@ -0,0 +1,16 @@
...
+
+F3D_BIN_PATH = Path(__file__).parent / "bin" / "f3d"
+if sys.platform == "win32":
+    F3D_BIN_PATH = F3D_BIN_PATH.with_suffix(".exe")
+
+def run(args):
+    return subprocess.call([F3D_BIN_PATH] + args)
+
+def main():
+    return SystemExit(run(sys.argv[1:]))
```
snoyer, line 13:
```suggestion
def main():
    exit(subprocess.call([F3D_BIN_PATH] + sys.argv[1:]))
```

setup.py
```diff
@@ -0,0 +1,57 @@
+import pathlib
+
+from skbuild import setup
+
+def exclude_static_libraries(cmake_manifest):
+    return list(filter(lambda name: not (name.endswith(".a") and not name.endswith(".lib")), cmake_manifest))
```
snoyer, line 6: `not (name.endswith(".a") and not name.endswith(".lib"))` seems redundant at best and a bug/typo at worst, the same name won't end with `.lib` if is already ends with `.a`

Is it meant to check that it doesn't end in neither `.a` nor `.lib`?
```suggestion
    return list(filter(lambda name: not name.endswith(".a") and not name.endswith(".lib"), cmake_manifest))
```
or same with list comprehension
```suggestion
    return [name for name in cmake_manifest if not name.endswith(".a") and not name.endswith(".lib")]
```

setup.py
```diff
@@ -0,0 +1,57 @@
...
+    cmake_args=[  # TODO: enable/disable modules with env variables?
+        "-DF3D_BINDINGS_PYTHON:BOOL=ON",
+        "-DBUILD_SHARED_LIBS:BOOL=OFF",
+        "-DF3D_PLUGINS_STATIC_BUILD:BOOL=ON"
+        "-DF3D_PLUGIN_BUILD_ALEMBIC:BOOL=ON",
+        "-DF3D_PLUGIN_BUILD_ASSIMP:BOOL=ON",
+        "-DF3D_PLUGIN_BUILD_DRACO:BOOL=ON",
+        "-DF3D_PLUGIN_BUILD_EXODUS:BOOL=ON",
+        "-DF3D_PLUGIN_BUILD_OCCT:BOOL=ON",
+        "-DF3D_MODULE_EXR:BOOL=ON",
```
snoyer, line 53: Depending on who the temporary (as all of that would eventually be solved with a proper PyPI build) "target audience" is you may want to scale it down to stuff that we can expect to be obtainable without having to be compiled "by hand"?

For example on Fedora I can get the Assimp deps easy enough with the package manager, but the default VTK that doesn't have EXR.  Not sure about the other plugins.

Meakk: I agree, I think we should remove all optional dependencies and allow activation somehow (env var?)

setup.py
Meakk: I think the idea is to exclude static libs from the final wheel. Not sure about the code, it's coming from the initial PR from @sitic

python/packaging/__main__.py
Meakk: What's the difference?

snoyer: not much really :) However now I'm questioning a lot more about it...

`__main__.py` runs the F3D app a process from a python script, either when running `python -m f3d` or just `f3d` that would have been setup in the local/virtualenv `bin/` directory by `setup.py` (`entry_points={"console_scripts": ["f3d=f3d.__main__:main"]}`).
That works because `setup.py` actually builds the `pybind` module _and the whole f3d app_ and installs it in `site-packages`.

But is that really the goal, or should the python module only contain the libf3d python bindings and not the F3D app?

setup.py
```diff
@@ -0,0 +1,57 @@
...
+    ],
```
snoyer, line 54:
```suggestion
    cmake_args=[
        "-DF3D_BINDINGS_PYTHON:BOOL=ON",
        "-DBUILD_SHARED_LIBS:BOOL=OFF",
        "-DF3D_PLUGINS_STATIC_BUILD:BOOL=ON"
        *(f"-D{key}={val}" for key, val in os.environ.items() if key.startswith("F3D")),
    ],
```
would allow `F3D_PLUGIN_BUILD_ASSIMP=ON F3D_MODULE_EXR=ON pip install .`

setup.py
snoyer: env var suggestion in separate comment

setup.py
snoyer: Actually depending on what the python module should contain (only lib bindings? plugins? app? configs? ...) there may be more to filter out, see other comment

python/packaging/__main__.py
Meakk: Yes, currently it builds everything. We could remove the executable but the initial PR description #319 suggest to keep it.  
We could even rely on a pre-installed libf3d on build only the bindings, it's supported

setup.py
Meakk: Thanks

python/packaging/__main__.py
snoyer: ah yes I've skimmed over the original PR too fast and missed this key bit:

> I advocate for including the f3d binary in the python distribution wheel, to me pip the fastest & easiest way to install software (especially if I only need it for a specific project).

The goal is to use pip as a general package manager / distribution method to deliver both the whole F3D app and the libf3d bindings. I'm not totally sold on the soundness of the idea but it is definitely valid, moving on :)

PR #872, opened 2023-06-23
Build TBB in CI

Meakk:
TBB is currently automatically built by OSPRay in the CI.  
However if the raytracing is disabled, we still need TBB for VTK SMP backend, and for USD soon.  
Currently, we get TBB with brew for macOS where raytracing is disabled, but that's not great.

Let's build it ourself.

### DIFF
--- a/.github/actions/ospray-sb-install-dep/action.yml
+++ b/.github/actions/ospray-sb-install-dep/action.yml
@@ -9,7 +9,7 @@ runs:
       uses: actions/cache@v3
       with:
         path: dependencies/ospray_install
-        key: ospray-sb-v2.7.1-${{runner.os}}-0
+        key: ospray-sb-v2.7.1-${{runner.os}}-1
 
     - name: Checkout ospray
       if: steps.cache-ospray.outputs.cache-hit != 'true'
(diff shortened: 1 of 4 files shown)


.github/actions/ospray-sb-install-dep/action.yml
```diff
@@ -48,7 +48,7 @@ runs:
         -DCMAKE_BUILD_TYPE=Release
         -DCMAKE_INSTALL_PREFIX:PATH=../ospray_install
         -DDOWNLOAD_ISPC=ON
-        -DDOWNLOAD_TBB=ON
+        -DDOWNLOAD_TBB=OFF
         -DINSTALL_DEPENDENCIES=ON
         -DINSTALL_IN_SEPARATE_DIRECTORIES=OFF
```
mwestphal, line 53: TBB_ROOT_DIR ?

mwestphal, line 53: -DCMAKE_PREFIX_PATH:PATH=$(pwd)/../install/

Meakk: It's not working because the variable does not propagate through the subprojects of the superbuild.  
I'll try to use the env variable TBB_ROOT sued in their FindTBB.cmake script.

mwestphal (approved):

PR #899, opened 2023-07-06
Fix F3D build when added as a subproject

Meakk:
If F3D is added as a subproject using `add_subdirectory`, many things are not working because of the usage of `CMAKE_SOURCE_DIR` which refer to the top level directory. Let's use `F3D_SOURCE_DIR` instead.  
FYI the variable `F3D_SOURCE_DIR` is defined automatically when calling `project(F3D)`
Since CMake 3.21 there is a [PROJECT_IS_TOP_LEVEL](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html). We can use that too for the plugins.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -641,9 +641,9 @@ jobs:
         mkdir install
 
     - name: Install VTK dependency
-      uses: ./source/.github/actions/vtk-install-dep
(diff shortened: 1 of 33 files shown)


mwestphal:
Ok but why should F3D support that ?

mwestphal:
If we want to support that we need to test it

mwestphal (approved):

PR #904, opened 2023-07-08
Fix an initialized var with f3d version logic

mwestphal:


cmake/f3dVersion.cmake
```diff
@@ -61,6 +61,7 @@ function(f3d_determine_version source_dir git_command var_prefix)
 endfunction()
 
 # Extracts components from a version string. See f3d_determine_version() for usage.
+set(CMAKE_MATCH_5 "") # Init the patch extra as it can be undefined
```
Meakk, line 64: Should it be moved inside the function?

mwestphal: definitely

Meakk (approved):

PR #916, opened 2023-07-15
Add support for "No file render" testing

mwestphal:
- Add a dedicated env var to change F3D behavior in order to test no file render
- Use that var to add two new tests


application/F3DStarter.cxx
```diff
@@ -320,10 +321,14 @@ int F3DStarter::Start(int argc, char** argv)
       }
     }
 
+    auto forceNoFileRenderVar =
+      F3DSystemTools::GetVectorEnvironnementVariable("CTEST_F3D_NO_FILE_RENDER_TESTING");
+    bool forceNoFileRender = forceNoFileRenderVar.size() > 0 && forceNoFileRenderVar[0] == "1";
```
snoyer, line 326:
```suggestion
    bool forceNoFileRender = !forceNoFileRenderVar.empty() && forceNoFileRenderVar[0] == "1";
```

application/F3DStarter.cxx
```diff
@@ -320,10 +321,14 @@ int F3DStarter::Start(int argc, char** argv)
       }
     }
 
+    auto forceNoFileRenderVar =
+      F3DSystemTools::GetVectorEnvironnementVariable("CTEST_F3D_NO_FILE_RENDER_TESTING");
```
Meakk, line 325: What do you get a vector instead of a simple string?  
Moreover, do we care about the content? I guess just checking if the variable is defined is enough?

application/testing/CMakeLists.txt
```diff
@@ -94,8 +94,13 @@ function(f3d_test)
     set_tests_properties(f3d::${F3D_TEST_NAME} PROPERTIES FIXTURES_REQUIRED f3d::${F3D_TEST_DEPENDS}_FIXTURE)
   endif()
 
+  list(APPEND f3d_test_env_vars "CTEST_F3D_PROGRESS_BAR_TESTING=1")
+  if (F3D_TEST_NO_FILE_RENDER_TESTING)
+    list(APPEND f3d_test_env_vars "CTEST_F3D_NO_FILE_RENDER_TESTING=1")
```
Meakk, line 99: about the naming, the `TESTING` suffix sound a bit redundant with the `CTEST` prefix.

application/testing/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # F3D Testing
 function(f3d_test)
 
-  cmake_parse_arguments(F3D_TEST "DEFAULT_LIGHTS;LONG_TIMEOUT;HDRI;INTERACTION;NO_BASELINE;NO_RENDER;NO_OUTPUT;WILL_FAIL" "NAME;CONFIG;DATA;RESOLUTION;THRESHOLD;REGEXP;REGEXP_FAIL;DEPENDS" "ARGS" ${ARGN})
+  cmake_parse_arguments(F3D_TEST "DEFAULT_LIGHTS;LONG_TIMEOUT;HDRI;INTERACTION;NO_BASELINE;NO_RENDER;NO_OUTPUT;WILL_FAIL;NO_FILE_RENDER_TESTING" "NAME;CONFIG;DATA;RESOLUTION;THRESHOLD;REGEXP;REGEXP_FAIL;DEPENDS" "ARGS" ${ARGN})
```
Meakk, line 4: what about just `NO_DATA` instead of `NO_FILE_RENDER_TESTING`. Since we specify the data with `DATA`, it makes more sense to me.

application/testing/CMakeLists.txt
Meakk: Here are my suggestions:
- `CTEST_F3D_FORCE_PROGRESS_BAR`
- `CTEST_F3D_NO_FILE_RENDER`

application/testing/CMakeLists.txt
mwestphal: That doesnt really explain what does the flag do, does it ? One would need to look at the code to understand.

application/testing/CMakeLists.txt
mwestphal: good suggestion

application/F3DStarter.cxx
mwestphal: fair enough

application/testing/CMakeLists.txt
mwestphal: NO_DATA_FORCE_RENDER ?

Meakk: From a test dev perspective, you don't really care about the details, it's just calling `f3d_test(DATA foo.bar ...)` or `f3d_test(NO_DATA ...)`.  
So I prefer just `NO_DATA`, but `NO_DATA_FORCE_RENDER` is fine too if you prefer.

application/F3DStarter.cxx
```diff
@@ -320,10 +321,12 @@ int F3DStarter::Start(int argc, char** argv)
       }
     }
 
+    char* noDataForceRender = std::getenv("CTEST_F3D_NO_DATA_FORCE_RENDER");
```
Meakk, line 324:
```suggestion
    bool noDataForceRender = vtksys::SystemTools::HasEnv("CTEST_F3D_NO_DATA_FORCE_RENDER");
```

mwestphal: nope, no access to vtksys in application

Meakk: ok

Meakk (approved):

PR #919, opened 2023-07-16
Add HDRI hide feature and improve HDRI performance

mwestphal:
- Add a `--hdri-hide` option to hide the HDRI skybox while using a HDRI
- Add a test
- Improve HDRI computation to rely on cache only when possible
- Improve HDRI cache performance by computing hashsum on file instead of content

### DIFF
--- a/application/F3DOptionsParser.cxx
+++ b/application/F3DOptionsParser.cxx
@@ -331,6 +331,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp2, "roughness", "", "Roughness coefficient (0.0-1.0)", options.getAsDoubleRef("model.material.roughness"), HasDefault::YES, MayHaveConfig::YES, "<roughness>");
     this->DeclareOption(grp2, "metallic", "", "Metallic coefficient (0.0-1.0)", options.getAsDoubleRef("model.material.metallic"), HasDefault::YES, MayHaveConfig::YES, "<metallic>");
     this->DeclareOption(grp2, "hdri", "", "Path to an image file that will be used as a light source", options.getAsStringRef("render.background.hdri"), LocalHasDefaultNo, MayHaveConfig::YES, "<file path>");
+    this->DeclareOption(grp2, "hdri-hide", "", "When using a HDRI, use it as a light source but hide the background skybox", options.getAsBoolRef("render.background.hdri.hide"), HasDefault::YES, MayHaveConfig::YES);
     this->DeclareOption(grp2, "texture-matcap", "", "Path to a texture file containing a material capture", options.getAsStringRef("model.matcap.texture"), LocalHasDefaultNo, MayHaveConfig::YES, "<file path>");
     this->DeclareOption(grp2, "texture-base-color", "", "Path to a texture file that sets the color of the object", options.getAsStringRef("model.color.texture"), LocalHasDefaultNo, MayHaveConfig::YES, "<file path>");
     this->DeclareOption(grp2, "texture-material", "", "Path to a texture file that sets the Occlusion, Roughness and Metallic values of the object", options.getAsStringRef("model.material.texture"), LocalHasDefaultNo, MayHaveConfig::YES, "<file path>");
--- a/library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
+++ b/library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
@@ -648,7 +648,8 @@ void vtkF3DRenderer::ConfigureHDRI()
       hdriTexture->SetInputConnection(textureReader->GetOutputPort());
 
       // 8-bit textures are usually gamma-corrected
-      if (textureReader->GetOutput() && textureReader->GetOutput()->GetScalarType() == VTK_UNSIGNED_CHAR)
+      if (textureReader->GetOutput() &&
+        textureReader->GetOutput()->GetScalarType() == VTK_UNSIGNED_CHAR)
       {
         hdriTexture->UseSRGBColorSpaceOn();
       }
--- a/library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
+++ b/library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
@@ -493,6 +493,21 @@ void vtkF3DRenderer::ConfigureGridUsingCurrentActors()
 }
 
 //----------------------------------------------------------------------------
+void vtkF3DRenderer::ShowHDRISkybox(bool show)
+{
+  if (this->HDRISkyboxVisible != show)
+  {
+    this->HDRISkyboxVisible = show;
+
+    // TODO merely changing the Skybox visibility could be optimized
+    // instead of configuring the HDRI lighting again
+    this->HDRIConfigured = false;
+    this->TextActorsConfigured = false;
+    this->RenderPassesConfigured = false;
+  }
+}
+
+//----------------------------------------------------------------------------
 void vtkF3DRenderer::SetHDRIFile(const std::string& hdriFile)
 {
   // Check HDRI is different than current one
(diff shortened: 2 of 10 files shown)


mwestphal:
@snoyer @Meakk please review

application/F3DOptionsParser.cxx
```diff
@@ -331,6 +331,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp2, "roughness", "", "Roughness coefficient (0.0-1.0)", options.getAsDoubleRef("model.material.roughness"), HasDefault::YES, MayHaveConfig::YES, "<roughness>");
     this->DeclareOption(grp2, "metallic", "", "Metallic coefficient (0.0-1.0)", options.getAsDoubleRef("model.material.metallic"), HasDefault::YES, MayHaveConfig::YES, "<metallic>");
     this->DeclareOption(grp2, "hdri", "", "Path to an image file that will be used as a light source", options.getAsStringRef("render.background.hdri"), LocalHasDefaultNo, MayHaveConfig::YES, "<file path>");
+    this->DeclareOption(grp2, "hdri-hide", "", "When using a HDRI, use it as a light source but hide the background skybox", options.getAsBoolRef("render.background.hdri.hide"), HasDefault::YES, MayHaveConfig::YES);
```
Meakk, line 334: If I'm correct, it's the first option we are using to disable something. Why not something like `show-skybox`? Is it because it would change the default behavior for the previous version users?

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -624,91 +632,124 @@ void vtkF3DRenderer::ConfigureHDRI()
     }
 
     this->GetEnvMapPrefiltered()->HalfPrecisionOff();
+    needHDRIComputation = !lutCacheExists || !shCacheExists || !specCacheExists;
```
Meakk, line 635: In theory it should not happen, but if `lutCacheExists` is false but `shCacheExists` and `specCacheExists` are true, then you do not need HDRI computation.

```suggestion
    needHDRIComputation = !shCacheExists || !specCacheExists;
```

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -624,91 +632,124 @@ void vtkF3DRenderer::ConfigureHDRI()
...
-      table->AddColumn(this->SphericalHarmonics);
+      // TODO Improve vtkOpenGLRenderer to support not having a texture
+      vtkNew<vtkImageData> img;
+      hdriTexture->SetInputData(img);
 
-      vtkNew<vtkXMLTableWriter> writer;
-      writer->SetInputData(table);
-      writer->SetFileName(shCachePath.c_str());
-      writer->Write();
+#if VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 2, 20221220)
```
Meakk, line 664: is it required?

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
mwestphal: good point, I need to check the logic is fine though

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
mwestphal: SphericalHarmonics is only defined if  VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 2, 20221220)

application/F3DOptionsParser.cxx
Meakk: As long as we have a default config with `show-skybox`, I guess we're fine?  
I like our philosophy "disable everything by default and let the user enable what they want".

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
Meakk: Are you sure? I implemented it way before December 2022 :)

mwestphal: yes but without cache support (which seems to require 20221220), what is the point of calling Modified ?

Meakk: You are not going through this code path without cache support because `needHDRIComputation` is true.

application/F3DOptionsParser.cxx
Meakk: I'm afraid we'll run into difficulties if we want to change the options later.  
But I'll not push further, let's stick with that for now.

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -492,6 +495,21 @@ void vtkF3DRenderer::ConfigureGridUsingCurrentActors()
   this->ResetCameraClippingRange();
 }
 
+//----------------------------------------------------------------------------
+void vtkF3DRenderer::ShowHDRISkybox(bool show)
```
Meakk, line 499: Can we have a hotkey to toggle it? That would be awesome!

testing/baselines/TestHDRIHide.png
Meakk, line 1: Can you add a test with `--no-background` too?

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
mwestphal: that would be trivial, but I need to improve perf of HDRI, this will come in the next PR

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
mwestphal: I need to check something because without it 9.0.0 would not build. It does not change anything to check an earlier date than 20221220, at least this way it is coherant that this call is about the cache mechanism

testing/baselines/TestHDRIHide.png
mwestphal: you mean `--hdri` `--no-background` ?

Meakk: Yes, just to check if outputting with no background also work if the skybox is hidden.

application/F3DOptionsParser.cxx
mwestphal: https://github.com/f3d-app/f3d/issues/927

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
mwestphal: are you sure about this ?

```
//------------------------------------------------------------------------------
// Concrete open gl render method.
void vtkOpenGLRenderer::DeviceRender()
{
  vtkTimerLog::MarkStartEvent("OpenGL Dev Render");

  bool computeIBLTextures = !(this->Pass && this->Pass->IsA("vtkOSPRayPass")) &&
    this->UseImageBasedLighting && this->EnvironmentTexture;
  if (computeIBLTextures)
  {
    this->GetEnvMapLookupTable()->Load(this);
    this->GetEnvMapPrefiltered()->Load(this);

    bool useSH = this->UseSphericalHarmonics;

    if (useSH && this->EnvironmentTexture->GetCubeMap())
    {    
      vtkWarningMacro(
        "Cannot compute spherical harmonics of a cubemap, fall back to irradiance texture");
      useSH = false;
    }    

```

to me it looks like a Render is needed, is it not ?

testing/baselines/TestHDRIHide.png
mwestphal: it doesn't work.

 - `--hdri --no-background` -> hdri is visible
 - `--hdri --hdri-hide --no-background` -> black background

Meakk: ok, then it needs to be fixed. Let me know if you need help.

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
Meakk: ok

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
Meakk: Right. Just in case, here's the MR: https://gitlab.kitware.com/vtk/vtk/-/merge_requests/6708

mwestphal:
superseeded by https://github.com/f3d-app/f3d/pull/932

Issue #885, opened 2023-07-01
Add shortcut/hotkeys to rotate camera view


lknknm:
**Is your feature request related to a problem? Please describe.**
As described on the issue #14 and suggested on the PR #863, I'm proposing the following structure to the camera view manipulation hotkeys.

**Describe the solution you'd like**
We could make it similar to Blender so it's familiar with Blender users, but also with some minor changes:
- [x] `1`, `3` and `7`: Axis movement (Front, Right, Top View);
- [x] `9`: Front isometric view;
- [ ] `Shift`, `Alt` or `Ctrl` modifiers: Invert views (Rear, Left, Bottom, Rear Isometric) - ideally, I'd suggest `Shift` key so we could use it as a standard to "reverse" a command as in `L` (light up), `Shift+L` (light down);
- [x] `5`: Orthographic toggle.
- [x] `4` and `6`: rotate camera left and right;
- [ ] `2` and `8` rotate camera top or bottom;

Aditionally, we could also use a camera transition animation for the feature. 

**Describe alternatives you've considered**
In the future we can implement user options to change hotkeys preferences, as previously discussed.

**Contribution**
I'm currently working on the initial stage of this feature under the PR #863.

Da3n0n:
can we have alt + mmb + Drag to snap to relative Ortho view so like exactly same as blender :P

Da3n0n:
unreal engine uses CTRL + MMB Drag to View orthographic view from almost any direction and blender uses Alt + MMB Drag both are similar but i would say the relative implementation from blender is better than UE because you cant view the back using UE directional method because drag down is always bottom view and the relative method seems to be better IMO

Da3n0n:
only just realized someone else has opened this my bad sorry !

mwestphal:
Lets wait for #443

mwestphal:
This can now be implemented.

dreadbalaur:
Hopefully hotkeys `2` and `8` can be implemented soon. If not, I propose remapping one of the current camera keys to allow for ventral and left view cameras.

mwestphal:
2 and 8 can be implemented anytime, as soon as someone pick up this issue :)

mwestphal:
Closing this in favor of individual issues as the level of complexity is actual quite different.
 - https://github.com/f3d-app/f3d/issues/2342
 - https://github.com/f3d-app/f3d/issues/2343

PR #924, opened 2023-07-21
Add Orthographic View Toggle

lknknm:
As suggested on the Issue #885, this PR has the intention to continue developing the camera hotkeys feature.
This will implement the Ortographic View bound to the `5` key of the keyboard/numeric keyboard, using the `vtkCamera::ParallelProjectionOn/Off()` methods.


library/private/camera_impl.h
```diff
@@ -57,6 +57,7 @@ class camera_impl : public camera
   camera& yaw(angle_deg_t angle) override;
   camera& elevation(angle_deg_t angle) override;
   camera& pitch(angle_deg_t angle) override;
+  camera& ortoview() override;
```
Meakk, line 60: English is not my native language but it seems like a `h` is missing.  
Please, someone confirm?
image

library/src/camera_impl.cxx
```diff
@@ -214,6 +214,17 @@ camera& camera_impl::pitch(angle_deg_t angle)
...
 }
 
+//----------------------------------------------------------------------------
+camera& camera_impl::ortoview()
+{
+  vtkCamera* cam = this->GetVTKCamera();
+  if (cam->GetParallelProjection() == 0)
+    cam->ParallelProjectionOn();
+  else
+    cam->ParallelProjectionOff();
```
Meakk, line 224:
```suggestion
  cam->SetParallelProjection(!cam->GetParallelProjection());
```

library/src/camera_impl.cxx
```diff
@@ -214,6 +214,17 @@ camera& camera_impl::pitch(angle_deg_t angle)
   return *this;
 }
 
+//----------------------------------------------------------------------------
+camera& camera_impl::ortoview()
```
Meakk, line 218: Maybe a better name would be `toggleOrthographic()` or `togglePerspective()` to clearly understand calling it again revert to the original state.

Meakk (changes requested):
Please add the Python bindings. You don't have to add the Java and JS bindings tho, except if you really want :)  
You also need tests in both C++ and Python since you are extending the public API.  
An interactive test (with `add_test`) is also required.

mwestphal:
superseeded by https://github.com/f3d-app/f3d/pull/1235

PR #925, opened 2023-07-22
Always use the same renderer

mwestphal:
- The window now handle a single renderer (with coloring) to avoid recreating renderer which can have a very high cost when using HDRIs
-  Add changelog


library/src/window_impl.cxx
```diff
@@ -50,10 +50,11 @@ class window_impl::internals
 
   std::unique_ptr<camera_impl> Camera;
   vtkSmartPointer<vtkRenderWindow> RenWin;
-  vtkSmartPointer<vtkF3DRenderer> Renderer;
+  vtkNew<vtkF3DRendererWithColoring> Renderer;
```
Meakk, line 53: So `vtkF3DRenderer` is never instantiated now? Then maybe we should merge it with `vtkF3DRendererWithColoring`

mwestphal: > So vtkF3DRenderer is never instantiated now?

Correct

> Then maybe we should merge it with vtkF3DRendererWithColoring

I don't think that is worth it. I like the conceptual separation.

Meakk: fair enough

Meakk (approved):

mwestphal:
CI is not clean though ^^

Meakk (approved):

library/src/window_impl.cxx
```diff
@@ -205,36 +206,19 @@ point3_t window_impl::getDisplayFromWorld(const point3_t& worldPoint) const
...
-  {
-    this->Internals->Renderer = vtkSmartPointer<vtkF3DRendererWithColoring>::New();
-  }
-  else if (!withColoring && (renWithColor || !this->Internals->Renderer))
-  {
-    this->Internals->Renderer = vtkSmartPointer<vtkF3DRenderer>::New();
-  }
+  // Hide axis to make sure the renderer can be deleted if needed
+  this->Internals->Renderer->ShowAxis(false);
+  this->Internals->RenWin->RemoveRenderer(this->Internals->Renderer);
```
mwestphal, line 221: this is actually wrong

PR #926, opened 2023-07-23
Fix hdri lighting

mwestphal:
- Fix an issue with HDRI lighting when reloading or changing to another file

### DIFF
--- a/application/testing/CMakeLists.txt
+++ b/application/testing/CMakeLists.txt
@@ -317,10 +317,8 @@ if(VTK_VERSION VERSION_GREATER_EQUAL 9.2.20221220)
   f3d_test(NAME TestHDRIToneMapping DATA suzanne.ply ARGS -t --hdri=${F3D_SOURCE_DIR}/testing/data/palermo_park_1k.hdr HDRI)
   f3d_test(NAME TestInteractionHDRIMove DATA suzanne.ply ARGS --hdri=${F3D_SOURCE_DIR}/testing/data/palermo_park_1k.hdr HDRI INTERACTION THRESHOLD 60) #Shift+MouseRight;
   f3d_test(NAME TestInteractionHDRIBlur DATA suzanne.ply ARGS --hdri=${F3D_SOURCE_DIR}/testing/data/palermo_park_1k.hdr INTERACTION HDRI DEFAULT_LIGHTS) #U
+  f3d_test(NAME TestInteractionHDRIReload DATA suzanne.ply ARGS --hdri=${F3D_SOURCE_DIR}/testing/data/palermo_park_1k.hdr INTERACTION HDRI DEFAULT_LIGHTS) #Up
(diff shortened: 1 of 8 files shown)


mwestphal:
@Meakk @snoyer please review

application/testing/CMakeLists.txt
```diff
@@ -317,9 +317,11 @@ if(VTK_VERSION VERSION_GREATER_EQUAL 9.2.20221220)
...
   f3d_test(NAME TestInteractionHDRIMove DATA suzanne.ply ARGS --hdri=${F3D_SOURCE_DIR}/testing/data/palermo_park_1k.hdr HDRI INTERACTION THRESHOLD 60) #Shift+MouseRight;
   f3d_test(NAME TestInteractionHDRIBlur DATA suzanne.ply ARGS --hdri=${F3D_SOURCE_DIR}/testing/data/palermo_park_1k.hdr INTERACTION HDRI DEFAULT_LIGHTS) #U
-  f3d_test(NAME TestInteractionHDRIReload DATA suzanne.ply ARGS --hdri=${F3D_SOURCE_DIR}/testing/data/palermo_park_1k.hdr INTERACTION HDRI DEFAULT_LIGHTS) #Up
   f3d_test(NAME TestInteractionHDRIChange DATA multi ARGS --hdri=${F3D_SOURCE_DIR}/testing/data/palermo_park_1k.hdr CONFIG ${F3D_SOURCE_DIR}/testing/configs/complex.json INTERACTION DEFAULT_LIGHTS) #Left
 
+  f3d_test(NAME TestHDRIReloadOutput DATA suzanne.ply ARGS --hdri=${F3D_SOURCE_DIR}/testing/data/palermo_park_1k.hdr HDRI NO_BASELINE DEFAULT_LIGHTS)
+  f3d_test(NAME TestInteractionHDRIReload DATA suzanne.ply ARGS --hdri=${F3D_SOURCE_DIR}/testing/data/palermo_park_1k.hdr --ref=${CMAKE_BINARY_DIR}/Testing/Temporary/TestHDRIReloadOutput.png DEPENDS TestHDRIReloadOutput INTERACTION HDRI NO_BASELINE DEFAULT_LIGHTS) #Up
```
Meakk, line 323: Not a fan of this. I got it, you are generating a baseline and making sure we have the same result. But why not taking the result of `TestHDRIReloadOutput`, save it as the baseline like we do usually?

mwestphal: because what I'm testing here is that the image after and before reload is the same.

> save it as the baseline like we do usually?

Because this what we were doing before and we did not catch the issue. The baseline was just incorrect.

Meakk: So, just update the baseline?

mwestphal: I want to test that the reload output is the same as the not reload output. Dont you think we need to test that ?

mwestphal: If that is not clear here is what was tested before:

 - TestHDRI -> Test that rendering with an HDRI produce an expected baseline
 - TestHDRIReload -> Test that rendering with an HDRI after reloading produce an expected baseline, baseline can be different
 
 Here is what I'm testing now:
 
 - TestHDRI -> Test that rendering with an HDRI produce an expected baseline
 - TestHDRIReload -> Test that rendering with an HDRI after reloading produce that same baseline as TestHDRI

Meakk (approved):

PR #929, opened 2023-07-27
Fix an issue with dropping HDRIs after a model

mwestphal:
- Fix a drop HDRI issue
- Add logic to vtkF3DPolyDataMapper to check for HDRI change in GetNeedToRebuildShaders
- Add a test and changelog


application/testing/CMakeLists.txt
```diff
@@ -559,7 +559,17 @@ if(VTK_VERSION VERSION_GREATER_EQUAL 9.1.20220519) # Drop file test uses stream
     configure_file("${F3D_SOURCE_DIR}/testing/recordings/TestInteractionDropHDRI.log.in"
       "${CMAKE_BINARY_DIR}/TestInteractionDropHDRI.log")
 
-    f3d_test(NAME TestInteractionDropHDRI ARGS "--interaction-test-play=${CMAKE_BINARY_DIR}/TestInteractionDropHDRI.log" HDRI)#X;DropEvent dragon.vtu palermo.hdr;
+    f3d_test(NAME TestInteractionDropHDRI ARGS "--interaction-test-play=${CMAKE_BINARY_DIR}/TestInteractionDropHDRI.log" HDRI)#X;DropEvent dragon.vtu;DropEven palermo.hdr;
+
+    configure_file("${F3D_SOURCE_DIR}/testing/recordings/TestInteractionDropHDRIInvert.log.in"
```
Meakk, line 564: Can be done in another PR but we have a lot of cases like that now, so maybe we could have a flag in `f3d_test` like `CONFIG_INTERACTION_FILE` which automatically configures the file `"${F3D_SOURCE_DIR}/testing/recordings/${F3D_TEST_NAME}.log.in` and call `--interaction-test-play` on it.

Meakk (approved):

application/testing/CMakeLists.txt
mwestphal: very good point

PR #932, opened 2023-07-29
Improve HDRI logic and add new HDRI related options and bindings

mwestphal:
- Decorrelate HDRIFile, HDRI Skybox and IBL
 - Add a simple deprecation framework for F3D and libf3d options
 - Deprecate `--hdri`
 - Add `--hdri-file`, `--hdri-ambient`, `--hdri-skybox`
 - Add bindings to control ambient and skybox
 - Add tests and doc
 - Add dynamic cache path support


library/VTKExtensions/Rendering/vtkF3DRenderer.h
```diff
@@ -267,8 +278,13 @@ class vtkF3DRenderer : public vtkOpenGLRenderer
 
   bool HasValidHDRIFile = false;
   std::string HDRIFile;
+  bool HasValidHDRIHash = false;
   std::string HDRIHash;
+  bool HasValidHDRITexture = false;
   vtkNew<vtkTexture> HDRITexture;
+  bool HasValidHDRILUT = false;
+  bool HasValidHDRISH = false;
+  bool HasValidHDRISpec = false;
```
mwestphal, line 287: `HasValidHDRI*` is different from `HDRI*Configured` in the sense that valid means it is valid while configured means will be set to true whatever happens when Configure method is called.

It is needed to ensure we can turn on/off IBL or Skybox without recomputing anything while still being able to change the HDRI and actually recompute the IBL textures cache and such.

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -512,14 +533,34 @@ void vtkF3DRenderer::SetHDRIFile(const std::string& hdriFile)
...
     this->HDRISkyboxConfigured = false;
   }
 }
 
+//----------------------------------------------------------------------------
+void vtkF3DRenderer::SetUseImageBasedLighting(bool use)
+{
+  if (use != this->GetUseImageBasedLighting())
+  {
+    this->Superclass::SetUseImageBasedLighting(use);
```
mwestphal, line 553: I could define my own variable instead of relying on UseImageBasedLighting, but that doesnt seems necessary

library/src/window_impl.cxx
```diff
@@ -285,11 +285,28 @@ void window_impl::UpdateDynamicOptions()
...
+  bool hdriAmbient = this->Internals->Options.getAsBool("render.hdri.ambient");
+  bool hdriSkybox = this->Internals->Options.getAsBool("render.background.skybox");
+#ifndef F3D_NO_DEPRECATED
+  std::string legacyHDRI = this->Internals->Options.getAsString("render.background.hdri");
+  if (!legacyHDRI.empty())
+  {
+    hdriFile = legacyHDRI;
+    hdriAmbient = true;
+    hdriSkybox = true;
+    // TODO This is a bit spammy, how could we deprecate libf3d options cleanly ?
```
mwestphal, line 301: ideas @Meakk ?

application/testing/CMakeLists.txt
```diff
@@ -306,26 +311,57 @@ endif()
 
 # HDRI test needs https://gitlab.kitware.com/vtk/vtk/-/merge_requests/9767
```
LucasGivord, line 312: so at least vtk 9.2.4?

doc/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## Ongoing development
 
 For F3D users:
+ - Added new options `--hdri-file`, `hdri-ambient`, `hdri-skybox` to have more control on the HDRI behavior
```
LucasGivord, line 6: no need of these extra `--` just `hdri-file` should be good for the changelog, right?

(or add it for the ambient `--hdri-ambient` and skybox `--hdri-skybox`)

doc/user/INTERACTIONS.md
```diff
@@ -42,8 +42,10 @@ Other options can be toggled directly by pressing the following hotkeys:
 * `Z`: the display of the FPS counter.
 * `R`: raytracing.
 * `D`: the denoiser when raytracing.
-* `U`: background blur when using a HDRi.
+* `U`: background blur, useful with an HDRI.
 * `K`: trackball interaction mode.
+* `F`: HDRI ambient lighting.
+* `J`: the display of the HDRI skybox.
```
LucasGivord, line 48: is there any logic behind choosing `F` and `J` as shortcut?

library/VTKExtensions/Rendering/vtkF3DPolyDataMapper.cxx
```diff
@@ -265,18 +265,20 @@ bool vtkF3DPolyDataMapper::GetNeedToRebuildShaders(
   vtkOpenGLHelper& cellBO, vtkRenderer* ren, vtkActor* actor)
 {
   bool ret = this->Superclass::GetNeedToRebuildShaders(cellBO, ren, actor);
-  if (!ret)
+
+  // TODO: Put this logic in VTK
```
LucasGivord, line 269: add an issue to track this remark ?

library/VTKExtensions/Rendering/vtkF3DRenderer.h
```diff
@@ -173,6 +175,14 @@ class vtkF3DRenderer : public vtkOpenGLRenderer
   void ConfigureHDRISkybox();
   ///@}
 
+  ///@{
+  /**
+   * Methods to check if certain HDRI caches are available
```
LucasGivord, line 180: imo you can specify which hdri caches could be considered ?

library/src/image.cxx
```diff
@@ -1,5 +1,6 @@
 #include "image.h"
 
+#include "export.h"
```
LucasGivord, line 3: why?

library/src/options.cxx
```diff
@@ -137,9 +137,11 @@ options::options()
   this->Internals->init("render.effect.ambient-occlusion", false);
   this->Internals->init("render.effect.tone-mapping", false);
 
+  this->Internals->init("render.hdri.file", std::string());
+  this->Internals->init("render.hdri.ambient", false);
   this->Internals->init("render.background.color", std::vector<double>{ 0.2, 0.2, 0.2 });
-  this->Internals->init(
-    "render.background.hdri", std::string()); // XXX This overrides background.color
+  this->Internals->init("render.background.hdri", std::string());
```
LucasGivord, line 143: no warn about it will be deprecated here?

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -296,12 +300,33 @@ void vtkF3DRenderer::ConfigureRenderPasses()
...
-  vtkOSPRayRendererNode::SetBackgroundMode(
-    this->HasValidHDRIFile ? vtkOSPRayRendererNode::Environment : vtkOSPRayRendererNode::Backplate,
-    this);
+  vtkOSPRayRendererNode::BackgroundMode mode = vtkOSPRayRendererNode::Backplate;
+#else
+  int mode = 1;
+#endif
+  if (this->HasValidHDRIFile)
+  {
+    if (this->GetUseImageBasedLighting())
```
LucasGivord, line 309: directly `if (this->HasValidHDRIFile && this->GetUseImageBasedLighting())`

application/testing/CMakeLists.txt
mwestphal: > if(VTK_VERSION VERSION_GREATER_EQUAL 9.2.20221220)

Should be thine one, did you check ?

doc/CHANGELOG.md
mwestphal: good point. wdyt @Meakk ? Keep the dashes ?

doc/user/INTERACTIONS.md
mwestphal: Before this change, only W,F and J were available. I just picked randomly. BTW we have an issue to rework interactions and all the bindings (#443 )

library/VTKExtensions/Rendering/vtkF3DPolyDataMapper.cxx
mwestphal: good point, I have identified many changes that should go in VTK, I will open an issue to track.

library/VTKExtensions/Rendering/vtkF3DRenderer.h
mwestphal: you mean in the comment ? No need, no doxygen doc is being built for this class.

library/src/image.cxx
mwestphal: we are using it below (F3D_NO_DEPRECATED). RN its coming from image.h but I think we should include what we need.

library/src/options.cxx
mwestphal: I cant warn but I should not add it if F3D_NO_DEPRECATED is defined. good find.

@Meakk do you think we should include deprecation logic in f3d::options directly ?

doc/CHANGELOG.md
Meakk: I'd omit the dashes

application/testing/CMakeLists.txt
```diff
@@ -662,13 +703,13 @@ f3d_test(NAME TestInvalidFullScene DATA invalid.gltf ARGS --verbose REGEXP "Load
...
 # Test non existent HDRI, do not add a dummy.png
-f3d_test(NAME TestNonExistentHDRI DATA cow.vtp ARGS --hdri=${F3D_SOURCE_DIR}/testing/data/dummy.png REGEXP "HDRI file does not exist" NO_BASELINE)
+f3d_test(NAME TestNonExistentHDRI DATA cow.vtp ARGS --hdri-file=${F3D_SOURCE_DIR}/testing/data/dummy.png REGEXP "HDRI file does not exist" NO_BASELINE)
 
 # Test non existent interaction file, do not add a TestNonExistentInteraction
 f3d_test(NAME TestNonExistentInteraction DATA cow.vtp INTERACTION REGEXP "Interaction record file to play does not exist" NO_BASELINE)
 
 # Test invalid provided HDRI
-f3d_test(NAME TestInvalidHDRI DATA cow.vtp ARGS --hdri=${F3D_SOURCE_DIR}/testing/data/invalid.png REGEXP "Cannot open HDRI file" NO_BASELINE)
+f3d_test(NAME TestInvalidHDRI DATA cow.vtp ARGS --hdri-file=${F3D_SOURCE_DIR}/testing/data/invalid.png REGEXP "Cannot open HDRI file" NO_BASELINE)
```
Meakk, line 712: same here

doc/CHANGELOG.md
```diff
@@ -13,10 +15,12 @@ For F3D users:
  - Improved overall text contrast for any background color
  - Improved performance when changing model and using a HDRI
  - Improved HDRI cache performance
+ - Deprecated `--hdri` option
 
 For libf3d users:
  - Reworked image API to support many file formats to read (EXR, HDR) and write (PNG, JPG, TIF, BMP)
  - Deprecated previous image API
+ - Deprecated `render.background.hdri` in favor of new options
```
Meakk, line 23: List the new options here.

library/VTKExtensions/Rendering/vtkF3DCachedLUTTexture.h
```diff
@@ -25,11 +25,18 @@ class vtkF3DCachedLUTTexture : public vtkPBRLUTTexture
    */
   void Load(vtkRenderer*) override;
 
+  /**
```
Meakk, line 28: doxygen group tag missing

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -582,33 +648,63 @@ void vtkF3DRenderer::ConfigureHDRITexture()
...
-        // 8-bit textures are usually gamma-corrected
-        if (reader->GetOutput() && reader->GetOutput()->GetScalarType() == VTK_UNSIGNED_CHAR)
+#if VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 2, 20221220)
+        if (this->GetUseImageBasedLighting() && !this->HasValidHDRIHash)
         {
-          this->HDRITexture->UseSRGBColorSpaceOn();
+          // Compute HDRI MD5
+          this->HDRIHash = ::ComputeFileHash(this->HDRIFile);
+          this->HasValidHDRIHash = true;
+          this->CreateCacheDirectory();
```
Meakk, line 658: Since it's already called by `SetCachePath`, do you need it here as well?

library/src/image.cxx
Meakk: It is needed in `image.h` tho because we are using `F3D_EXPORT` so it will never be needed here.

library/src/options.cxx
Meakk: Yes I think so. In the `set()` function, we are already testing if the option is valid. We could test if it's deprecated too.

library/src/window_impl.cxx
Meakk: see above. I think we should catch it when `set`

Meakk (changes requested):

library/src/image.cxx
mwestphal: untils its not and the ifndef logic breaks somewhere else in the file.

application/F3DOptionsParser.cxx
```diff
@@ -330,7 +331,12 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp2, "opacity", "", "Opacity", options.getAsDoubleRef("model.color.opacity"), HasDefault::YES, MayHaveConfig::YES, "<opacity>");
     this->DeclareOption(grp2, "roughness", "", "Roughness coefficient (0.0-1.0)", options.getAsDoubleRef("model.material.roughness"), HasDefault::YES, MayHaveConfig::YES, "<roughness>");
     this->DeclareOption(grp2, "metallic", "", "Metallic coefficient (0.0-1.0)", options.getAsDoubleRef("model.material.metallic"), HasDefault::YES, MayHaveConfig::YES, "<metallic>");
-    this->DeclareOption(grp2, "hdri", "", "Path to an image file that will be used as a light source", options.getAsStringRef("render.background.hdri"), LocalHasDefaultNo, MayHaveConfig::YES, "<file path>");
+#ifndef F3D_NO_DEPRECATED
+    this->DeclareOption(grp2, "hdri", "", "Path to an image file that will be used as a light source and skybox (deprecated)", options.getAsStringRef("render.background.hdri"), LocalHasDefaultNo, MayHaveConfig::YES, "<file path>");
```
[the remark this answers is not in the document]
mwestphal: why would we want to hide it ? If the executable accept it it should be present in the help

application/F3DOptionsParser.cxx
```diff
@@ -772,3 +778,15 @@ void F3DOptionsParser::LoadPlugins(const F3DAppOptions& appOptions) const
 {
   return this->ConfigOptions->LoadPlugins(appOptions);
 }
+
+//----------------------------------------------------------------------------
+void F3DOptionsParser::WarnForDeprecatedOptions(const F3DAppOptions&, const f3d::options& options)
```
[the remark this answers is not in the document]
mwestphal: in F3DOptionsParser::GetOptions ? This will be very spammy

application/testing/CMakeLists.txt
```diff
@@ -44,7 +49,7 @@ function(f3d_test)
   add_test(NAME "f3d::${F3D_TEST_NAME}" COMMAND $<TARGET_FILE:f3d> ${F3D_TEST_ARGS} COMMAND_EXPAND_LISTS)
 
   set(_timeout "30")
-  if(F3D_TEST_LONG_TIMEOUT OR F3D_TEST_INTERACTION OR F3D_TEST_HDRI)
+  if(F3D_TEST_LONG_TIMEOUT OR F3D_TEST_INTERACTION OR F3D_TEST_HDRI_FLAG)
```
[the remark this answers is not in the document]
mwestphal: its not fine at all on macos. We could merge the flags though.

application/F3DOptionsParser.cxx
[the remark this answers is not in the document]
Meakk: If someone is requesting for the help, I don't see the point of suggesting deprecated options.

application/F3DOptionsParser.cxx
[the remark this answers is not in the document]
Meakk: Maybe in `F3DStarter::Start`, just after the call to `GetOptions`?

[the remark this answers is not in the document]
mwestphal: then we need to check the options, why not in a dedicated method ? Indeed it could be namespaced but I like that its in F3DOptionParser instead of being in F3DStarter

application/F3DOptionsParser.cxx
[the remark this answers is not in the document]
mwestphal: if someone is debugging a bash script with --hdri in it, they will be happy to find the it in the help

library/src/options.cxx
mwestphal: that seems doable but the warnings will be spammy

doc/CHANGELOG.md
mwestphal: please note this is not how 2.1.0 release note was written.

application/F3DOptionsParser.cxx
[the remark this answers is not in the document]
mwestphal: > Maybe in F3DStarter::Start, just after the call to GetOptions?

We need to check just after loading the file, as we need to check for deprecated usage in the config files

I stand by the idea of putting that in a dedicated method.

application/testing/CMakeLists.txt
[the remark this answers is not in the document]
mwestphal: We are disable HDRI test on anything but commit CI:

` ${{ matrix.vtk_version == 'commit' && '-DF3D_TESTING_ENABLE_HDRI_TESTS=ON' || '-DF3D_TESTING_ENABLE_HDRI_TESTS=OFF' }}`

I will try a run to see if this is ok to enable on all VTK version but I seem to remember it was not for performance reason.

application/testing/CMakeLists.txt
```diff
@@ -662,13 +703,13 @@ f3d_test(NAME TestInvalidFullScene DATA invalid.gltf ARGS --verbose REGEXP "Load
 f3d_test(NAME TestInvalidTexture DATA cow.vtp ARGS --texture-material=${F3D_SOURCE_DIR}/testing/data/invalid.png REGEXP "Cannot open texture file" NO_BASELINE)
 
 # Test non existent HDRI, do not add a dummy.png
-f3d_test(NAME TestNonExistentHDRI DATA cow.vtp ARGS --hdri=${F3D_SOURCE_DIR}/testing/data/dummy.png REGEXP "HDRI file does not exist" NO_BASELINE)
+f3d_test(NAME TestNonExistentHDRI DATA cow.vtp ARGS --hdri-file=${F3D_SOURCE_DIR}/testing/data/dummy.png REGEXP "HDRI file does not exist" NO_BASELINE)
```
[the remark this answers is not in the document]
mwestphal: that would disable the test on many CI, I dont think we want that.

application/testing/CMakeLists.txt
mwestphal: same

library/src/options.cxx
mwestphal: Done, the output looks like this:

```
Option render.background.hdri is deprecated
Option render.background.hdri is deprecated
Option render.background.hdri is deprecated
--hdri option is deprecated, please use --hdri-file, --hdri-ambient and --hdri-skybox instead.
```

application/testing/CMakeLists.txt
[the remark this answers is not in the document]
mwestphal: Actually the test are not even added with older version. I removed the CMake option.

library/VTKExtensions/Rendering/vtkF3DPolyDataMapper.cxx
mwestphal: https://github.com/f3d-app/f3d/issues/935

library/src/image.cxx
Meakk: Well, "until it's not" is not really an argument. We'll fix it when it's not.  
Anyway, not a big deal. Keep that change if you prefer.

library/src/options.cxx
Meakk: Why is the option set three times?

mwestphal: cause we parse the options (and call the getAsStringRef) three times:
 - command line args
 - global config
 - regex config

application/F3DOptionsParser.cxx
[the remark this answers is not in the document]
Meakk: Good point. Let's keep it then.

library/src/image.cxx
mwestphal: all I mean is that we should include what is being used in the file.

application/F3DOptionsParser.cxx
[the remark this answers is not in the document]
Meakk: OK, then what's the point of declaring a new function with an argument you are not using?

application/testing/CMakeLists.txt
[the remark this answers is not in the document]
Meakk: Maybe we should reduce the size of HDRIs in the testing files to make sure the tests run faster and complete quicker on macOS runners. Then we can get rid of the flag to disable HDRI tests.

application/F3DOptionsParser.cxx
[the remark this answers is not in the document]
mwestphal: not absolutely needed for now, true, but will be needed once we deprecate options from F3DAppOptions. Ok to remove if you prefer.

application/testing/CMakeLists.txt
```diff
@@ -44,7 +49,7 @@ function(f3d_test)
   add_test(NAME "f3d::${F3D_TEST_NAME}" COMMAND $<TARGET_FILE:f3d> ${F3D_TEST_ARGS} COMMAND_EXPAND_LISTS)
 
   set(_timeout "30")
-  if(F3D_TEST_LONG_TIMEOUT OR F3D_TEST_INTERACTION OR F3D_TEST_HDRI)
+  if(F3D_TEST_LONG_TIMEOUT OR F3D_TEST_INTERACTION)
```
Meakk, line 52: Can you remember to me why interaction tests are running slower?

application/testing/CMakeLists.txt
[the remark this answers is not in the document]
mwestphal: TestHDRI without caches is taking 30s on linux. We would need to use a really really small HDRI to be able to run these tests on MacOS, I'm afraid it would reduce the quality of testing.

library/VTKExtensions/Rendering/vtkF3DCachedLUTTexture.cxx
```diff
@@ -51,13 +56,18 @@ void vtkF3DCachedLUTTexture::Load(vtkRenderer* ren)
     vtkImageData* img = reader->GetOutput();
     int dims[3];
     img->GetDimensions(dims);
+    if (dims[0] != dims[1])
+    {
+      vtkWarningMacro("LUT cache has unexpected dimensions");
```
Meakk, line 61: should we have an assert here instead?

library/src/options.cxx
```diff
@@ -29,10 +32,16 @@ class options::internals
   };
 
   template<typename T>
-  void init(const std::string& name, const T& value)
+  void init(const std::string& name, const T& value, [[maybe_unused]] bool deprecated = false)
   {
     static_assert(IsTypeValid<T, OptionVariant>::value);
     this->Options[name] = value;
+#ifndef F3D_NO_DEPRECATED
+    if (deprecated)
```
Meakk, line 40: Is the `deprecated` flag useful here? It seems like it's already checked by the build time condition.

library/VTKExtensions/Rendering/vtkF3DCachedLUTTexture.cxx
mwestphal: nope, this case can happen if a user starts manually puting cache files in the cache path

application/testing/CMakeLists.txt
mwestphal: They just do. They are just running slow. Like one interaction every second.

library/src/options.cxx
mwestphal: the deprecated flag is here to add the option to the deprecated options set.
the F3D_NO_DEPRECATED is here to add this block of code or not.

application/F3DOptionsParser.cxx
```diff
@@ -772,3 +778,15 @@ void F3DOptionsParser::LoadPlugins(const F3DAppOptions& appOptions) const
...
+{
+#ifndef F3D_NO_DEPRECATED
+  if (options.getAsString("render.background.hdri") != "")
```
Meakk, line 786:
```suggestion
  if (!options.getAsString("render.background.hdri").empty())
```

application/F3DOptionsParser.cxx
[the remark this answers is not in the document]
Meakk: yes I think it's better to remove it add we'll add it later if needed.

application/testing/CMakeLists.txt
Meakk: weird

library/src/options.cxx
Meakk: Right, I misread the code.  
Then I think it would be better (easier to read, and more performant) to just create a function in `f3d::internals` called like:
```cpp
bool IsDeprecated(const std::string& name)
{
  // compile time list of deprecated options
  constexpr std::string_view deprecated[] = { "render.background.hdri" };

  auto it = std::find(std::begin(deprecated), std::end(deprecated), name);
  return it != std::end(deprecated);
}
```
And call `internals->IsDeprecated(name)` instead of `this->DeprecatedOptions.count(name) > 0`

mwestphal: I suppose this is compile time find like this ?

Meakk: the find is not compile time because `name` argument is not known at compile time.  
But the list is so the compiler can optimize it out if it's relevant. And no need to store the list in heap.

library/src/options.cxx
LucasGivord: imo it's important so even if it's a bit spammy it's needed and can be acceptable like that, wdyt @Meakk ?

library/src/image.cxx
LucasGivord: I missed it thanks @mwestphal, it's fine for me like that

LucasGivord (approved):
that's nice, approved !

library/src/options.cxx
Meakk: yeah that's fine

Meakk (approved):

PR #938, opened 2023-08-02
Install dependencies

Meakk:
- Move `libf3d` archive (.lib files on Windows) in SDK component
- Add an optional `dependencies` component used to copy runtime dependencies in the install directory
- Specify individual components to install when building python wheels


library/CMakeLists.txt
```diff
@@ -193,10 +193,19 @@ endif()
...
     RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT library
     LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT library)
 
+  # Install dependencies, but exclude system libraries
+  set(F3D_DEPENDENCIES_DIR "" CACHE STRING "Semicolon separated directories for dependencies look up")
+  install(RUNTIME_DEPENDENCY_SET libf3dDeps
+    COMPONENT dependencies EXCLUDE_FROM_ALL
+    PRE_EXCLUDE_REGEXES "api-ms-" "ext-ms-"
+    POST_EXCLUDE_REGEXES ".*system32/.*" "^/usr/lib.*" "^/lib.*" "^/var/lib.*"
+    DIRECTORIES "${F3D_DEPENDENCIES_DIR}")
```
mwestphal, line 207: I'm a bit confused, what files are installed here ?

Meakk: From [CMake](https://cmake.org/cmake/help/latest/command/install.html#runtime-dependency-set):

> The dependencies of targets belonging to a runtime dependency set are installed in the RUNTIME destination and component on DLL platforms, and in the LIBRARY destination and component on non-DLL platforms.

So basically, `objdump`/`dumpbin`/`otool` is run internally to determine dependencies of `libf3d` and copies all of them in `<install>/bin` (Windows) or `<install>/lib` (Linux)  
System libraries are excluded by the regexp.

It's useful when building Python wheels for example where dependencies need to be packaged in the install directory so they are included in the wheel.

mwestphal: I had no idea CMake could do that. I suppose it does not fix the rpath and such making it not relocatable.

In any case, lets thouroughly document this component, I dont want people to use that for anything else but its intended purpose.

doc/dev/BUILD.md
```diff
@@ -60,14 +60,26 @@ Installation can be done through CMake, by running the following command:
...
-* `sdk`: Install the [libf3d](../libf3d/README_LIBF3D.md) SDK.
+Here is the list of all the components:
+
+Name|Installed by default|Operating system|Description
+------|------|------|------
+`application`|YES|ALL|F3D application
+`configuration`|NO|ALL|Default configuration files, `config` and `thumbnail`.
+`library`|YES|ALL|libf3d library binaries
+`plugin`|YES|ALL|libf3d plugins
+`dependencies`|NO|ALL|libf3d runtime dependencies (system libraries excluded)
```
mwestphal, line 77: how does one know if they need to install it ? Maybe explain what it could help with ?

doc/dev/BUILD.md
```diff
@@ -60,14 +60,26 @@ Installation can be done through CMake, by running the following command:
...
+`sdk`|NO|ALL|libf3d SDK (headers and CMake config files)
+`licenses`|YES|ALL|F3D and third party licenses
+`documentation`|YES|Linux|`man` documentation
+`shellext`|YES|Windows/Linux|Desktop integration
```
mwestphal, line 81: linux ?

doc/dev/BUILD.md
```diff
@@ -60,14 +60,26 @@ Installation can be done through CMake, by running the following command:
...
+`pythonmodule`|YES|ALL|Python bindings
+`java`|YES|ALL|Java bindings
```
mwestphal, line 83: should the bindings be exclude from all ? Should be optional imo.

mwestphal: pythonmodule -> python ?

mwestphal (changes requested):
changes needed

doc/dev/BUILD.md
```diff
@@ -60,14 +60,26 @@ Installation can be done through CMake, by running the following command:
...
+`mimetypes`|NO|Linux|Plugins mimetype XML files for integration with Freedesktop
+`assets`|YES|Linux|Assets for integration with Freedesktop
```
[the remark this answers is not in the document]
Meakk: nop

doc/dev/BUILD.md
Meakk: Yes, the Freedesktop integration is under this component

doc/dev/BUILD.md
Meakk: Added more description

doc/dev/BUILD.md
Meakk: I changed the name. It's not optional because the component is not even created if `F3D_BINDINGS_PYTHON` is not enabled anyway.

mwestphal: Right.

mwestphal (approved):

PR #946, opened 2023-08-19
OCC BRep format

snoyer:
> [Open Cascade BRep format](https://dev.opencascade.org/doc/overview/html/specification__brep_format.html) could be useful for CAD users working with OCC-based procedural generation (eg. [`cadquery`](https://github.com/CadQuery/cadquery), [`build123d`](https://github.com/gumyr/build123d)) to F3D pipelines.
> 
> example file here: https://github.com/tpaviot/pythonocc-demos/blob/master/assets/models/Motor-c.brep

I'm not sure I'm doing it right but I copy-pasted my way through to loading a `.brep` file.
Motor-c brep

The tests don't pass on my machine (including the existing `.stp` file, even though the test files load and are visually correct in the app) but they look like they should so here we go anyway :)

### DIFF
--- a/plugins/occt/module/Testing/TestF3DOCCTReader.cxx
+++ b/plugins/occt/module/Testing/TestF3DOCCTReader.cxx
@@ -1,3 +1,5 @@
+#include <cassert>
+
 #include <vtkMultiBlockDataSet.h>
 #include <vtkNew.h>
 #include <vtkTestUtilities.h>
@@ -8,14 +10,23 @@
 
 int TestF3DOCCTReader(int vtkNotUsed(argc), char* argv[])
 {
-  std::string filename = std::string(argv[1]) + "data/f3d.stp";
-  vtkNew<vtkF3DOCCTReader> reader;
-  reader->RelativeDeflectionOn();
-  reader->SetLinearDeflection(0.1);
-  reader->SetAngularDeflection(0.5);
-  reader->ReadWireOn();
-  reader->SetFileName(filename);
-  reader->Update();
-  reader->Print(cout);
-  return reader->GetOutput()->GetNumberOfBlocks() > 0 ? EXIT_SUCCESS : EXIT_FAILURE;
+  for (const auto file : {
+         "data/f3d.stp",
+         "data/f3d.igs",
+         "data/f3d.brep",
+       })
+  {
+    std::string filename = std::string(argv[1]) + file;
+    vtkNew<vtkF3DOCCTReader> reader;
+    reader->RelativeDeflectionOn();
+    reader->SetLinearDeflection(0.1);
+    reader->SetAngularDeflection(0.5);
+    reader->ReadWireOn();
+    reader->SetFileName(filename);
+    reader->Update();
+    reader->Print(cout);
+    assert(reader->GetOutput()->GetNumberOfBlocks() > 0);
+  }
+
+  return EXIT_SUCCESS;
 }
--- a/plugins/occt/module/Testing/TestF3DOCCTReader.cxx
+++ b/plugins/occt/module/Testing/TestF3DOCCTReader.cxx
@@ -4,10 +4,9 @@
 
 #include "vtkF3DOCCTReader.h"
 
-#include <cassert>
 #include <iostream>
 
-void testReader(const std::string& filename, const vtkF3DOCCTReader::FILE_FORMAT& format)
+bool testReader(const std::string& filename, const vtkF3DOCCTReader::FILE_FORMAT& format)
 {
   vtkNew<vtkF3DOCCTReader> reader;
   reader->RelativeDeflectionOn();
@@ -18,13 +17,14 @@ void testReader(const std::string& filename, const vtkF3DOCCTReader::FILE_FORMAT
   reader->SetFileFormat(format);
   reader->Update();
   reader->Print(cout);
-  assert(reader->GetOutput()->GetNumberOfBlocks() > 0);
+  return reader->GetOutput()->GetNumberOfBlocks() > 0;
 }
 
 int TestF3DOCCTReader(int vtkNotUsed(argc), char* argv[])
 {
-  testReader(std::string(argv[1]) + "data/f3d.stp", vtkF3DOCCTReader::FILE_FORMAT::STEP);
-  testReader(std::string(argv[1]) + "data/f3d.igs", vtkF3DOCCTReader::FILE_FORMAT::IGES);
-  testReader(std::string(argv[1]) + "data/f3d.brep", vtkF3DOCCTReader::FILE_FORMAT::BREP);
-  return EXIT_SUCCESS;
+  return (testReader(std::string(argv[1]) + "data/f3d.stp", vtkF3DOCCTReader::FILE_FORMAT::STEP) &&
+           testReader(std::string(argv[1]) + "data/f3d.igs", vtkF3DOCCTReader::FILE_FORMAT::IGES) &&
+           testReader(std::string(argv[1]) + "data/f3d.brep", vtkF3DOCCTReader::FILE_FORMAT::BREP))
+    ? EXIT_SUCCESS
+    : EXIT_FAILURE;
 }
--- a/plugins/occt/module/Testing/TestF3DOCCTReader.cxx
+++ b/plugins/occt/module/Testing/TestF3DOCCTReader.cxx
@@ -4,19 +4,27 @@
 
 #include "vtkF3DOCCTReader.h"
 
+#include <cassert>
 #include <iostream>
 
-int TestF3DOCCTReader(int vtkNotUsed(argc), char* argv[])
+void testReader(const std::string& filename, const vtkF3DOCCTReader::FILE_FORMAT& format)
 {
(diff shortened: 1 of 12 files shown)


plugins/occt/module/Testing/TestF3DOCCTReader.cxx
```diff
@@ -8,14 +10,23 @@
...
+    vtkNew<vtkF3DOCCTReader> reader;
+    reader->RelativeDeflectionOn();
+    reader->SetLinearDeflection(0.1);
+    reader->SetAngularDeflection(0.5);
+    reader->ReadWireOn();
+    reader->SetFileName(filename);
+    reader->Update();
+    reader->Print(cout);
+    assert(reader->GetOutput()->GetNumberOfBlocks() > 0);
+  }
```
mwestphal, line 29: testing all format here are not absolutely needed, as you need to add test in `application/testing/CMakeLists.txt` anyway.

plugins/occt/module/Testing/TestF3DOCCTReader.cxx
```diff
@@ -8,14 +10,23 @@
...
+         "data/f3d.stp",
+         "data/f3d.igs",
+         "data/f3d.brep",
+       })
+  {
+    std::string filename = std::string(argv[1]) + file;
+    vtkNew<vtkF3DOCCTReader> reader;
+    reader->RelativeDeflectionOn();
+    reader->SetLinearDeflection(0.1);
+    reader->SetAngularDeflection(0.5);
```
mwestphal, line 23: you would need a `reader->SetFileFormat(vtkF3DOCCTReader::FILE_FORMAT::BREP);`  and same for iges.

mwestphal (changes requested):

plugins/occt/module/vtkF3DOCCTReader.cxx
```diff
@@ -260,7 +270,10 @@ class vtkF3DOCCTReader::vtkInternals
     polydata->SetLines(linesCells);
 
 #if F3D_PLUGIN_OCCT_XCAF
-    polydata->GetCellData()->SetScalars(colors);
+    if (colors->GetSize() > 0)
+    {
+      polydata->GetCellData()->SetScalars(colors);
+    }
```
mwestphal, line 276: not a big fan of a logic based on colors size. What about checking if color tools is available as soon as possible and not allocating colors at all if not needed ?

plugins/occt/module/vtkF3DOCCTReader.cxx
```diff
@@ -549,6 +596,9 @@ void vtkF3DOCCTReader::PrintSelf(ostream& os, vtkIndent indent)
   os << indent << "AngularDeflection: " << this->AngularDeflection << "\n";
   os << indent << "RelativeDeflection: " << (this->RelativeDeflection ? "true" : "false") << "\n";
   os << indent << "ReadWire: " << (this->ReadWire ? "true" : "false") << "\n";
-  os << indent << "FileFormat: " << (this->FileFormat == FILE_FORMAT::STEP ? "STEP" : "IGES")
+  os << indent << "FileFormat: "
+     << (this->FileFormat == FILE_FORMAT::BREP      ? "BREP"
+            : this->FileFormat == FILE_FORMAT::STEP ? "STEP"
+                                                    : "IGES")
```
mwestphal, line 602: a switch case on top may be a better approach, especially if we add more of them in the future.

plugins/occt/module/vtkF3DOCCTReader.cxx
```diff
@@ -458,6 +471,40 @@ int vtkF3DOCCTReader::RequestData(
...
+      {
+        output->SetBlock(1, polydata);
+      }
+      return 1;
+    }
+    else
+    {
+      vtkErrorWithObjectMacro(this, "Failed opening file " << this->GetFileName());
+      return 0;
+    }
```
mwestphal, line 505: you should be able to test (at least cover it) this easily with a test at VTK module level `TestF3DOCCTReader.cxx`

mwestphal (changes requested):
Nice job! needs some changes still

plugins/occt/module/vtkF3DOCCTReader.h
```diff
@@ -30,14 +30,15 @@ class vtkF3DOCCTReader : public vtkMultiBlockDataSetAlgorithm
...
+    BREP,
     STEP,
     IGES
   };
 
   ///@{
   /**
    * Set the file format to read.
-   * It can be either STEP or IGES.
+   * It can be either BREP, STEP or IGES.
```
mwestphal, line 41: documentation needs an update too, looks for IGES and STEP in the doc.

plugins/occt/module/vtkF3DOCCTReader.h
```diff
@@ -30,14 +30,15 @@ class vtkF3DOCCTReader : public vtkMultiBlockDataSetAlgorithm
...
    * Default is FILE_FORMAT::STEP
```
mwestphal, line 42: also add a changelog entry

plugins/occt/module/vtkF3DOCCTReader.cxx
snoyer: the only initialization I can see is: https://github.com/f3d-app/f3d/blob/895480c57ae56ed8800f64591317db364aa7e29b/plugins/occt/module/vtkF3DOCCTReader.cxx#L90-L92

then insertions in 2 place, both conditional on `if (this->ColorTool)`:
https://github.com/f3d-app/f3d/blob/895480c57ae56ed8800f64591317db364aa7e29b/plugins/occt/module/vtkF3DOCCTReader.cxx#L151
https://github.com/f3d-app/f3d/blob/895480c57ae56ed8800f64591317db364aa7e29b/plugins/occt/module/vtkF3DOCCTReader.cxx#L257

Is there some costly allocation I'm missing somewhere? 

Or are we talking about refactoring the whole thing to de-tangle the colors from the other stuff happening in between so it can be guarded by a single `if (this->ColorTool)`? because that sound like a lot of work for a janky file format that nobody uses ;)

plugins/occt/module/vtkF3DOCCTReader.cxx
snoyer: When I try opening the `.brep` file as `STEP` format to trigger the code block the test picks up the `... vtkF3DOCCTReader.cxx:504    ERR| ...` message as a failure.
Do I need a separate `.cxx` file with some CMake magic to expect failure as success?

plugins/occt/module/vtkF3DOCCTReader.cxx
mwestphal: > Is there some costly allocation I'm missing somewhere?

Nope

> Or are we talking about refactoring the whole thing to de-tangle the colors 

Yes. It is just a suggestion, Perfectly ok not to do it :)

plugins/occt/module/vtkF3DOCCTReader.cxx
mwestphal: Right, I forgot about that.

> Do I need a separate .cxx file with some CMake magic to expect failure as success?

Hum, we do have some examples of that in the application testing. If you had a invalid.brep (just an empty file) and try opening it with F3D, do you hit this code ?

If that is so, you can copy the `TestInvalidFullScene` test, where the test result does not matter as long as the text is present.

snoyer: empty file seems to do the trick, thanks!

plugins/occt/module/Testing/TestF3DOCCTReader.cxx
```diff
@@ -4,18 +4,27 @@
...
   reader->RelativeDeflectionOn();
   reader->SetLinearDeflection(0.1);
   reader->SetAngularDeflection(0.5);
   reader->ReadWireOn();
   reader->SetFileName(filename);
+  reader->SetFileFormat(format);
   reader->Update();
   reader->Print(cout);
-  return reader->GetOutput()->GetNumberOfBlocks() > 0 ? EXIT_SUCCESS : EXIT_FAILURE;
+  assert(reader->GetOutput()->GetNumberOfBlocks() > 0);
```
mwestphal, line 21: that wont test in release mode, you need a return.

plugins/occt/module/vtkF3DOCCTReader.cxx
```diff
@@ -260,7 +270,10 @@ class vtkF3DOCCTReader::vtkInternals
     polydata->SetLines(linesCells);
 
 #if F3D_PLUGIN_OCCT_XCAF
-    polydata->GetCellData()->SetScalars(colors);
+    if (colors->GetSize() > 0)
```
mwestphal, line 273: please add a comment stating that this may be empty if OCCT does not provide a ColorTool

mwestphal (changes requested):
small changes needed

mwestphal (approved):
Nice, lets wait for @Meakk review though before merging.

Meakk (approved):

PR #950, opened 2023-08-20
Add Default HDRI and use it in config file

### LINKED ISSUE #880: Provide a pre-baked HDRI for better default lighting
When not providing an HDRI, lighting in F3D can look dark with reflective surface, this is expected with points lights and PBR reflective materials.

To improve that, we should add an option to light the scene with a prebaked HDRI and make it as cheap as possible to do so.
This option should be turned on in the options.

(alternatively we could always light with it when no HDRI are provided but that would not be very minimalist IMO)

mwestphal:
- Add a embedded default HDRI loaded as needed
 - Use `--hdri-ambient` in default config file to use the default HDRI by default
 - Add and update tests
 - Fix #880


mwestphal:
Small issue in CI:

a

macos takes roughly 2.5 minutes to render with the default HDRI.  
With vtk-commit this is fine thanks to the cache and actually improve our testing because HDRI was not tested on macos before.
With 9.2.6, since cache is not supported, its not great as the check install step render two times.

It's still acceptable (16 minutes CI) but we may want to turn default-hdri for the check install to avoid that maybe and have a dedicated check install with default HDRI for vtk commit only.

Wdyt ? @Meakk

mwestphal:
> Small issue in CI:
> 
> a
> 
> macos takes roughly 2.5 minutes to render with the default HDRI. With vtk-commit this is fine thanks to the cache and actually improve our testing because HDRI was not tested on macos before. With 9.2.6, since cache is not supported, its not great as the check install step render two times.
> 
> It's still acceptable (16 minutes CI) but we may want to turn default-hdri for the check install to avoid that maybe and have a dedicated check install with default HDRI for vtk commit only.
> 
> Wdyt ? @Meakk

Thinking back on this, I think this is critical that we actually do test the default HDRI on all systems and VTK version we support, so lets keep it.

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -626,93 +649,138 @@ void vtkF3DRenderer::ConfigureHDRI()
...
+#if VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 2, 20221220)
+  if (!this->HasValidHDRIHash && this->GetUseImageBasedLighting() && this->HasValidHDRIReader)
   {
-    if (!vtksys::SystemTools::FileExists(this->HDRIFile, true))
+    if (this->UseDefaultHDRI)
     {
-      F3DLog::Print(
-        F3DLog::Severity::Warning, std::string("HDRI file does not exist ") + this->HDRIFile);
+      // Use a random HDRIHash, no need to generate one
+      this->HDRIHash = "fa96df14ed5268a6dbe81478abf8aac2";
```
Meakk, line 706: That's odd. Do you really need to set a hash here? What happens if `HDRIHash` is empty?

testing/baselines/TestConfigFileUpperCase.png
Meakk, line 1: what's wrong with that? The direction of the camera changed

Meakk (changes requested):

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
mwestphal: then you'd need to be able to create an empty directory to store the cache, which is not supported.

Alternatively I could go through the HDRIHashInvalid code path in that case, but that makes it more complex. This is the simplest implementation.

library/VTKExtensions/Rendering/CMakeLists.txt
```diff
@@ -1,3 +1,13 @@
+include("f3dEmbed")
+f3d_embed_file(
+  INPUT "${F3D_SOURCE_DIR}/resources/defaultHDRI.png"
```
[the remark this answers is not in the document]
mwestphal: interesting idea, Ill take a look

testing/baselines/TestConfigFileUpperCase.png
mwestphal: I changed the config file used, it was the wrong one. I's mentionned in the PR description.

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
Meakk: It means you are creating a cache directory for the default HDRI?

mwestphal: yes of course, Id not need a random hash if not. As I said, I could consider that the default HDRI cannot have a hash and force IBL texture computation if you think this is a better approach.

mwestphal: that should work indeed.

Meakk (approved):

Issue #933, opened 2023-07-29
HDRI Skybox is pixelated with raytracing and blur has no effect on it


mwestphal:
HDRI Skybox is pixelated with raytracing and blur has no effect on it

`f3d cow.vtp --hdri=path/to/file.hdr -r`:

`f3d cow.vtp --hdri=path/to/file.hdr -r -u`:

withou raytracing it looks fine

`f3d cow.vtp --hdri=path/to/file.hdr -r`:

`f3d cow.vtp --hdri=path/to/file.hdr -r -u`:

mwestphal:
It is related to the presence of the EnvironnementTexture. If it is removed, the bug disappear

PR #957, opened 2023-08-27
Update vtk commit further

mwestphal:

PR #959, opened 2023-08-28
Fix OSPRay color space

Meakk:
- Fix OSPRay color space change introduced in https://gitlab.kitware.com/vtk/vtk/-/merge_requests/8577
- Rely on OpenGL to draw the background in order to have a smooth skybox and blur support

Fix https://github.com/f3d-app/f3d/issues/933


library/VTKExtensions/Rendering/vtkF3DRenderPass.cxx
```diff
@@ -248,7 +249,20 @@ void vtkF3DRenderPass::Blend(const vtkRenderState* s)
     vtkShaderProgram::Substitute(FSSource, "//VTK::FSQ::Decl", ssDecl.str());
 
     std::stringstream ssImpl;
+
     ssImpl << "  vec4 mainSample = texture(texMain, texCoord);\n";
+
+#if VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 0, 20230824)
```
mwestphal, line 255: without a VTK commit update, this is not tested.

mwestphal: you can take #957

PR #973, opened 2023-09-05
Fix wasm CI

Meakk:
Currently, WASM CI is run 4 times for nothing.  
Fixed it by removing the matrix


mwestphal:
copypaste failure ^^

mwestphal (approved):

PR #974, opened 2023-09-05
Refactor CI and enable Apple Silicon

Meakk:


Meakk:
@mwestphal please review

.github/workflows/ci.yml
```diff
@@ -115,9 +137,44 @@ jobs:
...
+# MacOS CI: Build and test, cross-vtk build matrix
+#----------------------------------------------------------------------------
+  macos:
+    if: github.event.pull_request.draft == false
+    needs: cache_lfs
+
+    strategy:
+      fail-fast: false
+      matrix:
+        vtk_version: [commit, v9.2.6, v9.1.0, v9.0.0]
```
mwestphal, line 152: missing bundle ci

Meakk: right

.github/actions/alembic-install-dep/action.yml
```diff
@@ -41,6 +46,7 @@ runs:
         -DCMAKE_PREFIX_PATH:PATH=../install/
         -DUSE_BINARIES=OFF
         -DUSE_TESTS=OFF
+        -DCMAKE_OSX_ARCHITECTURES=${{ inputs.cpu }}
```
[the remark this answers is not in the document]
Meakk: So in the end I think we should stick with the CMake define

.github/workflows/ci.yml
```diff
@@ -115,9 +137,49 @@ jobs:
...
+      with:
+        path: 'source'
+        fetch-depth: 0
+        lfs: false
+
+    - name: Copy LFS Data
+      uses: ./source/.github/actions/lfs-copy
+      with:
+        lfs_sha: ${{ needs.cache_lfs.outputs.lfs_sha}}
+        workflow_label: 'consumer'
```
mwestphal, line 174: we could consider putting that into the action that needs it instead of duplcating it.

.github/actions/vtk-install-dep/action.yml
```diff
@@ -107,7 +111,7 @@ runs:
         -DVTK_MODULE_ENABLE_VTK_FiltersGeneral=YES
         -DVTK_MODULE_ENABLE_VTK_FiltersGeometry=YES
         -DVTK_MODULE_ENABLE_VTK_IOCityGML=YES
-        -DVTK_MODULE_ENABLE_VTK_IOExodus=YES
+        -DVTK_MODULE_ENABLE_VTK_IOExodus=${{ inputs.cpu == 'x86_64' && 'YES' || 'NO' }}
```
mwestphal, line 114: ideally open an issue about exodus not being support on arm

mwestphal: and link it in this file

.github/actions/vtk-install-dep/action.yml
```diff
@@ -17,6 +17,10 @@ inputs:
     description: 'Label to control egl'
     required: false
     default: 'no-egl'
+  cpu:
+    description: 'Label to control cpu architecture'
```
mwestphal, line 21: Not a `label` (its a concept I created, not someting from github). description could be:

`CPU architecture to build for`

.github/actions/generic-ci/action.yml
```diff
@@ -240,20 +257,21 @@ runs:
...
+      if: inputs.cpu == 'x86_64'
       shell: bash
       working-directory: ${{github.workspace}}
       run: |
         cmake --install build_plugins --config Release
         cmake --install build_plugins --config Release --component configuration
 
     - name: Check Install
-      if: inputs.vtk_version != 'v9.0.0' && inputs.vtk_version != 'v9.1.0'
+      if: inputs.vtk_version != 'v9.0.0' && inputs.vtk_version != 'v9.1.0' && inputs.cpu == 'x86_64' && inputs.cpu == 'x86_64'
```
mwestphal, line 274: duplicated check, also open an issue about the fact the arm cant be tested

.github/actions/generic-ci/action.yml
```diff
@@ -229,7 +245,8 @@ runs:
     - name: Set F3D binary path MacOS bundle
       if: |
         runner.os == 'macOS' &&
-        inputs.bundle_label != 'bundle'
+        inputs.bundle_label != 'bundle' &&
+        inputs.cpu == 'x86_64'
```
mwestphal, line 249: check not needed

.github/actions/generic-ci/action.yml
```diff
@@ -240,20 +257,21 @@ runs:
...
 
     - name: Check F3D_PLUGINS_PATH using plugin example
-      if: inputs.vtk_version != 'v9.0.0' && inputs.vtk_version != 'v9.1.0'
+      if: inputs.vtk_version != 'v9.0.0' && inputs.vtk_version != 'v9.1.0' && inputs.cpu == 'x86_64'
       shell: bash
       working-directory: ${{github.workspace}}/install
       run: F3D_PLUGINS_PATH=$(pwd)/../build_plugins/example-plugin${{ runner.os == 'Windows' && '/Release' || null }} ${{ env.F3D_BIN_PATH }} ../source/examples/plugins/example-plugin/data.expl --load-plugins=example --output=../install_example_plugin_output.png --ref=../source/.github/baselines/install_example_plugin_output.png --resolution=300,300 --verbose
 
     - name: Install plugin examples
+      if: inputs.cpu == 'x86_64'
```
mwestphal, line 266: why arm cant be installed ?  Or maybe its just not needed ?

.github/actions/generic-ci/action.yml
```diff
@@ -180,6 +194,7 @@ runs:
         cmake --install . --component configuration
 
     - name: Build and configure libf3d examples
+      if: inputs.cpu == 'x86_64'
```
mwestphal, line 197: This could be checked for arm I think

.github/actions/generic-ci/action.yml
```diff
@@ -150,6 +163,7 @@ runs:
         path: ${{github.workspace}}\build\bin_Release
 
     - name: Test
+      if: inputs.cpu == 'x86_64'
```
mwestphal, line 166: afaiu arm just cannot be tested. Wdyt of adding a new inputs var to control this behavior instead of checking the cpu arch ? It would be more clear in terms of intent.

.github/actions/generic-ci/action.yml
```diff
@@ -103,14 +114,14 @@ runs:
...
+        -DBUILD_TESTING=${{ inputs.cpu == 'x86_64' && 'ON' || 'OFF' }}
         -DCMAKE_BUILD_TYPE=Release
         -DCMAKE_INSTALL_PREFIX:PATH=../install
         -DCMAKE_INSTALL_RPATH_USE_LINK_PATH=ON
         -DCMAKE_POLICY_DEFAULT_CMP0091=NEW
         -DCMAKE_PREFIX_PATH:PATH=$(pwd)/../dependencies/install/
-        -DF3D_BINDINGS_JAVA=ON
-        -DF3D_BINDINGS_PYTHON=ON
+        -DF3D_BINDINGS_JAVA=${{ inputs.cpu == 'x86_64' && 'ON' || 'OFF' }}
+        -DF3D_BINDINGS_PYTHON=${{ inputs.cpu == 'x86_64' && 'ON' || 'OFF' }}
```
mwestphal, line 124: any reason why the bindings cannot be supported on arm ?

mwestphal: right, the dependencies. Do you know if python bindings will work in the superbuild ?

.github/actions/alembic-install-dep/action.yml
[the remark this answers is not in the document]
mwestphal: you are right

.github/actions/vtk-install-dep/action.yml
Meakk: That's not new though. HDF5 fails to build on Android (ARM) and wasm too.  
I think the reason is HDF5 generate an executable and execute it during the configuration, but execution fails because we are cross-compiling.

.github/actions/generic-ci/action.yml
Meakk: Maybe it works. `pybind11` is a header-only library. I'll try to enable it.

.github/actions/generic-ci/action.yml
Meakk: You mean doing `inputs.cpu == 'arm64'` instead?  
I disagree. The runners are `x86_64`, so we need to check the architecture is the same as the runners.

.github/actions/generic-ci/action.yml
Meakk: Hmm. That means we will have two inputs that will be enabled strictly at the same time. I'm not against it but it will add bloating code.  
What I can do is adding a new step which set an env variable like `ENABLE_TESTING` depending on criteria (will only depend on the cpu arch for now)

.github/actions/generic-ci/action.yml
mwestphal: Sorry,  I should have been clearer. You should be able to build and configure libf3d example on arm I think.

.github/actions/generic-ci/action.yml
mwestphal: that could work

.github/actions/generic-ci/action.yml
mwestphal: ok, too bad we cant test it in the CI, but well, not going to build python ourselve here

.github/actions/vtk-install-dep/action.yml
mwestphal: yes but we are not shipping anything android or wasm, but we will ship arm, so user we want to know that exodus will not work in that case.

Meakk: ok

.github/actions/generic-ci/action.yml
Meakk: Well, it works :)  
Not tested tho

.github/actions/generic-ci/action.yml
Meakk: used ga outputs instead of envvar

.github/actions/generic-ci/action.yml
```diff
@@ -150,6 +176,7 @@ runs:
         path: ${{github.workspace}}\build\bin_Release
 
     - name: Test
+      if: steps.vars.outputs.run_tests == true
```
mwestphal, line 179: I would have expected `env.run_tests`

mwestphal (approved):
small comment to adress

.github/actions/generic-ci/action.yml
mwestphal: ha right! I never used that before. Looks like a good design!

PR #977, opened 2023-09-08
CI: Small fixes and improvements

mwestphal:
- Add a check for all required inputs in actions (https://github.com/actions/runner/issues/1070)
 - Add missing lfs_sha input in nightly workflow
 - Fix small macos related issues
 - Refactor nightly workflow by OS

### DIFF
--- a/.github/workflows/nightly_vtk_master.yml
+++ b/.github/workflows/nightly_vtk_master.yml
@@ -67,33 +67,14 @@ jobs:
         workflow_label: 'producer'
 
 #----------------------------------------------------------------------------
-# MAIN CI: Build and test with a cross-platform build matrix
+# Windows CI: Build and test
 #----------------------------------------------------------------------------
-  ci:
+  windows:
     needs: [cache_lfs, check_nightly]
     if: ${{ needs.check_nightly.outputs.should_run == 'true' }}
     strategy:
       fail-fast: false
-      matrix:
-        os: [ubuntu, windows, macos]
-        build_type: [standard]
-        include:
-          - raytracing_label: raytracing
-          - egl_label: no-egl
-          - os: macos
-            raytracing_label: no-raytracing #598
-          - build_type: headless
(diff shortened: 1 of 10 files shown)


.github/workflows/nightly_vtk_master.yml
```diff
@@ -67,58 +67,103 @@ jobs:
...
-          - os: macos
-            raytracing_label: no-raytracing #598
           - build_type: headless
-            os: ubuntu
-            raytracing_label: raytracing
             exclude_deprecated_label: no-exclude-deprecated
-            bundle_label: no-bundle
             egl_label: egl
+          - build_type: exclude_deprecated
+            exclude_deprecated_label: exclude-deprecated
```
Meakk, line 112: do we have to test `exclude-deprecated`? I mean, we can choose a single linux configuration (let's say `commit, egl, no-exclude-deprecated`) and just run this one. As long as all configurations are tested in the regular CI that's fine I guess?

Meakk (approved):
Please consider the comment. If you disagree, feel free to merge it :)

.github/workflows/nightly_vtk_master.yml
mwestphal: > do we have to test exclude-deprecated

Yes we do.

>  I mean, we can choose a single linux configuration 

You mean testing multiple "config flag" in a single CI job ? No, that would be bad design. We should switch "config flag" one by one out of the default for good testing. It would be even better to test the whole matrix but that would be too many builds.

mwestphal: Ha, I misunderstood your comment, you actually meant:

> do we have to test exclude-deprecated *in the nightly VTK build*

Let me think about it.

mwestphal: You are right, we dont need to test `exclude-deprecated` in the nnightly VTK CI

PR #978, opened 2023-09-08
Load from memory API

Meakk:


library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
```diff
@@ -0,0 +1,64 @@
...
+vtkF3DMemoryMesh::vtkF3DMemoryMesh()
+{
+  this->SetNumberOfInputPorts(0);
+}
+
+//------------------------------------------------------------------------------
+vtkF3DMemoryMesh::~vtkF3DMemoryMesh() = default;
+
+//------------------------------------------------------------------------------
+void vtkF3DMemoryMesh::PrintSelf(ostream& os, vtkIndent indent)
```
mwestphal, line 20: remove or test if it matters

library/VTKExtensions/Readers/vtkF3DMemoryMesh.h
```diff
@@ -0,0 +1,47 @@
+/**
+ * @class vtkF3DMemoryMesh
+ * @brief create vtkPolyData from STL vectors.
+ *
+ * Simple source which convert and copy STL vectors provided by the user
```
mwestphal, line 5: I would not mention STL, this is much simpler than the actual STL format

library/public/loader.h
```diff
@@ -61,6 +61,16 @@ class F3D_EXPORT loader
    */
   virtual loader& loadGeometry(const std::string& filePath, bool reset = false) = 0;
 
+  /**
+   * Load a geometry from memory buffers.
+   * Reset the scene before loading if a full scene was loaded previously or if reset is set to
+   * false, do not reset if only loaded geometries previously. Geometries loader using this method
+   * will be available in a default scene and use all default scene related options. Throw a
+   * load_failure_exception on failure.
```
mwestphal, line 69: Lets improve the doc, what I've written is unclear and also false at times.

```
   * Calling this method will reset the scene before loading if a full scene was loaded previously or if the reset argument is set to true, It will not reset if only geometries were loaded previously. 
   * Geometries loaded using this method will be available in a default scene and use all default scene related options. 
   * Throw a load_failure_exception on failure.
   ```
   
   Update the other method doc as well.

File: python/testing/test_loader.py [hunk 4]
```diff
@@ -0,0 +1,25 @@
+import pytest
+import tempfile
+import f3d
+
+
+def test_load_memory():
+    dataset = "./testing/data/cow.vtp"
```
mwestphal, line 7: not needed

library/testing/TestSDKLoader.cxx
```diff
@@ -206,5 +206,16 @@ int TestSDKLoader(int argc, char* argv[])
     return EXIT_FAILURE;
   }
 
+  // Load from memory
+  try
+  {
+    load.loadGeometry({ 0.f, 0.f, 0.f, 0.f, 1.f, 0.f, 1.f, 0.f, 0.f }, { 0, 1, 2 });
```
mwestphal, line 212: we may need a TestDSKLoadGeometryFromMemory that do actual image comparison. I know there is a python test but python tests should be considered optional and focus on python bindings.

library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
```diff
@@ -0,0 +1,64 @@
...
+void vtkF3DMemoryMesh::SetTriangles(const std::vector<unsigned int>& triangles)
+{
+  vtkNew<vtkCellArray> cells;
+  cells->AllocateExact(triangles.size() / 3, triangles.size());
+
+  for (size_t i = 0; i < triangles.size(); i += 3)
+  {
+    vtkIdType t[] = { static_cast<vtkIdType>(triangles[i]),
+      static_cast<vtkIdType>(triangles[i + 1]), static_cast<vtkIdType>(triangles[i + 2]) };
+    cells->InsertNextCell(3, t);
```
mwestphal, line 49: SetCells should be more efficient

library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
```diff
@@ -0,0 +1,64 @@
...
+//------------------------------------------------------------------------------
+void vtkF3DMemoryMesh::SetPoints(const std::vector<float>& positions)
+{
+  vtkNew<vtkPoints> points;
+  points->SetNumberOfPoints(positions.size() / 3);
+
+  for (size_t i = 0; i < positions.size() / 3; i++)
+  {
+    points->SetPoint(i, &positions[3 * i]);
+  }
```
mwestphal, line 34: you could access the internal array and use ranges to improve performance when TBB is enabled

library/VTKExtensions/Readers/vtkF3DMemoryMesh.h
```diff
@@ -0,0 +1,47 @@
...
+class vtkF3DMemoryMesh : public vtkPolyDataAlgorithm
+{
+public:
+  static vtkF3DMemoryMesh* New();
+  vtkTypeMacro(vtkF3DMemoryMesh, vtkPolyDataAlgorithm);
+  void PrintSelf(ostream& os, vtkIndent indent) override;
+
+  /**
+   * Set contiguous list of positions.
+   * Length of the list must be a multiple of 3.
```
mwestphal, line 24: clarify that this call do an actual copy of the data

library/VTKExtensions/Readers/vtkF3DMemoryMesh.h
```diff
@@ -0,0 +1,47 @@
...
+   */
+  void SetPoints(const std::vector<float>& positions);
+
+  /**
+   * Set contiguous list of triangles by vertex indices.
+   * Length of the list must be a multiple of 3.
```
mwestphal, line 30: clarify that this call do an actual copy of the data

library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
Meakk: can you give me an example of ranges usage?

mwestphal: I would need to spend more time on this if you need an actual snippet of code but should be something based on:

C++11 for-range loops with vtk::DataArrayTupleRange : https://www.kitware.com/c11-for-range-support-in-vtk/

and

vtKSMP::Transform: https://vtk.org/doc/nightly/html/classvtkSMPTools.html#a16e88039a1e33aaacef235e92779c4ee

library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
mwestphal: you can probably use range and smp transform here too

library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
Meakk: ok I'll figure out

library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
```diff
@@ -0,0 +1,74 @@
...
+vtkF3DMemoryMesh::~vtkF3DMemoryMesh() = default;
+
+//------------------------------------------------------------------------------
+void vtkF3DMemoryMesh::SetPoints(const std::vector<float>& positions)
+{
+  vtkNew<vtkFloatArray> arr;
+  arr->SetNumberOfComponents(3);
+  arr->SetNumberOfTuples(positions.size() / 3);
+
+  std::copy(positions.cbegin(), positions.cend(), arr->GetPointer(0));
```
mwestphal, line 30: I guess thats not SMP accelerated ? Should it be ?

mwestphal: Its not necesarry, I'm just wondering if there is a reason for you not to try to do it.

library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
```diff
@@ -0,0 +1,74 @@
...
+  vtkNew<vtkIdTypeArray> offsets;
+  vtkNew<vtkIdTypeArray> connectivity;
+  offsets->SetNumberOfTuples(cellSize.size() + 1);
+  connectivity->SetNumberOfTuples(cellIndices.size());
+
+  // fill offsets
+  offsets->SetTypedComponent(0, 0, 0);
+  std::inclusive_scan(cellSize.cbegin(), cellSize.cend(),
+    vtk::DataArrayValueRange(offsets).begin() + 1,
+    [](unsigned int a, unsigned int b) { return static_cast<vtkIdType>(a + b); });
```
mwestphal, line 52: same

library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
```diff
@@ -0,0 +1,74 @@
...
+void vtkF3DMemoryMesh::SetCells(
+  const std::vector<unsigned int>& cellSize, const std::vector<unsigned int>& cellIndices)
+{
+  vtkNew<vtkIdTypeArray> offsets;
+  vtkNew<vtkIdTypeArray> connectivity;
+  offsets->SetNumberOfTuples(cellSize.size() + 1);
+  connectivity->SetNumberOfTuples(cellIndices.size());
+
+  // fill offsets
+  offsets->SetTypedComponent(0, 0, 0);
```
mwestphal, line 49: what is that for ?

library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
mwestphal: (SMP question)

library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
```diff
@@ -0,0 +1,74 @@
...
+  std::inclusive_scan(cellSize.cbegin(), cellSize.cend(),
+    vtk::DataArrayValueRange(offsets).begin() + 1,
+    [](unsigned int a, unsigned int b) { return static_cast<vtkIdType>(a + b); });
+
```
mwestphal, line 53: not covered by tests ?

python/testing/test_loader.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+import tempfile
+import f3d
+
+
+def test_load_memory():
+    reference = "./testing/baselines/TestSDKLoadMemory.png"
+    output = tempfile.gettempdir() + "/TestSDKLoadMemory.png"
```
mwestphal, line 8: copy the lfs file into a dedicated baseline with the same hash

library/src/loader_impl.cxx
```diff
@@ -317,6 +323,46 @@ loader& loader_impl::loadScene(const std::string& filePath)
...
+  size_t nbPoints = positions.size() / 3;
+
+  auto it = std::find_if(
+    cellIds.cbegin(), cellIds.cend(), [=](unsigned int idx) { return idx >= nbPoints; });
+  if (it != cellIds.cend())
+  {
+    throw loader::load_failure_exception("Cell vertex at index " +
+      std::to_string(std::distance(cellIds.cbegin(), it)) +
+      " is greater than the maximum vertex index (" + std::to_string(nbPoints) + ")");
+  }
```
mwestphal, line 354: should be easy to test

library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
Meakk: I saw no smp copy functions.  
What about getting rid of smp entirely and use C++17 parallel algorithms?

library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
Meakk: Set the first offset to 0. The inclusive scan does not do it.

library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
Meakk: I pushed a commit for C++17 algorithms everywhere instead of VTK SMP, let me know what you think.

library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
Meakk: That's because there's a single cell. I've fixed it.

library/VTKExtensions/Readers/vtkF3DMemoryMesh.cxx
mwestphal: ill check but sad to not use tbb when we can

Meakk: Well, as long as it's parallelized, we don't care if we are using STD or TBB, do we?

Meakk: Well, it looks like parallel algorithms are not well supported by some compilers, including Apple Clang...  
I'll revert these changes.

mwestphal: we do, TBB performs better.

Meakk: Well, for copying data, I'm suspect it wouldn't make a difference.  
Anyway, I migrated toward `vtkSMPTools::For`. I can't use `vtkSMPTools::Transform` because it's not available in VTK 9.0

mwestphal (approved):

mwestphal:
` vtkF3DMemoryMesh.cxx:60:8: error: 'inclusive_scan' is not a member of 'std'`

From: 

https://github.com/f3d-app/f3d-superbuild/actions/runs/6153139319/job/16696523040

PR #983, opened 2023-09-12
Rework load from memory

Meakk:


snoyer:
While not strictly necessary it'd be nice to be able to just pass `face_sizes = { 3 }` for a tri-mesh without having to fill a whole vector of all `3`s.

One way to allow this without having a separate code path would be to make `face_sizes` periodic.
Instead of having:
- precondition: `face_indices.size() == sum(face_sizes)`
- face count: `face_indices.size()`
- face[i] size: `face_sizes[i]`

you'd have:
- precondition: `face_indices.size() % sum(face_sizes) == 0`
- face count: `face_indices.size() / sum(face_sizes) * face_sizes.size()`
- face[i] size: `face_sizes[i%face_sizes.size()]`

The vast majority of uses would be the edge cases where `face_sizes.size() == face_indices.size()` for abitrary mixed meshes or `face_sizes.size() == 1` for regular meshes.
But here's an example of a mesh with mixed but periodic face sizes for which you'd be able to pass `face_sizes = { 3,4,3 }` instead of `face_sizes = { 3,4,3, 3,4,3, 3,4,3, 3,4,3 }`
```
OFF
10 12 0
0 0 0
0 0 10
-1 -1 2
-1 +1 2
+1 +1 2
+1 -1 2
-1 -1 8
-1 +1 8
+1 +1 8
+1 -1 8
3 0 2 3
4 2 3 7 6
3 1 6 7
3 0 3 4
4 3 4 8 7
3 1 7 8
3 0 4 5
4 4 5 9 8
3 1 8 9
3 0 5 2
4 5 2 6 9
3 1 9 6
```

Maybe that's overly complicated for little gain...

Meakk:
Yes, I think it's a bit complicated.  
In Python it would be trivial, just use `[3, 4, 3] * 3`.  
For C++ that's a bit more complicated but should be doable using `std::copy_n`.  
For C++, if you have only triangles you could even use `std::vector<unsigned int>(nbTriangles, 3)` constructor.

mwestphal (approved):

PR #986, opened 2023-09-13
Fix exr can read check and segfault

mwestphal:


Meakk (approved):

library/VTKExtensions/Readers/vtkF3DEXRReader.cxx
```diff
@@ -104,6 +104,12 @@ void vtkF3DEXRReader::ExecuteDataWithInformation(vtkDataObject* output, vtkInfor
   }
 
   vtkFloatArray* scalars = vtkFloatArray::SafeDownCast(data->GetPointData()->GetScalars());
+  if (!scalars)
+  {
+    vtkErrorMacro(<< "Could not find expected scalar array");
```
Meakk, line 109: Is it possible to cover that? Especially if it's the reason of the crash in th SB

mwestphal: definitely!

Meakk (approved):

PR #1039, opened 2023-10-04
Update Changelog and install doc

mwestphal:


doc/user/INSTALLATION.md
```diff
@@ -42,3 +43,4 @@ You can also find packages for the following repositories:
 
 - [FreeBSD](https://cgit.freebsd.org/ports/tree/graphics/f3d)
 - [Spack](https://packages.spack.io/package.html?name=f3d)
+- [pypi](https://test.pypi.org/project/f3d/)
```
Meakk, line 46: It would be great to have a Python category and specify it can be installed by running `pip install f3d`.  
Overall I think the command should be specified for all packages

mwestphal: I agree, lets improve that in the coming weeks. The point of this PR is just to list package and update links. Out of scope.

mwestphal: https://github.com/f3d-app/f3d/issues/1041

PR #1048, opened 2023-10-10
Convert Alembic importer to reader

Meakk:


plugins/alembic/module/vtkF3DAlembicReader.cxx
```diff
@@ -0,0 +1,148 @@
...
+    if (pmesh.getSchema().getNumSamples() > 0)
+    {
+      pmesh.getSchema().get(samp);
+
+      Alembic::AbcGeom::P3fArraySamplePtr positions = samp.getPositions();
+      Alembic::AbcGeom::Int32ArraySamplePtr indices = samp.getFaceIndices();
+      Alembic::AbcGeom::Int32ArraySamplePtr counts = samp.getFaceCounts();
+
+      size_t P_size = positions->size();
+      size_t counts_size = counts->size();
```
mwestphal, line 41: perfect time to follow guidelines

plugins/alembic/module/vtkF3DAlembicReader.cxx
```diff
@@ -0,0 +1,148 @@
...
+
+      for (size_t i = 0; i < P_size; i++)
+      {
+        points->InsertNextPoint(
+          positions->get()[i].x, positions->get()[i].y, positions->get()[i].z);
```
mwestphal, line 46: switching to SetPoint should be trivial

plugins/alembic/module/vtkF3DAlembicReader.cxx
```diff
@@ -0,0 +1,148 @@
...
+      }
+
+      size_t face_index = 0;
+      for (size_t i = 0; i < counts_size; i++)
+      {
+        auto polyface_vertex_count = counts->get()[i];
+        polys->InsertNextCell(polyface_vertex_count);
+        for (auto j = 0; j < polyface_vertex_count; j++)
+        {
+          polys->InsertCellPoint(indices->get()[face_index++]);
```
mwestphal, line 56: using a faster API should be trivial

plugins/alembic/module/vtkF3DAlembicReader.cxx
```diff
@@ -0,0 +1,148 @@
...
+      append->AddInputData(ProcessIPolyMesh(polymesh));
+      nextParentObject = polymesh;
+    }
+
+    // Recursion
+    if (nextParentObject.valid())
+    {
+      for (size_t i = 0; i < nextParentObject.getNumChildren(); ++i)
+      {
+        IterateIObject(append, nextParentObject, nextParentObject.getChildHeader(i));
```
mwestphal, line 89: this

plugins/alembic/module/vtkF3DAlembicReader.cxx
```diff
@@ -0,0 +1,148 @@
...
+    }
+  }
+
+  void ImportRoot(vtkAppendPolyData* append)
+  {
+    Alembic::Abc::IObject top = this->Archive.getTop();
+
+    for (size_t i = 0; i < top.getNumChildren(); ++i)
+    {
+      IterateIObject(append, top, top.getChildHeader(i));
```
mwestphal, line 100: this

plugins/alembic/module/vtkF3DAlembicReader.cxx
```diff
@@ -0,0 +1,148 @@
...
+    }
+  }
+
+  void ReadScene(const std::string& filePath)
```
mwestphal, line 104: rename to PrepareArchive ?

plugins/alembic/module/vtkF3DAlembicReader.cxx
```diff
@@ -0,0 +1,148 @@
...
+
+//----------------------------------------------------------------------------
+int vtkF3DAlembicReader::RequestData(
+  vtkInformation*, vtkInformationVector**, vtkInformationVector* outputVector)
+{
+  vtkPolyData* output = vtkPolyData::GetData(outputVector);
+
+  this->Internals->ReadScene(this->FileName);
+
+  vtkNew<vtkAppendPolyData> append;
```
mwestphal, line 134: does that work when working with textures ?

mwestphal:
doc update missing

plugins/alembic/module/vtkF3DAlembicReader.cxx
Meakk: yes it will use the same textures for all submeshes.

mwestphal: is that ok ? would be beneficial to have @nyue input on this

Meakk: Well, that's the case with all geometry-only formats.  
As far as I know, Alembic supports texture coordinates but no materials. The texture coordinates are supposed to be used in a later stage in Blender/Maya or any final rendering software by crafting materials there.  
The goal of f3d here is to provide a way to visualize these texture coordinates by using a checkboard texture for example.

nyue: Technically, Alembic does have a [prototype](https://github.com/alembic/alembic/tree/master/lib/Alembic/AbcMaterial) material support but most studio working with Alembic have their own way to deal with what they call *Look Development* in their pipeline and the industry is gravitating towards [MaterialX](https://materialx.org/).

I believe for a start, what @Meakk solution provides is sufficient.

mwestphal: Alright then!

Meakk:
> doc update missing

what doc?

mwestphal:
> > doc update missing
> 
> what doc?

I was thinking about this:

https://github.com/f3d-app/f3d/blob/master/doc/user/USAGE.md?plain=1#L28

But it was never listed as a full scene format anyway. soo all good.

mwestphal (approved):

PR #1049, opened 2023-10-11
fix `grid-absolute` for off-center models

snoyer:
`absolute-grid` option currently works as intended when models are off-center vertically (in the up direction) but not when they are off-center "sideways" (along the grid plane), see example below:

absolute-grid-bug

This PR changes the behavior to the following:

absolute-grid-fix

axes are still in the same position at the origin, but the actor and fading radius of the grid are centered below the model.

There is still room for improvement, for example if the model is very far from the origin the axes may not be visible at all (outside of radius), solving that would require more involved framing/fading patterns of the grid plane.

This fix should at least make the feature not broken anymore.


Meakk (approved):
Indeed, fading based on the model bbox looks better than fading based on the grid origin.

PR #1052, opened 2023-10-15
Check and link with libatomic for `armel` architecture

mzf-guest:
Debian build of F3D v2.2.1 is failing for `armel` architecture (Buildlog: [f3d_2.2.1+dfsg-1_armel.log](https://github.com/f3d-app/f3d/files/12910154/f3d_2.2.1%2Bdfsg-1_armel.log)). The issue is that gcc does not automatically link with `libatomic` for some architectures  (gcc bug [here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81358)).

The issue have been reported many times to the CMake team ([here](https://gitlab.kitware.com/cmake/cmake/-/issues/23021), [here](https://gitlab.kitware.com/cmake/cmake/-/issues/21174) or [here](https://gitlab.kitware.com/cmake/cmake/-/issues/20895)), and the CMake team just suggests to check it explicitly with the `check_cxx_source_compiles` command. 

So I've pushed this quick patch (heavily inspired by the one from [libtorrent](https://github.com/arvidn/libtorrent/pull/5748/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR632)), which will be suitable for Debian package migration.
However, its uses `link_libraries` CMake command for simplicity and you may want to use `target_link_libraries` command at a better place than the top level CMakeLists.txt.

Feel free to suggest and edit !

Best,
François

### DIFF
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -191,6 +191,14 @@ if (APPLE OR UNIX)
   endif ()
 endif ()
 
+if(UNIX AND NOT APPLE)
+  option(F3D_LINUX_LIBRARY_LINK_ATOMIC "Link with libatomic" OFF)
+  mark_as_advanced(F3D_LINUX_LIBRARY_LINK_ATOMIC)
+  if(F3D_LINUX_LIBRARY_LINK_ATOMIC)
+    target_link_libraries(libf3d PRIVATE atomic)
+  endif()
+endif()
+
 # Testing
 if(BUILD_TESTING)
   add_subdirectory(testing)
--- a/doc/user/LIMITATIONS_AND_TROUBLESHOOTING.md
+++ b/doc/user/LIMITATIONS_AND_TROUBLESHOOTING.md
@@ -53,6 +53,10 @@ Your data probably contains some translucent data for some reason, turn on trans
 
 With some C++ STD library version, explicit linking to `stdc++fs` is not supported. We provide a CMake option `F3D_LINUX_APPLICATION_LINK_FILESYSTEM` that you can set to `OFF` to workaround this issue.
 
+> I have a link error related to undefined reference to symbol of `libatomic`.
+
+The GCC flag `-latomic` is not being added automatically with specific architectures, like `armel` and `RISCV64`. We provide a CMake option `F3D_LINUX_LIBRARY_LINK_ATOMIC` that you can set to `ON` to workaround this issue.
+
 > Thumbnails are not working in my file manager.
 
  * Check that your file manager supports the thumbnailer mechanism.


CMakeLists.txt
```diff
@@ -67,6 +67,40 @@ if (F3D_USE_EXTERNAL_NLOHMANN_JSON)
...
+		int main() {
+			x.fetch_add(1, std::memory_order_relaxed);
+			return 0;
+		}
+	]=])
+	string(REPLACE "std::atomic<int>" "std::atomic<std::int64_t>" ATOMICS64_TEST_SOURCE "${ATOMICS_TEST_SOURCE}")
+
+	check_cxx_source_compiles("${ATOMICS_TEST_SOURCE}" HAVE_CXX_ATOMICS_WITHOUT_LIB)
+	check_cxx_source_compiles("${ATOMICS64_TEST_SOURCE}" HAVE_CXX_ATOMICS64_WITHOUT_LIB)
+	if((NOT HAVE_CXX_ATOMICS_WITHOUT_LIB) OR (NOT HAVE_CXX_ATOMICS64_WITHOUT_LIB))
```
Meakk, line 88: Can we have cases where it works for `int` but not `int64_t` or the other way around?  
What should we do in this case?

CMakeLists.txt
```diff
@@ -67,6 +67,40 @@ if (F3D_USE_EXTERNAL_NLOHMANN_JSON)
   find_package(nlohmann_json REQUIRED)
 endif ()
 
+# check if we need to link with libatomic (not needed on MSVC)
+if ((NOT Windows) AND (NOT APPLE))
```
Meakk, line 71: Maybe we should test if the compiler is GCC.  
We had a similar issue where sometimes we had to link explicitly with `filesystem`.  
Maybe we should do something like this for it too.

CMakeLists.txt
```diff
@@ -67,6 +67,40 @@ if (F3D_USE_EXTERNAL_NLOHMANN_JSON)
...
+	# TODO: migrate to CheckSourceCompiles in CMake >= 3.19
```
Meakk, line 72: We have a prerequisite on CMake 3.21 already so it's fine to migrate on `CheckSourceCompiles` now.

Meakk (changes requested):
Here are some required changes.  
I'm happy to take over but I have no idea how to test it.  
I can blindly make the changes and maybe you can confirm it works on your side before merging?

CMakeLists.txt
Meakk: Here's what we are doing for filesystem: https://github.com/f3d-app/f3d/blob/b40c3c4b05e8db307642194878d10e7e2132134e/application/CMakeLists.txt#L33  
Simpler but not automatic, so maybe your suggestion is better.  
Wdyt @mwestphal ?

mwestphal: I like simpler, manual solution, especially for a usecase as specific as this one.

mwestphal:
@mzf-guest do you need help finishing this one ?

mzf-guest:
> @mzf-guest Can we do anything to help move this forward ?

Thanks for the ping. As far as I understand you are expecting a dedicated CMake variable (let's name it `FORCE_LIB_ATOMICS_LINKING` or `EXPLICIT_LIB_ATOMICS_LINK`, please advise) and let the user enable it when relevant?

If yes, it should be easy to implement and I'll try to push it shortly.

Meakk:
> > @mzf-guest Can we do anything to help move this forward ?
> 
> Thanks for the ping. As far as I understand you are expecting a dedicated CMake variable (let's name it `FORCE_LIB_ATOMICS_LINKING` or `EXPLICIT_LIB_ATOMICS_LINK`, please advise) and let the user enable it when relevant?
> 
> If yes, it should be easy to implement and I'll try to push it shortly.

We have something like that for `filesystem`:
```cmake
if(UNIX AND NOT APPLE)
  option(F3D_LINUX_APPLICATION_LINK_FILESYSTEM "Link with libstdc++fs" ON)
  mark_as_advanced(F3D_LINUX_APPLICATION_LINK_FILESYSTEM)
  if(F3D_LINUX_APPLICATION_LINK_FILESYSTEM)
    target_link_libraries(f3d PRIVATE stdc++fs)
  endif()
endif()
```

So we probably need something consistent like:
```cmake
if(UNIX AND NOT APPLE)
  option(F3D_LINUX_LIBRARY_LINK_ATOMIC "Link with libatomic" ON)
  mark_as_advanced(F3D_LINUX_LIBRARY_LINK_ATOMIC)
  if(F3D_LINUX_LIBRARY_LINK_ATOMIC)
    target_link_libraries(libf3d PRIVATE atomic)
  endif()
endif()
```

However, looking at the link error, it seems like the file producing the error is `vtkF3DMemoryMesh.cxx` which uses `vtkSMPTools` internally. So maybe the correct fix is to patch VTK to expose linkage to `libatomic` via its VTK-config.cmake file?  
Do you know which VTK SMP backend is enabled on this system?

> /usr/bin/ld: ../lib/libVTKExtensionsReaders.a(vtkF3DMemoryMesh.cxx.o): undefined reference to symbol '__atomic_fetch_add_8@@LIBATOMIC_1.0'
> /usr/bin/ld: /usr/lib/arm-linux-gnueabi/libatomic.so.1: error adding symbols: DSO missing from command line

mzf-guest:
Hi @Meakk 

thanks for the code suggestion, I've just checked with an `armel` machine and it works! However, I would let the option at OFF by default and enable it only with problematic architecture. Do you agree?

> Do you know which VTK SMP backend is enabled on this system?

TBB

> So maybe the correct fix is to patch VTK to expose linkage to libatomic via its VTK-config.cmake file?

Makes sense as I can find several projects (like [libpcl](https://sources.debian.org/src/pcl/1.13.0%2Bdfsg-3/CMakeLists.txt/#L61) ) which are also patching the libatomic linkage themselves. Feel free to open an upstream issue and link to this discussion.

Thanks!

mwestphal:
Hi @mzf-guest , we will release soon, do you want to merge something soon then ?

mzf-guest:
> Hi @mzf-guest , we will release soon, do you want to merge something soon then ?

Yes, it will be great but not mandatory for the Debian package because you release more often than the 2-3 year Debian release cycle.

I can read that the `f3d::TestSkinningManyBones` CI test on macos is failing for this PR, but I don't understand why from my modifications. Could you please help?

mwestphal:
> I can read that the f3d::TestSkinningManyBones CI test on macos is failing for this PR, but I don't understand why from my modifications. Could you please help

Fixed in master

Meakk:
I'm wondering if we are doing the right thing here. Isn't it something we can control with `CMAKE_EXE_LINKER_FLAGS`?
https://cmake.org/cmake/help/latest/variable/CMAKE_EXE_LINKER_FLAGS.html

I don't know if it's a workaround or the correct fix tho. Any thoughts @mwestphal @mzf-guest?

mwestphal:
Using the `CMAKE_EXE_LINKER_FLAGS` may work but that'd be the hack. Current implementation is cleaner imo.

library/CMakeLists.txt
```diff
@@ -191,6 +191,14 @@ if (APPLE OR UNIX)
   endif ()
 endif ()
 
+if(UNIX AND NOT APPLE)
+  option(F3D_LINUX_LIBRARY_LINK_ATOMIC "Link with libatomic" OFF)
```
mwestphal, line 195: Add a mention of it in https://f3d.app/doc/user/LIMITATIONS_AND_TROUBLESHOOTING.html ?

https://github.com/f3d-app/f3d/blob/master/doc/user/LIMITATIONS_AND_TROUBLESHOOTING.md

mwestphal (changes requested):

library/CMakeLists.txt
mzf-guest: Done!

mwestphal: perfect!

mwestphal (approved):

Meakk (approved):

Issue #1018, opened 2023-09-30
When creating an engine, the window title is VTK


Meakk:
**Describe the bug**
When creating an engine in Python, the window title is VTK

**To Reproduce**

```py
import f3d
e=f3d.Engine()
```

![image](https://github.com/f3d-app/f3d/assets/6069752/3ee9ff62-9987-4921-abc7-b06779fa8c0b)

**Expected behavior**
The Window title should be F3D

**System Information:**
 - OS: Windows

**F3D Information**
2.2 RC2

mwestphal:
more libf3d than python I believe.

Meakk:
Probably yes, but I haven't check in C++ so I cannot tell for sure.

shaikan:
Hi 
I would like to work on this issue.
thanks,

Meakk:
Great! Thank you @shaikan  
Please submit a PR when you are ready and don't hesitate to reach out on discord if you need help.

mwestphal:
Also do not hesitate to reach out on discord!
https://discord.f3d.app

mwestphal:
Need any help @shaikan ?

shaikan:
Hi 
sorry I was busy with office work so couldn't work on this issue.
I am setting up the build environment right now.
 Will seek help if I get stuck.

shaikan:
So I checked and here is my analysis.
The renaming of the window to `F3D - A fast and minimalist 3D viewer` is done by `F3DStarter`.
The actual `vtkRenderWindow` gets created inside `window_impl` at which time it gets the default name from VTK i.e. "Visualization Toolkit..".
There are two things here.

1. In C++ F3D application, there is a slight hiccup while creating the window which causes it to show with the default name for a fraction of a second before it gets resized and renamed.
2. In case of Python, the script directly creates the engine object and hence the window never gets renamed and shows with the original VTK name.

I propose that setting the window name and icon should be shifted from f3d application to inside libf3d
either inside `engine` or the root of window creation `window_impl`.
This would make the window display behaviour same across c++ and python.

Meakk:
Ok that makes sense.  
We have a public API to set the window title `f3d::window::setWindowName` and as you said, it's called by the application which is fine to me.  
I think the issue is the lack of default value in libf3d.  
It does not have to be `F3D - A fast and minimalist 3D viewer`, maybe something simple like just `libf3d` or `f3d` or `f3d window`...

mwestphal:
[...]

PR #1057, opened 2023-10-26
Fix set default window name inside lib

shaikan:
Fix for #1018 
Added code to set default title for window inside libf3d.


Meakk (approved):
Looks good to me! Thanks!

mwestphal (approved):

PR #1065, opened 2023-11-12
Autoplay

technologeli:
Brief description of changes:
- Created boolean `scene.animation.autoplay`, `false` by default
- Added CLI option `--animation-autoplay`
- When `scene.animation.autoplay` is true, files with animations will automatically start their animation.

### DIFF
--- a/doc/libf3d/OPTIONS.md
+++ b/doc/libf3d/OPTIONS.md
@@ -17,6 +17,7 @@ See the exhaustive list below, but note that this may change in the future.
 
 Option|Type<br>Default<br>Trigger|Description|F3D option
 :---:|:---:|:---|:---:
+scene.animation.autoplay|bool<br>false<br>load|Automatically start animation.|\-\-animation-autoplay
 scene.animation.index|int<br>0<br>load|Select the animation to load.<br>Any negative value means all animations (glTF only).<br>The default scene always has at most one animation.|\-\-animation-index
 scene.animation.speed-factor|double<br>1<br>render|Set the animation speed factor to slow, speed up or even invert animation.|\-\-animation-speed-factor
 scene.animation.frame-rate|double<br>60<br>render|Set the animation frame rate used to play the animation interactively.|\-\-animation-frame-rate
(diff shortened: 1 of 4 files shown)


mwestphal:
So, if I get this right, it means that autoplay will automatically play a file on load. Eg, when changing file, it will start playing the animation. Is that correct @technologeli ?

Wdyt @Meakk ?

Meakk:
Looks like the correct approach to me. 👍 
I don't know if it's possible to cover that with a test, any input @mwestphal ?

mwestphal:
Definitely possible! Animation is already tested with an interaction test. Just removing the interaction with the same test should provide the same output.

mwestphal:
@technologeli let me know if you need help with adding an test.

Meakk:
Adding a simple application test with `--animation-autoplay` option should be enough, right @mwestphal?

@technologeli see https://github.com/f3d-app/f3d/blob/f037d908c081cc2960999f7c53eaf86f89950816/application/testing/CMakeLists.txt#L140C37-L140C37

Adding a new line like:
```cmake
f3d_test(NAME TestGLTFReaderWithAnimationAutoplay DATA BoxAnimated.gltf ARGS --geometry-only --animation-autoplay DEFAULT_LIGHTS)
```

Then you should be able to run `ctest -R TestGLTFReaderWithAnimationAutoplay -VV` to get the baseline file to copy to `testing/baselines`.  
When the test pass, you can push the test and the coverage CI job should be happy.

technologeli:
Thank you for the pointers, I had a test written but was unsure of how to get it passing while using `ctest`.

doc/libf3d/OPTIONS.md
```diff
@@ -17,6 +17,7 @@ See the exhaustive list below, but note that this may change in the future.
 
 Option|Type<br>Default<br>Trigger|Description|F3D option
 :---:|:---:|:---|:---:
+scene.animation.autoplay|bool<br>false<br>load|Automatically start animation.|\-\-animation-autoplay
```
mwestphal, line 20: should be added to https://github.com/f3d-app/f3d/blob/master/doc/user/OPTIONS.md as well

mwestphal: and to the changelog

mwestphal:
> I had a test written but was unsure of how to get it passing while using ctest.

`ctest -R Autoplay` :)

doc/libf3d/OPTIONS.md
technologeli: Got it! Should I add a bullet point for both "For F3D users" and "For libf3d users" since both are affected? Also just to confirm, this is for v2.2.1 in the changelog?

mwestphal: > Should I add a bullet point for both "For F3D users" and "For libf3d users" since both are affected? 

Yes

> this is for v2.2.1 in the changelog?

Nope, please put it in `## Ongoing development` with the same logic.

mwestphal (approved):

mwestphal:
Thanks a lot for your contribution! Do not hesitate to reach out if you need any help fixing another issue!

PR #1068, opened 2023-11-14
Add VTK 9.3 in CI

Meakk:


mwestphal:
you can drop the two last "check" action from `.github/actions/generic-ci/action.yml`  as they are very expansive with vtk 9.2.6. Testing it only wiht 9.3 is enough.

mwestphal:
I do not know why raytracing would behave differently on 9.3.0. Investigation needed.

Meakk:
> you can drop the two last "check" action from `.github/actions/generic-ci/action.yml` as they are very expansive with vtk 9.2.6. Testing it only wiht 9.3 is enough.

What do you mean?

mwestphal:
in `.github/actions/generic-ci/action.yml` there is two steps named `Check F3D_PLUGINS_PATH using plugin example`  and `Check Install`. These steps uses the install  which contains the default configuration which contains the default HDRI.

On macOS CI, displaying an HDRI is very slow without the cache (because github provide not a lot of CPU on macOS), which means that these two steps takes 2/3 minutes each, which is why macOS + VTK v9.2.6 is always the last CI to finish.

With older VTK version, these steps are skipped but I decided to keep these steps for 9.2.6 because it is important to check that the install of F3D is working (at least) with the last release of VTK.

With VTK 9.3.0 out, we can drop these two steps with 9.2.6 to increase CI speed.

hth :)

mwestphal:
Do you want me to take care of this @Meakk ?

Meakk:
Any idea why raytracing is failing on 9.3?

mwestphal:
> Any idea why raytracing is failing on 9.3

Needs investigation

library/VTKExtensions/Rendering/vtkF3DRenderPass.cxx
```diff
@@ -275,7 +275,8 @@ void vtkF3DRenderPass::Blend(const vtkRenderState* s)
 
     ssImpl << "  vec4 mainSample = texture(texMain, texCoord);\n";
 
-#if VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 2, 20230824)
+// This check is not compatible with VTK 9.3 pre-release version build number < 20230824
+#if VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 3, 0)
```
mwestphal, line 279: I'm sorry but I don't get it. 

https://gitlab.kitware.com/vtk/vtk/-/merge_requests/8577 has not been backported to release and corresponding commits are not present in 9.3.0 afaics.

Meakk: You're right, the CI is failing, let me try something else

Meakk:
@mwestphal looks better now, please take another look

plugins/usd/module/vtkF3DUSDImporter.cxx
```diff
@@ -295,7 +295,7 @@ class vtkF3DUSDImporter::vtkInternals
         {
           pxr::UsdGeomCapsule capsulePrim = pxr::UsdGeomCapsule(prim);
 
-#if VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 3, 20230921)
+#if VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 3, 0)
```
mwestphal, line 298: Which MR does this requires ? would be nice to add a link.

mwestphal (approved):

plugins/usd/module/vtkF3DUSDImporter.cxx
mwestphal: Thanks, this should never have been backported imo.

PR #1078, opened 2023-11-25
Handle commas in filenames

### LINKED ISSUE #1070: Comma in filename prevent file to be opened by F3D
**Describe the bug**
A filename containing a comma cannot be opened by F3D as F3D understand it as being a separator

**To Reproduce**
 1. Open the file using `f3d --dry-run exam,ple.glb`
 2. Error

**Expected behavior**
F3D should be able to open the file with some form of escaped char or quotes

**System Information:**
 - OS: Windows and Linux

**F3D Information**
Paste the content of `f3d --version`: 2.2.1 and master

**Additional context**
Related to #434

snoyer:
This should work around `cxxopts` splitting filenames on commas by retrieving the filenames from the unmatched arguments rather than into a `vector<string>` option.

~~This comes at the cost of the `--input` named argument but I'm unclear if having the filenames as both positional and named args is useful, so hopefully it's fine?~~

~~The `TestHelpPositional` test fails as, contrary to what its name suggest, it checks for `--input` in the help text,
Should it have been looking for `file1 file2 ...` this whole time? or should it have been named something like `TestHelpInputArg` in the first place?~~

edit: after @mwestphal request for backward compatibility:
- the `--input` argument is kept but deprecated as it would be a lot of work to make it work with commas
- the positional filenames are handled through the unmatched arguments instead of `parse_positional` and commas are ok there

would fix #1070


application/F3DOptionsParser.cxx
```diff
@@ -413,11 +409,18 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
 
       auto unmatched = result.unmatched();
 
-      if (unmatched.size() > 0)
+      inputs.clear();
```
snoyer, line 412: This `clear()` feels out of place but it's needed because `ConfigurationOptions::GetOptions` is called multiple times with `parseCommandLine=true` and I don't have a good enough understanding of the overall life-cycle of the app to come up with a better solution.

application/F3DOptionsParser.cxx
```diff
@@ -289,11 +289,10 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
   try
   {
     cxxopts::Options cxxOptions(this->ExecutableName, F3D::AppTitle);
-    cxxOptions.positional_help("file1 file2 ...");
+    cxxOptions.custom_help("[OPTIONS...] file1 file2 ...");
 
     // clang-format off
     auto grp0 = cxxOptions.add_options("Applicative");
-    this->DeclareOption(grp0, "input", "", "Input files", inputs, LocalHasDefaultNo, MayHaveConfig::YES , "<files>");
```
mwestphal, line 296: that removes the `--input` option. this would require deprecation

application/F3DOptionsParser.cxx
```diff
@@ -289,11 +289,10 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
   try
   {
     cxxopts::Options cxxOptions(this->ExecutableName, F3D::AppTitle);
-    cxxOptions.positional_help("file1 file2 ...");
+    cxxOptions.custom_help("[OPTIONS...] file1 file2 ...");
```
mwestphal, line 292: what does this do ?

application/F3DOptionsParser.cxx
mwestphal: indeed. I dont have a better solution to suggest, but adding a comment would go a long way

application/F3DOptionsParser.cxx
mwestphal: related test is broken btw

snoyer: `positional_help` has no effect when there's no positional arguments so I used `custom_help` to mimic the behavior.
What do you mean about the test?

mwestphal: if you check the CI (or run tests locally) you will see:

```
Errors while running CTest
Output from these tests are in: /__w/f3d/f3d/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
1/1 Test #372: f3d::TestHelpPositional ..........***Failed  Required regular expression not found. Regex=[--input
]  0.04 sec

```

It is about a regex that check for the presence of `input`, which obviously is not here anymore.

application/F3DOptionsParser.cxx
mwestphal: We can, but we need to deprecate it first. So we need to keep supporting it somehow, even if it doesnt support commas.

application/testing/CMakeLists.txt
```diff
@@ -176,6 +176,8 @@ f3d_test(NAME TestLightIntensityDarker DATA cow.vtp ARGS --light-intensity=0.2 D
 f3d_test(NAME TestLightIntensityBrighterFullScene DATA WaterBottle.glb ARGS --light-intensity=5.0 THRESHOLD 60 DEFAULT_LIGHTS)
 f3d_test(NAME TestLightIntensityDarkerFullScene DATA WaterBottle.glb ARGS --light-intensity=0.2 DEFAULT_LIGHTS)
 f3d_test(NAME TestUTF8 DATA "(ノಠ益ಠ )ノ.vtp" DEFAULT_LIGHTS)
+f3d_test(NAME TestFilenameCommas DATA a,b,c.d REGEXP "a,b,c\.d\" does not exist")
+f3d_test(NAME TestFilenameSpaces DATA "a b c.d" REGEXP "a b c\.d\" does not exist")
```
mwestphal, line 180: I would be more confortable with an actual test that open a file named like this. maybe put spaces and commas in a single file ?

application/F3DOptionsParser.cxx
```diff
@@ -274,6 +274,8 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
   std::vector<std::string>& inputs, std::string filePathForConfigBlock, bool allOptionsInitialized,
   bool parseCommandLine)
 {
+  inputs.clear(); /* that would not be necessary if this function wasn't called multiple times */
```
mwestphal, line 277: It is needed to call it multipe time, maybe better to say:

"Required because this function is called multiple times"

application/F3DOptionsParser.cxx
```diff
@@ -402,22 +406,37 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
...
       auto result = cxxOptions.parse(this->Argc, this->Argv);
 
-      auto unmatched = result.unmatched();
+#ifndef F3D_NO_DEPRECATED
+      for (const std::string& input : deprecatedInputs)
+      {
+        /* `deprecatedInputs` may contain an empty string instead of being empty itself */
+        if (!input.empty())
+        {
+          f3d::log::warn("--input option is deprecated, please use positional arguments instead.");
```
mwestphal, line 421: can it be empty though ?

mwestphal (changes requested):
Only small changes required

application/F3DOptionsParser.cxx
snoyer: > It is needed to call it multipe time

[[citation needed](#)] ;p

But you are right about the phrasing

application/F3DOptionsParser.cxx
snoyer: Actually we'd prefer it empty and we want to warn of the deprecation if it is not.
But we cannot do `if(!deprecatedInputs.empty()) warn()` because sometimes (I don't understand why) `deprecatedInputs` comes up as containing an empty string (`{"", }`) instead of empty (`{}`, which also happens, as expected). So we have to check for any non empty element and warn then.

TBH I haven't wrapped my head around why/how the arguments are parsed multiple times and I'm just mitigating the mysterious (to me) side-effects. Hopefully It's not breaking anything along the way.

mwestphal:
@Meakk ptal

Meakk (approved):
Nice workaround, it's a shame we have to do that but I can live with it.

testing/baselines/TestFilenameCommasSpaces.png
mwestphal, line 1: Arent we suposed to have a filename visible on this baseline ?

mwestphal: actually, no, I got baited by the name of the test, although displaying the filename in the test would kinda make sense.

Alex-Kent:
For anyone using an older version on *nix, here's a partial work-around.  This can be specified as the program to use to open a file when double-clicking on it from one's file manager.  It takes a single argument (the file's path).  Adjust f3d's command line options to taste within the script as they can't be passed on the command line.  Place the script into e.g. `/usr/local/bin/f3d-wrapper`

```
#!/bin/bash

IFS=$(echo -en "\n\b")

d=$( dirname $1 )
f=$( basename $1 )

cd "$d"
f3d "$f" --bg-color=0.1,0.1,0.1 --roughness=0.5 --ssao --tone-mapping
```

Issue #350, opened 2022-08-10
Log should be behind model window by default (Windows)


astandarduser:
Each time I open a model the log window is always directly covering the model window, not sure if it's possible currently but would be great to have it set behind the model window regardless of errors

mwestphal:
Indeed, that would be nice. We will take a look if this is possible.

astandarduser:
Deleted MTL's are probably a big reason, always end up deleting them

astandarduser:
I know many artists do the same

mwestphal:
What error do you have exactly ? Do you think F3D should avoid warning in that case ?

astandarduser:
I'll get it for you tomorrow, it's always about missing mtls as I mentioned. It's worth keeping the warning so people know why their texture isn't missing but the logging window should still show up behind regardless, that's the case for all 3d software I have used

astandarduser:
> What error do you have exactly ? Do you think F3D should avoid warning in that case ?

ERROR: In D:\a\f3d\f3d\dependencies\vtk\IO\Import\vtkOBJImporter.cxx, line 867
vtkOBJPolyDataProcessor (000002633F28DE60):  material 'lambert5SG' appears in OBJ but not MTL file?

mwestphal:
Ok, I think this is a fair warning, you may want to edit your obj to avoid it. Anyway, we will take a look at putting the window behind.

jiangsheng:
vtkF3DWin32OutputWindow::Initialize calls vtkWin32OutputWindow::Initialize() which calls ShowWindow(win, SW_SHOW); at the end. The ShowWindow call activates the vtkOutputWindow window and steals focus.

mwestphal:
Thanks for the investigation! I will see of the right way to fix this in VTK.

PR #1088, opened 2023-12-06
Put F3D window on top of output window

mwestphal:
Fix: https://github.com/f3d-app/f3d/issues/350

When initializing the output window (on the first show only), put the F3D window on top of the output window so it is less disruptive.


mwestphal:
@jiangsheng ptal

Meakk (approved):
I assume you tested locally?

mwestphal:
> I assume you tested locally?

Of course

PR #1106, opened 2023-12-18
Remove CMake deprecation check for 9.0

Meakk:


library/VTKExtensions/Applicative/Testing/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ if(WIN32 AND F3D_WINDOWS_GUI)
 endif()
 
 if(VTK_VERSION VERSION_LESS_EQUAL 9.1.0)
+  set(CMAKE_WARN_DEPRECATED OFF)
```
mwestphal, line 10: should we set it back on after ?

mwestphal (changes requested):
discussion

library/VTKExtensions/Applicative/Testing/CMakeLists.txt
Meakk: Not sure, I can try but since it's only affecting the current scope I'd say it's fine.

mwestphal: right

mwestphal (approved):

PR #1110, opened 2023-12-20
Fix USD up axis

Meakk:
Make all USD files +Y up by applying a rotation to the root.  
`suzanne.usd` is defined as +Z so this dataset is used to test the fix instead of specifying +Z explicitly in f3d.

### DIFF
--- a/plugins/usd/module/vtkF3DUSDImporter.cxx
+++ b/plugins/usd/module/vtkF3DUSDImporter.cxx
@@ -129,6 +129,7 @@ public:
         pxr::UsdGeomXform xform = pxr::UsdGeomXform(prim);
 
         auto mat = this->GetLocalTransform(xform, timeCode);
+        vtkMatrix4x4::Multiply4x4(currentMatrix, mat, mat);
 
         this->ImportNode(renderer, prim.GetPrototype(), path.AppendChild(prim.GetName()), mat);
       }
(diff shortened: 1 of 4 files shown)


plugins/usd/module/vtkF3DUSDImporter.cxx
```diff
@@ -539,8 +541,20 @@ class vtkF3DUSDImporter::vtkInternals
     pxr::UsdSkelBakeSkinning(this->Stage->Traverse());
     this->Stage->Save();
 
-    vtkNew<vtkMatrix4x4> identity;
-    this->ImportNode(renderer, this->Stage->GetPseudoRoot(), pxr::SdfPath("/"), identity);
+    vtkNew<vtkMatrix4x4> rootTransform;
+
+    pxr::TfToken up = pxr::UsdGeomGetStageUpAxis(this->Stage);
+
+    if (up == pxr::UsdGeomTokens->z)
```
mwestphal, line 548: so it can only be +Z or +Y ? No other possibilities ?

Or maybe I dont get what this condition achieve.

Meakk: Yes, no other possibilities. Here's the documentation about it:

> The stage up axis is encoded as stage metadatum upAxis, whose legal values are "Y" and "Z", as represented by UsdGeomTokens->y and UsdGeomTokens->z.

mwestphal: alright. You mentionned wanting to add some king of logging, lets add it ?

plugins/usd/module/vtkF3DUSDImporter.cxx
```diff
@@ -539,8 +540,22 @@ class vtkF3DUSDImporter::vtkInternals
...
+    {
+      vtkWarningWithObjectMacro(
+        renderer, << "This USD file up axis is +Z, internally rotate it to align with +Y axis");
```
mwestphal, line 550: use F3DLog

Meakk: Can we use F3DLog from VTK classes?

Meakk: Hmmm it seems so... Should we?

mwestphal: F3DLog has been created for logging from VTKExtensions class in F3D.

vtk macro should not be used as they were not supposed to be shown to the user. (they are currently shown in debug mode because it sadly appears to be needed).

Meakk: We need the VTK extension for that, see #1115

mwestphal (approved):

Issue #1103, opened 2023-12-17
F3D exits with error code 0 (success) despite an error having occurred (invalid json config file)


mrdeveloperdude:
**Describe the bug**

When running f3d with an invalid json file as input to the --config commandline parameter, it exits with an error message output on stderr but returns exit code 0 to the shell, which signifies success.

The error message that f3d outputs looks like this:

```
"/tmp/myconfig.json" is not a file of a supported file format.
This is a headless build of F3D, interactive rendering is not supported

```

This constitutes 2 distinct problems;

1. F3D exists with 0 when clearly there is an error preventing normal operation
2. F3D tries to default to interactive mode upon failing to read the json and subsequently fails since we are using the headless version which has no support for interactive mode.

**To Reproduce**
 1. Create an invalid json file on disk named test.json
 2. Run headless version of F3D (other versions may be affected as well) like so: `f3d --config=test.json`

**Expected behavior**
I expect the program to report the error and exit with a non-zero exit code such as 1

**System Information:**
 - Debian 11 x64 inside docker
 - F3D v2.2.1 headless

Meakk:
Probably related to https://github.com/f3d-app/f3d/issues/310
Are you using the occt plugin?

mwestphal:
Please should your "myconfig.json" file

mwestphal:
here is the file that was invalid for @mrdeveloperdude , but it is working well for me.

[Test_files_F3D.zip](https://github.com/f3d-app/f3d/files/13698057/Test_files_F3D.zip)


Lets consider that for this issue, the .json is **just** and we should improve error management in this context

snoyer:
Original info from the Discord discussion:
> Commandline looks like this:
`/usr/bin/f3d --config= /tmp/1FGYnjiFIDanGgzNtCGOBT0bCfaO1fzdu_1702839294626.json /tmp/1FGYnjiFIDanGgzNtCGOBT0bCfaO1fzdu_1702839294626.stl`
.stl file is present and should be valid
The error output produced by f3d looks like this:
`"/tmp/1FGYnjiFIDanGgzNtCGOBT0bCfaO1fzdu_1702839294626.json" is not a file of a supported file format`
`This is a headless build of F3D, interactive rendering is not supported`
(produced on stderr)

`"foo.bar" is not a file of a supported file format` is the message you get when opening an invalid _scene/model/3d_ file.

[...]

PR #1112, opened 2023-12-21
Detect incorrect implicit values usage

mwestphal:
- All options without an implicit value now error out if used with an implicit value
- Add a F3DExFailure exception
 - Fix incorrect return code when using incorrect options

### DIFF
--- a/application/F3DException.h
+++ b/application/F3DException.h
@@ -13,4 +13,13 @@ public:
   }
 };
 
+class F3DExFailure : public std::runtime_error
+{
+public:
+  explicit F3DExFailure(const std::string& what = "")
+    : std::runtime_error(what)
+  {
+  }
+};
+
 #endif
--- a/application/F3DOptionsParser.cxx
+++ b/application/F3DOptionsParser.cxx
@@ -438,6 +438,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
       inputs = deprecatedInputs;
 #endif
 
+      // Check that all options are matched
       auto unmatched = result.unmatched();
       bool found_unknown_option = false;
       for (std::string unknownOption : unmatched)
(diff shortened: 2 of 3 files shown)


snoyer:
That will catch a bunch of stuff at a syntactic level (eg. `--config` should be an error and this PR makes it one) :+1:
But for #1103 there's still a problem at the semantics level (`--config=` and `--config=""` are syntactically valid but nonsensical and should probably generate a warning)

mwestphal:
if a user specify `--config=""` let them specify it. I dont see a problem with that.

It let user keep the default behavior of F3D while still specifying a `--config`. I can imagine a scenario where this is valid.

One could argue that if a user can set `--config= ` involontarly, then they will probably be able to set `--config=""` involontaly as well, but I cant detect it since it is a correct value imo.

snoyer:
The app should be able to generate a warning when it receives an empty string where it expects a filename (and right now it doesn't because it uses an empty string to represent a not-set optional value).
It's not actually in the scope for this PR (it only mentions #1103 but doesn't claim to fix it, my bad) so it doesn't really matter here. 

---

back on the topic of this PR there might be a slight drawback to this whole argparse hackery:
```F3D - A fast and minimalist 3D viewer
Usage:
  bin/f3d [OPTIONS...] file1 file2 ...

 Applicative options:
      --input [=<files>(=f3d_application_reserved)]
                                Input files
      --output [=<png file>(=f3d_application_reserved)]
                                Render to file
      --no-background           No background when render to file
  -h, --help                    Print help
      --version                 Print version details
      --readers-list            Print the list of readers
      --config [=<filePath/filename/fileStem>(=f3d_application_reserved)]
                                Specify the configuration file to use. 
                                absolute/relative path or filename/filestem 
                                to search in configuration file locations.
      --dry-run                 Do not read the configuration file
      --no-render               Do not render anything and quit right after 
                                loading the first file, use with --verbose 
                                to recover information about a file.
      --max-size [=<size in Mib>(=f3d_application_reserved)]
                                Maximum size in Mib of a file to load, 
                                negative value means unlimited (default: 
                                -1)
      --load-plugins [=<paths or names>(=f3d_application_reserved)]
                                List of plugins to load separated with a 
                                comma
...
```
that's a lot of `f3d_application_reserved` leaking out through the help text :/ unfortunately there may not be an easy way to fix it without refactoring the whole `DeclareOption` mechanism :(

mwestphal:
Indeed, dropping this. Cannot be implemented like this anyway.

PR #1113, opened 2023-12-21
Correct options usage in tests

mwestphal:


Meakk (approved):
I'm not against changing that, so go ahead and merge it.  
However, using space should work for options without implicit value.

Issue #590, opened 2023-01-31
Improve/Augment examples


mwestphal:
Currently, there are a few examples in https://github.com/f3d-app/f3d/tree/master/examples, but this is very lacking, lets add many useful examples !

mwestphal:
Adding python example should also be considered

Ni-g-3l:
Hey I'm currently playing with python bindings. And I miss a lot some example to understand how it works exactly. 

I'm trying to reproduce actual CPP examples in python (I can put it in a PR if you want). But it's very slow because I miss Python API documentation like class, methods, and attribute (like Read the docs or Furo). May be we could try to add python API doc generated by sphinx to actual F3D documentation (in Bindings section).

And then what kind of example would you add ? (may be I can add it to the PR too in both Python and CPP)

mwestphal:
> May be we could try to add python API doc generated by sphinx to actual F3D documentation (in Bindings section).
 
 Definitely! If you want to work on that you'd be very welcome! You can look at the python tests, its the best we have as a documentation for now.
 
>  And then what kind of example would you add ?

Anything really, copying the C++ examples would be a start

mwestphal:
Python examples have been added, Ill update the issue to have a more coherent plan for examples in general.

mwestphal:
@snoyer @Meakk : please provide more ideas and concrete solutions :)

mwestphal:
Examples are more and more outdated, setting to improve for 3.0 if possible

Meakk:
I think tests within the examples add confusion for the user.  
We should either remove them and test manually the examples before the release, or add all example tests in a separate folder.

mwestphal:
> I think tests within the examples add confusion for the user. We should either remove them and test manually the examples before the release, or add all example tests in a separate folder.

What do you mean add all examples tests in a separate folder ?
Maybe we could add a testing layer in separate files somehow ?

Meakk:
Something like:
```
examples/
├─ cxx/
│  ├─ testing/
│  ├─ example1/
│  ├─ example2/
├─ python/
│  ├─ testing/
│  ├─ example1/
│  ├─ example2/
```

mwestphal:
Sure, that works for me

mwestphal:
Fixed

PR #1120, opened 2023-12-23
Port CPP example to python

Ni-g-3l:
Hey !

As we discussed in Discord  this week and in order to start answering to this #590 , I ported Cpp examples to python in order to familiarize me with python bindings. 

Feel free to comment if I missed something or make the contribution process is not like this ! I will be happy to follow contribution guidelines If someone send me a link !


mwestphal:
Thanks for your contribution @Ni-g-3l !

Something super nice would be to find a way to automatize the examples testing the same way we test the C++ examples, in the continuous integration. Are you familiar with github actions ?

Ni-g-3l:
Hello,

No I'm not but it's good day to learn how it works. If you can point me the file that setup Examples CPP tests I can see to create equivalent in python. I'v just found Generic CI which call CPP example tests but I didn't found where Tests are declare

mwestphal:
I'm sure @snoyer and @Meakk will have some inputs too. I will also think about the right way to do this before pointing you to a direction.

At first glance, adding a logic in python CI to run the examples you created would be nice. But I'm not familiar with how examples are usually handled in python.

Fixing the current CI is already a good first step! Do not hesitate if you have any questions.

examples/libf3d/cpp/multi-geom/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ endif()
 # Simple testing
 if(BUILD_TESTING)
   enable_testing()
-  add_test(NAME test_multi-geom COMMAND "$<TARGET_FILE:multi-geom>" "${CMAKE_CURRENT_SOURCE_DIR}/data/")
+  add_test(NAME test_multi-geom COMMAND "$<TARGET_FILE:multi-geom>" "${CMAKE_CURRENT_SOURCE_DIR}/../../data/datasets")
```
mwestphal, line 18: why changing the location ?

examples/libf3d/cpp/multi-geom/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ endif()
...
   set_tests_properties(test_multi-geom PROPERTIES
-    PASS_REGULAR_EXPRESSION "Scene bounding box: -0.487464,1,-0.487464,1,-0.5,1")
+    PASS_REGULAR_EXPRESSION "Scene bounding box: -4.44584,5.99809,-3.63704,2.75972,-1.70141,1.70141")
```
mwestphal, line 20: why would the bounding box change ?

examples/libf3d/python/render_image.py
```diff
@@ -0,0 +1,20 @@
...
+    if len(sys.argv) > 4:
+        sys.exit(1)
+
+    try:
+        f3d.Engine.autoload_plugins()
+
+        eng = f3d.Engine(f3d.Window.NATIVE_OFFSCREEN)
+        eng.loader.load_geometry(sys.argv[1])
+
+        img = eng.window.setSize(300, 300).renderToImage()
```
snoyer, line 14:
```suggestion
        eng.window.size = 300, 300
        img = eng.render_to_image()
```
C++ style `setSize` and `renderToImage` are deprecated

examples/libf3d/python/multi_geom.py
```diff
@@ -0,0 +1,26 @@
...
+    # Load static plugins
+    f3d.Engine.autoload_plugins()
+
+    # Create a native window engine
+    eng = f3d.Engine(f3d.Window.NATIVE)
+
+    # Load all files from provided directory as geometries
+    files = [f for f in Path(sys.argv[1]).iterdir() if f.is_file()]
+    for file in files:
+        eng.loader.load_geometry(str(file))
```
snoyer, line 18:
```suggestion
    for file in files:
        try:
            eng.loader.load_geometry(str(file))
        except RuntimeError as e:
            print(e) # or something
```
would avoid crashing the whole program if a file is unsupported (maybe f3d should raise a more helpful `IOError` here actually?)

examples/libf3d/cpp/multi-geom/CMakeLists.txt
Ni-g-3l: To avoid data duplication and use it into python examples tests. But If you prefer I can move python porting to related folders

examples/libf3d/cpp/multi-geom/CMakeLists.txt
Ni-g-3l: Because the data folder change it embed the cow.vtp now.

examples/libf3d/cpp/multi-geom/CMakeLists.txt
mwestphal: examples should be considered independant from the repo if needed, that is why it has its own data directory.

Ni-g-3l: Ok I make the move ! Do you prefer create external python folder for python examples or add each python scripts in the dedicated example folder ?

mwestphal: imo `examples/libf3d/cpp` and `examples/libf3d/python` makes sense.

examples/libf3d/python/multi_geom.py
mwestphal: `load_failure_exception` already exists

snoyer: > `load_failure_exception` already exists

indeed `load_failure_exception : public exception`, and `pybind` maps c++'s `exception` to python's `RuntimeError`.
May be worth investigating https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html#registering-custom-translators so we can map it to `IOError` which is more specific

mwestphal: alright, definitely an improvement to the bindings to consider

examples/libf3d/python/render-image/render_image.py
```diff
@@ -0,0 +1,21 @@
...
+        sys.exit(1)
+
+    try:
+        f3d.Engine.autoload_plugins()
+
+        eng = f3d.Engine(f3d.Window.NATIVE_OFFSCREEN)
+        eng.loader.load_geometry(sys.argv[1])
+
+        eng.window.size = 300, 300
+        img = eng.render_to_image()
```
snoyer, line 15:
```suggestion
        img = eng.window.render_to_image()
```
I messed up my copy-pasting in my previous suggestion :facepalm:, `render_to_image` is a window method.

mwestphal:
Hi @Ni-g-3l , need any help moving forward ?

mwestphal:
> but It didn't find f3d Module

You mean the test is not able to import f3d, is that correct ?

Ni-g-3l:
Yes exactly !

mwestphal:
I think you should be able to position the PYTHONPATH for each test.
https://www.scivision.dev/cmake-ctest-set-environment-variable/

What do you think @Meakk @snoyer ?

snoyer:
> I think you should be able to position the PYTHONPATH for each test. [scivision.dev/cmake-ctest-set-environment-variable](https://www.scivision.dev/cmake-ctest-set-environment-variable/)
> 
> What do you think @Meakk @snoyer ?

Can't we just assume `ctest` is running in an environment where `f3d` has been installed already, ie. `${Python_EXECUTABLE}` can do `import f3d` successfully? Isn't that what the CI does already for the actual testing?

If the question is how to run the examples through `ctest` locally, I'd expect installing the local `f3d` with `pip install ../path-to-f3d-source-dir` (preferably in a virtual env) and running `ctest` to work. However can't say I fully understand the situation at hand...

mwestphal:
> Can't we just assume ctest is running in an environment where f3d has been installed already

I dont think we can do that. ctest should work without the need to modify local env.

Ni-g-3l:
Yes it's exactly the problem, I'm able to run python command in ctest but the python env doesn't embed f3d. Example tests are run at the same time of cpp tests, where I can put the f3d install the python env in order to run test properly ? maybe here ? https://github.com/Ni-g-3l/f3d/blob/feat/add_python_examples/examples/libf3d/python/CMakeLists.txt in the if 'enable_testing' condition ? or would you prefer create new env for each test ?

mwestphal:
@Meakk should find_package F3D provide a path to python module that we could reuse in the testing?

mwestphal:
Specifficaly, I'm thinking of adding a `f3d_PYTHON_DIR` which is set only if `f3d_BINDINGS_PYTHON` is true.

Meakk:
Can we used predefined variables? I think `f3d_ROOT_DIR` should be available

mwestphal:
here is all that we provide for now:

```
#   f3d_BUILD_APPLICATION          Will be enabled if F3D application was built
#   f3d_MODULE_EXTERNAL_RENDERING  Will be enabled if F3D was built with external rendering support
#   f3d_MODULE_RAYTRACING          Will be enabled if F3D was built with raytracing support
#   f3d_MODULE_EXR                 Will be enabled if F3D was built with OpenEXR images support
#   f3d_BINDINGS_PYTHON            Will be enabled if F3D was built with python bindings
#   f3d_BINDINGS_JAVA              Will be enabled if F3D was built with java bindings
#   f3d_INCLUDE_DIR                Absolute path to the include directory of public f3d library classes
#   f3d_CONFIG_DIR                 Path to F3D configuration directory, can be absolute or relative
#   f3d_PLUGINS_INSTALL_DIR        Relative path to F3D plugins install directory

```

Meakk:
Yes but there are other predefined variables by CMake

mwestphal:
```
find_package(f3d REQUIRED)
message(WARNING ${f3d_ROOT_DIR})
```

Do not output anything.

Meakk:
Here are the available variables after the find_package related to f3d:

```
-- f3d_BINDINGS_JAVA=ON
-- f3d_BINDINGS_PYTHON=ON
-- f3d_BUILD_APPLICATION=ON
-- f3d_CONFIG=/usr/lib/cmake/f3d/f3dConfig.cmake
-- f3d_CONFIG_DIR=/etc/f3d
-- f3d_CONSIDERED_CONFIGS=/usr/lib/cmake/f3d/f3dConfig.cmake
-- f3d_CONSIDERED_VERSIONS=2.2.1
-- f3d_DIR=/usr/lib/cmake/f3d
-- f3d_FOUND=1
-- f3d_INCLUDE_DIR=/usr/include
-- f3d_MODULE_EXR=OFF
-- f3d_MODULE_EXTERNAL_RENDERING=ON
-- f3d_MODULE_RAYTRACING=ON
-- f3d_PLUGINS_INSTALL_DIR=lib
-- f3d_VERSION=2.2.1
-- f3d_VERSION_COUNT=3
-- f3d_VERSION_MAJOR=2
-- f3d_VERSION_MINOR=2
-- f3d_VERSION_PATCH=1
-- f3d_VERSION_TWEAK=0
```

A new variable in the package config file could work.  

However, I think the examples look too much to unit tests in my opinion. Ideally, we shouldn't even rely on CMake. If the idea is to make sure the examples run fine, everything should be done in the CI directly. The presence of CMake file will only confuse potential users.  
Honestly, only a `.sh` script (in the CI directory) calling the python files and checking the return code is more than enough.

mwestphal:
> A new variable in the package config file could work.

@Ni-g-3l do you know how to do that ? look at `cmake/f3dConfig.cmake.in`

mwestphal:
@Ni-g-3l We discussed a bit. Lets drop any CI or CMake logic for now 🙂. We will improve that later.

examples/libf3d/CMakeLists.txt
```diff
@@ -1,14 +1,11 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.19)
 
-project(libf3d-examples)
+project(examples)
```
mwestphal, line 3: im not sure we want to change the name of this project ? there is already a project on top called example

examples/libf3d/CMakeLists.txt
```diff
@@ -1,14 +1,11 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.19)
```
mwestphal, line 1: is that needed ?

examples/libf3d/CMakeLists.txt
```diff
@@ -1,14 +1,11 @@
...
 
 if(BUILD_TESTING)
   enable_testing()
 endif()
 
-add_subdirectory(check-engine)
-add_subdirectory(multi-geom)
-add_subdirectory(render-image)
-add_subdirectory(render-interact)
+add_subdirectory(cpp)
```
mwestphal, line 11: just keep this, there rest is not needed

mwestphal:
Could you add a simple doc pointer to the source so python users can find the example from https://f3d.app/doc/libf3d/BINDINGS.html#python ?

Ni-g-3l:
Hey ! I will do it tomorrow when I come back home !

examples/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.19)
+cmake_minimum_required(VERSION 3.5)
```
mwestphal, line 1: this is not supposed to change. The "plugin" examples requires 3.19.

examples/libf3d/CMakeLists.txt
```diff
@@ -2,13 +2,4 @@ cmake_minimum_required(VERSION 3.5)
...
-
-if(BUILD_TESTING)
-  enable_testing()
-endif()
-
-add_subdirectory(check-engine)
-add_subdirectory(multi-geom)
-add_subdirectory(render-image)
-add_subdirectory(render-interact)
+add_subdirectory(cpp)
```
mwestphal, line 5: missing end of line

examples/libf3d/python/render-image/check.py
```diff
@@ -0,0 +1,16 @@
...
+
+    img_0 = f3d.Image(sys.argv[1])
+    img_1 = f3d.Image(sys.argv[2])
+
+    diff = f3d.Image()
+    error = 0.0
+
+    result = img_0.compare(img_1, 50, diff, error)
+
+    sys.exit(0 if result else 1)
```
mwestphal, line 16: since there is no automated tests anymore, do we need this one ?

mwestphal:
I dont think you need to add any new LFS data

examples/libf3d/python/render-image/check.py
Ni-g-3l: May be we can move it to another folder to give another example of f3d use ? We can call it img-comp or something like that

examples/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.19)
+cmake_minimum_required(VERSION 3.19cd)
```
mwestphal, line 1: that doesnt look right :)

examples/libf3d/python/render-image/baselines/cow.png
mwestphal, line 1: I think you can remove all LFS data, including the images.

Ni-g-3l: How I can do that ? I just remove it from my repo ?

mwestphal: sorry, I was not very clear.

You change is adding a few data files, you can remove them, eg:

examples/libf3d/python/render-interact/data/cow.vtp

Ni-g-3l: Ok I think I understand the problem, but how I can remove it from repository ?

mwestphal: to do that, simply 

```
git rm examples/libf3d/python/render-interact/data/cow.vtp
git commit
git push
```

Ni-g-3l: Hello, I've just done it, but now python examples havn't  any data to use in order to test ?

Meakk (approved):

mwestphal:
Thanks for your contribution @Ni-g-3l ! I will update the issue to reflect what was done and what more could be done.

PR #1147, opened 2024-01-08
Upload Android binaries

Meakk:


.github/actions/android-ci/action.yml
```diff
@@ -75,3 +75,16 @@ runs:
...
       run: cmake --build . --parallel 2 --target install
+
+    - name: Upload native library
+      uses: actions/upload-artifact@v4
+      with:
+        path: ./build/lib/libf3d-java.so
+        name: f3d-android-native-${{inputs.arch}}
+
+    - name: Upload Java package
+      if: inputs.arch == 'arm64-v8a' # avoid duplication
```
mwestphal, line 86: what duplication ?

.github/workflows/ci.yml
```diff
@@ -359,6 +359,7 @@ jobs:
       uses: ./source/.github/actions/android-ci
       with:
         arch: ${{matrix.arch}}
+        vtk_version: v9.3.0
```
mwestphal, line 362: why not commit ?

Meakk: Vtk::token

.github/actions/android-ci/action.yml
Meakk: F3d.jar is the same regardless of the architecture. I don't want to pollute artifacts with duplicate archives

mwestphal: ok got it

.github/workflows/ci.yml
mwestphal: ha right

mwestphal: please add a link to the issue on top of the step so that we can fix it when we can

mwestphal (approved):

PR #1148, opened 2024-01-09
v2.3.0 release split

mwestphal:


Meakk (approved):
Didn't really review but I agree with the split ^^

PR #1164, opened 2024-01-12
Update VCPKG baseline and improve documentation

Meakk:


doc/dev/GETTING_STARTED.md
```diff
@@ -86,9 +86,14 @@ mkdir build
 
  * Open git bash
 
+Check the VCPKG baseline in `vcpkg.json` at the root of f3d source directory.  
+It should be `c9140a3b500812ad3206317885860d9553b93f13` (please let us know if it's not)
```
mwestphal, line 90: What does it mean to "check the VCPKG baseline" ?

Keep in mind this guide is for complete beginners. If VCPKG doesnt work out of the box, we may consider modifying the guide to build VTK manually.

Meakk: > modifying the guide to build VTK manually

Definitely not. I don't want to write a guide for every single dependency.

> Check the VCPKG baseline in `vcpkg.json` at the root of f3d source directory.

I'll be more clear, but that's something a beginner can check easily.

Meakk: Let me know if it's better now.

mwestphal: On windows, VTK is the only needed dependency afaik

Meakk: And plenty of other optional dependencies. I still think VCPKG is way more beginner-friendly.

mwestphal: I agree

doc/dev/GETTING_STARTED.md
```diff
@@ -86,9 +86,14 @@ mkdir build
 
  * Open git bash
 
+Check the value of `builtin-baseline` in the `vcpkg.json` file at the root of f3d source directory.  
+Make sure it's `c9140a3b500812ad3206317885860d9553b93f13` (please let us know if it's not)
```
mwestphal, line 90: I don't think that comment is important? As long as we ensure the hash in the guide is the same as the one in the JSON file unit will work, right ?

Meakk: We may forgot right? There's nothing in the CI that ensure it's the case

Meakk: Can we come up with a bash command to read a value in a json file and check if the value is found in the documentation file?

mwestphal: To me its similar to forgetting that archlinux dependencies are optional for vtk, we just need to put good info in the guide. If we change it in one location, we need to change it in the other location. Its just doc.

Meakk: Fair enough. I'll remove the comment, but you have been warned ^^

mwestphal: move this forward ?

Meakk: done

mwestphal (approved):

PR #1165, opened 2024-01-12
Skip hidden and proxy trees in USD

Meakk:


Meakk:
@snoyer @mwestphal please take a look.  
We need to find a small asset for testing that.

mwestphal (approved):

PR #1172, opened 2024-01-14
Minor picking fixes

snoyer:
Focal point picking with `--grid` and `--verbose` logs this message (still works fine, though):
```
ERROR: In vtkExecutive.cxx, line 344
vtkCompositeDataPipeline (0x563398629000): Attempt to get input information vector from input port index 0 for algorithm vtkF3DOpenGLGridMapper (0x5633987555e0), which has 0 input ports.
```
I don't know whether we never checked with both these options or if this only started showing up with recent VTK logging changes.

Making the grid not pickable makes the message go away.
I also simplified some conditional logic for the picking (amazing what reading the docs can do!)


mwestphal:
please target release instead of master so this change can get into 2.3.0 :)

Also update the changelog accordingly.

mwestphal:
https://github.com/f3d-app/f3d/pull/1173

PR #1173, opened 2024-01-14
Fix grid picking

mwestphal:
Focal point picking with `--grid` and `--verbose` logs this message (still works fine, though):
```
ERROR: In vtkExecutive.cxx, line 344
vtkCompositeDataPipeline (0x563398629000): Attempt to get input information vector from input port index 0 for algorithm vtkF3DOpenGLGridMapper (0x5633987555e0), which has 0 input ports.
```
I don't know whether we never checked with both these options or if this only started showing up with recent VTK logging changes.

Making the grid not pickable makes the message go away.
I also simplified some conditional logic for the picking (amazing what reading the docs can do!)

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -26,6 +26,7 @@ For F3D users:
  - Fixed dependency issue with the headless .deb binary release
  - Fixed a crash when using a file with more than 254 bones (Requires OpenGL 4.3)
  - Fixed an issue with Unicode filenames on the title bar on Windows
+ - Fixed an issue where focal point picking would pick the grid
  - Changed `--verbose` into a string based option, eg: `--verbose=quiet` or `--verbose=debug`. `--verbose` is still supported.
  - Changed `--no-render` behavior so that it doesn't impact verbosity anymore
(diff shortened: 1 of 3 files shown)


mwestphal:
@snoyer

doc/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ For F3D users:
  - Fixed dependency issue with the headless .deb binary release
  - Fixed a crash when using a file with more than 254 bones (Requires OpenGL 4.3)
  - Fixed an issue with Unicode filenames on the title bar on Windows
+ - Fixed an issue where focal point picking would pick the grid
```
snoyer, line 29: Maybe something more like
```suggestion
 - Fixed an issue where focal point picking would generate error messages
```
not to make it sound like the grid was actually getting picked ? (which it wasn't and the user-facing behavior hasn't changed, it's just quiet now)

mwestphal: right, I should have read more carefully :)

snoyer (approved):
I think it's fine but that's not very objective now is it? ;)

Meakk (approved):

PR #1208, opened 2024-01-21
Add full scene support in json metadata

Meakk:
We'll use the json file in Android to determine the list of supported extensions. On top of that we need to know if we can call `loadScene` or `loadGeometry`. This PR adds a field `full_scene` which is true if `loadScene` is supported.


mwestphal:
I suppose this is needed for f3d-android, is that correct ? Could you explain a bit the usecase in the PR description ?

Meakk:
Done

mwestphal (approved):

PR #1217, opened 2024-01-23
Partially revert a8fdf79e6e7b4006e71ae8a5db0592796a0a1477

pbsds:
Addresses https://github.com/NixOS/nixpkgs/issues/262328 by partially reverting a8fdf79e6e7b4006e71ae8a5db0592796a0a1477, found by bisecting between v2.1.0 and v2.2.0

I don't know what the issue was or how to properly fix it, as i don't speak cmake

### DIFF
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -177,17 +177,11 @@ if(F3D_STRICT_BUILD AND MSVC)
   target_compile_definitions(libf3d PRIVATE _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS)
 endif()
 
-if(APPLE)
-  set_target_properties(libf3d PROPERTIES INSTALL_RPATH "@loader_path")
-elseif(UNIX)
-  set_target_properties(libf3d PROPERTIES INSTALL_RPATH "$ORIGIN")
-endif()
-
 if (APPLE OR UNIX)
   # Make sure rpath is correctly set for libf3d
   get_target_property(target_type VTK::CommonCore TYPE)
   if (target_type STREQUAL SHARED_LIBRARY)
-    set_target_properties(libf3d PROPERTIES BUILD_RPATH "$<TARGET_FILE_DIR:VTK::CommonCore>")
+    set_target_properties(libf3d PROPERTIES INSTALL_RPATH "$<TARGET_FILE_DIR:VTK::CommonCore>")
   endif ()
 endif ()
 


pbsds:
For the interested, [here is our build instructions](https://github.com/NixOS/nixpkgs/blob/nixos-23.11/pkgs/applications/graphics/f3d/default.nix#L20-L24), and here are the resulting CMAKE arguments during build:

```
-DCMAKE_FIND_USE_SYSTEM_PACKAGE_REGISTRY=OFF 
-DCMAKE_FIND_USE_PACKAGE_REGISTRY=OFF 
-DCMAKE_BUILD_TYPE=Release 
-DBUILD_TESTING=OFF 
-DCMAKE_INSTALL_LOCALEDIR=/nix/store/sgil9qqlypzrafijywifrzby2jvgn0rh-f3d-2.2.1/share/locale 
-DCMAKE_INSTALL_LIBEXECDIR=/nix/store/sgil9qqlypzrafijywifrzby2jvgn0rh-f3d-2.2.1/libexec 
-DCMAKE_INSTALL_LIBDIR=/nix/store/sgil9qqlypzrafijywifrzby2jvgn0rh-f3d-2.2.1/lib 
-DCMAKE_INSTALL_DOCDIR=/nix/store/sgil9qqlypzrafijywifrzby2jvgn0rh-f3d-2.2.1/share/doc/F3D 
-DCMAKE_INSTALL_INFODIR=/nix/store/sgil9qqlypzrafijywifrzby2jvgn0rh-f3d-2.2.1/share/info 
-DCMAKE_INSTALL_MANDIR=/nix/store/pwzc62pjf29h3rxp7z4nvvad326xkw6v-f3d-2.2.1-man/share/man 
-DCMAKE_INSTALL_OLDINCLUDEDIR=/nix/store/sgil9qqlypzrafijywifrzby2jvgn0rh-f3d-2.2.1/include 
-DCMAKE_INSTALL_INCLUDEDIR=/nix/store/sgil9qqlypzrafijywifrzby2jvgn0rh-f3d-2.2.1/include 
-DCMAKE_INSTALL_SBINDIR=/nix/store/sgil9qqlypzrafijywifrzby2jvgn0rh-f3d-2.2.1/sbin 
-DCMAKE_INSTALL_BINDIR=/nix/store/sgil9qqlypzrafijywifrzby2jvgn0rh-f3d-2.2.1/bin 
-DCMAKE_INSTALL_NAME_DIR=/nix/store/sgil9qqlypzrafijywifrzby2jvgn0rh-f3d-2.2.1/lib 
-DCMAKE_OSX_SYSROOT= 
-DCMAKE_FIND_FRAMEWORK=LAST 
-DCMAKE_STRIP=/nix/store/xq8920m5mbd83vdlydwli7qsh67gfm5v-gcc-wrapper-13.2.0/bin/strip 
-DCMAKE_RANLIB=/nix/store/xq8920m5mbd83vdlydwli7qsh67gfm5v-gcc-wrapper-13.2.0/bin/ranlib 
-DCMAKE_AR=/nix/store/xq8920m5mbd83vdlydwli7qsh67gfm5v-gcc-wrapper-13.2.0/bin/ar 
-DCMAKE_C_COMPILER=gcc 
-DCMAKE_CXX_COMPILER=g++ 
-DCMAKE_INSTALL_PREFIX=/nix/store/sgil9qqlypzrafijywifrzby2jvgn0rh-f3d-2.2.1 
-DCMAKE_INSTALL_LIBDIR=lib 
-DCMAKE_INSTALL_INCLUDEDIR=include 
-DCMAKE_INSTALL_BINDIR=bin 
-DF3D_LINUX_GENERATE_MAN=ON 
```

Meakk:
I'm a bit surprised by the install path `-DCMAKE_INSTALL_PREFIX=/nix/store/sgil9qqlypzrafijywifrzby2jvgn0rh-f3d-2.2.1`  
Does the package stays there? Isn't there a common path like `/usr/lib` where all the libraries are located?  
Where is VTK installed?

mwestphal:
@bcdarwin may be of help

bcdarwin:
> I'm a bit surprised by the install path `-DCMAKE_INSTALL_PREFIX=/nix/store/sgil9qqlypzrafijywifrzby2jvgn0rh-f3d-2.2.1` Does the package stays there? Isn't there a common path like `/usr/lib` where all the libraries are located? Where is VTK installed?

Hi @Meakk! I'm the packager/maintainer of f3d in nixpkgs. Nixpkgs is a indeed a slightly unusual distribution as you've seen, and we do have ways of fixing the linking issue on our side (either by applying this patch ourself or by retroactively calling `patchElf` after the build), so you and other `f3d` maintainers to feel obligated to deal with this issue, but rather see it as Nix having exposed some assumptions made by `f3d`'s (and unfortunately also `VTK`'s) about the build environment (typically things such as `CMAKE_...` directories being relative paths, dependencies being located in a common directory, or a single dependency's `lib/` and `share/` being located in a common directory).

As you mentioned, Nix puts each package into one or several directories in `/nix/store/...` to allow isolation, reproducibility, and concurrent installation of multiple versions of a package. There are indeed cases where multiple packages are symlinked into a single location but that's generally not the case for linking libraries. VTK is located in a location such as `/nix/store/fwy9cs7v4q1mx57i46d7y5pc0bwfbql9-vtk-9.2.6`, but this will change whenever VTK or its dependencies is updated; we of course use the usual compiler/linker variables/flags to pass these locations.

Finally it's possible that this issue is not directly related to `f3d` but arises from the somewhat unfortunate stanza
```
"-DCMAKE_INSTALL_LIBDIR=lib" "-DCMAKE_INSTALL_INCLUDEDIR=include" "-DCMAKE_INSTALL_BINDIR=bin"
```
included in our build to work around a VTK issue. However, the attached patch appears to resolve the issue, so do consider it if it doesn't break your existing build and isn't totally unexpected from a CMake perspective (unfortunately neither @pbsds nor I are CMake experts).

Thanks again for having taken a look.

mwestphal:
> However, the attached patch appears to resolve the issue, so do consider it if it doesn't break your existing build 

Well it revert a feature that we need for a cleaner python integration. We will need to understand a bit better why you need it.

mwestphal:
The original error is : `f3d: error while loading shared libraries: libvtkInteractionWidgets.so.1: cannot open shared object file: No such file or directory`

I supose that on nixos you rely on rpath ?

bcdarwin:
> The original error is : `f3d: error while loading shared libraries: libvtkInteractionWidgets.so.1: cannot open shared object file: No such file or directory`
> 
> I supose that on nixos you rely on rpath ?

Correct. In this case though VTK is not added to the rpath for some reason. (As I mentioned, we can always fall back to using `patchelf` to fix up the rpath later.)

Meakk:
Alternatively we can just remove `INSTALL_RPATH` and let the packagers set `CMAKE_INSTALL_RPATH` themselves?

mwestphal:
@Meakk fine by me

mwestphal:
So in any case, this is not the correct solution. @pbsds @bcdarwin feel free to open an issue if needed.

PR #1219, opened 2024-01-24
Image to terminal text

snoyer:
Printing an `f3d::image` in a terminal using escape sequences and block characters: once a meme, soon a feature?

### DIFF
--- a/library/public/image.h
+++ b/library/public/image.h
@@ -171,7 +171,7 @@ public:
   /**
    * Convert to colored text using ANSI escape sequences for printing in a terminal.
    * Assumes terminal support for:
-   * - block character (`▄`, `▀`, `█`)
+   * - unicode block characters (U+2580, U+2584, U+2588)
    * - SGR escape codes (`ESC[0m`, `ESC[39m`, `ESC[49m`)
    * - 24-bit escape codes (`ESC[38;2;{r};{g};{b}m`, `ESC[48;2;{r};{g};{b}m`)
    * Throw an exception if the type is not byte RGB or RGBA.
--- a/library/public/image.h
+++ b/library/public/image.h
@@ -170,10 +170,20 @@ public:
 
   /**
    * Convert to colored text using ANSI escape sequences for printing in a terminal.
+   * Block and half-block characters are used to represent two pixels per character (vertically)
+   * meaning that each text line represent two rows of pixels.
+   * Byte encoded RGB and RGBA images are supported (alpha channel values are binarized).
    * Assumes terminal support for:
-   * - unicode block characters (U+2580, U+2584, U+2588)
-   * - SGR escape codes (`ESC[0m`, `ESC[39m`, `ESC[49m`)
+   * - unicode block characters (`U+2580`, `U+2584`, `U+2588`)
+   * - SGR escape codes (`ESC[0m`, `ESC[49m`)
    * - 24-bit escape codes (`ESC[38;2;{r};{g};{b}m`, `ESC[48;2;{r};{g};{b}m`)
+   * Throw an exception if the type is not byte RGB or RGBA.
+   */
+  const f3d::image& toTerminalText(std::ostream& stream) const;
+
+  /**
+   * Convert to colored text using ANSI escape sequences for printing in a terminal.
+   * See `toTerminalText(std::ostream& stream)`.
    * Throw an exception if the type is not byte RGB or RGBA.
    */
   std::string toTerminalText() const;
--- a/library/public/image.h
+++ b/library/public/image.h
@@ -169,6 +169,16 @@ public:
   std::vector<unsigned char> saveBuffer(SaveFormat format = SaveFormat::PNG) const;
 
   /**
+   * Convert to colored text using ANSI escape sequences for printing in a terminal.
+   * Assumes terminal support for:
+   * - block character (`▄`, `▀`, `█`)
+   * - SGR escape codes (`ESC[0m`, `ESC[39m`, `ESC[49m`)
+   * - 24-bit escape codes (`ESC[38;2;{r};{g};{b}m`, `ESC[48;2;{r};{g};{b}m`)
+   * Throw an exception if the type is not byte RGB or RGBA.
+   */
+  std::string toTerminalText() const;
+
+  /**
    * An exception that can be thrown by the image when there.
    * is an error on write.
    */
--- a/library/src/image.cxx
+++ b/library/src/image.cxx
@@ -428,11 +428,13 @@ std::string image::toTerminalText() const
     }
   };
 
-  const std::string EMPTY_BLOCK = " ";
-  const std::string TOP_BLOCK = "\u2580";    // U+2580
-  const std::string BOTTOM_BLOCK = "\u2584"; // U+2584
-  const std::string FULL_BLOCK = "\u2588";   // U+2588
-  const std::string EOL = "\n";
+  constexpr std::string_view EMPTY_BLOCK = " ";
+  // clang-format off
+  constexpr std::string_view TOP_BLOCK = u8"\u2580";    // U+2580
+  constexpr std::string_view BOTTOM_BLOCK = u8"\u2584"; // U+2584
+  constexpr std::string_view FULL_BLOCK = u8"\u2588";   // U+2588
+  // clang-format on
+  constexpr std::string_view EOL = "\n";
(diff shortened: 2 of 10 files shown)


library/src/image.cxx
```diff
@@ -351,6 +353,145 @@ std::vector<unsigned char> image::saveBuffer(SaveFormat format) const
...
+    }
+    else if (bg && currentBg > -1)
+    {
+      stream << "\033[49m"; // reset background
+      currentBg = -1;
+    }
+  };
+
+  const std::string EMPTY_BLOCK = " ";
+  const std::string BOTTOM_BLOCK = "▄"; // U+2584
```
michaelmigliore, line 434: I'm not feeling super comfortable with the idea of adding unicode character directly. what about that?
```suggestion
  constexpr std::string_view  = u8"\u2584";
```

library/public/image.h
```diff
@@ -168,6 +168,16 @@ class F3D_EXPORT image
    */
   std::vector<unsigned char> saveBuffer(SaveFormat format = SaveFormat::PNG) const;
 
+  /**
+   * Convert to colored text using ANSI escape sequences for printing in a terminal.
+   * Assumes terminal support for:
+   * - block character (`▄`, `▀`, `█`)
```
michaelmigliore, line 174: same here

```suggestion
   * - unicode block character (U+2584, U+2580, U+2588)
```

library/src/image.cxx
snoyer: `u8"\u1234"` is c++11 so we'd need to update the clang format (currently as per `Standard: Cpp03` it inserts a space).
I used `"\u"` without the `u8` assuming that'd be fine but it doesn't work on windows.

Meakk: > disabling clang-format for that section seems reasonable for now?

Sounds good to me

library/src/image.cxx
```diff
@@ -351,6 +353,145 @@ std::vector<unsigned char> image::saveBuffer(SaveFormat format) const
...
+  // clang-format off
+  constexpr std::string_view TOP_BLOCK = u8"\u2580";    // U+2580
+  constexpr std::string_view BOTTOM_BLOCK = u8"\u2584"; // U+2584
+  constexpr std::string_view FULL_BLOCK = u8"\u2588";   // U+2588
+  // clang-format on
+  constexpr std::string_view EOL = "\n";
+  const unsigned char alphaCutoff = 127;
+
+  unsigned char a1, a2;
+  for (int y = 0; y < height; y += 2)
```
Meakk, line 441: What happens if the height is not even?

library/src/image.cxx
```diff
@@ -351,6 +353,145 @@ std::vector<unsigned char> image::saveBuffer(SaveFormat format) const
...
+  };
+
+  constexpr std::string_view EMPTY_BLOCK = " ";
+  // clang-format off
+  constexpr std::string_view TOP_BLOCK = u8"\u2580";    // U+2580
+  constexpr std::string_view BOTTOM_BLOCK = u8"\u2584"; // U+2584
+  constexpr std::string_view FULL_BLOCK = u8"\u2588";   // U+2588
+  // clang-format on
+  constexpr std::string_view EOL = "\n";
+  const unsigned char alphaCutoff = 127;
```
Meakk, line 438: Every variable known at compile time should be constexpr

library/src/image.cxx
```diff
@@ -351,6 +353,145 @@ std::vector<unsigned char> image::saveBuffer(SaveFormat format) const
...
+    - opaque alpha value when the image has no alpha channel.
+    Rendering with half blocks means 1 line of text represents 2 rows of pixels
+    so we _will_ attempt to access a line past the bottom if the height is not even.
+  */
+  const auto getPixel = [&content, width, height, depth](int x, int y, unsigned char* alpha)
+  {
+    if (x >= 0 && x < width && y >= 0 && y < height)
+    {
+      const size_t i = depth * ((height - 1 - y) * width + x);
+      *alpha = depth > 3 ? content[i + 3] : (unsigned char)255;
```
Meakk, line 382: Instead of casting a int, just use 0xff

mwestphal:
I wonder if we could put a screenshot of this in the doc somewhere ? Not sure where though.

library/src/image.cxx
snoyer: `getPixel` returns transparent black for the non-existing pixels and the bottom row is drawn with only top half blocks with no background

library/src/image.cxx
snoyer: isn't `0xff` the same int tho?

Meakk: It is, but you won't have to cast it to `unsigned char` since it's typed to that already.

library/src/image.cxx
```diff
@@ -351,6 +353,145 @@ std::vector<unsigned char> image::saveBuffer(SaveFormat format) const
...
+  const int height = dims[1];
+  const unsigned char* content = static_cast<unsigned char*>(this->getContent());
+
+  /* Function to retrieve pixels so we can return:
+    - transparent black values for out-of-bounds coords,
+    - opaque alpha value when the image has no alpha channel.
+    Rendering with half blocks means 1 line of text represents 2 rows of pixels
+    so we _will_ attempt to access a line past the bottom if the height is not even.
+  */
+  const auto getPixel = [&content, width, height, depth](int x, int y, unsigned char* alpha)
```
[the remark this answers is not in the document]
snoyer: nice I'll go with the pair then :D. I'll also do the cutoff comparison in there to get rid of the `unsigned char a` altogether and just have `bool transparent` instead

library/public/image.h
```diff
@@ -168,6 +168,16 @@ class F3D_EXPORT image
    */
   std::vector<unsigned char> saveBuffer(SaveFormat format = SaveFormat::PNG) const;
 
+  /**
+   * Convert to colored text using ANSI escape sequences for printing in a terminal.
+   * Assumes terminal support for:
+   * - unicode block characters (U+2580, U+2584, U+2588)
```
Meakk, line 174: Please add a comment explaining that two pixel lines are written per terminal line

Meakk: And also something about the alpha cutoff.

library/src/image.cxx
```diff
@@ -351,6 +353,149 @@ std::vector<unsigned char> image::saveBuffer(SaveFormat format) const
...
+    - opaque alpha value when the image has no alpha channel.
+    Rendering with half blocks means 1 line of text represents 2 rows of pixels
+    so we _will_ attempt to access a line past the bottom if the height is not even.
+  */
+  const auto getPixel = [=](int x, int y)
+  {
+    if (x >= 0 && x < width && y >= 0 && y < height)
+    {
+      const size_t i = depth * ((height - 1 - y) * width + x);
+      const int rgb = content[i + 0] << 16 | content[i + 1] << 8 | content[i + 2];
```
Meakk, line 384: Why are you packing the values in an `int` instead of a `char[3]`?  
The bit manipulation looks useless to me, unless I'm missing something.

snoyer: we need to compare the RGB values to the previous/current background and foreground (that are either RGB or "not set") to optimize the escape sequences so it's easier to compare `int`s that can be unset to `-1` than to drag 3 `uchar`s + 1 `bool` around.
I guess the _right_ way to do it would be to go full `optional<triple<uchar, uchar, uchar>>` but that seems a bit heavy :/
What's your take?

library/src/image.cxx
```diff
@@ -351,6 +353,149 @@ std::vector<unsigned char> image::saveBuffer(SaveFormat format) const
...
+    }
+    return std::make_pair(0x000000, true);
+  };
+
+  /* Functions to manipulate the terminal colors using escape sequences.
+    Keep track of the foreground and background states to avoid redundant sequences.
+  */
+  int currentFg = -1;
+  int currentBg = -1;
+  const auto setFg = [&stream, &currentFg](int rgb)
```
Meakk, line 396: Since everything is captured by reference, let's simplify that to
```suggestion
  const auto setFg = [&](int rgb)
```
(and the other ones below)

library/src/image.cxx
Meakk: Aaaaah got it.

Meakk (approved):
Awesome! :)

PR #1226, opened 2024-01-27
adapt step size to CoC

### LINKED ISSUE #1213: Bokeh blur shader needs adaptive iterations
When using large circle of confusion radius the default 10 iterations of the bokeh blur shader are not sufficient.

This first leads to "inverted Y" appearing when the value goes reasonably above the default `--blur-coc=20`, for example in the following capture with `--blur-coc=40`:
f3d-coc40

Then the effects breaks down as the radius gets even larger, eg. `--blur-coc=200`:
f3d-coc200

This can be mitigated if not outright fixed by increasing the number of iterations. Extrapolating the default `step=0.1` for `coc=20` we can guess that we'd need `step = 2.0 / coc`.
with a recompiled shader with `const float step = 2.0 / 200.0;` and `--blur-coc=200` we get:
f3d-coc200-recompiled

 
The number of iterations/step in the shader must be a constant so the shader needs to be recompiled with updated code when the circle of confusion radius radius is changed.

Adding a `float step` parameter to the function generating the shader code and invalidating said shader for recompilation at appropriate times sounds like the solution but I'm not clear enough on the life cycle of the app/renderer to manage it. Hopefully someone else (hi, @Meakk!) can get around to it ;)

snoyer:
Turns out render passes are re-instantiated when the options have changed so we don't need to cache/invalidate the shader internally and we only need to compute the correct step size before compiling it.
@Meakk can you confirm? I feel like I'm just winging it here

would fix #1213


mwestphal:
Please add a test :)

snoyer:
> Please add a test :)

Coverage looks fine to me, do you mean "fix the existing test that wasn't catching the issue before" ? ;p

I believe we need an hdri that shows more obvious bokeh that the one currently in the testing data, should we add one or swap the existing one? an if we swap should it be a separate PR?

mwestphal:
> I believe we need an hdri that shows more obvious bokeh that the one currently in the testing data, should we add one or swap the existing one? an if we swap should it be a separate PR?

Perfectly ok to add one, as long as is a low res one.

Meakk:
Ok then it's OK. Rebuilding the whole rendering pipeline when an option changes sounds inefficient to me tho. We'll have to think about it later.

Meakk (approved):
Perfect! It's also not changing the result for the default COC value which is great!

Meakk:
Yes, that sounds good to me but we are looping from 0 to 1 so a step value of 2 does not make sense. I'd say we need at least 4 iterations, so what about `2.0 / max(CoC, 8.0)`?

snoyer:
> what about `2.0 / max(CoC, 8.0)`?

I was going to do something clunky like `max(2.0 / max(CoC, minCoC), maxStep)` but yours is better :+1: I'll do that

PR #1235, opened 2024-01-29
Orthographic projection

kidharb:
Added orthographic projection by pressing 5


kidharb:
The following tests are failing, I am investigating

 	168 - f3d::TestInteractionCheatsheetRaytracing (Failed)
	169 - f3d::TestInteractionCheatsheetWhiteBGRaytracing (Failed)
	170 - f3d::TestInteractionCheatsheetBlackBGRaytracing (Failed)
	171 - f3d::TestInteractionCheatsheetScalarsRaytracing (Failed)
	172 - f3d::TestInteractionNoFileCheatsheetRaytracing (Failed)

mwestphal:
Yes, the baselines needs to be updated surely

doc/user/OPTIONS.md
```diff
@@ -88,16 +88,17 @@ Options|Default|Description
...
+------|---------|------
+\-\-camera-position=\<X,Y,Z\>|         |Set the camera position, overrides --camera-direction and camera-zoom-factor.
+\-\-camera-focal-point=\<X,Y,Z\>|         |Set the camera focal point.
+\-\-camera-view-up=\<X,Y,Z\>|         |Set the camera view up vector. Will be orthogonalized.
+\-\-camera-view-angle=\<angle\>|         |Set the camera view angle, a strictly positive value in degrees.
+\-\-camera-direction=\<X,Y,Z\>|         |Set the camera direction, looking at the focal point.
+\-\-camera-zoom-factor=\<factor\>|         |Set the camera zoom factor relative to the autozoom on data, a strictly positive value.
+\-\-camera-azimuth-angle=\<angle\>| 0.0     |Apply an azimuth transformation to the camera, in degrees, added after other camera options.
+\-\-camera-elevation-angle=\<angle\>| 0.0     |Apply an elevation transformation to the camera, in degrees, added after other camera options.
+\-\-camera-orthographic|         |Enable orthographic projection.
```
mwestphal, line 101: this change is not needed

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -1338,6 +1339,22 @@ void vtkF3DRenderer::ShowEdge(bool show)
   }
 }
 
+//----------------------------------------------------------------------------
+void vtkF3DRenderer::SetUseOrthographicProjection(bool use)
+{
+  vtkCamera* camera = GetActiveCamera();
+  if (use)
```
mwestphal, line 1346: how do you turn it off then ?

library/src/camera_impl.cxx
```diff
@@ -263,4 +263,4 @@ vtkCamera* camera_impl::GetVTKCamera()
 {
   return this->Internals->VTKRenderer->GetActiveCamera();
 }
-};
+}
```
mwestphal, line 266: I think this is needed

library/src/interactor_impl.cxx
```diff
@@ -319,6 +320,10 @@ class interactor_impl::internals
         self->SetViewOrbit(ViewType::VT_RIGHT, self);
         render = true;
         break;
+      case '5':
```
mwestphal, line 323: you should add an interactive test for this

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
kidharb: Point noted, we need an else part to capture the use=FALSE case

doc/user/OPTIONS.md
kidharb: I will remove these options

library/src/interactor_impl.cxx
kidharb: Definitely, the codecov is also failing because there is not test

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
mwestphal: look at other turn on/off feature like the one just below :)

library/src/camera_impl.cxx
kidharb: This is needed as it closes the `namespace f3d::detail
{`

on line 9. Should I fix the indentation so it does not look like it is there by mistake?

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -1244,6 +1244,7 @@ void vtkF3DRenderer::ConfigureCheatSheet()
     cheatSheetText << " DOWN : Add files from dir of current file\n";
     cheatSheetText << "\n 1: Front View camera\n";
     cheatSheetText << " 3: Right View camera\n";
+    cheatSheetText << " 5: Orthographic Projection\n";
```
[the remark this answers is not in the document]
kidharb: Changed to "Toggle Orthographic Projection"

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
kidharb: Since this is toggle, we do not need to keep track of the state here

library/src/interactor_impl.cxx
kidharb: I added the test cases for orthographic projection and updated the baseline

doc/user/OPTIONS.md
```diff
@@ -88,16 +88,13 @@ Options|Default|Description
...
-\-\-camera-zoom-factor=\<factor\>||Set the camera zoom factor relative to the autozoom on data, a strictly positive value.
-\-\-camera-azimuth-angle=\<angle\>|0.0|Apply an azimuth transformation to the camera, in degrees, added after other camera options.
-\-\-camera-elevation-angle=\<angle\>|0.0|Apply an elevation transformation to the camera, in degrees, added after other camera options.
+Options| Default |Description
+------|---------|------
+\-\-camera-position=\<X,Y,Z\>|         |Set the camera position, overrides --camera-direction and camera-zoom-factor.
+\-\-camera-focal-point=\<X,Y,Z\>|         |Set the camera focal point.
+\-\-camera-view-up=\<X,Y,Z\>|         |Set the camera view up vector. Will be orthogonalized.
+\-\-camera-view-angle=\<angle\>|         |Set the camera view angle, a strictly positive value in degrees.
+\-\-camera-direction=\<X,Y,Z\>|         |Set the camera direction, looking at the focal point.
```
mwestphal, line 97: remove not needed change

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
mwestphal: also its better to use the same design as other boolean option

library/src/camera_impl.cxx
mwestphal: you lost me there. there is no need for this change as far as i can see

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -1338,6 +1339,22 @@ void vtkF3DRenderer::ShowEdge(bool show)
...
+  vtkCamera* camera = GetActiveCamera();
+  if (use)
+  {
+    const double* position = camera->GetPosition();
+    const double* focal = camera->GetFocalPoint();
+    const double angle = camera->GetViewAngle();
+    double distance = std::sqrt(vtkMath::Distance2BetweenPoints(position, focal));
+    double parallelScale = distance * tan(0.5 * vtkMath::RadiansFromDegrees(angle));
+    camera->SetParallelScale(parallelScale);
+  }
```
snoyer, line 1354: So here we compute the `parallelScale` from the `distance` (to the focal point) when turning parallel projection on?
But we don't do the inverse by repositioning the camera at a `distance` (from the focal point) computed from `parallelScale` when we turn parallel projection off?

It feels like there should be an `else` block going something like
```cpp
double distance = camera->GetParallelScale() / tan(angle/2);
// new_pos = focal + distance * normed(pos - focal)
camera->setPosition(new_pos)
```
@Meakk to confirm?

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
snoyer, line 1346:
```suggestion
  vtkCamera* camera = GetActiveCamera();
  if (use == camera->GetParallelProjection())
    return;
  if (use)
```
can we shortcut the computation if we're already in the right mode? or is there some cases where they need to be done anyway?

mwestphal: We can, but lets use an internal state instead.

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
Meakk: To me it feels like we need a `--camera-parallel-scale` for orthographic mode, just like we have `--camera-view-angle` for perspective mode.  
I don't know where the formula above comes from, but it looks like a heuristic to me.

testing/recordings/TestInteractionCameraHotkeys.log
```diff
@@ -3,6 +3,9 @@ RenderEvent 0 0 0 0 0 0 0
 KeyPressEvent 0 0 0 51 1 3 0
 CharEvent 0 0 0 51 1 3 0
 KeyReleaseEvent 0 0 0 51 1 3 0
+KeyPressEvent 0 0 0 53 1 5 0
```
Meakk, line 6: Should we create a new test instead of modifying an existing one?

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
snoyer: I'm willing to believe the heuristic is good enough, and I think it's needed so the camera doesn't jump back and forth when pressing the toggle key. But it would need to work both ways and I'm not sure it would as is. needs playing around with to find out

testing/recordings/TestInteractionCameraHotkeys.log
kidharb: Yes I agree. We should also create separate tests for each camera position (1,3,7,9)

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
kidharb: I will investigate using the internal state

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -1244,6 +1244,7 @@ void vtkF3DRenderer::ConfigureCheatSheet()
     cheatSheetText << " DOWN : Add files from dir of current file\n";
     cheatSheetText << "\n 1: Front View camera\n";
     cheatSheetText << " 3: Right View camera\n";
+    cheatSheetText << " 5: Toggle Orthographic Projection\n";
```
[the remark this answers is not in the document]
kidharb: This should tie up with getting the  internal state and will be worked on when that is implemented

mwestphal:
There was an issue that required rewriting history. Please cherry pick your commits on f3d-app/f3d master.

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -1340,6 +1341,39 @@ void vtkF3DRenderer::ShowEdge(bool show)
...
+    double direction[3];
+    vtkMath::Subtract(position, focal, direction);
+    vtkMath::Normalize(direction);
+    vtkMath::MultiplyScalar(direction, distance);
+    double newPosition[3];
+    vtkMath::Add(focal, direction, newPosition);
+    camera->SetPosition(newPosition);
+  }
+  camera->SetParallelProjection(use);
+  this->ResetCameraClippingRange();
```
snoyer, line 1374: I won't claim the logic here doesn't work but it feels more complicated than it has to be.
Can't it go something like this?
```c++
SetUseOrthographicProjection(bool use){

  // if the internal state is already the same as the target state there's nothing to do
  if(this->UseOrthographicProjection == use) 
    return;

  // grab camera variables
  vtkCamera* camera = GetActiveCamera();

  if (use) // switching to ortho
  { 
     // set new parallel scale from current distance
  }
  else // switching to perspective
  { 
    // set new distance from current parallel scale
  }

  camera->SetParallelProjection(use); // actually do the switch
  this->UseOrthographicProjection = use; // update internal state

  this->ResetCameraClippingRange(); // needed to avoid rendering issues
}
```

kidharb: It is rather complicated. Are you saying that it would be better to add 2 helper functions and call those from the 'if then else'?

snoyer: no need for helper functions, it's just about the checking/setting of the internal state.

In your current code you do:
```cpp
  if (this->UseOrthographicProjection != use)
  {
    this->UseOrthographicProjection = use;
    // set new parallel scale from current distance
  }
  else
  {
    this->UseOrthographicProjection = !use;
    // set new distance from current parallel scale
  }
  camera->SetParallelProjection(use);
```
and I think that is confusing 2 things:
- checking the internal state: we _want_ to compare the current state and the target state at the top of the function so we can bail early if there's nothing to do
- checking the target state: we _need_ to do this to know which mode we're switching to and adjust the camera accordingly

mwestphal: I fully agree with @snoyer analysis here. :)

testing/recordings/TestInteractionCameraHotkeys.log
kidharb: I think it makes sense to add to this test suite since this is testing the camera hotkeys. 
However, let's also create another issue to separate each interaction into its own test, as there is value in checking the baseline image after each key is pressed instead of only checking that the last baseline image is correct.

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -1340,6 +1341,43 @@ void vtkF3DRenderer::ShowEdge(bool show)
...
+  // if the internal state is already the same as the target state there's nothing to do
+  if (this->UseOrthographicProjection == use)
+    return;
+
+  vtkCamera* camera = GetActiveCamera();
+  const double angle = vtkMath::RadiansFromDegrees(camera->GetViewAngle());
+  const double* position = camera->GetPosition();
+  const double* focal = camera->GetFocalPoint();
+
+  if (this->UseOrthographicProjection != use)
```
snoyer, line 1356:
```suggestion
  if (use)
```
you've already done the comparison of `this->UseOrthographicProjection` and `use` at the top to return early.
At this point you only need to check the value of `use` knowing that `use == true` (`if(use)`) means going from perspective to ortho and `use == false` (`else`) means going from ortho to perspective

testing/recordings/TestInteractionCameraHotkeys.log
kidharb: I have created a new test only for Orthographic Projection and removed the test case from CameraHotKeys

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
kidharb: Thanks, you are a much better programmer than me :-P

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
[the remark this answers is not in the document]
mwestphal: you still need to report if its on or off here

library/VTKExtensions/Rendering/vtkF3DRenderer.h
```diff
@@ -77,6 +77,14 @@ class vtkF3DRenderer : public vtkOpenGLRenderer
   void SetRaytracingSamples(int samples);
   ///@}
 
+  ///@{
+  /**
+   * Set/Get SetUseOrthographicProjection
+   */
+  void SetUseOrthographicProjection(bool use);
+  vtkGetMacro(UseOrthographicProjection, bool);
```
mwestphal, line 85: if you dont need the getter, do not declare it.

library/src/camera_impl.cxx
```diff
@@ -263,4 +263,4 @@ vtkCamera* camera_impl::GetVTKCamera()
 {
   return this->Internals->VTKRenderer->GetActiveCamera();
 }
-};
+};
```
mwestphal, line 266: missing end of line

testing/recordings/TestInteractionOrthographicProjection.log
```diff
@@ -0,0 +1,3 @@
+KeyPressEvent 0 0 0 53 1 5 0
+CharEvent 0 0 0 53 1 5 0
+KeyReleaseEvent 0 0 0 53 1 5 0
```
mwestphal, line 3: check other log file, this files is not complete I believe

library/VTKExtensions/Rendering/vtkF3DRenderer.cxx
```diff
@@ -1340,6 +1341,43 @@ void vtkF3DRenderer::ShowEdge(bool show)
...
+void vtkF3DRenderer::SetUseOrthographicProjection(bool use)
+{
+  // if the internal state is already the same as the target state there's nothing to do
+  if (this->UseOrthographicProjection == use)
+    return;
+
+  vtkCamera* camera = GetActiveCamera();
+  const double angle = vtkMath::RadiansFromDegrees(camera->GetViewAngle());
+  const double* position = camera->GetPosition();
+  const double* focal = camera->GetFocalPoint();
```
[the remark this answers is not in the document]
kidharb: That is correct, I have added a toggle test case

library/VTKExtensions/Rendering/vtkF3DRenderer.h
kidharb: Will remove, thanks

testing/recordings/TestInteractionOrthographicProjection.log
```diff
@@ -0,0 +1,6 @@
+# StreamVersion 1.2
+ExposeEvent 0 599 0 0 0 0 0
+RenderEvent 0 599 0 0 0 0 0
+KeyPressEvent 1129 70 0 53 1 5 0
+CharEvent 1129 70 0 53 1 5 0
+KeyReleaseEvent 1129 70 0 53 1 5 0
```
mwestphal, line 6: missing end of line

library/VTKExtensions/Rendering/vtkF3DRenderer.h
```diff
@@ -77,6 +77,13 @@ class vtkF3DRenderer : public vtkOpenGLRenderer
   void SetRaytracingSamples(int samples);
   ///@}
 
+  ///@{
+  /**
+   * Set/Get SetUseOrthographicProjection
```
mwestphal, line 82: Set instead of Set/Get

library/VTKExtensions/Rendering/vtkF3DRenderer.h
```diff
@@ -77,6 +77,13 @@ class vtkF3DRenderer : public vtkOpenGLRenderer
   void SetRaytracingSamples(int samples);
   ///@}
 
+  ///@{
```
mwestphal, line 80: remove group doxygen doc

mwestphal:
almost there @kidharb ! just a few more changes and we are good for integration :)

testing/recordings/TestInteractionOrthographicProjection.log
kidharb: Removed newline

library/VTKExtensions/Rendering/vtkF3DRenderer.h
kidharb: Removed doxygen tags

library/VTKExtensions/Rendering/vtkF3DRenderer.h
kidharb: Corrected Comment to "Set"

application/testing/CMakeLists.txt
```diff
@@ -695,6 +695,8 @@ f3d_test(NAME TestInteractionGroupGeometriesLoadParentDirectory DATA mb/mb_0_0.v
 f3d_test(NAME TestInteractionInvertZoom DATA suzanne.ply ARGS --invert-zoom INTERACTION DEFAULT_LIGHTS)
 f3d_test(NAME TestInteractionCameraHotkeys DATA cow.vtp INTERACTION DEFAULT_LIGHTS)
 f3d_test(NAME TestInteractionZoomToMouse DATA cow.vtp INTERACTION DEFAULT_LIGHTS)
+f3d_test(NAME TestInteractionOrthographicProjection DATA cow.vtp INTERACTION DEFAULT_LIGHTS)
+f3d_test(NAME TestInteractionZoomToggleOrthographicProjection DATA cow.vtp INTERACTION DEFAULT_LIGHTS)
```
mwestphal, line 699: which one of the test is pressing 5 two times ?

library/VTKExtensions/Rendering/vtkF3DRenderer.h
mwestphal: not correctly removed, Set/Get Use trackball should have them. SetUseOrthographicProjection shoult not.

mwestphal:
Missing an update of INTERACTIONS.md

application/testing/CMakeLists.txt
kidharb: TestInteractionZoomToggleOrthographicProjection

It does a zoom out and then presses 5 and then 5 again. I think that is wrong, it need to be

Press 5
Zoom out
Press 5

kidharb: The updated test case performs the following

     1. Zooms out by 1 mouse scroll wheel rotation
     2. Presses "5" to enable orthographic projection
     3. Zooms out a further 5 rotations of the scroll wheel
     4. Presses "5" to disable orthographic projection
 
The test case was implemented in this way so that it tests both the toggle mode in conjunction with the zoom mode.

I also did not move the mouse during the test so there are no unnecessary events (as before)

mwestphal, line 699: perfect, please add a comment after your test describging what it does, see above, most of the interaction have one.

library/src/window_impl.cxx
```diff
@@ -417,6 +417,12 @@ void window_impl::UpdateDynamicOptions()
     this->Internals->Options.getAsBool("render.grid.absolute"));
   this->Internals->Renderer->ShowGrid(this->Internals->Options.getAsBool("render.grid.enable"));
 
+  if (this->Internals->Options.getAsInt("scene.camera.index") < 0)
```
mwestphal, line 420: more like `== -1` imo

library/src/options.cxx
```diff
@@ -216,6 +216,9 @@ options::options()
   this->Internals->init("model.volume.enable", false);
   this->Internals->init("model.volume.inverse", false);
 
+  // Camera projection
+  this->Internals->init("camera.orthographic", false);
```
mwestphal, line 220: doc for this should be added in doc/libf3d/OPTIONS.md

doc/user/INTERACTIONS.md
```diff
@@ -56,6 +56,7 @@ Note that some hotkeys can be available or not depending on the file being loade
 Camera Hotkeys:
 * <kbd>1</kbd>: front view camera.
 * <kbd>3</kbd>: right view camera.
+* <kbd>5</kbd>: toggle orthographic projection.
```
mwestphal, line 59: @Meakk do you think we should have a command line option for this or only interaction is fine ?

mwestphal: can be done in another PR though

mwestphal (changes requested):
small changes needed

library/src/options.cxx
kidharb: It looks like the `OPTIONS.md` is for command line options. 
This feature is not a command line option at the moment.

Meakk (approved):

mwestphal (approved):

mwestphal:
You need to rebase on master.

mwestphal:
CI is unrelated, merging.

PR #1250, opened 2024-02-01
Fix grid clipping

### LINKED ISSUE #1247: Shallow clipping on flat models
**Describe the bug**
when viewing a "flat" mesh, eg. with small z values compared to x or y the clipping is quite shallow in the z direction and it cuts the grid

**To Reproduce**
1.  open a "flat" mesh, eg. with small z values compared to x or y 
2.  observe the clipping is quite shallow in the z direction and it cuts the grid

**Expected behavior**
the clipping is not quite shallow in the z direction and does not cut the grid

**F3D Information**
```Version: 2.3.0-RC5.
Build date: 2024-01-21 14:18:24.
Build system: Linux 64-bits.
Compiler: GNU 8.4.0.
External rendering module: ON.
Raytracing module: ON.
VTK version: 9.3.0-1702-geb008dfe7b (date: 20240111).
```

**Additional context**
- [screen cast of the problem](https://github.com/f3d-app/f3d/assets/489715/e963f5a1-0b1e-440a-8a13-8b04d8070e4b)
- [zipped `.glb` file to reproduce](https://github.com/f3d-app/f3d/files/14114069/flat.glb.zip)

snoyer:
The grid isn't clipped properly because it's not taken into account at all: https://github.com/f3d-app/f3d/blob/b7e8fc8edaaf146a8dc28ace999ef30a21d56258/library/VTKExtensions/Rendering/vtkF3DRenderer.cxx#L527

We cannot just swap for `GridActor->UseBoundsOn()` because that would mess up all the bounds computations, especially for camera resets.

We can hijack `ResetCameraClippingRange()` to temporarily turn on the bounds on the grid actor but is that the right thing to do?

would fix #1247, also makes the grid bbox flat as pointed out by @Meakk


library/VTKExtensions/Rendering/Testing/TestF3DOpenGLGridMapper.cxx
```diff
@@ -4,18 +4,66 @@
 #include "vtkF3DOpenGLGridMapper.h"
 
 #include <iostream>
+#include <sstream>
+
+std::string boundsToStr(const double* bounds)
```
Meakk, line 9: Functions should start with an upper case letter

Meakk: However, for this, what about using a class `BoundsTest` and with a constructor taking the name and the bounds, and implement `operator==` for comparison and `operator<<` for streaming?

Meakk (commented):
Usual nitpicking but the fix look good to me

library/VTKExtensions/Rendering/Testing/TestF3DOpenGLGridMapper.cxx
snoyer: `std::array` already has `==` right?
how about
```cpp
template<class T, std::size_t N>
ostream& operator<<(std::ostream& o, const std::array<T, N>& arr)
{
  std::copy(arr.cbegin(), arr.cend(), std::ostream_iterator<T>(o, " "));
  return o;
}

bool CheckBounds(const std::string& name, vtkF3DOpenGLGridMapper* mapper, //
  double x0, double x1, double y0, double y1, double z0, double z1)
{
  std::array<double, 6> actual;
  mapper->GetBounds(actual.data());
  const std::array<double, 6> expected = { x0, x1, y0, y1, z0, z1 };
  if (actual != expected)
  {
    std::cerr << "wrong bounds for " << name << " failed:" << std::endl;
    std::cerr << "  got " << actual << std::endl;
    std::cerr << "  expected " << expected << std::endl;
    return false;
  }
  return true;
}
```?

File: library/VTKExtensions/Rendering/Testing/TestF3DOpenGLGridMapper.cxx [hunk 2]
```diff
@@ -3,19 +3,76 @@
...
+    std::cerr << "  expected " << expected << std::endl;
+    return false;
+  }
+  return true;
+}
+
 int TestF3DOpenGLGridMapper(int argc, char* argv[])
 {
-  vtkNew<vtkF3DOpenGLGridMapper> mapper;
+  const double safeMargin = 1e-4;
```
Meakk, line 44:
```suggestion
  constexpr double safeMargin = 1e-4;
```

library/VTKExtensions/Rendering/Testing/TestF3DOpenGLGridMapper.cxx
```diff
@@ -3,19 +3,76 @@
...
+
+  {
+    vtkNew<vtkF3DOpenGLGridMapper> mapper;
```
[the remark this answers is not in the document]
snoyer: I'm trying to pretend we're _unit_ testing here ;p should even have one scope and mapper per check really

But yes we could reuse the same mapper. We could even remove the first test with the default up direction and just keep the 3 explicit ones if you want. Let me know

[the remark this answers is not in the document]
Meakk: that's ok :)

Meakk (approved):

mwestphal:
nice fix!

PR #1265, opened 2024-02-05
Fixed script extension and file path location

MarcosDanielTorres:
- On windows it should be .bat instead of .sh
- vcpkg was installed under `dev`


mwestphal (approved):
LGTM but I defer to  @Meakk

PR #1266, opened 2024-02-05
add javascript to insert download buttons

snoyer:
in order to have big "get F3D button" for the current OS above the release table:

Include a script in the header, the script is loaded on every page but will perform different actions depending on which page it's in. 
The idea is that the javascript only enhances the existing html, while being totally optional. We don't want to start adding <script> tags in the markdown documents.

In the case of the install page, the script retrieves the download links from the table in the page (html generated from the markdown file) and adds buttons for the ones where the "Platform" column matches the guessed client OS

edit: temporarily deployed at https://snoyer.github.io/f3d/doc/user/INSTALLATION.html
should look something like:
screenshot

### DIFF
--- a/_includes/head_custom.html
+++ b/_includes/head_custom.html
@@ -1,2 +1,2 @@
 <link rel="shortcut icon" href="https://raw.githubusercontent.com/f3d-app/f3d/master/resources/logo.ico" type="image/x-icon">
-<script type="application/javascript" src="{{ "/assets/main.js" | prepend:site.baseurl}}"></script>
+<script type="application/javascript" src="{{site.baseurl}}/assets/main.js"></script>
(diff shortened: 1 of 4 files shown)


mwestphal:
add a screenshot in description ? :)

_includes/head_custom.html
```diff
@@ -1 +1,2 @@
 <link rel="shortcut icon" href="https://raw.githubusercontent.com/f3d-app/f3d/master/resources/logo.ico" type="image/x-icon">
+<script type="application/javascript" src="{{site.baseurl}}/assets/main.js"></script>
```
mwestphal, line 2: missing eol

assets/main.js
```diff
@@ -0,0 +1,90 @@
...
+  } else if (/Android/.test(userAgent)) {
+    return "Android";
+  } else if (/Linux/.test(platform)) {
+    return "Linux";
+  } else {
+    return null;
+  }
+}
+
+window.addEventListener("load", onload);
```
mwestphal, line 90: can clang-format format this ?

snoyer: it can, but "we'll see later" as per @Meakk on Discord

mwestphal (approved):

Meakk (approved):

PR #1273, opened 2024-02-09
add metadata to `f3d::image`

snoyer:
Add support for metadata key,value pairs on `f3d::image` objects.

Saving and loading is currently only implemented for PNG format. It may be technically possible for other formats (eg. JPEG using Exif data) assuming VTK supports it.

Metadata would be used by future features to embed data in F3D's render, such as camera parameters, rendering options, ...

### DIFF
--- a/.github/actions/alembic-install-dep/action.yml
+++ b/.github/actions/alembic-install-dep/action.yml
@@ -15,7 +15,7 @@ runs:
       uses: actions/cache@v4
       with:
         path: dependencies/alembic_install
-        key: alembic-v1.8.5-${{runner.os}}-${{inputs.cpu}}-2
+        key: alembic-v1.8.5-${{runner.os}}-${{inputs.cpu}}-3
 
     - name: Checkout Alembic
       if: steps.cache-alembic.outputs.cache-hit != 'true'
--- a/.github/actions/assimp-install-dep/action.yml
+++ b/.github/actions/assimp-install-dep/action.yml
@@ -15,7 +15,7 @@ runs:
       uses: actions/cache@v4
       with:
         path: dependencies/assimp_install
-        key: assimp-v5.3.1-${{runner.os}}-${{inputs.cpu}}-2
+        key: assimp-v5.3.1-${{runner.os}}-${{inputs.cpu}}-3
 
     - name: Checkout ASSIMP
       if: steps.cache-assimp.outputs.cache-hit != 'true'
--- a/.github/actions/blosc-install-dep/action.yml
+++ b/.github/actions/blosc-install-dep/action.yml
@@ -15,7 +15,7 @@ runs:
       uses: actions/cache@v4
       with:
         path: dependencies/blosc_install
-        key: blosc-1.21.5-${{runner.os}}-${{inputs.cpu}}-2
+        key: blosc-1.21.5-${{runner.os}}-${{inputs.cpu}}-3
 
     # Dependents: openvdb vtk
     - name: Checkout blosc
--- a/.github/actions/boost-install-dep/action.yml
+++ b/.github/actions/boost-install-dep/action.yml
@@ -15,7 +15,7 @@ runs:
       uses: actions/cache@v4
       with:
         path: dependencies/boost_install
-        key: boost-1.83.0-${{runner.os}}-${{inputs.cpu}}-3
+        key: boost-1.83.0-${{runner.os}}-${{inputs.cpu}}-4
 
     # Dependents: usd openvdb vtk
     - name: Checkout Boost
--- a/.github/actions/draco-install-dep/action.yml
+++ b/.github/actions/draco-install-dep/action.yml
@@ -15,7 +15,7 @@ runs:
       uses: actions/cache@v4
       with:
         path: dependencies/draco_install
-        key: draco-1.5.7-${{runner.os}}-${{inputs.cpu}}-0
+        key: draco-1.5.7-${{runner.os}}-${{inputs.cpu}}-1
 
     - name: Checkout Draco
       if: steps.cache-draco.outputs.cache-hit != 'true'
--- a/.github/actions/generic-ci/action.yml
+++ b/.github/actions/generic-ci/action.yml
@@ -124,11 +124,6 @@ runs:
       working-directory: ${{github.workspace}}
       run: echo "PATH=$env:PATH;$(pwd)\dependencies\install\bin\;$(pwd)\dependencies\install\lib\;$(pwd)\install\bin\;"| Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
 
-    - name: Determine if testing should be enabled
-      id: vars
-      shell: bash
-      run: echo "run_tests=${{ inputs.cpu == 'x86_64' && 'true' || 'false' }}" >> $GITHUB_OUTPUT
-
       # occt coloring not supported on macOS because of #792
       # Exodus not supported on Apple Silicon because of #976
       # CMAKE_MODULE_PATH is required because of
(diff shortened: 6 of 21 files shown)


snoyer:
```
library/src/image.cxx:56:23: style: Variable 'key' is not assigned a value. [unassignedVariable]
    for (const auto& [key, value] : this->Metadata)
                      ^
```
supposedly a fixed bug https://trac.cppcheck.net/ticket/10368

mwestphal:
this approach looks fine to me. Wdyt @Meakk ?

mwestphal:
There was an issue that required rewriting history. Please cherry pick your commits on f3d-app/f3d master.

snoyer:
re-opening for no reason, nothing happened, there was no git accident, at all, ever

library/src/image.cxx
```diff
@@ -343,11 +391,15 @@ std::vector<unsigned char> image::saveBuffer(SaveFormat format) const
...
     case SaveFormat::PNG:
-      return this->Internals->SaveBuffer<vtkPNGWriter>();
+    {
+      vtkSmartPointer<vtkPNGWriter> writer = vtkSmartPointer<vtkPNGWriter>::New();
+      this->Internals->writePngMetadata(writer);
+      return this->Internals->SaveBuffer<vtkPNGWriter>(writer);
+    }
     case SaveFormat::JPG:
-      return this->Internals->SaveBuffer<vtkJPEGWriter>();
+      return this->Internals->SaveBuffer<vtkJPEGWriter>(vtkSmartPointer<vtkJPEGWriter>::New());
```
Meakk, line 400: I think the compiler can deduce the type now

```suggestion
      return this->Internals->SaveBuffer(vtkSmartPointer<vtkJPEGWriter>::New());
```

library/src/image.cxx
```diff
@@ -496,6 +548,42 @@ std::string image::toTerminalText() const
...
+  else
+  {
+    throw std::out_of_range(key);
+  }
+}
+
+//----------------------------------------------------------------------------
+std::set<std::string> image::allMetadata() const
+{
+  std::set<std::string> keys;
```
Meakk, line 581: Any reason to use a set instead of a vector here?

python/F3DPythonBindings.cxx
```diff
@@ -133,7 +133,21 @@ PYBIND11_MODULE(pyf3d, module)
     .def("save_buffer", getFileBytes, py::arg("format") = f3d::image::SaveFormat::PNG)
     .def("_repr_png_",
       [&](const f3d::image& img) { return getFileBytes(img, f3d::image::SaveFormat::PNG); })
-    .def("to_terminal_text", [](const f3d::image& img) { return img.toTerminalText(); });
+    .def("to_terminal_text", [](const f3d::image& img) { return img.toTerminalText(); })
+    .def("set_metadata", &f3d::image::setMetadata)
```
Meakk, line 137: Can we use def_property here?

library/src/image.cxx
snoyer: I think I tried but it complained, I'll double check

library/src/image.cxx
snoyer: It makes me feel smart to go "ahah the keys are unique and unordered, they're a _set_ not a _list_" but apart from that not much.
I guess if there were _lots_ of keys a set implementation could be faster but we won't ever have many keys and a vector may even be more performant at the lower count. But it's small potatoes either way.

So up to you, can be a vector if you prefer

python/F3DPythonBindings.cxx
snoyer: I don't think so because we're not exposing `metadata` as a dict/map, `get_metadata` and `set_metadata` are still function calls with at least a `key` parameter

library/src/image.cxx
snoyer: it can infer if we change the function prototype to use a smart pointer
```diff
-  std::vector<unsigned char> SaveBuffer(WriterType* writer)
+  std::vector<unsigned char> SaveBuffer(vtkSmartPointer<WriterType> writer)
```
would that be good?

library/src/image.cxx
Meakk: `std::set` is ordered btw, you may want to use `std::unordered_set` if you really want it to be unsorted.  
However, usually `set` is used when pushing elements in it and we want to make sure it's always sorted and unique. However, insertion complexity is `log(n)`, and while the search is, in theory, faster, `vector` outperforms it with a small number of elements because it is contiguous in memory.  
Sorry to insist, but yeah I'd prefer `std::vector` here.

python/F3DPythonBindings.cxx
Meakk: Understood now :)

library/src/image.cxx
```diff
@@ -45,6 +50,41 @@ class image::internals
...
+      }
+    }
+  }
+
+  void readPngMetadata(vtkPNGReader* pngReader)
+  {
+    int beginEndIndex[2];
+    for (size_t i = 0; i < pngReader->GetNumberOfTextChunks(); ++i)
+    {
+      const vtkStdString key = pngReader->GetTextKey((int)i);
```
Meakk, line 72:
```suggestion
      const vtkStdString key = pngReader->GetTextKey(static_cast<int>(i));
```

library/src/image.cxx
```diff
@@ -45,6 +50,41 @@ class image::internals
 
     return result;
   }
+
+  void writePngMetadata(vtkPNGWriter* pngWriter)
```
Meakk, line 54: Internal functions are uppercase :P

```suggestion
  void WritePngMetadata(vtkPNGWriter* pngWriter)
```

library/src/image.cxx
```diff
@@ -314,8 +360,12 @@ void image::save(const std::string& path, SaveFormat format) const
   switch (format)
   {
     case SaveFormat::PNG:
-      writer = vtkSmartPointer<vtkPNGWriter>::New();
-      break;
+    {
+      vtkSmartPointer<vtkPNGWriter> pngWriter = vtkSmartPointer<vtkPNGWriter>::New();
```
Meakk, line 364: Is it working?

```suggestion
      vtkNew<vtkPNGWriter> pngWriter;
```

library/src/image.cxx
```diff
@@ -496,6 +550,42 @@ std::string image::toTerminalText() const
...
+std::string image::getMetadata(const std::string& key) const
+{
+  if (this->Internals->Metadata.count(key))
+  {
+    return this->Internals->Metadata[key];
+  }
+  else
+  {
+    throw std::out_of_range(key);
+  }
```
Meakk, line 577: else not necessary

```suggestion
  throw std::out_of_range(key);
```

library/testing/TestSDKImage.cxx
```diff
@@ -5,10 +5,14 @@
 #include <functional>
 #include <iostream>
 #include <random>
+#include <set>
 #include <sstream>
 
 int TestSDKImage(int argc, char* argv[])
 {
+  const std::string TESTING(argv[1]);
+  const std::string TMP(argv[2]);
```
Meakk, line 14: Let's stick with lowercase style
```suggestion
  const std::string testingDir(argv[1]);
  const std::string tmpDir(argv[2]);
```

library/testing/TestSDKImage.cxx
```diff
@@ -258,20 +262,102 @@ int TestSDKImage(int argc, char* argv[])
...
+    img.setMetadata("foo", "bar");
+    img.setMetadata("hello", "world");
+    if (img.getMetadata("foo") != "bar" || img.getMetadata("hello") != "world")
+    {
+      std::cerr << "setMetadata() or getMetadata() failed" << std::endl;
+      return EXIT_FAILURE;
+    }
+
+    const std::vector<std::string> keys = img.allMetadata();
+    if (std::set<std::string>(keys.begin(), keys.end()) !=
```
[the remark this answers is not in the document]
snoyer: Order of the keys is not guaranteed so `.allMetadata()` may not always `==` to the hardcoded test vector?
Unless I'm wrong (again!) about the workings of various `std` containers, which is definitely a possibility :/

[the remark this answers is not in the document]
Meakk: you are right :)

Meakk (approved):

PR #1275, opened 2024-02-09
Updating codecov action

mwestphal:


.github/actions/coverage-ci/action.yml
```diff
@@ -105,7 +105,7 @@ runs:
         lcov --remove coverage.info "*Test*" -o coverage.info
 
     - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@v3
+      uses: codecov/codecov-action@v4
       with:
         files: ${{github.workspace}}/source/coverage.info
-        version: "v0.1.15" # works better according to https://github.com/codecov/codecov-action/issues/598
+        token: ${{ secrets.CODECOV_TOKEN }}
```
snoyer, line 111: > To make a secret available to an action, you must set the secret as an input or environment variable in the workflow file.

https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#accessing-your-secrets

mwestphal: good point!

PR #1279, opened 2024-02-10
fix: print error when test record file doesn't exist

### LINKED ISSUE #1231: F3D does not error out when specifying non-existent path on --interaction-test-record
**Describe the bug**
Opening f3d with --interaction-test-record and giving it a path which is a non-existing directory bugs out the program. Escape does not work and the killing the process with Ctrl + C does not generate the log file. Other toggles seems to work correctly

**To Reproduce**
 1. Open the file using `f3d example.glb --interactive-test-record path/does/not/exist/abc.log`
 2. Try closing the program using Escape

**Expected behavior**
Print an error stating that the path does not exist and exit

**System Information:**
 - OS: Gentoo Linux
 - GPU and GPU driver: Intel Corporation Alder Lake-P Integrated Graphics Controller, i915

**F3D Information**
F3D 2.2.1

F3D - A fast and minimalist 3D viewer
Version: 2.2.1-55-g77596024.
Build date: 2024-01-23 03:42:40.
Build system: Linux 64-bits.
Compiler: GNU 13.2.1.
External rendering module: OFF.
Raytracing module: OFF.
VTK version: 9.3.0.
Copyright (C) 2019-2021 Kitware SAS.
Copyright (C) 2021-2023 Michael Migliore, Mathieu Westphal.
License BSD-3-Clause.
By Michael Migliore, Mathieu Westphal and Joachim Pouderoux.

**Additional context**
The attached video shows this behaviour when specifying an non-existent path

https://github.com/f3d-app/f3d/assets/14817339/00466046-2219-4015-be61-ccdca2fb359d

chirag-droid:
## Fixes

#1231 

## Changes

F3D now errors out when the record file doesn't exist.

### DIFF
--- a/library/src/interactor_impl.cxx
+++ b/library/src/interactor_impl.cxx
@@ -683,7 +683,6 @@ bool interactor_impl::recordInteraction(const std::string& file)
 {
   if (file.empty())
   {
-    // TODO improve VTK to check file opening
     log::error("No interaction record file provided");
     return false;
   }
--- a/library/src/interactor_impl.cxx
+++ b/library/src/interactor_impl.cxx
@@ -717,7 +717,6 @@ bool interactor_impl::recordInteraction(const std::string& file)
   this->Internals->Recorder->SetInteractor(this->Internals->VTKInteractor);
 #endif
 
-  std::string cleanFile = vtksys::SystemTools::CollapseFullPath(file);
   this->Internals->Recorder->SetFileName(cleanFile.c_str());
   this->Internals->Recorder->On();
   this->Internals->Recorder->Record();
--- a/library/src/interactor_impl.cxx
+++ b/library/src/interactor_impl.cxx
@@ -695,14 +695,14 @@ bool interactor_impl::recordInteraction(const std::string& file)
   // Check if the parent directory exists
   if (!vtksys::SystemTools::FileExists(parentDirectory))
   {
-    log::error("Interaction log file directory does not exist ", parentDirectory);
+    log::error("Interaction record directory does not exist ", parentDirectory);
     return false;
   }
 
   // Check if we can write to the directory
   if (!vtksys::SystemTools::TestFileAccess(parentDirectory, vtksys::TEST_FILE_WRITE))
   {
-    log::error("Don't have permission to write to ", parentDirectory);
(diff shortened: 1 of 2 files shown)


Meakk:
The CI jobs are failing though. It needs to be fixed first.

library/src/interactor_impl.cxx
```diff
@@ -688,6 +688,12 @@ bool interactor_impl::recordInteraction(const std::string& file)
     return false;
   }
 
+  if (!vtksys::SystemTools::FileExists(file))
+  {
+    log::error("Interaction log file to record interactions to does not exist", file);
```
0x006E, line 693: Rather than checking if the file exists,I think we should check if the path is writable or not, because when new interaction tests are created, we specifiy filenames that may not exist yet.

chirag-droid: I am working on this

chirag-droid:
Should I create a temp file in parent directory and check if it succeeds? or is there a better way

chirag-droid:
wait what happened..

mwestphal:
for the inexistent path, just pick pick an `invalid` dir in the testing dir.

For the path without permissions, you can create a file without write permission in the testing dir and try writing to it.

chirag-droid:
@mwestphal The tests are failing in Windows and Linux GitHub runners. For some reason changing directory permission is not working as expected. On Mac however the permissions are set correctly.

mwestphal:
I tried locally on my linux and this is working fine, strange.

mwestphal:
maybe try putting no permissions at all ?

chirag-droid:
> I tried locally on my linux and this is working fine, strange.

The issue is only happening inside github runners. specifically windows and linux.

In both of these platform cmake fails to set appropriate permission. The test expects a non writable directory.

I have tried running raw `chmod` command and it fails too.

If i remove this test then coverage test will fail

mwestphal:
Let's remove this test. Coverage reduction is ok when justified

chirag-droid:
https://github.com/f3d-app/f3d/blob/b40fba3e10401e34816926818a29e47571926777/library/src/interactor_impl.cxx#L684-L689

This todo can be removed now?

mwestphal (approved):

PR #1288, opened 2024-02-14
add dark title bar support on windows

### LINKED ISSUE #895: Add dark title bar support on Windows
Dark Title bar Implementation.

https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/apply-windows-themes#enable-a-dark-mode-title-bar-for-win32-applications

from @Da3n0n, must be done in VTK Windows code.

chirag-droid:
Fix #895 

## Changes

Added some helper functions to fetch User theme from registry and also to query build numbers to check if a feature is supported.

Sets Immersive Dark Mode based on user theme.

### DIFF
--- a/library/src/window_impl.cxx
+++ b/library/src/window_impl.cxx
@@ -63,7 +63,9 @@ public:
   bool IsWindowsBuildNumberOrGreater(int buildNumber)
   {
     std::string value{};
-    bool result = vtksys::SystemTools::ReadRegistryValue("HKEY_LOCAL_MACHINE\\Software\\Microsoft\\Windows NT\\CurrentVersion;CurrentBuildNumber", value);
+    bool result = vtksys::SystemTools::ReadRegistryValue(
+      "HKEY_LOCAL_MACHINE\\Software\\Microsoft\\Windows NT\\CurrentVersion;CurrentBuildNumber",
+      value);
 
     if (result == true)
     {
@@ -86,16 +88,18 @@ public:
 
   /**
    * Helper function to fetch a DWORD from windows registry.
-   * 
+   *
    * @param hKey A handle to an open registry key
    * @param subKey The path of registry key relative to 'hKey'
    * @param value The name of the registry value
    * @param dWord Variable to store the result in
    */
-  bool ReadRegistryDWord(HKEY hKey, const std::wstring& subKey, const std::wstring& value, DWORD& dWord)
+  bool ReadRegistryDWord(
+    HKEY hKey, const std::wstring& subKey, const std::wstring& value, DWORD& dWord)
   {
     DWORD dataSize = sizeof(DWORD);
-    LONG result = RegGetValueW(hKey, subKey.c_str(), value.c_str(), RRF_RT_REG_DWORD, nullptr, &dWord, &dataSize);
+    LONG result = RegGetValueW(
+      hKey, subKey.c_str(), value.c_str(), RRF_RT_REG_DWORD, nullptr, &dWord, &dataSize);
 
     return result == ERROR_SUCCESS;
   }
@@ -118,7 +122,7 @@ public:
     }
 
     result = ReadRegistryDWord(HKEY_CURRENT_USER, subKey, L"SystemUsesLightTheme", value);
-    
+
     if (result && value == 0)
     {
       return true;
@@ -301,11 +305,10 @@ void window_impl::Initialize(bool withColoring)
   this->Internals->Initialized = true;
 
 #ifdef _WIN32
-  HWND hwnd = static_cast<HWND>(this->Internals->RenWin->GetGenericWindowId());
-  BOOL useDarkMode = this->Internals->IsWindowsInDarkMode();
-
   if (this->Internals->IsWindowsBuildNumberOrGreater(IMMERSIVE_DARK_MODE_SUPPORTED_SINCE))
   {
+    HWND hwnd = static_cast<HWND>(this->Internals->RenWin->GetGenericWindowId());
+    BOOL useDarkMode = this->Internals->IsWindowsInDarkMode();
     DwmSetWindowAttribute(hwnd, DWMWA_USE_IMMERSIVE_DARK_MODE, &useDarkMode, sizeof(useDarkMode));
   }
 #endif
--- a/library/src/window_impl.cxx
+++ b/library/src/window_impl.cxx
@@ -22,8 +22,6 @@
 #include <vtkWindowToImageFilter.h>
 #include <vtksys/SystemTools.hxx>
 
-#include <charconv>
-
 #if F3D_MODULE_EXTERNAL_RENDERING
 #include <vtkExternalOpenGLRenderWindow.h>
 #endif
--- a/library/src/window_impl.cxx
+++ b/library/src/window_impl.cxx
@@ -22,8 +22,19 @@
 #include <vtkWindowToImageFilter.h>
 #include <vtksys/SystemTools.hxx>
 
+#include <charconv>
+
 #if F3D_MODULE_EXTERNAL_RENDERING
 #include <vtkExternalOpenGLRenderWindow.h>
+#endif
+
+#ifdef _WIN32
+#include <Windows.h>
+#include <dwmapi.h>
+
+#define DWMWA_USE_IMMERSIVE_DARK_MODE 20
+
+constexpr auto MAX_VALUE_NAME = 16383;
 #endif
 
 namespace f3d::detail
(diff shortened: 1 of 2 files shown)


library/src/window_impl.cxx
```diff
@@ -44,6 +55,97 @@ class window_impl::internals
     return this->CachePath;
   }
 
+#if _WIN32
+  /**
+   * Helper function to detect if the
+   * Windows Build Number is equal or greater to a number
+   */
+  BOOL IsWindowsBuildNumberOrGreater(int buildNumber)
```
snoyer, line 63:
```suggestion
  bool IsWindowsBuildNumberOrGreater(int buildNumber)
```

library/src/window_impl.cxx
```diff
@@ -44,6 +55,97 @@ class window_impl::internals
...
+      "SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", 0, KEY_READ, &versionHKey);
+
+    // Query value from the reg key
+    if (result == ERROR_SUCCESS)
+    {
+      // Receives code indicating the type of data stored in the reg value.
+      // CurrentBuildNumber is of type REG_SZ
+      DWORD type;
+      DWORD dataSize = MAX_VALUE_NAME;
+      CHAR data[MAX_VALUE_NAME]{};
```
snoyer, line 79: is that a 16kB buffer to hold a text representation of an `int`?

library/src/window_impl.cxx
```diff
@@ -44,6 +55,97 @@ class window_impl::internals
...
+
+      RegGetValueA(versionHKey, "", "CurrentBuildNumber", RRF_RT_REG_SZ, &type, data, &dataSize);
+
+      if (result == ERROR_SUCCESS && type == REG_SZ)
+      {
+        std::from_chars(data, data + dataSize, windowsBuildNumber);
```
snoyer, line 85: need to check error code or are we sure it will parse to an `int`?

library/src/window_impl.cxx
```diff
@@ -44,6 +55,97 @@ class window_impl::internals
...
+    }
+
+    RegCloseKey(versionHKey);
+    return windowsBuildNumber >= buildNumber;
+  }
+
+  /**
+   * Helper function to detect user theme
+   */
+  BOOL IsWindowsInDarkMode()
```
snoyer, line 104:
```suggestion
  bool IsWindowsInDarkMode()
```

library/src/window_impl.cxx
```diff
@@ -215,6 +317,17 @@ void window_impl::Initialize(bool withColoring)
   this->Internals->WithColoring = withColoring;
   this->Internals->Renderer->Initialize(this->Internals->Options.getAsString("scene.up-direction"));
   this->Internals->Initialized = true;
+
+#ifdef _WIN32
+  HWND hwnd = static_cast<HWND>(this->Internals->RenWin->GetGenericWindowId());
+
+  BOOL useDarkMode = this->Internals->IsWindowsInDarkMode();
+
+  if (this->Internals->IsWindowsBuildNumberOrGreater(22000))
```
snoyer, line 326: magic `22000` should be a constant or have a comment static what it means

library/src/window_impl.cxx
```diff
@@ -215,6 +317,17 @@ void window_impl::Initialize(bool withColoring)
...
+  {
+    DwmSetWindowAttribute(hwnd, DWMWA_USE_IMMERSIVE_DARK_MODE, &useDarkMode, sizeof(useDarkMode));
+  }
```
snoyer, line 329:
```suggestion
  if (this->Internals->IsWindowsBuildNumberOrGreater(22000))
  {
    const HWND hwnd = static_cast<HWND>(this->Internals->RenWin->GetGenericWindowId());
    const bool useDarkMode = this->Internals->IsWindowsInDarkMode();
    DwmSetWindowAttribute(hwnd, DWMWA_USE_IMMERSIVE_DARK_MODE, &useDarkMode, sizeof(useDarkMode));
  }
```
these variables aren't needed outside the `if` block

snoyer (commented):
I don't Windows so I may be wrong on all counts here :) better wait for @mwestphal and @Meakk to confirm

chirag-droid:
Also windows documentation is very lacking for win32.

The [official docs](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/ne-dwmapi-dwmwindowattribute) mention immersive dark mode is only available after win 11 build 22000.

But MicaForEveryone app don't follow these.

I didn't find any docs on how to get Windows build number. So I went with checking registry value. which is used by `winver` too.

Same with theme.

win32 is super old

library/src/window_impl.cxx
chirag-droid: So, I looked up registry value and `BuildNumber` is a string type. `REG_SZ`. So I am creating a buffer for it.

This constant is taken from a windows example. 

Found an article on windows registry using modern c++. https://learn.microsoft.com/en-us/archive/msdn-magazine/2017/may/c-use-modern-c-to-access-the-windows-registry

library/src/window_impl.cxx
chirag-droid: I think there need to check error code

library/src/window_impl.cxx
chirag-droid: `BOOL` is a windows specific type. I am not sure if its ok to change it to bool.

library/src/window_impl.cxx
```diff
@@ -44,6 +55,97 @@ class window_impl::internals
...
+    // Query value from the reg key
+    if (result == ERROR_SUCCESS)
+    {
+      // Receives code indicating the type of data stored in the reg value.
+      // CurrentBuildNumber is of type REG_SZ
+      DWORD type;
+      DWORD dataSize = MAX_VALUE_NAME;
+      CHAR data[MAX_VALUE_NAME]{};
+
+      RegGetValueA(versionHKey, "", "CurrentBuildNumber", RRF_RT_REG_SZ, &type, data, &dataSize);
```
Meakk, line 81: Can we use the vtksys wrapper to simplify the code a bit? See https://github.com/Kitware/VTK/blob/697646db54ce4de95ff5adb500799e49feee4935/Utilities/KWSys/vtksys/SystemTools.hxx.in#L844

chirag-droid: sure

library/src/window_impl.cxx
chirag-droid: I am now using vtksys wrapper to handle string values

library/src/window_impl.cxx
```diff
@@ -26,6 +26,17 @@
...
 
+#ifdef _WIN32
+#include <Windows.h>
+#include <dwmapi.h>
+
+#ifndef DWMWA_USE_IMMERSIVE_DARK_MODE
+#define DWMWA_USE_IMMERSIVE_DARK_MODE 20
+#endif
+
+constexpr auto IMMERSIVE_DARK_MODE_SUPPORTED_SINCE = 22000;
```
chirag-droid, line 37: Should this be changed to 19041 just like in MicaForEveryone app? windows documentation says 22000 which is win11.

library/src/window_impl.cxx
```diff
@@ -44,6 +55,83 @@ class window_impl::internals
...
+      DWORD dataSize = sizeof(DWORD);
+
+      RegQueryValueExA(hKey, "SystemUsesLightTheme", NULL, &type, (LPBYTE)&value, &dataSize);
+
+      if (result == ERROR_SUCCESS)
+      {
+        RegCloseKey(hKey);
+        return value == 0;
+      }
+    }
```
chirag-droid, line 128: Create a helper function to remove repetitive code

library/src/window_impl.cxx
Meakk: 19041 is windows 10?

library/src/window_impl.cxx
```diff
@@ -44,6 +55,78 @@ class window_impl::internals
     return this->CachePath;
   }
 
+#if _WIN32
+  /**
+   * Helper function to detect if the
+   * Windows Build Number is equal or greater to a number
+   */
+  bool IsWindowsBuildNumberOrGreater(int buildNumber)
```
Meakk, line 63: These functions should be `static`

library/src/window_impl.cxx
```diff
@@ -44,6 +55,78 @@ class window_impl::internals
...
+
+    if (result == true)
+    {
+      try
+      {
+        return std::stoi(value) >= buildNumber;
+      }
+      catch (const std::invalid_argument& e)
+      {
+        f3d::log::debug("Error parsing CurrentBuildNumber", e.what());
```
Meakk, line 78: Should be an error message, not a debug

library/src/window_impl.cxx
chirag-droid: Yes

library/src/window_impl.cxx
```diff
@@ -44,6 +55,78 @@ class window_impl::internals
...
+   * @param hKey A handle to an open registry key
+   * @param subKey The path of registry key relative to 'hKey'
+   * @param value The name of the registry value
+   * @param dWord Variable to store the result in
+   */
+  bool ReadRegistryDWord(
+    HKEY hKey, const std::wstring& subKey, const std::wstring& value, DWORD& dWord)
+  {
+    DWORD dataSize = sizeof(DWORD);
+    LONG result = RegGetValueW(
```
Meakk, line 101: Can't you use the vtksys wrapper here?

library/src/window_impl.cxx
chirag-droid: I am changing it to 19041 so whenever @mwestphal gets time to verify this can be directly merged

library/src/window_impl.cxx
```diff
@@ -215,6 +298,15 @@ void window_impl::Initialize(bool withColoring)
   this->Internals->WithColoring = withColoring;
   this->Internals->Renderer->Initialize(this->Internals->Options.getAsString("scene.up-direction"));
   this->Internals->Initialized = true;
+
+#ifdef _WIN32
+  if (this->Internals->IsWindowsBuildNumberOrGreater(IMMERSIVE_DARK_MODE_SUPPORTED_SINCE))
```
Meakk, line 303: Maybe move this block to its own internal function

Meakk (commented):
Here are a few comments

library/src/window_impl.cxx
chirag-droid: I tried it only works if the registry value is a string type not for DWORD

Meakk: Ok

library/src/window_impl.cxx
chirag-droid: On older versions of windows this will likely fail and is just ignored. So, I moved it to debug. Let me know what you think.

Meakk: Ok then I'm ok with the debug level.

Meakk (approved):

Issue #441, opened 2022-09-28
F3d


green-forg:

green-forg:
<img width="790" alt="image" src="https://user-images.githubusercontent.com/65800174/192673226-79be5307-381b-4295-b3ea-1285ae219970.png">

mwestphal:
Which version of F3D are you using? You may want to disable the thumbnailer.

mwestphal:
probably a duplicate of #300

green-forg:
Thanks a lot for your answer! I just updated it to the newest version to see if it will help. The thumbnailer is quite useful to me, however, I didn't open any file manager in the background, so it is probably not the reason for that?

mwestphal:
Without opening a file manager, I do not see why f3d would run. Please try to reproduce the issue after rebooting your computer.

green-forg:
@mwestphal It has been working just fine after the upgrade, I'll ask for help if the problem happens again. Appreciate it a lot!

Issue #300, opened 2022-05-28
Thumbnails generation on windows uses too much CPU


astandarduser:
Have just started using this software and it saves a whole lot of time finding things as an artist, just curious if there is a way to limit the CPU usage to around 70 percent when generating thumbnails?

mwestphal:
Hi, are you using Windows or Linux ?

astandarduser:
I am using Windows, there are also 4 instances of it running, although I installed it about 15 minutes ago

astandarduser:
So even without opening a new folder where it 100 percent CPU usage, it's still using a significant amount of the CPU 13% * 3 instances so 39% when doing nothing. Seems very power hungry when it shouldn't be doing much

astandarduser:
Update: after a restart it seems to be acting more like it should where the CPU use goes way down after a folder has thumbnails generated. But as mentioned in the topic it would still be good to put a limit on cpu usage when generating the thumbnails for the first time

mwestphal:
Hum, did you open multiple file manager or only one ? Can you share a archive containing the file you were generating thumbnails from ?

I will try to reproduce.

Thanks,

mwestphal:
Sounds more like a bug to me.

astandarduser:
I think you could be right that I opened multiple explorer windows, now it generates them around 20 percent. I guess you can wipe this thread. Looking forward to further developments. Not sure if it's possible but would be amazing for maya ascii support , ztl, and 3dxml

mwestphal:
> I think you could be right that I opened multiple explorer windows, now it generates them around 20 percent. I guess you can wipe this thread.

I still think there is an issue here, I remember seeing some slowdown sometimes, so identifying it more cleanly is nice, and at least maybe we can do something to avoid using too much CPU and slowing down windows.

> Not sure if it's possible but would be amazing for maya ascii support , ztl, and 3dxml

There is a dedicated issue for new file formats, coul you comment on it ?
https://github.com/f3d-app/f3d/issues/38

Thanks !

astandarduser:
Not sure how to vote on that but appreciate the link. And there are definitely bugs that need to be ironed out as it's currently using 20 percent CPU with no explorer windows open

mwestphal:
> Not sure how to vote on that but appreciate the link

You can just comment on the issue and I will add them in my top comment.

astandarduser:
[...]

PR #1291, opened 2024-02-15
fix: terminate f3d process spawned by F3DShellExtension if its still active after timeout

chirag-droid:
**Previous Behaviour**

If you previewed a thumbnail that took longer than 8 sec, the thumbnail would return in `E_FAIL` but wouldn't properly terminate the process associated with it. And continue running in the background.

**Current Behaviour**

The spawned process is properly terminated.

### DIFF
--- a/winshellext/F3DThumbnailProvider.cxx
+++ b/winshellext/F3DThumbnailProvider.cxx
@@ -188,6 +188,13 @@ IFACEMETHODIMP F3DThumbnailProvider::GetThumbnail(UINT cx, HBITMAP* phbmp, WTS_A
   // Clean up
   DWORD exitCode = EXIT_SUCCESS;
   GetExitCodeProcess(pi.hProcess, &exitCode);
+  
+  // Terminate the process if it is still active
+  if (exitCode == STILL_ACTIVE)
+  {
+    TerminateProcess(pi.hProcess, EXIT_FAILURE);
+    GetExitCodeProcess(pi.hProcess, &exitCode);
+  }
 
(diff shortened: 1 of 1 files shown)


chirag-droid:
related to #300 #441

Meakk (approved):
Looks good to me, but I'm not a Win32 expert. @mwestphal wdyt?

chirag-droid:
I am sure this only solves a part of the puzzle.

It needs a proper investigation. And its difficult without someone from windows dev team.

what happens when the user switches the file to preview?

And the function should return a default icon instead of returning error so windows saves it in cache.

mwestphal:
> f you previewed a thumbnail that took longer than 8 sec, 

Is this a Windows limitation ? in any case, the fix is good, there is no reason to leave the process running if Windows is not going ot do anything with it.

Dataset that are too slow to be opened but smaller than the max-size we should study to understand why they are slow, which is why @snoyer is doing in #1292 :)

mwestphal (approved):

chirag-droid:
> > f you previewed a thumbnail that took longer than 8 sec, 
> 
> Is this a Windows limitation ? in any case, the fix is good, there is no reason to leave the process running if Windows is not going ot do anything with it.

Its actually a timeout set in the code. The timeout can be changed by a compile definition.


In CMakeLists.txt it is set to 8000ms

chirag-droid:
https://github.com/f3d-app/f3d/blob/master/winshellext%2FCMakeLists.txt#L9

PR #1292, opened 2024-02-16
OCCT speedups and fixes

snoyer:
Performance improvements for the OCCT reader:
- color lookups turn out to add up a lot when they're done for every triangle of a re-meshed face.
- re-meshing can be batched instead of done for each face/edge individually.
- color lookups with `XCAFDoc_ColorTool` were also wrong in addition to being slow, using `XCAFPrs` to walk down the attribute tree and collect inherited style properties is more correct and faster

all in all the 18MB step file from https://www.printables.com/model/612925-creality-k1-corexy-motion-assembly-cad/files now opens in about ~~40sec~~ 15sec instead of almost 5min

before and after:

f3d-k1-before

f3d-k1-after

also fixes a minor bug where progress was reported up to 150%, meaning the progress bar would look full before the actual processing was finished.

### DIFF
--- a/plugins/occt/module/vtkF3DOCCTReader.cxx
+++ b/plugins/occt/module/vtkF3DOCCTReader.cxx
@@ -328,38 +328,36 @@ public:
     StyleMap collectedStyles;
     XCAFPrs::CollectStyleSettings(rootLabel, TopLoc_Location(), collectedStyles);
 
-    /* iterate styled shapes and assign style if leaf, otherwise collect if parent node */
+    /* iterate styled shapes and collect sorted by ascending shape type depth */
+    const auto cmp = [](const TopoDS_Shape& a, const TopoDS_Shape& b)
+    { return a.ShapeType() > b.ShapeType(); };
+    std::multimap<TopoDS_Shape, XCAFPrs_Style, decltype(cmp)> styledShapes(cmp);
+
     const TopAbs_ShapeEnum leafType = this->Parent->GetReadWire() ? TopAbs_EDGE : TopAbs_FACE;
-    std::vector<std::pair<TopoDS_Shape, XCAFPrs_Style> > parents;
     for (StyleMap::Iterator iter(collectedStyles); iter.More(); iter.Next())
     {
       const TopoDS_Shape& shape = iter.Key();
       if (shape.ShapeType() <= leafType)
       {
-        parents.emplace_back(std::make_pair(shape, iter.Value()));
+        styledShapes.insert({ shape, iter.Value() });
       }
     }
-
-    /* sort parents from deepest upwards */
-    std::sort(parents.begin(), parents.end(),
-      [](const std::pair<TopoDS_Shape, XCAFPrs_Style>& a,
-        const std::pair<TopoDS_Shape, XCAFPrs_Style>& b)
-      { return a.first.ShapeType() > b.first.ShapeType(); });
 
     /* pass down each parent style props to descendent edge/face leaves */
     const auto passDownToLeaves = [&](TopAbs_ShapeEnum type)
     {
-      for (const auto& parent : parents)
+      for (const auto& styledShape : styledShapes)
       {
-        for (TopExp_Explorer iter(parent.first, type); iter.More(); iter.Next())
+        for (TopExp_Explorer iter(styledShape.first, type); iter.More(); iter.Next())
         {
           try
           {
-            this->PassDownStyleProps(parent.second, inheritedStyles.ChangeFromKey(iter.Current()));
+            this->PassDownStyleProps(
+              styledShape.second, inheritedStyles.ChangeFromKey(iter.Current()));
           }
           catch (Standard_NoSuchObject&)
           {
-            inheritedStyles.Add(iter.Current(), parent.second);
+            inheritedStyles.Add(iter.Current(), styledShape.second);
           }
         }
       }
--- a/plugins/occt/module/vtkF3DOCCTReader.cxx
+++ b/plugins/occt/module/vtkF3DOCCTReader.cxx
@@ -244,23 +244,26 @@ public:
           std::swap(cell[0], cell[2]);
         }
         trianglesCells->InsertNextCell(3, cell);
+      }
 
 #if F3D_PLUGIN_OCCT_XCAF
-        if (this->ColorTool)
+      if (this->ColorTool)
+      {
+        std::array<unsigned char, 3> rgb = { 255, 255, 255 };
+        Quantity_Color aColor;
+        if (this->ColorTool->GetColor(face, XCAFDoc_ColorSurf, aColor) ||
+          this->ColorTool->GetColor(shape, XCAFDoc_ColorSurf, aColor))
         {
-          std::array<unsigned char, 3> rgb = { 255, 255, 255 };
-          Quantity_Color aColor;
-          if (this->ColorTool->GetColor(face, XCAFDoc_ColorSurf, aColor) ||
-            this->ColorTool->GetColor(shape, XCAFDoc_ColorSurf, aColor))
-          {
-            rgb[0] = static_cast<unsigned char>(255.0 * aColor.Red());
-            rgb[1] = static_cast<unsigned char>(255.0 * aColor.Green());
-            rgb[2] = static_cast<unsigned char>(255.0 * aColor.Blue());
-          }
+          rgb[0] = static_cast<unsigned char>(255.0 * aColor.Red());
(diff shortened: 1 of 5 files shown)


snoyer:
`TestIGES.png` baseline should not have needed an update now, but it should have been updated with the previous OCCT PR that did change the edges behavior. I guess the tiny variations from both PR just now added up past the test threshold?

Meakk (approved):
Looks good. Why do we have a baseline change tho?

snoyer:
> Looks good. Why do we have a baseline change tho?

Not sure to be honest. I think it's because of the edges showing, that should have needed an update with the last PR that fixed the edges but it didn't (only the STEP one needed to be updated).

But yeah I don't know why on this PR the IGES one failed and the STEP one didn't.

plugins/occt/module/vtkF3DOCCTReader.cxx
```diff
@@ -139,19 +164,24 @@ class vtkF3DOCCTReader::vtkInternals
...
+            Quantity_Color color = style.GetColorCurv();
+            rgb[0] = static_cast<unsigned char>(255.0 * color.Red());
+            rgb[1] = static_cast<unsigned char>(255.0 * color.Green());
+            rgb[2] = static_cast<unsigned char>(255.0 * color.Blue());
           }
-          colors->InsertNextTypedTuple(rgb.data());
         }
+        catch (Standard_NoSuchObject&)
+        {
+          /* ignore */
```
mwestphal, line 181: no need to warn or log ?

plugins/occt/module/vtkF3DOCCTReader.cxx
```diff
@@ -244,24 +266,32 @@ class vtkF3DOCCTReader::vtkInternals
...
+          Quantity_Color color = style.GetColorSurf();
+          rgb[0] = static_cast<unsigned char>(255.0 * color.Red());
+          rgb[1] = static_cast<unsigned char>(255.0 * color.Green());
+          rgb[2] = static_cast<unsigned char>(255.0 * color.Blue());
         }
-#endif
+      }
+      catch (Standard_NoSuchObject&)
+      {
+        /* ignore */
```
mwestphal, line 286: same

plugins/occt/module/vtkF3DOCCTReader.cxx
```diff
@@ -273,7 +303,7 @@ class vtkF3DOCCTReader::vtkInternals
     polydata->SetLines(linesCells);
 
 #if F3D_PLUGIN_OCCT_XCAF
-    /* colors may be left empty if this->ColorTool has not been initialized */
+    /* colors may be left empty ? */
```
mwestphal, line 306: do you mean that you do not understand why this code is there @snoyer ?

plugins/occt/module/vtkF3DOCCTReader.cxx
```diff
@@ -285,6 +315,104 @@ class vtkF3DOCCTReader::vtkInternals
...
+
+    /* iterate styled shapes and assign style if leaf, otherwise collect if parent node */
+    const TopAbs_ShapeEnum leafType = this->Parent->GetReadWire() ? TopAbs_EDGE : TopAbs_FACE;
+    std::vector<std::pair<TopoDS_Shape, XCAFPrs_Style> > parents;
+    for (StyleMap::Iterator iter(collectedStyles); iter.More(); iter.Next())
+    {
+      const TopoDS_Shape& shape = iter.Key();
+      if (shape.ShapeType() == leafType)
+      {
+        inheritedStyles.Add(shape, iter.Value());
```
mwestphal, line 339: do we need new data to cover this codepath ?

plugins/occt/module/vtkF3DOCCTReader.cxx
```diff
@@ -285,6 +315,104 @@ class vtkF3DOCCTReader::vtkInternals
...
+      const XCAFPrs_Style& defaultStyle = collectedStyles.FindFromKey(rootShape);
+      for (StyleMap::Iterator iter(inheritedStyles); iter.More(); iter.Next())
+      {
+        XCAFPrs_Style style = iter.Value();
+        this->PassDownStyleProps(defaultStyle, style);
+      }
+    }
+    catch (Standard_NoSuchObject&)
+    {
+      /* ignore */
```
mwestphal, line 391: no log or warn ?

plugins/occt/module/vtkF3DOCCTReader.cxx
```diff
@@ -285,6 +315,104 @@ class vtkF3DOCCTReader::vtkInternals
...
+    }
+
+    if (!child.IsSetColorSurf() && parent.IsSetColorSurf())
+    {
+      child.SetColorSurf(parent.GetColorSurfRGBA());
+    }
+
+    if (child.Material().IsNull() && !parent.Material().IsNull())
+    {
+      child.SetMaterial(parent.Material());
```
mwestphal, line 412: do we need new data to reach this codepath ?

mwestphal (changes requested):
some questions

plugins/occt/module/vtkF3DOCCTReader.cxx
snoyer: no, just means the part has no style set, i'll update the comment

plugins/occt/module/vtkF3DOCCTReader.cxx
snoyer: same

plugins/occt/module/vtkF3DOCCTReader.cxx
snoyer: there was something about `this->ColorTool` not being initialized (which I wasn't super clear about in the first place) but now I removed it entirely. Would need to investigate what situation this can be reached from...

plugins/occt/module/vtkF3DOCCTReader.cxx
snoyer: I thought it would be reached, I'll have to double check why it's not and possibly remove it if it's not actually useful

plugins/occt/module/vtkF3DOCCTReader.cxx
snoyer: that would be nice yeah. Guess I'll need to learn some more OCCT then

plugins/occt/module/vtkF3DOCCTReader.cxx
snoyer: no, it's an expected possibility that the root shape doesn't have a style set, I'll update the comment

plugins/occt/module/vtkF3DOCCTReader.cxx
mwestphal: thanks for checking

snoyer: alright so it's a cute optimization in theory but it seems unlikely to come up IRL.

I think the optimization isn't worth the code complexity in the end (also it depends on the `ReadWire` state  [which is on by default](https://github.com/search?q=repo%3Af3d-app%2Ff3d%20ReadWireOn&type=code)), so I'll just skip it and keep it simple

plugins/occt/module/vtkF3DOCCTReader.cxx
snoyer: We may not be able to hit this with STEP/IGES files as apparently they only support basic colors (or at least the OCC step exporter only exports colors)

https://www.opencascade.com/blog/pbr-in-occt-3d-viewer/
> - XBF document stores any combination of material and color labels. It should be noted, however, that assigning both color and material to the same part gives not well defined result and should be avoided.
> - STEP, IGES and similar formats storing only color information keep using a color table in an XCAF document. Visualization materials in the document are exported as colors.
> - JT reader/writer now translates material properties into a Common material definition in an XCAF document.

no idea what XBF and JT are and whether we can use them or not

Meakk (approved):
I don't pretend to understand everything but that looks good to me

plugins/occt/module/vtkF3DOCCTReader.cxx
mwestphal: improving the comment in that sense would be nice

plugins/occt/module/vtkF3DOCCTReader.cxx
mwestphal: fine if we cant, as you see fit.

mwestphal (approved):

PR #1302, opened 2024-03-07
maintain config declaration order

snoyer:
The config blocks are currently matched by regex in alphabetical order which is counter-intuitive.

Technically it makes sense because JSON specifies dictionaries as unordered collections and F3D uses `std::map` for storage, but this feels like unspecified behavior left to the implementation :(

This PR uses `nlohmann::ordered_json` and `std::vector<std::pair<...>>` to maintain the top-down declaration order form the config file.


Meakk (approved):
LGTM

mwestphal (approved):

testing/data/config-order.json
mwestphal, line 1: wrong location for this file

Issue #86, opened 2021-11-29
Option to automatically reload the active file


JulianGmp:
Hello there,
thank you for your work to provide this tool.

I was looking into this as a preview tool for text based 3D Modeling/CAD, and I'm suggesting adding a command line option to enable an automatic reload of the active file.
This way, when the loaded file gets overwritten by the 3D modeling script, the preview refreshes immediately. Ideally the camera should keep its position as well.

Best regards,
Julian

Meakk:
Hi @JulianGmp  
It's already possible, press the UP arrow key.  
Unfortunately, the camera position is reset right now.

JulianGmp:
I know that the up arrow key reloads the file. I work around the camera reset by setting a decent initial camera position through the command line.
But I think it would be really nice to have a command line option and have the window just auto update as soon as the file gets updated, but I realize that this might be a very niche use case that I have.

Meakk:
I see. It's actually a good idea, and looks very useful, but it may be tricky to implement. How to know if a file write is complete? How to trigger a refresh?  
Any idea @mwestphal ?

mwestphal:
That would require to monitor the writing time on the file with a dedicated thread, not something F3D should do imo.

mwestphal:
An alternative would be to reload the file every n seconds, that would be a much simpler approach, but could be done with a simple external script ?

JulianGmp:
Watching a file is dependant to the operating system, on Linux you would use inotify, but on other systems it's different. There are cross platform wrappers around like https://github.com/septag/dmon, but I agree that that sounds a bit overkill.

Checking if the file changed manually would be a much simpler solution, we could just poll it every second or so and reload the file if needed.

Using an external script does somewhat work, but it doesn't fix the problem with the reset camera on every load.


In fact, I started digging through the code (which is well documented I must say), and in `F3DLoader::LoadFile(int)` (`F3DLoader.cxx:389`) you actually create a new renderer object when loading a file.
I tried simply setting the camera object to the new renderer and it worked as I would expect, though all other render settings such as the grid enabled flag, were reset.
I _think_ we could simply reuse the same renderer, and only reset the camera if we load in a new file.
Or, if reusing the same renderer object can't be done, copy over the options to the new renderer.
That way, when going through multiple files, options such as grid enabled or SSAO will stick even when switching files.

This all might just be my personal opinion though :)

mwestphal:
[...]

PR #1308, opened 2024-03-10
Adding watch option using dmon

mwestphal:
This add support for an "--watch" cli option where the current file is reloaded automatically if changed on disk.
It relies on https://github.com/septag/dmon for detecting the changes.

In order to support this properly, F3DStarter is reworked with a proper event loops that currently supports:
 - Reloading the current file
 - Rendering

This also adds a test for it on linux/apple/windows

 
Script used:

```
while :
do
  sleep 1
  cp cowlow.vtp a.vtp
  sleep 1
  cp cow.vtp a.vtp
done
```

https://github.com/f3d-app/f3d/assets/3129530/3d4510b3-5303-476e-bba9-0e41394b30d0

Fix: #86

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -257,7 +257,7 @@ jobs:
     if: github.event.pull_request.draft == false
 
     runs-on: ubuntu-latest
-    container: mwestphal/f3d-ci
+    container: ghcr.io/f3d-app/f3d-ci
 
     # Add dummy F3D_PLUGINS_PATH for coverage
     env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -257,7 +257,7 @@ jobs:
     if: github.event.pull_request.draft == false
 
     runs-on: ubuntu-latest
-    container: ghcr.io/f3d-app/f3d-ci
+    container: mwestphal/f3d-ci
 
     # Add dummy F3D_PLUGINS_PATH for coverage
     env:
--- a/application/F3DOptionsParser.cxx
+++ b/application/F3DOptionsParser.cxx
@@ -311,6 +311,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp0, "dry-run", "", "Do not read the configuration file", appOptions.DryRun,  HasDefault::YES, MayHaveConfig::NO );
     this->DeclareOption(grp0, "no-render", "", "Do not render anything and quit right after loading the first file, use with --verbose to recover information about a file.", appOptions.NoRender,  HasDefault::YES, MayHaveConfig::YES );
     this->DeclareOption(grp0, "max-size", "", "Maximum size in Mib of a file to load, negative value means unlimited", appOptions.MaxSize,  HasDefault::YES, MayHaveConfig::YES, "<size in Mib>");
+    this->DeclareOption(grp0, "auto-reload", "", "Reload currently loaded file if it is modified on disk", appOptions.AutoReload,  HasDefault::YES, MayHaveConfig::YES );
     this->DeclareOption(grp0, "load-plugins", "", "List of plugins to load separated with a comma", appOptions.Plugins, LocalHasDefaultNo, MayHaveConfig::YES, "<paths or names>");
     this->DeclareOption(grp0, "scan-plugins", "", "Scan some directories for plugins (result can be incomplete)");
 
--- a/application/F3DOptionsParser.cxx
+++ b/application/F3DOptionsParser.cxx
@@ -311,7 +311,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp0, "dry-run", "", "Do not read the configuration file", appOptions.DryRun,  HasDefault::YES, MayHaveConfig::NO );
     this->DeclareOption(grp0, "no-render", "", "Do not render anything and quit right after loading the first file, use with --verbose to recover information about a file.", appOptions.NoRender,  HasDefault::YES, MayHaveConfig::YES );
     this->DeclareOption(grp0, "max-size", "", "Maximum size in Mib of a file to load, negative value means unlimited", appOptions.MaxSize,  HasDefault::YES, MayHaveConfig::YES, "<size in Mib>");
-    this->DeclareOption(grp0, "auto-reload", "", "Reload currently loaded file if it is modified on disk", appOptions.AutoReload,  HasDefault::YES, MayHaveConfig::YES );
+    this->DeclareOption(grp0, "watch", "", "Watch current file and automaticaly reload it whenever it is modified on disk", appOptions.Watch,  HasDefault::YES, MayHaveConfig::YES );
     this->DeclareOption(grp0, "load-plugins", "", "List of plugins to load separated with a comma", appOptions.Plugins, LocalHasDefaultNo, MayHaveConfig::YES, "<paths or names>");
     this->DeclareOption(grp0, "scan-plugins", "", "Scan some directories for plugins (result can be incomplete)");
 
--- a/application/F3DOptionsParser.cxx
+++ b/application/F3DOptionsParser.cxx
@@ -311,7 +311,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp0, "dry-run", "", "Do not read the configuration file", appOptions.DryRun,  HasDefault::YES, MayHaveConfig::NO );
     this->DeclareOption(grp0, "no-render", "", "Do not render anything and quit right after loading the first file, use with --verbose to recover information about a file.", appOptions.NoRender,  HasDefault::YES, MayHaveConfig::YES );
     this->DeclareOption(grp0, "max-size", "", "Maximum size in Mib of a file to load, negative value means unlimited", appOptions.MaxSize,  HasDefault::YES, MayHaveConfig::YES, "<size in Mib>");
-    this->DeclareOption(grp0, "watch", "", "Watch current file and automaticaly reload it whenever it is modified on disk", appOptions.Watch,  HasDefault::YES, MayHaveConfig::YES );
+    this->DeclareOption(grp0, "watch", "", "Watch current file and automatically reload it whenever it is modified on disk", appOptions.Watch,  HasDefault::YES, MayHaveConfig::YES );
     this->DeclareOption(grp0, "load-plugins", "", "List of plugins to load separated with a comma", appOptions.Plugins, LocalHasDefaultNo, MayHaveConfig::YES, "<paths or names>");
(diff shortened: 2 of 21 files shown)


application/F3DStarter.cxx
```diff
@@ -653,7 +686,9 @@ int F3DStarter::AddFile(const fs::path& path, bool quiet)
 
     if (it == this->Internals->FilesList.end())
     {
+      this->Internals->FilesListMutex.lock();
       this->Internals->FilesList.push_back(tmpPath);
+      this->Internals->FilesListMutex.unlock();
       return static_cast<int>(this->Internals->FilesList.size()) - 1;
```
snoyer, line 692: get size before unlocking
```suggestion
      this->Internals->FilesList.push_back(tmpPath);
      const size_t fileCount = this->Internals->FilesList.size();
      this->Internals->FilesListMutex.unlock();
      return static_cast<int>(fileCount) - 1;
```

mwestphal: line 690 is the only place where FilesList can be modfied. Concurrent read doesnt matter, does it ?

snoyer: I don't know, it _feels_ like if you're locking around the push you may want to get the updated size before another push has a chance to get in there.
But I you know for sure that specific piece of code isn't going to be concurrent with itself I guess it's fine either way

mwestphal: yes, this mutex is there ONLY for the dmon case. There are many other read of FileLists everywhere in the file, I'd rather not add mutex code there either. I can add a comment though.

Meakk: Let's use a `std::lock_guard` anyway

application/F3DStarter.cxx
```diff
@@ -132,15 +137,36 @@ class F3DStarter::F3DInternals
     }
   }
 
+  static void dmonFolderChanged(dmon_watch_id watch_id, dmon_action action, const char* rootdir,
+    const char* filename, const char*, void* userData)
+  {
+    F3DStarter* self = reinterpret_cast<F3DStarter*>(userData);
+    self->Internals->FilesListMutex.lock();
```
Meakk, line 144: again let's use a scoped lock (`std::lock_guard`)

application/F3DStarter.h
```diff
@@ -34,10 +34,15 @@ class F3DStarter
   void LoadFile(int index = 0, bool relativeIndex = false);
 
   /**
-   * Trigger a render
+   * Trigger a render on the next event loop
    */
   void Render();
 
+  /**
+   * Trigger a render now
```
Meakk, line 42:
```suggestion
   * Trigger a render immediately (must be called by the main thread)
```

application/F3DStarter.h
```diff
@@ -47,6 +52,19 @@ class F3DStarter
...
   std::unique_ptr<F3DInternals> Internals;
+
+  /**
+   * Internal method triggered when interacting with the application
+   * that load a file using relative index and handle camera restore
+   */
+  bool LoadRelativeFile(int relativeIndex = 0, bool restoreCamera = false);
+
+  /**
+   * Internal event loop that is triggered repeatdly to handle specific events:
```
Meakk, line 63:
```suggestion
   * Internal event loop that is triggered repeatedly to handle specific events:
```

application/testing/auto_reload.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
Meakk, line 1: should we use a Python script instead of bash to have a single script for all OS?

application/F3DOptionsParser.cxx
```diff
@@ -311,6 +311,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp0, "dry-run", "", "Do not read the configuration file", appOptions.DryRun,  HasDefault::YES, MayHaveConfig::NO );
     this->DeclareOption(grp0, "no-render", "", "Do not render anything and quit right after loading the first file, use with --verbose to recover information about a file.", appOptions.NoRender,  HasDefault::YES, MayHaveConfig::YES );
     this->DeclareOption(grp0, "max-size", "", "Maximum size in Mib of a file to load, negative value means unlimited", appOptions.MaxSize,  HasDefault::YES, MayHaveConfig::YES, "<size in Mib>");
+    this->DeclareOption(grp0, "auto-reload", "", "Reload currently loaded file if it is modified on disk", appOptions.AutoReload,  HasDefault::YES, MayHaveConfig::YES );
```
snoyer, line 314: It looks like "file watching" is the accepted terminology and `--watch` is already a thing out there ([`webpack`](https://webpack.js.org/api/cli/), [`tsc`](https://www.typescriptlang.org/docs/handbook/compiler-options.html), [`spawning(1)`](https://linux.die.net/man/1/spawning), [`qiv`](https://linux.die.net/man/1/qiv), [`kubectl`](https://kubernetes.io/docs/reference/kubectl/), ...)

For the description I'm not 100% but maybe using "whenever" instead of "if" is enough to convey the automatic/continuous/repeating idea?
```suggestion
    this->DeclareOption(grp0, "watch", "", "Reload the current file whenever it is modified on disk", appOptions.AutoReload,  HasDefault::YES, MayHaveConfig::YES );
```
also "currently loaded file" vs "current file" vs "active file"? idk

application/testing/auto_reload.sh
mwestphal: i'd rather write a powershell script instead of adding a dependency to python in our testing.

application/testing/test_watch.sh
```diff
@@ -0,0 +1,30 @@
...
+
+sleep 1
+cp $lowres_data $reloaded_data
+sleep 1
+
+grep -q "Number of points: 634" $log
+ret=$?
+
+kill $pid
+exit $ret
```
mwestphal, line 30: need a trap to improve the test

application/testing/test_watch.ps1
```diff
@@ -0,0 +1,30 @@
...
+
+$str = Select-String $log -Pattern "Number of points: 634"
+if ($str -ne $null) {
+  $ret = 0
+} else {
+  $ret = 1
+}
+
+Stop-Process -Id $id
+exit $ret
```
mwestphal, line 30: need a trap to improve the script

application/testing/CMakeLists.txt
```diff
@@ -897,3 +897,14 @@ if(NOT F3D_MACOS_BUNDLE)
...
   endif()
 endif()
+
+if(NOT VTK_OPENGL_HAS_EGL)
+  if(UNIX)
+    # Custom bash test for testing watch option
+    add_test (NAME f3d::TestWatch COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/test_watch.sh $<TARGET_FILE:f3d> ${F3D_SOURCE_DIR}/testing/data ${CMAKE_BINARY_DIR}/Testing/Temporary)
+  elseif(WIN32)
+    find_program(PWSH NAMES pwsh powershell)
+    add_test (NAME f3d::TestWatch COMMAND ${PWSH} ${CMAKE_CURRENT_SOURCE_DIR}/test_watch.ps1 $<TARGET_FILE:f3d> ${F3D_SOURCE_DIR}/testing/data ${CMAKE_BINARY_DIR}/Testing/Temporary)
```
mwestphal, line 907: could be improved with some cmake logic

application/F3DStarter.cxx
```diff
@@ -6,15 +6,29 @@
 #include "F3DOptionsParser.h"
 #include "F3DSystemTools.h"
 
+#define DMON_IMPL
+#ifdef WIN32
+#pragma warning(push)
+#pragma warning(disable : 4505)
+#include "dmon.h"
+// dmon includes Windows.h which defines 'ERROR' and conflicts with log.h
```
Meakk, line 14: Can you try to put `#define WIN32_LEAN_AND_MEAN` before `#include "dmon.h"` to see if `ERROR` is still defined?

snoyer:
Looks like you already have the implementation details covered so I'm left with nitpicking naming:

- `ForceRender()` just performs the render it doesn't take any extra steps to "force" it -> `Render()`
- `Render()` doens't actually perform the render, it only asks for the render to be done (or declares the current render outdated depending on how you want to look at it) -> `RequestRender()`, `InvalidateRender()`, ...
- `SoftRender` is a flag but is named like an action -> `RenderRequested`, `RenderNeeded`, `RenderFlag`, ... (same for `ReloadFile` -> `ReloadRequested`, ...)

tldr:

```cpp
void Render() { SoftRender = true; } // do the thing = don't actually  o_O
if (SoftRender) ForceRender(); // if soft then forceful  o_O
```
```cpp
void RequestRender() { RenderRequested = true; } // request the thing = mark it as requested  ^_^
if (RenderRequested) Render(); // do the thing if asked  ^_^
```

mwestphal:
As always, the easiest to find out how to do the right thing is just to do the wrong thing and wait for people to fix it for you, thanks @snoyer :)

application/F3DStarter.cxx
mwestphal: does not work sadly.

application/testing/test_watch.sh
```diff
@@ -0,0 +1,36 @@
...
+log=$tmp_dir/output.log
+$f3d_cmd --watch --verbose $reloaded_data > $log &
+pid=$!
+
+function cleanup()
+{
+  if [[ -z "${CTEST_F3D_COVERAGE-}" ]]; then
+    kill $pid
+  else
+    xdotool key Escape
```
mwestphal, line 27: @Meakk should I simplify this and require xdotool for testing all the time except on only when enabling F3D_COVERAGE ?

.github/workflows/ci.yml
```diff
@@ -257,7 +257,7 @@ jobs:
     if: github.event.pull_request.draft == false
 
     runs-on: ubuntu-latest
-    container: ghcr.io/f3d-app/f3d-ci
+    container: mwestphal/f3d-ci
```
Meakk, line 260: to change

application/testing/test_watch.sh
Meakk: The problem is `xdotool` won't work on macos.

mwestphal: very good point, lets keep as is.

application/testing/test_watch.ps1
mwestphal: lets not, its fine as is.

application/testing/CMakeLists.txt
```diff
@@ -899,3 +899,20 @@ if(NOT F3D_MACOS_BUNDLE)
     f3d_test(NAME TestScanPluginsCheckExodus ARGS --scan-plugins NO_RENDER NO_BASELINE REGEXP " - exodus")
   endif()
 endif()
+
+# Watch testing require onscreen renderong
```
Meakk, line 903:
```suggestion
# Watch testing require onscreen rendering
```

Meakk (approved):
small typo to fix and good to go!

PR #1309, opened 2024-03-11
Add colormap file support

Meakk:


application/F3DColorMapTools.cxx
```diff
@@ -0,0 +1,71 @@
...
+    cm.resize(4 * w);
+
+    for (int i = 0; i < w; i++)
+    {
+      cm[4 * i] = static_cast<double>(i) / w;
+      for (int j = 0; j < 3; j++)
+      {
+        cm[4 * i + j + 1] = img(i, 0, j);
+      }
+    }
```
snoyer, line 61: scale the pixel values to `[0,1]` here instead of within `image::operator(x,y,c)` ?
```suggestion
    for (int i = 0; i < w; i++)
    {
      cm[4 * i] = static_cast<double>(i) / (w - 1);
      for (int j = 0; j < 3; j++)
      {
        cm[4 * i + j + 1] = img(i, 0, j) / max;
      }
    }
```

with the channel type switch just above the loop
```cpp
    double max = 1.0;
    switch (img.getChannelType())
    {
      case f3d::image::ChannelType::BYTE:
        max = 255.0;
        break;
      case f3d::image::ChannelType::SHORT:
        max = 65535.0;
        break;
      default:
        break;
    }
```

and for some extra subjective and inconsequential readability considerations:

```suggestion
    constexpr int y = 0;
    for (int x = 0; x < w; ++x)
    {
      cm[4 * x + 0] = static_cast<double>(x) / (w - 1);
      cm[4 * x + 1] = img(x, y, 0) / max;
      cm[4 * x + 2] = img(x, y, 1) / max;
      cm[4 * x + 3] = img(x, y, 2) / max;
    }
```

library/src/image.cxx
```diff
@@ -353,6 +353,24 @@ bool image::operator!=(const image& reference) const
...
+  {
+    case ChannelType::BYTE:
+      return v / 255.0;
+    case ChannelType::SHORT:
+      return v / 65535.0;
+    default:
+      break;
+  }
+
+  return v;
```
snoyer, line 371:
```suggestion
  return this->Internals->Image->GetScalarComponentAsDouble(x, y, 0, channel);
```
Normalization feels specific to the palette usage, `image` API random access makes sense but should stay generic IMO

application/F3DColorMapTools.cxx
```diff
@@ -0,0 +1,70 @@
...
+{
+std::string Find(const std::string& str)
+{
+  if (fs::exists(str))
+  {
+    // already full path;
+    return str;
+  }
+
+  // todo: stem, find file in standard paths
```
mwestphal, line 20: todo ?

application/F3DColorMapTools.h
```diff
@@ -0,0 +1,19 @@
+/**
+ * @class   F3DColorMapReader
+ * @brief   A class used to convert images to libf3d colormap option
```
mwestphal, line 3: a namespace

application/F3DOptionsParser.cxx
```diff
@@ -379,6 +379,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp4, "cells", "c", "Use a scalar array from the cells", options.getAsBoolRef("model.scivis.cells"), HasDefault::YES, MayHaveConfig::YES);
     this->DeclareOption(grp4, "range", "", "Custom range for the coloring by array", options.getAsDoubleVectorRef("model.scivis.range"), HasDefault::YES, MayHaveConfig::YES, "<min,max>");
     this->DeclareOption(grp4, "bar", "b", "Show scalar bar", options.getAsBoolRef("ui.bar"), HasDefault::YES, MayHaveConfig::YES);
+    this->DeclareOption(grp4, "colormap-file", "", "Specify a colormap image (ignored if \"colormap\" is specified)", appOptions.UserColorMap, LocalHasDefaultNo, MayHaveConfig::YES, "<filePath/filename/fileStem>");
```
mwestphal, line 382: appOptions.ColorMapFile maybe ?

doc/user/COLOR_MAPS.md
```diff
@@ -0,0 +1,28 @@
+# Color maps
+
+## Image files
+
+It is possible to specify a color map using the `--colormap-file` option.
+TODO: full path, stem...
```
mwestphal, line 6: TODO

doc/user/COLOR_MAPS.md
```diff
@@ -0,0 +1,28 @@
...
+seismic|<img src="https://github.com/Meakk/f3d/blob/colormap-file/resources/colormaps/seismic.png?raw=true" alt="drawing" width="256" height="10" />
+viridis|<img src="https://github.com/Meakk/f3d/blob/colormap-file/resources/colormaps/viridis.png?raw=true" alt="drawing" width="256" height="10" />
+
+It is possible to create a custom color map by creating a simple RGB image in any supported format (PNG recommended), and any resolution (if the height is more than 1 row, only the first one is taken into account).
+
+## Custom values
+
+It is also possible to set values manually using `--colormap` option.
+
+TODO: detailed doc
```
mwestphal, line 28: TODO

library/src/image.cxx
mwestphal: I'm not convinced by the current approach either.

mwestphal: Maybe it makes sense, but lets make sure it makes sense because the API makes sense, not because we need it to behave like that to support that feature, iykwim.

Meakk: In rendering, passing integers (could be `char` or `short`) is considered as "normalized integers", and sampling buffers (or textures) results in a value between 0 and 1.  
If you think about it, a 8-bit channel in an image is just defining the brightness of the channel, it's a *quantized* value of the (0,1) range.  
What I'm doing here by dividing by 255 or 65535 is a dequantization (I think I made up this word, but you get it)  
32-bits floating point images can go beyond the (0,1) range, and it's called HDR.

See:
- https://www.khronos.org/opengl/wiki/Normalized_Integer
- https://pytorch.org/docs/stable/quantization.html

application/F3DConfigFileTools.h
```diff
@@ -14,7 +14,6 @@ namespace F3DConfigFileTools
 {
 std::filesystem::path GetUserConfigFileDirectory();
 std::filesystem::path GetBinaryConfigFileDirectory();
-std::filesystem::path GetSystemConfigFileDirectory();
```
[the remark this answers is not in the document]
Meakk: No, the function is defined but not implemented.

library/public/image.h
```diff
@@ -85,6 +85,13 @@ class F3D_EXPORT image
   bool operator!=(const image& reference) const;
   ///@}
 
+  /**
+   * Read one specific pixel and return all channel values.
+   * If the channel type is BYTE or SHORT, the values are normalized to [0, 1] range.
+   * \warning This function is slow, prefer getContent when possible.
```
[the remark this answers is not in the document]
Meakk: That's the point. It is slow when reading a large image and `getContent` should be used.

resources/colormaps/cividis.png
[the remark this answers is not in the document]
Meakk: @snoyer used `matplotlib` to generate them. It has the same license than Python (PSF). So I think it's fine? I mean you wouldn't add the Python license if you use Python to generate data?

doc/user/CONFIGURATION_FILE.md
```diff
@@ -68,9 +68,9 @@ off on the command line if needed, eg: `--point-sprites=false`.
 The configuration file or directory possible locations depends on your operating system.
 They are considered in the below order and only the first found will be used.
 
- * Linux: `${XDG_CONFIG_HOME}/.config/f3d/config(.json,.d)`, `~/.config/f3d/config(.json,d)`, `/etc/f3d/config(.json,.d)`, `/usr/share/f3d/configs/config(.json,.d)`, `[install_dir]/share/f3d/configs/config(.json,.d)`
+ * Linux: `${XDG_CONFIG_HOME}/f3d/config(.json,.d)`, `~/.config/f3d/config(.json,d)`, `/etc/f3d/config(.json,.d)`, `/usr/share/f3d/configs/config(.json,.d)`, `[install_dir]/share/f3d/configs/config(.json,.d)`
```
Meakk, line 71: I think `XDG_CONFIG_HOME` already contain the `.config`?

library/src/image.cxx
mwestphal: Looks like it is indeed a concept that exist and could be expected, lets make it clear in the doc what getContent and [] provides.

library/public/image.h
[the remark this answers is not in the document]
mwestphal: indeed, lets clarify the doc in that sense

resources/colormaps/cividis.png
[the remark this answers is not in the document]
mwestphal: hum that's a good question, let me check on that.

doc/user/CONFIGURATION_FILE.md
mwestphal: correct

library/src/image.cxx
snoyer: > In rendering, [...]

That's a better argument than "byte as double is weird" ;p

to sum up:

1. if `f3d::image` is stated to represent a generic image, ie. where a pixel is one or more arbitrary numerical value, then the pixel access shouldn't make assumptions about the meaning of the data. It would still need to cast to `vector<double>` because that can hold any of the possible underlying `byte`/`short`/`float` values but that's it, any transformation is up to the user to know about and deal with. Examples where you wouldn't want the API to mess with your data: indexed images where the values are actually ids, images where the values are raw physical data (DEM), anything where the quantization does not map to `[0,1]` (signed difference? normal map?).

2. if `f3d::image` is stated to represent a "rendering image", ie. where a pixel represents one or more color intensity (or otherwise normalized quantity), then it's fair to say the underlying encoding is an implementation detail and the pixel access API operates on/with said normalized values, that does mean making assumptions and assigning meaning to the data but if that's explicitly the nature/purpose of the class then all good. Users wanting to deal with esoteric bitmaps can still do so through the data buffer and the pixel access API is just not for them.

I'm personally partial to number 1, but, given the scope of F3D, number 2 is definitely valid.

resources/colormaps/cividis.png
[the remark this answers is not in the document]
mwestphal: each license for each colormap can be found here:

eg, for cividis: https://cmap-docs.readthedocs.io/en/latest/catalog/sequential/cividis%3Acividis/

Please respect these licenses terms and add them in the right document. Keep in mind that since we install them, it should end up in the installation too (contrary to the testing data).

The one colormap without license in your list is:
hot: so you should use either PSF or remove it: https://cmap-docs.readthedocs.io/en/latest/catalog/sequential/matlab%3Ahot/

library/src/image.cxx
Meakk: But "bytes as double is weird" is still one of my arguments 😛 
C++ is a typed language, and even if `f3d::image` abstracts the underlying type, trying to get values as `double` when it's not stored as such internally is conceptually an issue to me.  
If we want random access of not normalized pixel, I'd prefer to have `getPixelAsByte`, `getPixelAsShort` and `getPixelAsFloat` that throws if the requested type is invalid.

snoyer: > abstracts the underlying type [...] throws if the requested type is invalid.

We don't have the same definition of abstraction then :D
Coming from python it feels to me like setting yourself such runtime traps on purpose in a strongly and statically typed (and compiled!) language is a "worst of both world" situation. Obviously having the fully templated `image<Type, Size, Count>` would quickly become a drag for obvious reasons but if you care that much about types wouldn't you at least have `image<Type>` and throw on load so at least your accesses are type-safe? (or maybe I'm _that_ misguided in my understanding of c++)

Anyway we're in purely rhetorical territory now (because who doesn't love arguing moot points with strangers on the internet) so maybe we'll need to move this discussion somewhere else

Meakk: > We don't have the same definition of abstraction then :D

Maybe I should have said `f3d::image` is generic instead? 🤔 

We cannot really have an `image<Type>` because you don't know at compile time the type of your image if you are loading it from file. We could throw at opening if it's not the correct type though but that's a different design.

Meakk: Just read your message again and realized I'm describing exactly what you suggested 🤦  
Anyway, I'm not against it, but that's a big change, it needs in-depth discussions definitely out of the scope of this PR.  
Also, I don't know what it would look like in Python? 3 different Python classes for each specialization (like `f3d.Image8`, `f3d.Image16` and `f3d.Image32`)?

snoyer: no worries; and I'm not even saying it should be changed at all, just basically rambling out my confusion :)

For python idk, it's not statically typed so you could do the "throw on wrong type" thing or you just return whatever type and let the client check what it gets at runtime. For the actual data storage you'd probably just use numpy arrays for efficiency/performance and compatibility with existing ecosystem?

doc/user/COLOR_MAPS.md
```diff
@@ -0,0 +1,41 @@
+# Color maps
+
+## Image files
+
+It is possible to specify a color map using the `--colormap-file` option.
+The value of the option can be a full path to an image file, a filename only is which case, these directories are used to find the file (in this order):
```
mwestphal, line 6: The value of the option can be a path to an image file, a filename or a filestem.
If it is not a path, these directories are used to find the file (in this order):

mwestphal: any image file ? which extensions are supported ?

doc/user/COLOR_MAPS.md
```diff
@@ -0,0 +1,41 @@
...
+ * Linux: `${XDG_CONFIG_HOME}/f3d/colormaps`, `~/.config/f3d/colormaps`, `/etc/f3d/colormaps`, `/usr/share/f3d/colormaps`, `[install_dir]/share/f3d/colormaps`
+ * Windows: `%APPDATA%\f3d\colormaps`, `[install_dir]\share\f3d\configs\colormaps`
+ * macOS: `${XDG_CONFIG_HOME}/f3d/colormaps`, `~/.config/f3d/colormaps`, `/usr/local/etc/f3d/colormaps`, `f3d.app/Contents/Resources/configs/colormaps`
+In case the filename omits the extension, PNG files only are searched.
```
mwestphal, line 10: If it is a filestem, the `.png` extension will be added.

doc/user/COLOR_MAPS.md
```diff
@@ -0,0 +1,41 @@
...
+hot|<img src="https://github.com/Meakk/f3d/blob/colormap-file/resources/colormaps/hot.png?raw=true" class="cm" width="256" height="10" />
+inferno|<img src="https://github.com/Meakk/f3d/blob/colormap-file/resources/colormaps/inferno.png?raw=true" class="cm" width="256" height="10" />
+magma|<img src="https://github.com/Meakk/f3d/blob/colormap-file/resources/colormaps/magma.png?raw=true" class="cm" width="256" height="10" />
+plasma|<img src="https://github.com/Meakk/f3d/blob/colormap-file/resources/colormaps/plasma.png?raw=true" class="cm" width="256" height="10" />
+seismic|<img src="https://github.com/Meakk/f3d/blob/colormap-file/resources/colormaps/seismic.png?raw=true" class="cm" width="256" height="10" />
+viridis|<img src="https://github.com/Meakk/f3d/blob/colormap-file/resources/colormaps/viridis.png?raw=true" class="cm" width="256" height="10" />
+
+It is possible to create a custom color map by creating a simple RGB image in any supported format (PNG recommended), and any resolution (if the height is more than 1 row, only the first one is taken into account). The image must be copied in the user config directory:
```
mwestphal, line 26: which formats are supported ?

doc/user/COLOR_MAPS.md
mwestphal: why though ? look for file with the right filestem and try to load them, png or not ?

doc/user/COLOR_MAPS.md
mwestphal: mention that the first file found will be used.

resources/colormaps/licenses.md
```diff
@@ -0,0 +1,11 @@
+|Name      |Author                                              |License     |
+|----------|----------------------------------------------------|------------|
+|cividis   |Jamie Nunez                                         |BSD-3-Clause|
+|cubehelix |                                                    |BSD-2-Clause|
```
mwestphal, line 4: missing author for this BSD colormap

mwestphal (changes requested):
few small changes

doc/user/COLOR_MAPS.md
Meakk: so that means listing all "search dir" files and compare the filestem with the given one? Sounds expensive to me if the directory contains a lot of files.

mwestphal: it feels just a bit arbitrary to support filestem just for pngs and not other formats. Since the list of formats is small, you can just check every single possible extension instead ?

doc/user/COLOR_MAPS.md
Meakk: done

doc/user/COLOR_MAPS.md
Meakk: It was easier than what I thought. I rely on `f3d::image::getSupportedFormats`

.github/actions/generic-ci/action.yml
```diff
@@ -232,6 +232,7 @@ runs:
       run: |
         cmake --install . --component sdk
         cmake --install . --component configuration
+        cmake --install . --component colormaps
```
mwestphal, line 235: add an install test ?

application/F3DConfigFileTools.cxx
```diff
@@ -59,9 +59,9 @@ fs::path F3DConfigFileTools::GetBinaryConfigFileDirectory()
 
     // Add binary specific paths
 #if F3D_MACOS_BUNDLE
-    dirPath /= "Resources/configs";
+    dirPath /= "Resources";
 #else
-    dirPath /= "share/f3d/configs";
+    dirPath /= "share/f3d";
```
mwestphal, line 64: then that a strange name for that method is it not ?

application/F3DStarter.cxx
```diff
@@ -313,6 +314,27 @@ int F3DStarter::Start(int argc, char** argv)
     }
 #endif
   }
+
+  // Parse colormap
+  if (!this->Internals->AppOptions.ColorMapFile.empty())
```
mwestphal, line 319: should we check if colormap option is not empty ?

mwestphal (changes requested):
small changes needed

.github/actions/generic-ci/action.yml
Meakk: How? Are we checking things like mimetype or config installation somewhere?  
I think testing it is the SB responsibility?

Meakk: How? Are we checking things like mimetype or config installation somewhere?  
I think testing it is the SB responsibility?

application/F3DStarter.cxx
Meakk: Idk... Which one should take precedence?

.github/actions/generic-ci/action.yml
mwestphal: We are not testing mimetypes, but we are testing config installation alright. See generic-ci/action.yml, look for `install`.

It is tested by the superbuild as well, but it is different, the superbuild is testing the shipped package, not the installation, which is a step further.

application/F3DStarter.cxx
mwestphal: in the doc you say that colormap takes precedence, is it not implemented like that ?

.github/actions/generic-ci/action.yml
Meakk: found it, thanks
I extended the current test

application/F3DStarter.cxx
Meakk: I have updated the doc to clearly specify `--colormap-file` has precedence over `--colormap`

application/F3DOptionsParser.cxx
```diff
@@ -379,6 +379,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp4, "cells", "c", "Use a scalar array from the cells", options.getAsBoolRef("model.scivis.cells"), HasDefault::YES, MayHaveConfig::YES);
     this->DeclareOption(grp4, "range", "", "Custom range for the coloring by array", options.getAsDoubleVectorRef("model.scivis.range"), HasDefault::YES, MayHaveConfig::YES, "<min,max>");
     this->DeclareOption(grp4, "bar", "b", "Show scalar bar", options.getAsBoolRef("ui.bar"), HasDefault::YES, MayHaveConfig::YES);
+    this->DeclareOption(grp4, "colormap-file", "", "Specify a colormap image (ignored if \"colormap\" is specified)", appOptions.ColorMapFile, LocalHasDefaultNo, MayHaveConfig::YES, "<filePath/filename/fileStem>");
```
mwestphal, line 382: doc is wrong

application/F3DStarter.cxx
```diff
@@ -322,6 +323,27 @@ int F3DStarter::Start(int argc, char** argv)
...
+  {
+    std::string fullPath = F3DColorMapTools::Find(this->Internals->AppOptions.ColorMapFile);
+
+    if (!fullPath.empty())
+    {
+      auto cm = F3DColorMapTools::Read(fullPath);
+
+      if (!cm.empty())
+      {
+        this->Internals->Engine->getOptions().set("model.scivis.colormap", cm);
```
mwestphal, line 338: Something I dont like about this is that `colormap-file` only takes precedence if it is valid. You should add actual codepath to handle "both defined" case imo.

Meakk: The problem is there is a default value for the colormap so I cannot rely on the f3d::options value. And I don't have access to the cxxopts args here.

application/F3DOptionsParser.cxx
Meakk: good catch!

application/F3DStarter.cxx
mwestphal: how about cleaning up `colormap` if `colormap-file` is not empty ?

mwestphal (approved):

PR #1325, opened 2024-03-22
Fix fish completion when an option is specified in comments

Meakk:
Fix a fish issue caused by the string `--verbose` wrongly grepped as a long option although it's inside another option description:
```
      --no-render               Do not render anything and quit right after 
                                loading the first file, use with --verbose 
                                to recover information about a file.
```

image


Meakk:
please take a look @mwestphal

Issue #1101, opened 2023-12-17
Let F3D send it's output image in a stream rather than into a file on disk


mrdeveloperdude:
See related issue for [streaming input](https://github.com/f3d-app/f3d/issues/1100).

F3D currently accepts a png filename as a command-line argument for where to store it's output.

F3D will then write the output image into that file after rendering.

Writing the image to disk can incur significant IO penalty. Further, in many cases we won't need the image on disk at all, we simply want the image in memory for processing or redirection over a network or into an external storage system.

This feature request is for having a sanctified and documented way of streaming the image data out of F3D without having to save it to disk first to avoid said IO penalty.

There are ways to possibly do this (untested by issue author) by the use of tmpfs on linux platforms, and by passing /dev/stdout as the image filename.

There is also the possibility of using libf3d instead of the stand alone executable.

Here are some relevant comments from discord:

stream input is on the to-do list but requires changes in underlying libs so the ETA isn't entirely up to F3D (the changes are slowly but surely making their way into VTK as far as I know)
stream output should be doable on F3D's side

tracks that can be investigated without waiting for any new features:
using a ram disk, I'd be interested in seeing a benchmark of piping from stdout to stdin vs. writing and reading in a tmpfs directory (in theory piping should be better because it allows actual streaming, however in the case at hand we can't really render without buffering the whole input model and output image anyway so the streaming aspect won't matter and we just want fast/cheap IO)
using libf3d instead of the F3D application, for example through the python bindings, would let you read rendered images from memory and also allows to do multiple renders of the same model without reloading it

mwestphal:
This is already supported by the libf3d, but there is no mechanism to expose that in the F3D application yet.

The first question would be to identify how to do that in a cross platform way.

Meakk:
I confirm `--output /dev/stdout` works as intended.

snoyer:
As @mwestphal pointed out, @Meakk's newly merged `image::saveBuffer` can be used for this feature.

Here's a quick and dirty patch that could be a start for someone feeling like doing a proper PR:

```diff
diff --git a/application/F3DStarter.cxx b/application/F3DStarter.cxx
index 43a326c6..876b7f61 100644
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -14,6 +14,7 @@
 
 #include <cassert>
 #include <filesystem>
+#include <iostream>
 #include <set>
 
 namespace fs = std::filesystem;
[...]

PR #1326, opened 2024-03-22
Stream output file to stdout

Meakk:
Implement #1101

### DIFF
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -103,28 +103,28 @@ public:
     return false;
   }
 
-  static void SetVerboseLevel(const std::string& level)
+  static void SetVerboseLevel(const std::string& level, bool forceStdErr)
   {
     // A switch/case over verbose level
     if (level == "quiet")
     {
-      f3d::log::setVerboseLevel(f3d::log::VerboseLevel::QUIET);
+      f3d::log::setVerboseLevel(f3d::log::VerboseLevel::QUIET, forceStdErr);
     }
     else if (level == "error")
     {
-      f3d::log::setVerboseLevel(f3d::log::VerboseLevel::ERROR);
+      f3d::log::setVerboseLevel(f3d::log::VerboseLevel::ERROR, forceStdErr);
     }
     else if (level == "warning")
     {
-      f3d::log::setVerboseLevel(f3d::log::VerboseLevel::WARN);
+      f3d::log::setVerboseLevel(f3d::log::VerboseLevel::WARN, forceStdErr);
     }
     else if (level == "info")
     {
-      f3d::log::setVerboseLevel(f3d::log::VerboseLevel::INFO);
+      f3d::log::setVerboseLevel(f3d::log::VerboseLevel::INFO, forceStdErr);
     }
     else if (level == "debug")
     {
-      f3d::log::setVerboseLevel(f3d::log::VerboseLevel::DEBUG);
+      f3d::log::setVerboseLevel(f3d::log::VerboseLevel::DEBUG, forceStdErr);
     }
     else
     {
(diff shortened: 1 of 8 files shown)


Meakk:
@mwestphal @snoyer please take a look :)

application/F3DStarter.cxx
```diff
@@ -164,8 +165,16 @@ int F3DStarter::Start(int argc, char** argv)
...
+  const bool renderToStdout = this->Internals->AppOptions.Output == "-";
+  if (renderToStdout)
+  {
+    f3d::log::setVerboseLevel(f3d::log::VerboseLevel::ERROR);
+  }
+  else
+  {
+    // Set verbosity level early from command line
+    F3DInternals::SetVerboseLevel(this->Internals->AppOptions.VerboseLevel);
+  }
```
snoyer, line 177:
```suggestion
  std::streambuf* originalStdout = std::cout.rdbuf();
  if (renderToStdout)
  {
    std::cout.rdbuf(std::cerr.rdbuf());
  }

  // Set verbosity level early from command line
  F3DInternals::SetVerboseLevel(this->Internals->AppOptions.VerboseLevel);
```

application/F3DStarter.cxx
```diff
@@ -426,7 +438,15 @@ int F3DStarter::Start(int argc, char** argv)
...
+      if (renderToStdout)
+      {
+        const auto buffer = img.saveBuffer();
+        std::copy(buffer.begin(), buffer.end(), std::ostreambuf_iterator(std::cout));
+      }
+      else
+      {
+        img.save(this->Internals->AppOptions.Output);
+      }
       f3d::log::debug("Output image saved to ", this->Internals->AppOptions.Output);
```
snoyer, line 450:
```suggestion
      if (renderToStdout)
      {
        const auto buffer = img.saveBuffer();
        std::copy(buffer.begin(), buffer.end(), std::ostreambuf_iterator(originalStdout));
        f3d::log::debug("Output image data printed to stdout");
      }
      else
      {
        img.save(this->Internals->AppOptions.Output);
        f3d::log::debug("Output image saved to ", this->Internals->AppOptions.Output);
      }
```

snoyer (commented):
The key point here is that we want to ensure _nothing_ has been written to `stdout` before we use it to output render data, otherwise we will quietly output a corrupted PNG.

Setting the log level to `ERROR` early seemed to do a good enough job when I suggested it but I haven't fully tested it.
Possible issues include:
- if `ERROR` level messages go to `stdout` _and_ can we have non-critical error messages (ie. not exceptions and we still proceed to render)
- if there's any code that output to `stdout` directly (most likely external code not going through `f3d::log`)
- if there are cases where you'd want to render to stdout _and_ still see log messages somehow

Maybe it'd be better to not mess with the verbosity levels and instead _redirect_ `stdout` to `stderr` from the start of the app if we know we'll be rendering to stdout?
That way any logging would still get to the terminal through `stderr` and only the PNG data would go to `stdout` (`f3d foo.bar --verbose --output=- > /tmp/f3d.png` would output both the full log and a valid PNG).
Would that be just a dirty hack? Are there any critical implications I'm not seeing?

application/F3DStarter.cxx
mwestphal: interesting suggestion, how do other cli software do it ?

Meakk: Honestly I think we shouldn't even do anything about the verbose level automatically. Let's let the user specify the right verbosity level manually?
About external libraries potentially writing to std::cout, it's not our problem and we can still provide patches upstream if needed?

Meakk: And the default verbosity behavior is only writing in std::cerr anyway?

snoyer: > Honestly I think we shouldn't even do anything about the verbose level automatically.

That part definitely,  let's not change it magically.

> Let's let the user specify the right verbosity level manually?
> About external libraries potentially writing to std::cout, it's not our problem and we can still provide patches upstream if needed?

These 2 I guess can be argued both ways, we could do nothing and let the users find out if their output get corrupted, then either tell them they did it wrong in the first case, or go patch and rebuild in the second case.
Redirecting would be the "safe" way but I can't say if it's a good/nice/clean thing to do, or if anybody else does it that way as @mwestphal asked...

> And the default verbosity behavior is only writing in std::cerr anyway?

I haven't checked so I'll trust you on that, in which case it's fine as long as we trust libraries (looking at you OCCT...) to not randomly print directly.

Meakk: Redirect to stderr should be easy. We just have to override that: vtkOutputWindow::GetDisplayStream
So let's decide. I'm not against the idea.

mwestphal: that sounds like a good idea to me.

library/public/log.h
```diff
@@ -101,8 +101,10 @@ class F3D_EXPORT log
 
   /**
    * Set the verbose level.
+   * By default, only warnings and errors are written to stderr, debug and info are written to
+   * stdout. If forceStdErr is true, all messages including debug and info are written to stderr.
    */
-  static void setVerboseLevel(VerboseLevel level);
+  static void setVerboseLevel(VerboseLevel level, bool forceStdErr = false);
```
mwestphal, line 107: Does adding another staticc method makes sense ?

Meakk: That was my initial plan but it is calling the same function so exposing the display stream mode makes more sense imho

mwestphal: right I get your point.

mwestphal (approved):

PR #1328, opened 2024-03-27
Link app with pthread

mwestphal:


application/CMakeLists.txt
```diff
@@ -30,6 +30,10 @@ endif()
 add_executable(f3d ${F3D_SOURCE_FILES})
 target_link_libraries(f3d PUBLIC libf3d)
 
+if(UNIX)
```
Meakk, line 33: is it required on macos?

application/CMakeLists.txt
```diff
@@ -30,6 +30,10 @@ endif()
...
+  target_link_libraries(f3d PRIVATE pthread)
```
Meakk, line 34: I think the recommended method is:

```suggestion
  find_package(Threads REQUIRED)
  target_link_libraries(f3d PRIVATE Threads::Threads)
```

application/CMakeLists.txt
mwestphal: I do not know if it is needed for us but it may be needed in some cases: https://stackoverflow.com/questions/17841140/os-x-clang-pthread

application/CMakeLists.txt
mwestphal: https://stackoverflow.com/questions/5395309/how-do-i-force-cmake-to-include-pthread-option-during-compilation

indeed

Meakk (approved):

PR #1340, opened 2024-03-30
Alembic animation support

Meakk:


mwestphal (changes requested):
needs a data change but looks fine other then that :)

mwestphal (approved):

PR #1343, opened 2024-03-31
Generate BundleInfo.plist for macos

mwestphal:
A rather simple but a bit naive implementation for generating BundleInfo.plist during generation.

We could instead reparse the JSON but since this is quite simple, I've done it like that.

Also plugin developers maybe be happy to have access to this string if ever needed.


CMakeLists.txt
```diff
@@ -111,7 +111,7 @@ message(STATUS "VTK ${VTK_VERSION} found")
 # Shared options between application and library
 include(GNUInstallDirs)
 cmake_dependent_option(F3D_WINDOWS_GUI "Build a non-console Win32 application" ON "WIN32" OFF)
-cmake_dependent_option(F3D_MACOS_BUNDLE "Build a macOS bundle application" ON "APPLE" OFF)
+cmake_dependent_option(F3D_MACOS_BUNDLE "Build a macOS bundle application" ON "UNIX" OFF)
```
Meakk, line 114: ?

plugins/CMakeLists.txt
```diff
@@ -40,3 +40,7 @@ if (F3D_PLUGIN_BUILD_VDB)
     message(FATAL_ERROR "VDB plugin requires VTK >= 9.2.20220714")
   endif()
 endif()
+
+if(F3D_MACOS_BUNDLE)
```
Meakk, line 44: Should we put this variable as an internal cache variable instead of moving it from parent to parent?

Meakk: Or a global property if you prefer

mwestphal: indeed, global property would be better

cmake/f3dPlugin.cmake
```diff
@@ -370,4 +394,9 @@ macro(f3d_plugin_build)
   install(FILES "${F3D_PLUGIN_JSON_FILE}"
       DESTINATION "share/f3d/plugins"
       COMPONENT plugin)
+
+  if(F3D_MACOS_BUNDLE)
+    set_property(GLOBAL APPEND_STRING PROPERTY  F3D_MACOS_BUNDLE_XML ${F3D_MACOS_BUNDLE_XML_PLUGIN})
```
Meakk, line 399: What if the config step is run twice? Shouldn't we reset it somewhere?

mwestphal: global property are not stored in cache afaik, thet are just global.

Meakk: So that means it's empty every time cmake is configured?

mwestphal: yes

Meakk (approved):

PR #1347, opened 2024-04-01
Improve Alembic performance

Meakk:
Also restore the `RelWithDebInfo` since it's useful for performance analysis

On a large Alembic file, opening performance drops from 5.6 seconds to 530 ms.


mwestphal:
should go into release imo

mwestphal (approved):

PR #1377, opened 2024-04-14
Fix incorrect usage of internal variable PACKAGE_PREFIX_DIR

mwestphal:
See https://discourse.cmake.org/t/cmake-v3-29-1-regression-up-to-cmake-3-29-is-fine/10628/7 for more info about the "regression" in CMake and why this is needed.


mwestphal:
@thierry-FreeBSD please take a look :)

Meakk (approved):

Issue #792, opened 2023-04-29
Cell coloring fails on macOS


mwestphal:
**Describe the bug**
Cell coloring fails on macOS

**To Reproduce**
 0. Download sphere.vtp
 1. Open the file using `f3d --dry-run sphere.vtp`, S, C
 
![a](https://user-images.githubusercontent.com/3129530/235308499-2bd4fd9e-5056-402f-b9c7-a2807aecc74f.png)

**Expected behavior**
![a](https://user-images.githubusercontent.com/3129530/235308520-05c2e4f7-9797-4743-a923-587e1b286a80.png)

**System Information:**
 - OS: macOS
 - GPU and GPU driver: reproduced in a VM and in the github image

**F3D Information**
Paste the content of `f3d --version`: >2.0

**Additional context**
Sister VTK issue: https://gitlab.kitware.com/vtk/vtk/-/issues/18969

Meakk:
Seems to work fine here (macos arm)

![sphere](https://github.com/f3d-app/f3d/assets/6069752/3e70f613-0c8e-4483-92a1-62d6b0fc5eee)

mwestphal:
Can you try f3d.stp ?

mwestphal:
Turns out to be CI/VM specific

PR #1380, opened 2024-04-15
OCCT: Cleanup F3D_PLUGIN_OCCT_COLORING_SUPPORT

mwestphal:
With OCCT 7_8_0, Coloring is always supported, so this hides the CMake option to enable/disable the feature.

However, since it is not supported on macOS because of a bug, I set it to OFF in that case. Not enterely sure this is the right logic though.

https://github.com/f3d-app/f3d/issues/792


michaelmigliore (approved):
looks reasonable to me

Meakk (approved):

PR #1381, opened 2024-04-17
occt: Simplify static linking

mwestphal:


plugins/occt/module/CMakeLists.txt
```diff
@@ -16,9 +16,12 @@ vtk_module_include(f3d::vtkextOCCT
   PRIVATE
     $<BUILD_INTERFACE:${OpenCASCADE_INCLUDE_DIR}>)
 
-vtk_module_link(f3d::vtkextOCCT
-  PRIVATE
-    ${OpenCASCADE_LIBRARIES})
+set(_occt_link_libs "${OpenCASCADE_LIBRARIES}")
+if(CMAKE_LINK_GROUP_USING_RESCAN_SUPPORTED)
```
Meakk, line 20: change the minimum CMake version for `occt` plugin to 3.24

Meakk (changes requested):

plugins/occt/module/CMakeLists.txt
mwestphal: I dont think we want to do that. 3.24 is quite new, it is not strictly required to build the occt plugin.

mwestphal: And in anycase that check is needed, because rescan is a GCC linker thing only

Meakk: ok but then `CMAKE_LINK_GROUP_USING_RESCAN_SUPPORTED` won't be defined

Meakk: > ok but then `CMAKE_LINK_GROUP_USING_RESCAN_SUPPORTED` won't be defined

That would still work if undefined, but I believe we are checking if variables are undefined?

mwestphal: `if(VAR)` is not warning if VAR is not defined.

Meakk: Ok then

Issue #1194, opened 2024-01-20
Screenshot Interaction


snoyer:
Proposed feature:

Provide a key binding to render the current view to a file from within the app.

F3D already has an `--output <png file>` option to render as an image to a user-defined location, so implementing this feature **should** :crossed_fingers: be a matter of rewiring that to a key-binding and generating an appropriate filename.

Basic requirements:

- the destination filename should be computed from a configurable string template, eg. `F3D-{timestamp}.png`
- the default filename template should include the F3D version and OS (eg. `F3D-{version}-{os}-{timestamp}.png`) so users are able to communicate their F3D version and OS by just pressing the screenshot key and sending the image, without having to run `f3d --version` in a terminal..
- users must be able to find the file without having to know where the executable lives, saving it to the home directory seems sensible

Open questions:

- `{timestamp}` could mostly be a unique identifier to avoid overwriting previous images but in which case it does not _need_ to be a human readable date/time. Would be good if it was lexicographically sortable though. Maybe base64 unix timestamp could be nice and short? if using a human readable format it would be best to pick one that wouldn't leak timezone and OS locale info when sharing the file.
- instead, or in addition to, a timestamp should we check if a file of a given name exists and append some kind of counter to make the new name unique? (`F3D-blah.png`, `F3D-blah(2).png`, `F3D-blah(3).png`, ...)
- what should the string template syntax be? `{foo}`, `$(bar)`, `%baz`? 
- should the full image destination be configurable, or just the filename (with the directory hardcoded)?
- if the full destination is configurable in the filename template (eg, `{TEMP}/F3D-{timestamp}.png`, `{HOME}/F3D-{version}-{os}-{timestamp}.png`) how do we deal with path separators (use `/` even on Windows? always accept both `/` and `\`)?
- do we need more template variables? eg. `{filename}` so that `F3D-{filename}-{timestamp}.png` would expand to `F3D-my_model.glb-1234.png`

mwestphal:
any news on that @snoyer ? :)

Issue #596, opened 2023-02-01
Dump "camera" info should dump information about all options


mwestphal:
**Context**

F3D support printing information about the current state of the application by pressing ?, specfically about the configuration of the camera. While useful, this feature could become great if it was more rafined. With the addition of the commands, we can now do this

**Suggested Solution**
 - Modify `print_scene_info` to only call `PrintSceneDescription`
 - Add a `print_coloring_info` that calls `PrintColoringDescription`
 - Add a `print_mesh_info` that calls `DisplayImporterDescription`
 - Add a `print_options_info` that print the values of all options the same way they are printed when reading config file
 - Remove the `?` biding
 - Update doc about commands and bindings
 - Add testing using `SCRIPT`  tests

mwestphal:
Great ideas ! I wonder if we can manage cross platform clipboard. Afaik VTK does not support it.

mwestphal:
Updated with a suggested solution

mwestphal:
This is becoming possible with https://github.com/f3d-app/f3d/pull/1580.

Possible formatting suggested by @snoyer :

```
Updating Options:
  camera-direction: '-1,-0.5,-1' # # from 05_all.json₍₁₎:`.*`
  config: 'config_build' # from CLI options
  input: '/home/glow/dev/f3d/f3d/src/testing/data/dragon.vtu' # from CLI options
  axis: 'true' # from 05_all.json₍₁₎:`.*`
  scalars: '' # from 05_all.json₍₁₎:`.*`
  output: '/home/glow/dev/f3d/f3d/build/Testing/Temporary/TestDefaultConfigFileVTU.png' # from CLI options
  ref: '/home/glow/dev/f3d/f3d/src/testing/baselines/TestDefaultConfigFileVTU.png' # from CLI options
  anti-aliasing: 'true' # from 05_all.json₍₁₎:`.*`
  tone-mapping: 'true' # from 05_all.json₍₁₎:`.*`
  translucency-support: 'true' # from 05_all.json₍₁₎:`.*`
  grid: 'true' # from 05_all.json₍₁₎:`.*`
  hdri-ambient: 'true' # from 05_all.json₍₁₎:`.*`
  resolution: '300,300' # from CLI options
  animation-progress: 'true' # from 05_all.json₍₁₎:`.*`
  filename: 'true' # from 05_all.json₍₁₎:`.*`
  progress: 'true' # from 05_all.json₍₁₎:`.*`
[...]

PR #1391, opened 2024-04-24
add screenshot feature

snoyer:
Screenshot interaction feature as per #1194. Draft PR mostly because I'm not sure how to go about the testing.


application/F3DStarter.cxx
```diff
@@ -695,6 +703,141 @@ void F3DStarter::Render()
...
+    .setMetadata("camera", cameraMetadata.str())
+    .save(path, f3d::image::SaveFormat::PNG);
+
+  f3d::options& options = this->Internals->Engine->getOptions();
+  const std::string light_intensity_key = "render.light.intensity";
+  const double intensity = options.getAsDouble(light_intensity_key);
+  options.set(light_intensity_key, intensity * 5);
+  this->Render();
+  options.set(light_intensity_key, intensity);
+  this->Render();
```
snoyer, line 838: This flashes the model white for 1 frame to provide visual feedback.
Not sure if it's the right thing to do? if it is we can do it with a properly timed loop instead, otherwise we can just ditch it

application/F3DStarter.cxx
```diff
@@ -286,6 +287,13 @@ int F3DStarter::Start(int argc, char** argv)
           }
           return true;
         }
+
+        if (keySym == "F12")
```
mwestphal, line 291: any specific reason to put that in the app and not in the lib ?

snoyer: 100% because it was easier that way tbh :D
But it could go on the lib side instead, yes. The output filename templating could be reused by `--output` too possibly.

application/F3DOptionsParser.cxx
```diff
@@ -314,6 +314,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp0, "watch", "", "Watch current file and automatically reload it whenever it is modified on disk", appOptions.Watch,  HasDefault::YES, MayHaveConfig::YES );
     this->DeclareOption(grp0, "load-plugins", "", "List of plugins to load separated with a comma", appOptions.Plugins, LocalHasDefaultNo, MayHaveConfig::YES, "<paths or names>");
     this->DeclareOption(grp0, "scan-plugins", "", "Scan standard directories for plugins and display available plugins (result can be incomplete)");
+    this->DeclareOption(grp0, "screenshot-filename", "", "Screenshot filename", appOptions.ScreenshotFilename, LocalHasDefaultNo, MayHaveConfig::YES, "<filename>");
```
mwestphal, line 317: improve the doc and mention that its specific to the screenshot binding.

application/F3DOptionsParser.cxx
```diff
@@ -527,6 +528,11 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     f3d::log::error("Error parsing options: ", ex.what());
     throw;
   }
+
+  if (appOptions.ScreenshotFilename.empty())
+  {
+    appOptions.ScreenshotFilename = "{app}/{model}.png";
+  }
```
mwestphal, line 535: not sure why we would need this ?

application/F3DStarter.cxx
```diff
@@ -700,6 +708,123 @@ void F3DStarter::Render()
   f3d::log::debug("Render done");
 }
 
+//----------------------------------------------------------------------------
+void F3DStarter::SaveScreenshot(const std::string& filenameTemplate)
+{
+
+  const auto getScreenshotDir = []()
+  {
+    for (const char* const& candidate : { "XDG_PICTURES_DIR", "HOME", "USERPROFILE" })
```
mwestphal, line 717: Any source to justify writing in these locations ?

application/F3DStarter.cxx
```diff
@@ -700,6 +708,123 @@ void F3DStarter::Render()
...
+        {
+          return path;
+        }
+      }
+    }
+
+    return std::filesystem::current_path();
+  };
+
+  const auto applyFilenameTemplate = [&](const std::string& fn_template)
```
mwestphal, line 733: filenameTemplate

application/F3DStarter.cxx
```diff
@@ -700,6 +708,123 @@ void F3DStarter::Render()
...
+  {
+    std::string filename = fn_template;
+
+    {
+      std::regex re("\\{app\\}");
+      filename = std::regex_replace(filename, re, F3D::AppName);
+    }
+    {
+      std::regex re("\\{version\\}");
```
mwestphal, line 742: I dont see this being used except in the test, am I missing something ?

application/F3DOptionsParser.cxx
snoyer: it's the default value for the CLI argument but I don't know how to set it "properly"

application/F3DStarter.cxx
```diff
@@ -700,6 +708,123 @@ void F3DStarter::Render()
...
+        filename = std::regex_replace(filename, re, file.filename().string());
+      }
+    }
+
+    return std::filesystem::path(filename);
+  };
+
+  const auto findAvailableFilename = [&](const std::string& idealFilename)
+  {
+    const auto appendFilenameNumber = [](const std::filesystem::path& path, size_t n)
```
mwestphal, line 759: avoid single letter arg

application/F3DStarter.cxx
```diff
@@ -700,6 +708,123 @@ void F3DStarter::Render()
...
+    {
+      std::stringstream ss;
+      const std::string fn = path.string();
+      const std::string::size_type i = fn.find_last_of('.');
```
mwestphal, line 763: avoid single letter var :)

application/F3DStarter.cxx
```diff
@@ -700,6 +708,123 @@ void F3DStarter::Render()
...
+      }
+      else
+      {
+        ss << fn << "_" << n;
+      }
+      return ss.str();
+    };
+
+    std::string fn = idealFilename;
+    for (size_t i = 2; std::filesystem::exists(fn) && i < 1e8; ++i)
```
mwestphal, line 776: (in a short loop its fine)

mwestphal: why do we start at 2 though ?

application/F3DStarter.cxx
```diff
@@ -700,6 +708,123 @@ void F3DStarter::Render()
...
+    {
+      fn = appendFilenameNumber(idealFilename, i);
+    }
+    if (std::filesystem::exists(fn))
+    {
+      throw std::runtime_error("could not find available filenamne");
```
mwestphal, line 782: typo

application/F3DStarter.cxx
```diff
@@ -700,6 +708,123 @@ void F3DStarter::Render()
...
+      filename = std::regex_replace(filename, re, F3D::AppVersion);
+    }
+    {
+      std::regex re("\\{model\\}");
+      if (std::regex_search(filename, re))
+      {
+        const auto file = this->Internals->FilesList[this->Internals->CurrentFileIndex];
+        filename = std::regex_replace(filename, re, file.filename().string());
+      }
+    }
```
mwestphal, line 752: how about adding support for file numbering only when using a `{n}` semantic ?

application/F3DStarter.cxx
snoyer: not really :(
I know for example Gnome saves screenshots to `~/Pictures/Screenshot` which would be equivalent to `$XDG_PICTURES_DIR/Screenshots`. `XDG_PICTURES_DIR` may not be really standard so `HOME` would be a safer fallback. `USERPROFILE`: I'm winging it hoping it would work on Windows

application/F3DStarter.cxx
```diff
@@ -700,6 +708,123 @@ void F3DStarter::Render()
...
+
+    return std::filesystem::current_path();
+  };
+
+  const auto applyFilenameTemplate = [&](const std::string& fn_template)
+  {
+    std::string filename = fn_template;
+
+    {
+      std::regex re("\\{app\\}");
```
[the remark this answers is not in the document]
snoyer: I don't think so, I just pick `{foo}` arbitrarily for the templating syntax

application/F3DStarter.cxx
```diff
@@ -700,6 +708,123 @@ void F3DStarter::Render()
...
+    cameraMetadata << "{\n";
+    cameraMetadata << "  \"pos\": " << vec3toJson(state.pos) << ",\n";
+    cameraMetadata << "  \"foc\": " << vec3toJson(state.foc) << ",\n";
+    cameraMetadata << "  \"up\": " << vec3toJson(state.up) << ",\n";
+    cameraMetadata << "  \"angle\": " << state.angle << "\n";
+    cameraMetadata << "}\n";
+  }
+
+  window.renderToImage()
+    .setMetadata("camera", cameraMetadata.str())
```
mwestphal, line 816: should we set this when using `--output` too ? that would be more coherent imo.

application/F3DStarter.h
```diff
@@ -43,6 +43,11 @@ class F3DStarter
    */
   void Render();
 
+  /**
+   * Trigger a render and save a screenshot to disk according to a filename template.
+   */
+  void SaveScreenshot(const std::string& filenameTemplate);
```
mwestphal, line 49: this doc should be expanded a bit.

application/F3DOptionsParser.cxx
```diff
@@ -527,6 +528,11 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     f3d::log::error("Error parsing options: ", ex.what());
     throw;
   }
+
+  if (appOptions.ScreenshotFilename.empty())
+  {
+    appOptions.ScreenshotFilename = "{app}/{model}.png";
```
mwestphal, line 534: this whole {} semantic will need to be documented.

testing/recordings/TestScreenshotUser.log
```diff
@@ -0,0 +1,11 @@
...
+ExposeEvent 0 599 0 0 0 0 0
+RenderEvent 0 599 0 0 0 0 0
+KeyPressEvent 916 614 0 0 1 F12 0
+RenderEvent 916 614 0 0 1 F12 0
+CharEvent 916 614 0 0 1 F12 0
+KeyReleaseEvent 916 614 0 0 1 F12 0
+KeyPressEvent 916 614 0 0 1 F12 0
+RenderEvent 916 614 0 0 1 F12 0
+CharEvent 916 614 0 0 1 F12 0
+KeyReleaseEvent 916 614 0 0 1 F12 0
```
mwestphal, line 11: missing EOL

testing/recordings/TestScreenshot.log
```diff
@@ -0,0 +1,11 @@
...
+ExposeEvent 0 599 0 0 0 0 0
+RenderEvent 0 599 0 0 0 0 0
+KeyPressEvent 916 614 0 0 1 F12 0
+RenderEvent 916 614 0 0 1 F12 0
+CharEvent 916 614 0 0 1 F12 0
+KeyReleaseEvent 916 614 0 0 1 F12 0
+KeyPressEvent 916 614 0 0 1 F12 0
+RenderEvent 916 614 0 0 1 F12 0
+CharEvent 916 614 0 0 1 F12 0
+KeyReleaseEvent 916 614 0 0 1 F12 0
```
mwestphal, line 11: missing EOL

application/F3DStarter.cxx
snoyer: yeah I had it because we had discussed having a special case for "troubleshooting" screenshots where the filename would include the OS and version and the screenshot would have an overlay with more debug info. But we don't have such overlay yet so that part of the feature can't happen yet.
I can remove that part until it's actually needed.

application/F3DStarter.cxx
snoyer: fat fingers :D

application/F3DStarter.cxx
snoyer: can be done yes, would probably need to be on by default then?

application/F3DStarter.cxx
snoyer: could do, the templating could also be used. Would that mean moving some code to the lib or is that all in the app side still?

application/F3DOptionsParser.cxx
mwestphal: can you just set it in the header ?

application/F3DStarter.cxx
snoyer: I'm not adding a number if the filename is already available (which would be `1`)

application/F3DStarter.cxx
mwestphal: - Windows: https://superuser.com/questions/1797103/where-did-my-screenshots-on-windows-10-get-stored-to
- MacOS: https://support.apple.com/en-us/102646
- Linux: https://wiki.archlinux.org/title/XDG_user_directories

Gnome seems to save in Screenshots, KDE directly in $XDG_PICTURES_DIR

There doesnt seems to be a standard. Lets do the best we can and document it thouroughly

application/F3DStarter.cxx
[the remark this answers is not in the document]
mwestphal: what do you think @Meakk ?

application/F3DStarter.cxx
mwestphal: to be fair if we go with this semantic, we could do nice things, to be discussed.

application/F3DStarter.cxx
mwestphal: `--output` is an application side option, so most of it should be in the app side.

application/F3DStarter.cxx
```diff
@@ -165,6 +166,92 @@ class F3DStarter::F3DInternals
...
+      }
+    }
+
+    {
+      /* try and find non-existing filename; must be done last, assumes full path */
+      std::regex re("\\{n\\}");
+      if (std::regex_search(filename, re))
+      {
+        std::string fn = std::regex_replace(filename, re, "1");
+        for (size_t i = 2; std::filesystem::exists(fn) && i < 1000000; ++i)
```
mwestphal, line 240: `1e6` instead ?

application/F3DStarter.cxx
```diff
@@ -165,6 +166,92 @@ class F3DStarter::F3DInternals
...
+        {
+          fn = std::regex_replace(filename, re, std::to_string(i));
+        }
+        if (std::filesystem::exists(fn))
+        {
+          throw std::runtime_error("could not find available unique filename");
```
mwestphal, line 246: Yeah, I dont think we can reach this one in a reasonnable time frame

application/F3DStarter.cxx
```diff
@@ -165,6 +166,92 @@ class F3DStarter::F3DInternals
...
+        }
+        filename = fn;
+      }
+    }
+
+    return { filename };
```
mwestphal, line 252: what does the {} do ?

application/F3DStarter.cxx
snoyer: `1e6` is a `double` tho, does that matter?
Alternatively the internet tells me C++14 can do `1'000'000`

application/F3DStarter.cxx
snoyer: brace initializer to avoid repeating the function type (I had no idea, static analysis told me to)

application/F3DStarter.cxx
```diff
@@ -700,6 +798,49 @@ void F3DStarter::Render()
...
+
+  std::filesystem::path pathTemplate = std::filesystem::path(filenameTemplate);
+  std::filesystem::path fullPathTemplate =
+    pathTemplate.is_absolute() ? pathTemplate : getScreenshotDir() / pathTemplate;
+  std::filesystem::path path = this->Internals->applyFilenameTemplate(fullPathTemplate.string());
+
+  std::filesystem::create_directories(std::filesystem::path(path).parent_path());
+  f3d::log::info("saving screenshot to " + path.string());
+
+  f3d::image img = this->Internals->Engine->getWindow().renderToImage();
```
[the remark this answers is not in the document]
snoyer: good point, will do

application/F3DStarter.cxx
mwestphal: nice, sounds good to me.

application/F3DOptionsParser.cxx
snoyer: it would be better but it would I don't think `DeclareOption()` can handle it currently https://github.com/f3d-app/f3d/blob/e748359f371a820694609145839f5d856d5ecc01/application/F3DOptionsParser.cxx#L159-L172

application/F3DStarter.cxx
snoyer: `1'000'000` won't fly with `clang-format` and `Standard: Cpp03`... need to upgrade the config

mwestphal: alright, lets keep 1000000

application/F3DOptionsParser.cxx
mwestphal: I'm not following. To me, it should work out of the box as the string case is the simplest one that do not require any code. Also another option is really similar, `verbose`, and is initalized to `info` in the header.

application/F3DOptionsParser.cxx
mwestphal: still need a doc improvement :)

application/F3DStarter.cxx
```diff
@@ -165,6 +166,195 @@ class F3DStarter::F3DInternals
     }
   }
 
+  void addOutputImageMetadata(f3d::image& image)
```
mwestphal, line 169: I wonder what other metadata we should add ? 
Once we adresse #596 , should we store everything ?

application/F3DStarter.cxx
```diff
@@ -165,6 +166,195 @@ class F3DStarter::F3DInternals
...
+  }
+
+  /**
+   * Substitute the following variables in a filename template:
+   * - `{app}`: application name (eg. `F3D`)
+   * - `{version}`: application version (eg. `2.4.0`)
+   * - `{version_full}`: full application version (eg. `2.4.0-abcdefgh`)
+   * - `{model}`: current model filename without extension (eg. `foo` for `/home/user/foo.glb`)
+   * - `{model.ext}`: current model filename with extension (eg. `foo.glb` for `/home/user/foo.glb`)
+   * - `{model_ext}`: current model filename extension (eg. `glb` for `/home/user/foo.glb`)
```
mwestphal, line 198: why not "just": `{model}` and `{ext}` ?

mwestphal (changes requested):
Small changes needed. Improving the doc would also be nice and help with the release notes  :)

application/testing/CMakeLists.txt
```diff
@@ -215,6 +215,31 @@ if(NOT F3D_MACOS_BUNDLE)
...
 endif()
 
+# Screenshot Interaction
+function(f3d_ss_test)
+  cmake_parse_arguments(F3D_SS_TEST "" "NAME;TEMPLATE;EXPECTED;DEPENDS" "ARGS" ${ARGN})
+  f3d_test(NAME TestScreenshot${F3D_SS_TEST_NAME} DATA suzanne.ply ARGS --screenshot-filename=${F3D_SS_TEST_TEMPLATE} --dry-run --interaction-test-play=${F3D_SOURCE_DIR}/testing/recordings/TestScreenshot.log NO_BASELINE DEPENDS TestClearScreenshots)
+  f3d_test(NAME TestScreenshot${F3D_SS_TEST_NAME}File DATA suzanne.ply ARGS --ref=${F3D_SS_TEST_EXPECTED} DEPENDS TestScreenshot${F3D_SS_TEST_NAME} TestScreenshot${F3D_SS_TEST_DEPENDS} NO_BASELINE LONG_TIMEOUT)
+endfunction()
+
+if(F3D_TESTING_ENABLE_LONG_TIMEOUT_TESTS)
```
mwestphal, line 225: Here these screenshot actually long ? Since you are not using INTERACTION, you could put all these tests in normal testing.

application/F3DStarter.cxx
snoyer: My concerns on this one are:
- the file should go somewhere that's easy to find for regular users (home, desktop, ...)
- I don't want to try and independently replicate OS behaviors (unless there's a straightforward way)

That's why so far it tries:
- `XDG_PICTURES_DIR` first, that would be something like `$HOME/Pictures` or whatever, localized as per user settings, looks like the best option
- otherwise `HOME` or `USERPROFILE` to still land in a user owned location regardless of OS
- last resort fallback to current directory, no idea where that would be depending on the OS and how the app is started. Hopefully it doesn't come to that in a regular setup

Open to other options if you can think of any

mwestphal: I agree that there doesnt seem to be real standards and your approach is fine.

As always, just need good docs :)

application/F3DStarter.cxx
snoyer: because as soon as we release it with `{ext}` for the model's extension we'll find out we need a variable for another extension (hdri or something) :D

application/F3DStarter.cxx
mwestphal: Very good point I did not consider. In any case, out of scope of this PR.

application/F3DStarter.cxx
mwestphal: good point. fine as it is, but @Meakk may have an input here.

application/F3DOptionsParser.cxx
snoyer: You're right, I tried it at the beginning and it didn't work so I blamed `DeclareOption()` and moved on...
turns out with the right parameters it works, my bad

application/F3DOptionsParser.cxx
```diff
@@ -314,6 +314,7 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
     this->DeclareOption(grp0, "watch", "", "Watch current file and automatically reload it whenever it is modified on disk", appOptions.Watch,  HasDefault::YES, MayHaveConfig::YES );
     this->DeclareOption(grp0, "load-plugins", "", "List of plugins to load separated with a comma", appOptions.Plugins, LocalHasDefaultNo, MayHaveConfig::YES, "<paths or names>");
     this->DeclareOption(grp0, "scan-plugins", "", "Scan standard directories for plugins and display available plugins (result can be incomplete)");
+    this->DeclareOption(grp0, "screenshot-filename", "", "Screenshot filename", appOptions.ScreenshotFilename,  HasDefault::YES, MayHaveConfig::YES, "<filename>");
```
Meakk, line 317: extra space:

```suggestion
    this->DeclareOption(grp0, "screenshot-filename", "", "Screenshot filename", appOptions.ScreenshotFilename, HasDefault::YES, MayHaveConfig::YES, "<filename>");
```

application/F3DStarter.cxx
Meakk: that's fine to me

application/F3DStarter.cxx
```diff
@@ -165,6 +166,195 @@ class F3DStarter::F3DInternals
...
+        else
+        {
+          joined << fragment;
+        }
+      }
+      return joined.str();
+    };
+
+    /* apply incrementing numbering until file doesn't exist already */
+    for (size_t i = 1; i < 1000000; ++i)
```
Meakk, line 347: Let's document the limit

doc/user/INTERACTIONS.md
```diff
@@ -91,3 +92,10 @@ component is found.
 When changing the type of data to color with, the index of the array within the data will be kept if valid
 with the new data. If not, it will cycle until a valid array is found. After that, the component will be checked
 as specified above.
+
+## Taking Screenshots
+
+The destination filename used to save the screenshots (created by pressing <kbd>F12</kbd>) is configurable and can use template variables as described [on the options page](OPTIONS.md#filename-templating).
```
mwestphal, line 98: is configurable (using the `screenshot-filename` option)

doc/user/INTERACTIONS.md
```diff
@@ -91,3 +92,10 @@ component is found.
...
+
+Unless the configured filename template is an absolute path, images will be saved into the user's home directory
+(using the following environment variables: `XDG_PICTURES_DIR`, `HOME`, or `USERPROFILE`).
```
mwestphal, line 101: variables, in that order, :

doc/user/OPTIONS.md
```diff
@@ -18,6 +18,7 @@ Options|Default|Description
 \-\-watch||Watch current file and automatically reload it whenever it is modified on disk.
 \-\-load-plugins=\<paths or names\>||List of plugins to load separated with a comma. Official plugins are `alembic`, `assimp`, `draco`, `exodus`, `occt`, `usd`, `vdb`. See [usage](USAGE.md) for more info.
 \-\-scan-plugins||Scan standard directories for plugins and display their names, results may be incomplete. See [usage](USAGE.md) for more info.
+\-\-screenshot-filename||[Screenshot](INTERACTIONS.md#taking-screenshots) filename template. Can use [template variables](#filename-templating).
```
mwestphal, line 21: Filename to save [screenshot](INTERACTIONS.md#taking-screenshots) to.

mwestphal, line 21:
```suggestion
\-\-screenshot-filename||[Screenshot](INTERACTIONS.md#taking-screenshots) filename template. Can use [template variables](#filename-templating).
```

mwestphal (changes requested):
small changes needed

doc/user/OPTIONS.md
```diff
@@ -142,3 +143,23 @@ Some rendering options are not compatible between them, here is the precedence o
...
+
+- `{app}`: application name (ie. `F3D`)
+- `{version}`: application version (eg. `2.4.0`)
+- `{version_full}`: full application version (eg. `2.4.0-abcdefgh`)
+- `{model}`: current model filename without extension (eg. `foo` for `/home/user/foo.glb`)
+- `{model.ext}`: current model filename with extension (eg. `foo.glb` for `/home/user/foo.glb`)
+- `{model_ext}`: current model filename extension (eg. `glb` for `/home/user/foo.glb`)
+- `{date}`: current date in YYYYMMDD format
+- `{date:format}`: current date as per C++'s `std::put_time` format
+- `{n}`: auto-incremented number to make filename unique
```
mwestphal, line 159: (up to 1000000)

application/F3DStarter.cxx
```diff
@@ -165,6 +166,196 @@ class F3DStarter::F3DInternals
...
+   * - `{model}`: current model filename without extension (eg. `foo` for `/home/user/foo.glb`)
+   * - `{model.ext}`: current model filename with extension (eg. `foo.glb` for `/home/user/foo.glb`)
+   * - `{model_ext}`: current model filename extension (eg. `glb` for `/home/user/foo.glb`)
+   * - `{date}`: current date in YYYYMMDD format
+   * - `{date:format}`: current date as per C++'s `std::put_time` format
+   * - `{n}`: auto-incremented number to make filename unique
+   * - `{n:2}`, `{n:3}`, ...: zero-padded auto-incremented number to make filename unique
+   */
+  std::filesystem::path applyFilenameTemplate(
+    const std::string& templateString, size_t maxNumberingAttempts = 1000000)
```
mwestphal, line 205: not a big fan of having a param that we do not "use" apart the default value. should be a const expr instead.

doc/user/OPTIONS.md
snoyer: do we need "Path to a filename to save screenshots to"? or can we leave it at "filename to save screenshots to"?

doc/user/OPTIONS.md
snoyer: does it matter in the user documentation? The limit is only there because it's bad form to have unbound loops really. It's extremely unlikely to show up in nominal use.

doc/user/OPTIONS.md
mwestphal: fine by me

doc/user/OPTIONS.md
mwestphal: imo better safe than sorry. I think this is what @Meakk meant regarding documenting this though.

doc/user/OPTIONS.md
```diff
@@ -18,6 +18,7 @@ Options|Default|Description
 \-\-watch||Watch current file and automatically reload it whenever it is modified on disk.
 \-\-load-plugins=\<paths or names\>||List of plugins to load separated with a comma. Official plugins are `alembic`, `assimp`, `draco`, `exodus`, `occt`, `usd`, `vdb`. See [usage](USAGE.md) for more info.
 \-\-scan-plugins||Scan standard directories for plugins and display their names, results may be incomplete. See [usage](USAGE.md) for more info.
+\-\-screenshot-filename=\<png file\>||Filename to save [screenshots](INTERACTIONS.md#taking-screenshots) to. Can use [template variables](#filename-templating).
```
mwestphal, line 21: still missing the default value between the `||`.

mwestphal (approved):

Meakk (approved):

PR #1406, opened 2024-04-27
CI: Update and fix coverage workflow

mwestphal:
- Fix coverage action by specifying a codecov token
 - Update codecov action to v4 and adapt configuration

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,4 +1,5 @@
 codecov:
+  token: 7d2ecd8b-ac89-4c2d-ad05-119518130cb2
   notify:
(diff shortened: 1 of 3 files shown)


codecov.yml
```diff
@@ -1,4 +1,5 @@
 codecov:
+  token: 7d2ecd8b-ac89-4c2d-ad05-119518130cb2
```
mwestphal, line 2: visible token is required for fork workflow. I dont think we care that much if other starts pushing to our codecov.

Meakk (approved):

PR #1409, opened 2024-04-28
CI: Fix coverage again

mwestphal:
Here is the corrected version of codecov:

 - forked workflow use tokenless upload
 - non-forked workflow uses secrect token
 - the token is empty when run in forks, but codecov action does not even look at it


mwestphal:
I'll update the coverage token after we merge this one.

Meakk (approved):

Issue #1014, opened 2023-09-30
Path contains double backslashes in the Windows log window


Meakk:
**Describe the bug**
When trying to open an unsupported file, the path contains double backslashes.

**To Reproduce**
 1. Open a unsupported file
 2. Look at the log window

![image](https://github.com/f3d-app/f3d/assets/6069752/b97ebcd4-3727-464c-a6f8-430ab57c5ddd)

**Expected behavior**
Single backslash

**System Information:**
 - OS: Windows

**F3D Information**
2.2 RC2

snoyer:
using `generic_string()` solves the issue in so far as the output does not contain double back slashes (`\\`) anymore but the separators have been converted to forward slashes ('/') which can be used on Windows but are not preferred.

@shambhupatil can you check if using `filePath.string()` instead of `filePath.generic_string()` produces single backslashes?

shambhupatil:
Yes, path.string() does produce single backslash.

mwestphal:
That sounds like the right fix!

snoyer:
> Yes, path.string() does produce single backslash.

@shambhupatil nice! next step is the PR then :)

PR #1428, opened 2024-05-17
Remove double backslashes from file path when opening unsupported files.

shambhupatil:
Closes [#1014](https://github.com/f3d-app/f3d/issues/1014)

### DIFF
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -846,7 +846,7 @@ void F3DStarter::LoadFile(int index, bool relativeIndex)
         }
         else
         {
-          f3d::log::warn(filePath, " is not a file of a supported file format\n");
+          f3d::log::warn(filePath.string(), " is not a file of a supported file format\n");
           filenameInfo += " [UNSUPPORTED]";
         }
       }


shambhupatil:
I think CIs are failing because the file path should be displayed in quotes.

snoyer:
> I think CIs are failing because the file path should be displayed in quotes.

Looks like it, yes. The failing test
```
1/1 Test #429: f3d::TestUnsupportedFileText .....***Failed  Required regular expression not found. Regex=[.*unsupportedFile.dummy" is not a file of a supported file format
```
is defined here: https://github.com/f3d-app/f3d/blob/6133d0fa6189061f9fea453441dfbe086ba95087/application/testing/CMakeLists.txt#L877

You'll have to either:
- change the code to output the quotes
- change the test to not expect the quotes

I can't tell you which one is the way to go but @mwestphal and @Meakk will

mwestphal:
You need to update the test, but it's a bit unclear to me where the quote was coming from in the first place.

shambhupatil:
Okay, I will look into why the quotes were there. Could you tell me how to update the tests?

mwestphal (approved):

PR #1431, opened 2024-05-20
add camera pan method

### LINKED ISSUE #1430: Add a Pan method
When having an external window there is not method to pan.

A Pan method where you pass the delta x and y to pan vertically or horizontally.

**Describe alternatives you've considered**
Calculate the new camera position and focal point in the external program.

**Additional context**
[Here](https://github.com/Nokse22/Exhibit/blob/3f6b2118d0365da4eeab1e8cc87d760cb5c54dbd/src/window.py#L204-L212) is the solution thanks to @snoyer that I have in my app.

snoyer:
fixes #1430

### DIFF
--- a/library/public/camera.h
+++ b/library/public/camera.h
@@ -57,6 +57,7 @@ public:
    * Angles are in degrees.
    */
   virtual camera& dolly(double val) = 0;
+  virtual camera& pan(double right, double up) = 0;
   virtual camera& roll(angle_deg_t angle) = 0;
   virtual camera& azimuth(angle_deg_t angle) = 0;
   virtual camera& yaw(angle_deg_t angle) = 0;
--- a/library/public/camera.h
+++ b/library/public/camera.h
@@ -57,7 +57,7 @@ public:
    * Angles are in degrees.
    */
   virtual camera& dolly(double val) = 0;
-  virtual camera& pan(double right, double up) = 0;
+  virtual camera& pan(double right, double up, double forward = 0) = 0;
   virtual camera& roll(angle_deg_t angle) = 0;
   virtual camera& azimuth(angle_deg_t angle) = 0;
   virtual camera& yaw(angle_deg_t angle) = 0;
--- a/library/testing/TestSDKCamera.cxx
+++ b/library/testing/TestSDKCamera.cxx
@@ -7,6 +7,7 @@
 #include <iomanip>
 #include <iostream>
 #include <limits>
+#include <sstream>
 
 // TODO these methods should be put in types.h at some point.
 // https://github.com/f3d-app/f3d/issues/361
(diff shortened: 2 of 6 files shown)


library/public/camera.h
```diff
@@ -57,6 +57,7 @@ class F3D_EXPORT camera
    * Angles are in degrees.
    */
   virtual camera& dolly(double val) = 0;
+  virtual camera& pan(double right, double up) = 0;
```
mwestphal, line 60: I wonder if dolly and pan should be separately documented to clarify the arguments

python/testing/test_camera.py
```diff
@@ -86,6 +86,21 @@ def test_moves():
     camera.pitch(angle)
 
 
+def test_pan():
```
mwestphal, line 89: nice, but needs a C++ test :)

mwestphal (changes requested):
small changes needed

library/public/camera.h
snoyer: would changing `dolly(double val)` to `dolly(double forward)` be self-explanatory enough?

python/testing/test_camera.py
snoyer: for some reason I thought python testing counted towards coverage

library/public/camera.h
mwestphal: definitely an improvement.

python/testing/test_camera.py
mwestphal: The logic is that python is an optional dependency, so python tests are testing the bindings themselves, not the feature, which must be tested using c++

library/public/camera.h
mwestphal: unrelated to that PR though

library/testing/TestSDKCamera.cxx
```diff
@@ -27,6 +28,31 @@ bool comparePoint(const f3d::point3_t& vec1, const f3d::point3_t& vec2)
...
+  if (!compareDouble(actual[0], expected[0]) || !compareDouble(actual[1], expected[1]) ||
+    !compareDouble(actual[2], expected[2]))
+  {
+    std::stringstream ss;
+    ss << label << ": ";
+    ss << std::setprecision(12);
+    ss << "(" << actual[0] << "," << actual[1] << "," << actual[2] << ")";
+    ss << " != ";
+    ss << "(" << expected[0] << "," << expected[1] << "," << expected[2] << ")";
+    throw testFailure(ss.str());
```
mwestphal, line 52: any specific reason to go the exceptions instead of conditions (apart from your personnal preference) ?

snoyer: feels like it's less noise to throw on failure and have a single `try {...} catch { std::cerr<< what(); return EXIT_FAILURE; } return EXIT_SUCCESS;` than to have a bunch of `if(!...){ std::cerr<< ... ; return EXIT_FAILURE; }`
...so personal preference mostly I guess

mwestphal (approved):

Meakk (approved):

library/public/camera.h
snoyer, line 60: I'll add a forward parameter for completeness
```suggestion
  virtual camera& pan(double right, double up, double forward = 0) = 0;
```

mwestphal (approved):

Meakk (approved):

PR #1434, opened 2024-05-24
Removed recurrent logs from getReader

shambhupatil:
Moved debug logs from getReader() to loadScene() and loadGeometry() to reduce the same logs displayed many times.

### DIFF
--- a/library/src/loader_impl.cxx
+++ b/library/src/loader_impl.cxx
@@ -224,8 +224,13 @@ loader& loader_impl::loadGeometry(const std::string& filePath, bool reset)
   }
 
   f3d::reader* reader = f3d::factory::instance()->getReader(filePath);
-  if (!reader)
+  if (reader)
   {
+    log::debug("The best reader for \"" + filePath + "\" is \"" + reader->getName() + "\"");
+  }
+  else
+  {
+    log::debug("No reader found for \"" + filePath + "\"");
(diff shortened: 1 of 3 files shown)


library/src/loader_impl.cxx
```diff
@@ -224,8 +224,13 @@ loader& loader_impl::loadGeometry(const std::string& filePath, bool reset)
   }
 
   f3d::reader* reader = f3d::factory::instance()->getReader(filePath);
-  if (!reader)
+  if (reader)
+  {
+    log::debug("The best reader for \"" + filePath + "\" is \"" + reader->getName() + "\"");
```
mwestphal, line 229: thinking back on it, maybe it should be something like:

"found a reader for filepath: reader->getName()"

wdyt ?

mwestphal:
not sure what is hapening with codespell, ill take a look

library/src/loader_impl.cxx
shambhupatil: Yeah, no need to display reader name then?

mwestphal:
alright I fixed codespell, can you rebase on the last master @shambhupatil ?

mwestphal (approved):

mwestphal:
all good to me, can I merge @shambhupatil ?

PR #1437, opened 2024-05-26
Add tkinter.py to examples

JPLost:
Example for tkinter and customtkinter gui building to python examples folder


JPLost:
Small change can be added to line 15.
self.mEngine = f3d.Engine(f3d.Window.Type.EXTERNAL) → self.mEngine = f3d.Engine(f3d.Window.EXTERNAL)
This should be the newer binding behavior for Python

mwestphal:
Thanks @JPLost !

@Meakk I'm not sure what is the right place for this code.

`examples/libf3d/python` ?

Other external window "examples" are in "library/testing" but I dont think we want that here.

Meakk:
> @Meakk I'm not sure what is the right place for this code.

I think `examples/libf3d/python` is perfectly fine

snoyer (commented):
Few comments, nothing critical.

I would also rename to something like `minimal-tkinter` so the name of the file doesn't conflict with the name of the existing `tkinter` module (also there may be additional `tkinter`-based examples in the future?)

JPLost:
Updated according to your recommendations @snoyer.

examples/libf3d/python/tkinter/minimal_tkinter.py
```diff
@@ -0,0 +1,37 @@
...
+        )
+
+    def redraw(self):
+        self.engine.window.render()
+
+if __name__ == "__main__":
+    # Create main window and define size, position and title
+    root = tk.Tk()
+    root.geometry('640x480+100+100')
+    root.title('LanPtr3D')
```
snoyer, line 31: missed that one on the first pass, the old prototype title should be replaced with something related to F3D and/or the example name

mwestphal (approved):

mwestphal:
@snoyer what is the status of this ? still need a reformating ?

snoyer:
> @snoyer what is the status of this ? still need a reformating ?

it's not fully `black` formatted but that's neither documented nor enforced by the CI so we can't say it _needs_ to be

mwestphal:
fair enough, we may need to add that to CI at some point.

Thanks for your contribution @JPLost !

PR #1442, opened 2024-05-27
handle multiple configs

snoyer:
Refactoring to handle multiple config files.

This is needed because currently if a user config file is found the default and plugins' configs are ignored.

I'm not 100% sure about the detailed logic implications of my changes in terms of config search paths and the like, however the technical aspect should be fine.

@mwestphal please have a look


application/F3DConfigFileTools.cxx
```diff
@@ -115,16 +119,15 @@ fs::path F3DConfigFileTools::GetConfigPath(const std::string& configSearch)
...
-          return configPath;
+          paths.emplace_back(configPath);
         }
       }
     }
-    f3d::log::debug("No configuration file for \"", configSearch, "\" found");
-    return {};
   }
-  catch (const fs::filesystem_error&)
+  catch (const fs::filesystem_error&) // TODO this try/catch should be inside the loop?
```
mwestphal, line 127: yes, it should

application/F3DOptionsParser.cxx
```diff
@@ -675,6 +716,7 @@ void ConfigurationOptions::PrintReadersList()
 //----------------------------------------------------------------------------
 bool ConfigurationOptions::InitializeDictionaryFromConfigFile(const std::string& config)
 {
+  this->GlobalConfigEntries.clear();
   this->RegexConfigEntries.clear();
```
mwestphal, line 720: why is this needed ?

application/F3DOptionsParser.cxx
```diff
@@ -279,7 +260,67 @@ void ConfigurationOptions::GetOptions(F3DAppOptions& appOptions, f3d::options& o
...
+      std::regex re(pattern, std::regex_constants::icase);
+      std::smatch matches;
+      if (std::regex_match(filePathForConfigBlock, matches, re))
+      {
+        f3d::log::debug("using `", pattern, "` config from ", source.string());
+        update(conf);
+      }
+    }
+
+    this->CurrentFileConfig = tmpConfig;
```
mwestphal, line 322: what does this contains ?

application/F3DOptionsParser.cxx
snoyer: there was a single global block that would just get assigned https://github.com/f3d-app/f3d/blob/16fc29abfb227e76109ba6593d56f324921ee055/application/F3DOptionsParser.cxx#L786
now there's a list that gets appended to like the other one so I clear it too

application/F3DOptionsParser.cxx
snoyer: it contains the final configuration key/values (merged from global blocks and matched regex blocks) so that `GetOptionConfig()` can lookup directly into that instead of looping through all the blocks and doing the same matching over and over for every key like it does now

application/F3DOptionsParser.cxx
mwestphal: to be fair I fail to see why `this->RegexConfigEntries.clear();` is even needed. Unless I'm missing something `InitializeDictionaryFromConfigFile` is called only once ?

application/F3DOptionsParser.cxx
mwestphal: got it

application/F3DOptionsParser.cxx
snoyer: Argument 1: Personally I would argue it makes sense to clear just in case. Or we could refactor the function so it creates both list within and returns them. Or we document to say it appends to the lists but does not clear them.
Either way just removing the clears "because we know" it's only called one seems sketchy?

Argument 2: ask this guy https://github.com/f3d-app/f3d/blame/9ac4f1c7c1723ebbd21f1afb9984f1fe56e7b3b5/application/F3DOptionsParser.cxx#L674 ;p

mwestphal: Fair enough, maybe add a comment as to why we clear and that it is not absolutely needed  :)

mwestphal:
Alright, this is looking good. Feel free to take this one step further with the CI. Not sure if this can be tested though.

mwestphal:
```
/__w/f3d/f3d/source/application/F3DOptionsParser.cxx:258:49: error: the parameter 'filePathForConfigBlock' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param,-warnings-as-errors]
  std::vector<std::string>& inputs, std::string filePathForConfigBlock, bool allOptionsInitialized,
  ```

File: application/F3DOptionsParser.cxx [hunk 4]
```diff
@@ -274,12 +255,60 @@ void ConfigurationOptions::PrintPluginsScan()
...
+        }
+        f3d::log::debug("+ ", add);
+      }
+
+      tmpConfig[key] = value;
+    }
+  };
+
+  /* apply global entries */
+  for (const auto& [conf, source, pattern] : this->GlobalConfigEntries)
```
mwestphal, line 290: `pattern` here is always `global` right ?

Not saying we should change stuff, just checking I'm following.

application/F3DOptionsParser.cxx
```diff
@@ -274,12 +255,60 @@ void ConfigurationOptions::PrintPluginsScan()
...
+  std::vector<std::string>& inputs, const std::string& filePathForConfigBlock,
+  bool allOptionsInitialized, bool parseCommandLine)
 {
   inputs.clear(); /* needed because this function is called multiple times */
 
-  this->FilePathForConfigBlock = std::move(filePathForConfigBlock);
+  /* start with an empty config ... */
+  ConfigDict tmpConfig;
+
+  const auto update = [&](const ConfigDict& config)
```
mwestphal, line 266: add some doc in this lambda ?

application/F3DOptionsParser.cxx
```diff
@@ -65,34 +65,14 @@ class ConfigurationOptions
...
-          if (localIt != it.second.end())
-          {
-            configValue = localIt->second;
-            ret = true;
-          }
-        }
-      }
-    }
-    return ret;
+    // TODO ensure to return the actual default value if option key is not found
```
mwestphal, line 74: this TODO doesnt seems necessary, since this call is made with the default value

testing/configs/invalid_value.json
```diff
@@ -0,0 +1,6 @@
+{
+    "global":
+    {
+      "verbose": {"foo": "bar"}
+    }
+}
```
mwestphal, line 6: missing eol

application/F3DOptionsParser.cxx
snoyer: yes, I kept the same tuple typing with the pattern but for the global blocks it's basically unused, I'll add it to the comment

application/F3DOptionsParser.cxx
snoyer: that's leftover from my attempt at handling `null` values, I forgot to remove it

application/F3DConfigFileTools.cxx
```diff
@@ -74,23 +74,27 @@ fs::path F3DConfigFileTools::GetBinaryResourceDirectory()
...
 #endif
 #ifdef __linux__
-    dirsToCheck.emplace_back("/etc/f3d");
-    dirsToCheck.emplace_back("/usr/share/f3d/configs");
+    "/etc/f3d",
+    "/usr/share/f3d/configs",
 #endif
-    dirsToCheck.emplace_back(F3DConfigFileTools::GetBinaryResourceDirectory() / "configs");
+    F3DConfigFileTools::GetBinaryResourceDirectory() / "configs",
+    F3DConfigFileTools::GetUserConfigFileDirectory(),
```
mwestphal, line 92: I wonder if we should provide some feedbacks about the directories that have been checked ?

application/F3DOptionsParser.cxx
```diff
@@ -693,49 +724,55 @@ bool ConfigurationOptions::InitializeDictionaryFromConfigFile(const std::string&
...
+      f3d::log::debug("Using config file ", configPath.string());
+      actualConfigFilePaths.emplace_back(configPath);
     }
   }
-  else
+
+  if (actualConfigFilePaths.empty())
   {
-    actualConfigFilePaths.emplace(configPath);
+    f3d::log::error("Configuration file for \"", config, "\" could not be found");
```
mwestphal, line 774: One think I dont like with this is that when not specifying any `--config`, then `config` is empty resulting into:

`f3d::log::error("Configuration file for \"", config, "\" could not be found");`

Maybe this `config` var should be initialized to `config` much higher up.

application/F3DConfigFileTools.cxx
snoyer: Could be done, maybe that should be another PR that also cleans up the search logic? (currently is seems to be split between `F3DConfigFileTools::GetConfigPath` and the top of `ConfigurationOptions::InitializeDictionaryFromConfigFile`)

application/F3DOptionsParser.cxx
snoyer: TBH I didn't try to fully understand the `--config` logic, as per other comment I think there would be some refactoring to do to separate the config file search and parsing. 
Unless it's a blocker I'd rather leave that for someone else to do.

application/F3DConfigFileTools.cxx
mwestphal: yes, I agree.

application/F3DOptionsParser.cxx
mwestphal: not a blocker

snoyer: did the minimum of avoiding the bad log, proper fix of the logic would need a wider refactoring as already discussed

mwestphal (approved):

PR #1444, opened 2024-05-29
C Bindings

parkerstafford:
Review for Kam


mwestphal:
FYI @Meakk

mwestphal:
A few warnings and errors to fix in CI:

```
 cc1: warning: command-line option '-Woverloaded-virtual' is valid for C++/ObjC++ but not for C
cc1: error: command-line option '-Wsuggest-override' is valid for C++/ObjC++ but not for C [-Werror]
In file included from /__w/f3d/f3d/source/f3d/library/public/image.h:4,
                 from /__w/f3d/f3d/source/f3d/library/src/image_c_api.c:2:
/__w/f3d/f3d/source/f3d/library/public/exception.h:4:10: fatal error: stdexcept: No such file or directory
    4 | #include <stdexcept>
      |          ^~~~~~~~~~~
cc1: all warnings being treated a
```

But it looks good so far!

library/testing/test_image_c_api.c
mwestphal, line 1: is this file compiled ?

Meakk: It does not seem so, it needs to be added in `library/testing/CMakeLists.txt`

library/public/image_c_api.h
```diff
@@ -0,0 +1,53 @@
+#ifndef F3D_C_API_H
+#define F3D_C_API_H
+
+#include "export.h" // Ensure F3D_EXPORT is defined
```
Meakk, line 4: Is this type of comment valid in C? Or should you do something like
```suggestion
#include "export.h" /* Ensure F3D_EXPORT is defined */
```

library/public/image_c_api.h
```diff
@@ -0,0 +1,53 @@
...
+
+typedef struct {
+    float x;
+    float y;
+    float z;
+} f3d_vector3_t;
+
+typedef struct f3d_image f3d_image_t;
+
+F3D_EXPORT f3d_image_t* f3d_image_new(void);
```
Meakk, line 24: Let's add some comments, could be an adapted copy from `image.h`

library/src/image_c_api.c
```diff
@@ -0,0 +1,166 @@
+#include "image_c_api.h"
+#include "image.h"
+#include <cstring>
+
+struct f3d_image {
+    f3d::image img;
+};
+
+f3d_image_t* f3d_image_new(void) {
```
Meakk, line 9: I think this file (which includes and uses `image.h`) must be a C++ file. Only `image_c_api.h` have to be C only (and the unit test of course).

mwestphal:
hey @parkerstafford , need any help moving forward ?

mwestphal:
superseeded by https://github.com/f3d-app/f3d/pull/1892

PR #1446, opened 2024-05-30
Make Windows surface DPI aware

Meakk:
Here's a comparison with 150% scaling (left: release, right: this branch)
image

### DIFF
--- a/application/main.cxx
+++ b/application/main.cxx
@@ -51,6 +51,10 @@ int WINAPI wWinMain(HINSTANCE, HINSTANCE, LPWSTR, int)
 
   std::setlocale(LC_ALL, "en_US.utf8");
 
+#if WINVER >= 0x0605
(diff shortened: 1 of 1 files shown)


library/src/window_impl.cxx
```diff
@@ -153,6 +153,10 @@ class window_impl::internals
 window_impl::window_impl(const options& options, Type type)
   : Internals(std::make_unique<window_impl::internals>(options))
 {
+#if defined(_WIN32) && (WINVER >= 0x0605)
+  SetProcessDpiAwarenessContext(DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2);
```
mwestphal, line 157: maybe only for native ? I dont think its our responsability to set that otherwise ?

mwestphal (approved):

PR #1449, opened 2024-06-03
Improve WASM app

Meakk:
- Fix build guide with Windows
- Do not try to cache baked HDRI when running on WASM
- Add SSAO, HDRI ambient to rendering options
- Add dark/light mode switch

### DIFF
--- a/doc/dev/BUILD_WASM.md
+++ b/doc/dev/BUILD_WASM.md
@@ -4,7 +4,7 @@ F3D can be built in WebAssembly using emscripten in order to embed it into a web
 It is still experimental and only a small fraction of the libf3d public API is bound.
 An example can be seen at https://f3d.app/web
 
-This guide is describing how to build VTK and F3D with emscripten using Docker on Linux or Windows (WSL).
+This guide is describing how to build VTK and F3D with emscripten using Docker on Linux or Windows.
 
 # Building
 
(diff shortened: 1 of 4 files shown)


doc/dev/BUILD_WASM.md
```diff
@@ -4,7 +4,7 @@ F3D can be built in WebAssembly using emscripten in order to embed it into a web
 It is still experimental and only a small fraction of the libf3d public API is bound.
 An example can be seen at https://f3d.app/web
 
-This guide is describing how to build VTK and F3D with emscripten using Docker on Linux.
+This guide is describing how to build VTK and F3D with emscripten using Docker on Linux or Windows (WSL).
```
mwestphal, line 7: WSL is not Windows, not sure its worth mentionning.

mwestphal (changes requested):
small discussion

doc/dev/BUILD_WASM.md
Meakk: It's using the WSL layer silently through Docker. I think it's worth mentioning this guide is valid to run on Windows.

Meakk: I can remove ` (WSL)` if you think it's better tho

mwestphal: Do the command need to be run in a WSL terminal or does it work in any terminal assuming docker is installed and in the PATH ?

Meakk: Any terminal. If WSL is not installed, Docker will refuse to run the image.

mwestphal (approved):

PR #1458, opened 2024-06-07
add camera zoom method

### LINKED ISSUE #1429: Unable to zoom using `dolly` when it's orthographic
When it's in perspective I zoom using dolly, so moving the camera closer, but when it's orthographic it's not possible to zoom.

- A generic function to zoom, that handles both orthographic and perspective
- Exposing [VtkCamera.SetParallelScale](https://vtk.org/doc/nightly/html/classvtkCamera.html#ab9890abe84f27f544a68bb17133d1c73) to change the zoom in orthographic. This is what it's done in the app [here](https://github.com/f3d-app/f3d/blob/6133d0fa6189061f9fea453441dfbe086ba95087/vtkext/private/module/vtkF3DRenderer.cxx#L1365-L1367)
- A function only to zoom when orthographic similar to the other functions where you only pass the delta and not the final value
- Recompute the parallel scale internally when Dolly(), SetPosition(), and SetFocalPoint() are called if the camera is orthographic (and any other methods that affect the position-focal distance), this would not change the API

Also a function to be able to save the settings without rendering would be possible, because it would let me change to perspective, save, dolly and change back. Now it needs to render to be able to save.

**Additional context**
This is what I have done to make it work:
https://github.com/Nokse22/Exhibit/blob/3f6b2118d0365da4eeab1e8cc87d760cb5c54dbd/src/window.py#L190-L194

snoyer:
Add `f3d::camera::zoom(factor)` method by delegating to `vtkCamera::Zoom(factor)` to be able to "get closer" in both perspective and orthographic mode (where `dolly` is not applicable) 

fixes #1429

### DIFF
--- a/library/public/camera.h
+++ b/library/public/camera.h
@@ -52,18 +52,25 @@ public:
   ///@}
 
   ///@{ @name Manipulation
-  /**
-   * Standard camera manipulation methods.
-   * Angles are in degrees.
-   */
+  /// Standard camera manipulation methods. Angles are in degrees.
+
+  /** Divide the camera's distance from the focal point by the given value. */
   virtual camera& dolly(double val) = 0;
+  /** Move the camera along its horizontal, vertical, and forward axes */
   virtual camera& pan(double right, double up, double forward = 0) = 0;
+  /** Decrease the view angle (or the parallel scale in parallel mode) by the specified factor. */
   virtual camera& zoom(double factor) = 0;
+  /** Rotate the camera about its forward axis. */
   virtual camera& roll(angle_deg_t angle) = 0;
+  /** Rotate the camera about its vertical axis, centered at the focal point. */
   virtual camera& azimuth(angle_deg_t angle) = 0;
+  /** Rotate the camera about its vertical axis, centered the camera's position. */
   virtual camera& yaw(angle_deg_t angle) = 0;
+  /** Rotate the camera about its horizontal axis, centered at the focal point. */
   virtual camera& elevation(angle_deg_t angle) = 0;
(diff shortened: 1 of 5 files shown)


library/public/camera.h
```diff
@@ -58,6 +58,7 @@ class F3D_EXPORT camera
    */
   virtual camera& dolly(double val) = 0;
   virtual camera& pan(double right, double up, double forward = 0) = 0;
+  virtual camera& zoom(double factor) = 0;
```
mwestphal, line 61: some doc about the diff between zoom and dolly is important here imo.

snoyer: do we copy-paste the descriptions from corresponding `vtkCamera` methods?

mwestphal: that seems a bit overkill. I'm just suggesting sthing to help libf3d user figure out which method they should use between dolly and zoom if they are not super familiar with the concepts. Also "factor" is a bit opaque.

snoyer: > Also "factor" is a bit opaque.

Damn, should have stuck with `val` which is much better :D

The reasoning was that we basically have `dolly(x){ distance /= x }` and `zoom(x){ angle /= x }` where the argument is a multiplier (or rather the inverse of) meaning that the values should be around `1` (with `1` being a fixed-point/no-op). As opposed to `roll(x)`, `yaw(x)`, ... where the parameter is more like a delta that gets added.

Anyways, I'll add individual doc for these methods. What's the proper doxygen format to use here?

mwestphal (approved):

PR #1473, opened 2024-06-18
Set clang-format cpp style to modern C++

Meakk:
Let's bump C++ version to `Cpp11` (which means C++11 and above).  
It mainly impacts the closing bracket of nested template argument but I'll need it for a upcoming PR where I'm using raw strings.

### DIFF
--- a/library/src/options.cxx
+++ b/library/src/options.cxx
@@ -20,13 +20,13 @@ class options::internals
 {
 public:
   using OptionVariant =
-    std::variant<bool, int, double, std::string, std::vector<int>, std::vector<double> >;
+    std::variant<bool, int, double, std::string, std::vector<int>, std::vector<double>>;
 
   template<typename T, typename U>
   struct IsTypeValid;
 
   template<typename T, typename... Ts>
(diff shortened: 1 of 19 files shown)


snoyer:
here's a spot where `Cpp03` formatting had to be disabled but should work with `Cpp11` https://github.com/f3d-app/f3d/blob/b37f45513094f3b28eb7c264ab056281be6b130f/library/src/image.cxx#L515-L519

Meakk:
Indeed! Let me fix that.

Meakk:
@mwestphal please review

library/src/options.cxx
```diff
@@ -390,13 +390,13 @@ std::string options::getAsString(const std::string& name) const
 //----------------------------------------------------------------------------
 std::vector<int> options::getAsIntVector(const std::string& name) const
 {
-  return this->Internals->get<std::vector<int> >(name);
+  return this->Internals->get<std::vector<int>>(name);
```
mwestphal, line 393: this being uncovered is unexpected, is it not ?

mwestphal: or maybe it is.

mwestphal (approved):

PR #1476, opened 2024-06-20
Update VTK

Meakk:


application/testing/CMakeLists.txt
```diff
@@ -247,7 +247,7 @@ f3d_ss_test(NAME UserModelN TEMPLATE {model}_{n}.png EXPECTED ${_screenshot_user
 set_tests_properties(f3d::TestScreenshotUserModelN PROPERTIES ENVIRONMENT "XDG_PICTURES_DIR=${_screenshot_user_dir};HOME=${_screenshot_user_dir};USERPROFILE=${_screenshot_user_dir}")
 
 if(NOT APPLE OR VTK_VERSION VERSION_GREATER_EQUAL 9.3.0)
-  f3d_test(NAME TestTextureColor DATA WaterBottle.glb ARGS --geometry-only --texture-base-color=${F3D_SOURCE_DIR}/testing/data/albedo_mod.png --translucency-support DEFAULT_LIGHTS)
+  f3d_test(NAME TestTextureColor DATA WaterBottle.glb ARGS --geometry-only --texture-base-color=${F3D_SOURCE_DIR}/testing/data/albedo_mod.png --translucency-support DEFAULT_LIGHTS THRESHOLD 80)
```
mwestphal, line 250: I dont think this is needed ?

Meakk: It is. It breaks randomly on macos ARM.  
We had a discussion about it before and we decided a threshold increase is necessary

mwestphal: I've updated this baseline since. What I'm missing is that is this needed because of the VTK update or not ?

Meakk: Yes there were some visible artifacts randomly on macos arm

mwestphal: yes, but are those artifacts triggered by the VTK update ?

mwestphal (approved):

PR #1477, opened 2024-06-20
Add grid color option

Meakk:


webassembly/app.html
```diff
@@ -203,7 +213,7 @@ <h1 class="title">F3D Web</h1>
           });
 
         // load the file located in the virtual filesystem
-        openFile('f3d.vtp')
+        openFile('f3d.vtp');
```
mwestphal, line 216: what does `;` means in js ?

mwestphal (approved):

webassembly/app.html
Meakk: Same as C++, but it's optional as a new line does it for you.
I think it's good pratice to put it explicitly.
We may need a linter at some point.

Issue #1465, opened 2024-06-12
Add option to hide animation bar


snoyer:
The bar showing the progress of animations at the bottom of the window is useful in interactive mode but it would be nice to be able to disable/hide it.

There should be an option to control this like there is for the loading progress bar.

PR #1478, opened 2024-06-20
Add option to hide animation progress bar

spevnev:
I've added option to hide animation bar as requested in https://github.com/f3d-app/f3d/issues/1465

### DIFF
--- a/doc/libf3d/OPTIONS.md
+++ b/doc/libf3d/OPTIONS.md
@@ -94,6 +94,7 @@ ui.filename-info|string<br>-<br>render|Content of *filename info* to display.
 ui.font-file|string<br>-<br>render|Use the provided FreeType compatible font file to display text.<br>Can be useful to display non-ASCII filenames.|\-\-font-file
 ui.fps|bool<br>false<br>render|Display a *frame per second counter*.|\-\-fps
 ui.loader-progress|bool<br>false<br>load|Show a *progress bar* when loading the file.|\-\-progress
+ui.animation-progress|bool<br>false<br>load|Show a *progress bar* when playing the animation.|\-\-animation-progress
 ui.metadata|bool<br>false<br>render|Display the *metadata*.|\-\-metadata
 ui.dropzone|bool<br>false<br>render|Show a drop zone.
 ui.dropzone-info|string<br>-<br>render|Content of the drop zone text to display.
--- a/doc/libf3d/OPTIONS.md
+++ b/doc/libf3d/OPTIONS.md
@@ -94,7 +94,7 @@ ui.filename-info|string<br>-<br>render|Content of *filename info* to display.
 ui.font-file|string<br>-<br>render|Use the provided FreeType compatible font file to display text.<br>Can be useful to display non-ASCII filenames.|\-\-font-file
 ui.fps|bool<br>false<br>render|Display a *frame per second counter*.|\-\-fps
 ui.loader-progress|bool<br>false<br>load|Show a *progress bar* when loading the file.|\-\-progress
-ui.animation-progress|bool<br>false<br>load|Show a *progress bar* when playing the animation.|\-\-animation-progress
+ui.animation-progress|bool<br>true<br>load|Show a *progress bar* when playing the animation.|\-\-animation-progress
 ui.metadata|bool<br>false<br>render|Display the *metadata*.|\-\-metadata
 ui.dropzone|bool<br>false<br>render|Show a drop zone.
 ui.dropzone-info|string<br>-<br>render|Content of the drop zone text to display.
--- a/doc/libf3d/OPTIONS.md
+++ b/doc/libf3d/OPTIONS.md
@@ -94,7 +94,7 @@ ui.filename-info|string<br>-<br>render|Content of *filename info* to display.
 ui.font-file|string<br>-<br>render|Use the provided FreeType compatible font file to display text.<br>Can be useful to display non-ASCII filenames.|\-\-font-file
 ui.fps|bool<br>false<br>render|Display a *frame per second counter*.|\-\-fps
 ui.loader-progress|bool<br>false<br>load|Show a *progress bar* when loading the file.|\-\-progress
-ui.animation-progress|bool<br>true<br>load|Show a *progress bar* when playing the animation.|\-\-animation-progress
+ui.animation-progress|bool<br>false<br>load|Show a *progress bar* when playing the animation.|\-\-animation-progress
 ui.metadata|bool<br>false<br>render|Display the *metadata*.|\-\-metadata
 ui.dropzone|bool<br>false<br>render|Show a drop zone.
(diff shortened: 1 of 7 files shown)


library/src/animationManager.cxx
```diff
@@ -267,8 +267,11 @@ bool animationManager::LoadAtTime(double timeValue)
     // Set progress bar
     vtkProgressBarRepresentation* progressRep =
       vtkProgressBarRepresentation::SafeDownCast(this->ProgressWidget->GetRepresentation());
-    progressRep->SetProgressRate(
-      (this->CurrentTime - this->TimeRange[0]) / (this->TimeRange[1] - this->TimeRange[0]));
+    if (this->Options->getAsBool("ui.animation-progress"))
+      progressRep->SetProgressRate(
+        (this->CurrentTime - this->TimeRange[0]) / (this->TimeRange[1] - this->TimeRange[0]));
+    else
+      progressRep->SetProgressRate(0);
```
mwestphal, line 274: You may want to use `progressRep->SetVisibility(false);` instead

spevnev: You mean like this, right?
```
progressRep->SetProgressRate(
        (this->CurrentTime - this->TimeRange[0]) / (this->TimeRange[1] - this->TimeRange[0]));
else
      progressRep->SetVisibility(false);
```
I'll change that.

spevnev: Is there a way to reload configuration without relaunching?

spevnev: I know have realized that it is much better to simple leaved progress widget off during `animationManager::Initialize`.

mwestphal: hum, does that let you turn it on and off during execution ?

Im not sure that would be the case.

mwestphal: > Is there a way to reload configuration without relaunching?

If you mean is there a way to change the config without running from the start, yes there is. When you change a file, its own config block is read, which could be different from the other file and enable/disable the animation bar.

Also libf3d user may do other things.

This is why libf3d option have a "Trigger": https://f3d.app/doc/libf3d/OPTIONS.html

This is how the option is being "updated", some option are updated on load aka, when loading a file. some options are updated on "render", aka, when rendering again. No options should only be updated on "start".

mwestphal: If I remember well, "Initialize" is called on loadFile, which makes this option updated on "load", which is fine :)

doc/libf3d/OPTIONS.md
```diff
@@ -94,6 +94,7 @@ ui.filename-info|string<br>-<br>render|Content of *filename info* to display.
 ui.font-file|string<br>-<br>render|Use the provided FreeType compatible font file to display text.<br>Can be useful to display non-ASCII filenames.|\-\-font-file
 ui.fps|bool<br>false<br>render|Display a *frame per second counter*.|\-\-fps
 ui.loader-progress|bool<br>false<br>load|Show a *progress bar* when loading the file.|\-\-progress
+ui.animation-progress|bool<br>false<br>load|Show a *progress bar* when playing the animation.|\-\-animation-progress
```
mwestphal, line 97: please update https://f3d.app/doc/libf3d/OPTIONS.html accordingly

library/src/animationManager.cxx
```diff
@@ -55,7 +55,8 @@ bool animationManager::Initialize(
     progressRep->SetPadding(0.0, 0.0);
 #endif
 
-    this->ProgressWidget->On();
+    if (this->Options->getAsBool("ui.animation-progress"))
+      this->ProgressWidget->On();
```
mwestphal, line 59: if (this->Options->getAsBool("ui.animation-progress"))
    {
      this->ProgressWidget->On();
    }

mwestphal: actually this is incorrect. you should use setVisibility after turning the widget On iirc.

spevnev: > If I remember well, "Initialize" is called on loadFile, which makes this option updated on "load", which is fine :)

So it is okay to check this option only inside Initialize, but it should use SetVisibility instead of not turning the widget on, right?

mwestphal: yes indeed.

spevnev:
@mwestphal Why did the coverage check fail saying that no token has been found? It surely looks like it is unrelated to my changes.

mwestphal:
> @mwestphal Why did the coverage check fail saying that no token has been found? It surely looks like it is unrelated to my changes.

Indeed unrelated, codecov have some issues not yet fixed: https://github.com/codecov/feedback/issues/112#issuecomment-2082728236

doc/libf3d/OPTIONS.md
mwestphal, line 97: `bool<br>**true**<br>load`

spevnev: Should default be specified? If so, how?

library/src/options.cxx
```diff
@@ -182,6 +182,7 @@ options::options()
   this->Internals->init("ui.metadata", false);
   this->Internals->init("ui.font-file", std::string());
   this->Internals->init("ui.loader-progress", false);
+  this->Internals->init("ui.animation-progress", true);
```
mwestphal, line 185: @Meakk should we change that to false and add it in the default config instead ?

doc/libf3d/OPTIONS.md
spevnev: Also, I've just notice that there is default column in user/OPTIONS. I guess I should set true in there?

spevnev: Maybe change option to --hide-animation-progress then?

library/src/options.cxx
Meakk: Yes I agree.

spevnev: So the option has to be set in `src/resources/configs/config.d/10_global.json`?

mwestphal: Indeed! but you will need to update a few other tests that relies on the previous behavior.

doc/libf3d/OPTIONS.md
```diff
@@ -94,6 +94,7 @@ ui.filename-info|string<br>-<br>render|Content of *filename info* to display.
 ui.font-file|string<br>-<br>render|Use the provided FreeType compatible font file to display text.<br>Can be useful to display non-ASCII filenames.|\-\-font-file
 ui.fps|bool<br>false<br>render|Display a *frame per second counter*.|\-\-fps
 ui.loader-progress|bool<br>false<br>load|Show a *progress bar* when loading the file.|\-\-progress
+ui.animation-progress|bool<br>true<br>load|Show a *progress bar* when playing the animation.|\-\-animation-progress
```
mwestphal, line 97: default value is `false` then

mwestphal (changes requested):
small doc change needed

mwestphal (approved):

PR #1504, opened 2024-06-30
clamp animation time to available range

### LINKED ISSUE #1501: Avoid failing when provided animation time is outside of time range
**Describe the bug**
Loading a timestep outside of the temporal range fails in 

```
Provided time value: t is outside of animation time range: [t0, t1] 
```

Instead it could be a warning only and load the closest time instead.

snoyer:
fixes #1501, also makes the message slightly less verbose

### DIFF
--- a/application/testing/CMakeLists.txt
+++ b/application/testing/CMakeLists.txt
@@ -622,6 +622,10 @@ if(F3D_PLUGIN_BUILD_EXODUS)
   # Test no render animation time. Regex contains a part of the range of the ACCL field.
   f3d_test(NAME TestNoRenderAnimation DATA small.ex2 ARGS  --load-plugins=exodus --animation-time=0.003 REGEXP "-521950, 6.57485" NO_RENDER)
 
+  # Test animation time clamping
+  f3d_test(NAME TestAnimationTimeLimitsHigh DATA small.ex2 ARGS ARGS --load-plugins=exodus --animation-time=10 DEFAULT_LIGHTS)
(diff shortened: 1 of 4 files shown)


application/testing/CMakeLists.txt
```diff
@@ -358,7 +358,8 @@ if(VTK_VERSION VERSION_GREATER 9.0.1)
   f3d_test(NAME TestInteractionAnimationFrameRate DATA InterpolationTest.glb ARGS --animation-frame-rate=2 --animation-index=-1 INTERACTION NO_BASELINE)#Space;Wait;Space;
 
   # A verbose test that needs animation index support
-  f3d_test(NAME TestVerboseAnimationWrongAnimationTime DATA BoxAnimated.gltf ARGS --animation-time=10 --verbose REGEXP "Provided time value: 10 is outside of animation time range" NO_BASELINE)
+  f3d_test(NAME TestVerboseAnimationWrongAnimationTimeHigh DATA BoxAnimated.gltf ARGS --animation-time=10 --verbose REGEXP "animation time 10 is outside of range \\[0, 3\\.70833\\], using 3\\.70833" NO_BASELINE)
+  f3d_test(NAME TestVerboseAnimationWrongAnimationTimeLow DATA BoxAnimated.gltf ARGS --animation-time=-5 --verbose REGEXP "animation time -5 is outside of range \\[0, 3\\.70833\\], using 0" NO_BASELINE)
```
mwestphal, line 362: actually I would suggest to add a verbose test and a baseline test, in order to check that not only the text is displayed, but the right timestep is actually loaded

mwestphal (changes requested):
testing improvements needed

mwestphal (approved):

PR #1508, opened 2024-07-01
Prevent Jekyll from expanding `{{model}}.png` to `.png`

spevnev:


mwestphal (approved):
Assuming it actually works :)

PR #1523, opened 2024-07-07
Minimal screenshot

snoyer:
Allow saving "minimal"/"clean" screenshots with transparent background and no grid or overlays, as requested by a user on Discord.
Bound to `F11` instead of the initially proposed `shift+F12` because of current technical limitations.

### DIFF
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -490,6 +490,11 @@ int F3DStarter::Start(int argc, char** argv)
           this->SaveScreenshot(this->Internals->AppOptions.ScreenshotFilename);
           return true;
         }
+        if (keySym == "F11")
+        {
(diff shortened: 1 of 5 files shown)


mwestphal:
Not sure why CI is not clean, that looks like a missing DEFAULT_LIGHTS, but afair your screenshot testing mechanism should not require it ?

snoyer:
> Not sure why CI is not clean, that looks like a missing DEFAULT_LIGHTS, but afair your screenshot testing mechanism should not require it ?

Good point, it does compare the screenshot to a render so I'll add `DEFAULT_LIGHTS`. The question is more why hasn't it needed it before

mwestphal:
Yeah, I'm a bit confused here too. That being said, the image comparison algorithm is not great and can have false negative.

We plan to update it soon.

mwestphal (approved):

PR #1528, opened 2024-07-10
Improve LFS-related docs

spevnev:


doc/dev/BUILD.md
```diff
@@ -10,6 +10,7 @@ please take a look at our [getting started guide](GETTING_STARTED.md).
 * [VTK](https://vtk.org) >= 9.0.0 (9.2.6 recommended).
 * A C++17 compiler.
 * A CMake-compatible build system (Visual Studio, XCode, Ninja, Make, etc.).
+* Required for running tests, [git LFS](https://git-lfs.com/)
```
mwestphal, line 13: I dont think this is the right location for this, as this is really specific to people cloning the repository using git instead of downloading the tarbal. Not needed here imo.

mwestphal: but you could mention it next to the BUILD_TESTING option ?

doc/dev/TESTING.md
```diff
@@ -12,6 +12,8 @@ There are a few CMake options to configure the F3D testing framework:
 
 ## Running the tests
 
+Before running tests you need to fetch [git LFS](https://git-lfs.com/) data, which contains testing data and baselines. It can be done by running `git lfs pull` from the source directory.
```
mwestphal, line 15: it is done automatically if git lfs is installed when you clone.

doc/dev/BUILD.md
spevnev: So don't mention at all, then? Well, given links/mentions of git LFS it should be easy to figure it out, I guess

doc/dev/TESTING.md
spevnev: So maybe mention needing to have it installed here instead?

mwestphal: looks fine indeed.

mwestphal (approved):

PR #1536, opened 2024-07-14
Generalize up direction

### LINKED ISSUE #1533: Generalize "UP" vector to be any direction instead of axis aligned
the up direction could be generalized to be arbitrary instead of axis aligned, this has implication on the grid coloration.

snoyer:
would fix #1533


vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -1560,6 +1651,108 @@ void vtkF3DRenderer::ResetCameraClippingRange()
...
+            }
+            else
+            {
+              if (!extendBoxArbitrary(prop3d, box))
+              {
+                const std::string repr = std::string(prop3d->GetClassName());
+                F3DLog::Print(F3DLog::Severity::Warning,
+                  "Could not properly account for " + repr +
+                    " in non-axis-aligned bounds computation");
+                extendBoxAxisAligned(prop3d, box);
```
snoyer, line 1744: This is a fallback in the unlikely case there is no custom bounds implementation for a given `prop3d`. It should not happen if we have all the possible props/actors/mappers types handled so I don't think it can be covered by tests.

testing/baselines/TestGridX.png
Meakk, line 1: what happened here? Why is this new baseline required?

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -229,20 +234,86 @@ void vtkF3DRenderer::Initialize(const std::string& up)
...
+
+  /* if `up` is `(0,0,0)` make it `(0,1,0)` */
+  std::array<double, 3> up = upDir;
+  if (isEqual(up, { 0, 0, 0 }))
+  {
+    up[1] = 1.0;
+  }
+  vtkMath::Normalize(up.data());
+
+  /* make sure `right` is not `(0,0,0)` or the same as `up` */
```
Meakk, line 290: Not the same as `up` but colinear to `up`. If `right` is `p * up` for any `p` scalar, it won't work.  
It's better to check if the absolute value of the dot product between the two normalized vectors is below 0.99 or so.

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -507,15 +572,35 @@ void vtkF3DRenderer::ShowGrid(bool show)
...
+    };
+    upMatrix->DeepCopy(m);
+  }
+  vtkNew<vtkMatrix4x4> upMatrixInv;
+  upMatrixInv->DeepCopy(upMatrix);
+  upMatrixInv->Invert();
+  vtkNew<vtkTransform> upTransformInv;
+  upTransformInv->SetMatrix(upMatrixInv);
+  vtkNew<vtkTransform> upTransform;
+  upTransform->SetMatrix(upMatrix);
```
Meakk, line 596: A few remarks about that:
- You can use `vtkMatrix3x3` directly and use `MultiplyPoint` instead of relying on `vtkTransform` (which is very slow)
- The matrix is orthonormal, so you can just use the transpose instead of the inverse.

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -1560,6 +1651,108 @@ void vtkF3DRenderer::ResetCameraClippingRange()
   this->GridActor->SetUseBounds(gridUseBounds);
 }
 
+vtkBoundingBox vtkF3DRenderer::ComputeVisiblePropOrientedBounds(const vtkMatrix4x4* matrix)
+{
+  const auto isMatrixAxisAligned = [](const vtkMatrix4x4* m, const double tol = 1e-8)
```
Meakk, line 1656: Can't you just check that the sum of the asbolute value of all the elements (L1 norm) of the matrix is exactly 4? A flipped matrix would introduce a higher value I believe.

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -1560,6 +1651,108 @@ void vtkF3DRenderer::ResetCameraClippingRange()
...
+  {
+    vtkActor* actor = dynamic_cast<vtkActor*>(prop3d);
+    if (actor)
+    {
+      vtkPolyDataMapper* polyMapper = dynamic_cast<vtkPolyDataMapper*>(actor->GetMapper());
+      if (polyMapper)
+      {
+        vtkNew<vtkTransform> t;
+        t->Concatenate(*matrix->Element);
+        t->Concatenate(actor->GetMatrix());
```
Meakk, line 1700: Again, let's avoid `vtkTransform`. Just multiply the matrices to combine them.  
If the matrices can store a translation, you cannot use a 3x3 matrix and stick with the 4x4 matrix.  
In this case, you also need to use vec4, so make sure to use w=1 for points and w=0 for the fourth component.

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -1560,6 +1651,108 @@ void vtkF3DRenderer::ResetCameraClippingRange()
...
+      }
+    }
+    return false;
+  };
+
+  const bool isAxisAligned = isMatrixAxisAligned(matrix);
+  vtkBoundingBox box;
+
+  /* use `ComputeVisiblePropBounds()`'s logic to iterate `vtkProp3D`s contributing to the bounds */
+  {
```
Meakk, line 1719: is this scope needed?

vtkext/private/module/vtkF3DRenderer.cxx
Meakk: then let's assert?

Meakk (commented):
here are some comments

vtkext/private/module/vtkF3DRenderer.cxx
snoyer: couldn't a plugin use some exotic prop we haven't covered tho?

testing/baselines/TestGridX.png
snoyer: the drawing order of the axes has changed (blue over green vs the other way around)

vtkext/private/module/vtkF3DRenderer.cxx
snoyer: ah yes I assumed equal was enough because they'd be normalized but the can be opposite, good point!

vtkext/private/module/vtkF3DRenderer.cxx
snoyer: Maybe, I'm not sure TBH, I guess we would still be ok if there was some scaling in addition to swapping axes in the general case? But if we trust we're only using normalized vectors to build the matrix then your suggestion would work.

vtkext/private/module/vtkF3DRenderer.cxx
snoyer: It is indeed not

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -229,20 +234,87 @@ void vtkF3DRenderer::Initialize(const std::string& up)
...
+
+//----------------------------------------------------------------------------
+void vtkF3DRenderer::InitializeEnvironment(
+  const std::array<double, 3>& upDir, const std::array<double, 3>& rightDir)
+{
+  const auto isEqual = [](const std::array<double, 3>& u, const std::array<double, 3>& v)
+  {
+    constexpr double e = 1e-8;
+    return ::abs(u[0] - v[0]) < e && ::abs(u[1] - v[1]) < e && ::abs(u[2] - v[2]) < e;
+  };
```
Meakk, line 280: Since we only use it for comparing with the null vector, let's change that?
```suggestion
  const auto isNullVector = [](const std::array<double, 3>& u)
  {
    constexpr double e = 1e-8;
    return std::all_of(std::begin(u), std::end(u), [](double x){ return x < e; });
  };
```

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -530,26 +612,30 @@ void vtkF3DRenderer::ConfigureGridUsingCurrentActors()
         tmpUnitSquare = pow(10.0, round(log10(diag * 0.1)));
       }
 
-      double gridPos[3] = { 0, 0, 0 };
+      double center[3];
```
Meakk, line 615: Since `MultiplyDoublePoint` is expecting a `double[4]`, I think you'll have a potential out of bound issue there.  
Let's allocate `double center[4];` (and set the fourth component to 1 since it's a point, even if it won't make a difference here because there's no translation)

vtkext/private/module/vtkF3DRenderer.cxx
Meakk: fair enough.

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -1548,6 +1640,109 @@ void vtkF3DRenderer::ResetCameraClippingRange()
...
+    box.AddBounds(prop3d->GetBounds());
+
+    prop3d->PokeMatrix(nullptr);
+  };
+
+  /* Use custom logic to extend box.
+   * Should give the tightest bounds even when non-axis-aligned */
+  const auto extendBoxArbitrary = [&](vtkProp3D* prop3d, vtkBoundingBox& box)
+  {
+    vtkActor* actor = dynamic_cast<vtkActor*>(prop3d);
```
Meakk, line 1681: Use VTK style down cast, here are everywhere else.

```suggestion
    vtkActor* actor = vtkActor::SafeDownCast(prop3d);
```

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -518,26 +599,30 @@ void vtkF3DRenderer::ConfigureGridUsingCurrentActors()
...
+        // a small margin is added to the size to avoid z-fighting if large translucent
+        // triangles are exactly aligned with the grid bounds
+        constexpr double margin = 0.0001;
+        downShift += bbox.GetLength(1) / 2 + margin;
       }
 
+      double delta[3];
+      this->GetEnvironmentUp(delta);
+      vtkMath::MultiplyScalar(delta, downShift);
+      vtkMath::Subtract(gridPos, delta, gridPos);
```
Meakk, line 624: not sure here, but it seems like you are using the 4x4 matrix to rotate and then this code does the translation.  
You can encode rotation and translation within the same 4x4 matrix (last column for the translation)

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -548,13 +633,20 @@ void vtkF3DRenderer::ConfigureGridUsingCurrentActors()
...
       gridMapper->SetSubdivisions(this->GridSubdivisions);
-      gridMapper->SetUpIndex(this->UpIndex);
+
       if (this->GridAbsolute)
-        gridMapper->SetOriginOffset(-gridPos[0], -gridPos[1], -gridPos[2]);
+        gridMapper->SetOriginOffset(-center[0], -center[1], -center[2]);
+
+      double orientation[3];
+      vtkTransform::GetOrientation(orientation, upMatrixInv);
+      this->GridActor->SetOrientation(orientation);
```
Meakk, line 642: If properly built, you can just use `SetUserMatrix` directly.

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -226,20 +231,86 @@ void vtkF3DRenderer::Initialize(const std::string& up)
...
-    std::fill(this->RightVector, this->RightVector + 3, 0);
-    this->RightVector[this->UpIndex == 0 ? 1 : 0] = 1.0;
+    this->InitializeEnvironment(upDir, rightDir);
+  }
+  else if (std::regex_match(up, match, re2))
+  {
+    const std::array<double, 3> upDir = {
+      ::atof(match[1].str().c_str()), //
+      ::atof(match[4].str().c_str()), //
+      ::atof(match[7].str().c_str()), //
```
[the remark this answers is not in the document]
snoyer: will do

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -226,20 +231,86 @@ void vtkF3DRenderer::Initialize(const std::string& up)
...
+  /* make `front` orthogonal */
+  std::array<double, 3> front;
+  vtkMath::Cross(right.data(), up.data(), front.data());
+  vtkMath::Normalize(front.data());
+
+  /* ensure `right` is orthogonal */
+  vtkMath::Cross(up.data(), front.data(), right.data());
+  vtkMath::Normalize(right.data());
+
+  {
```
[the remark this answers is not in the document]
snoyer: no we don't. I think I left this one to keep the indentation and make the diff clearer. I'll remove it

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -500,10 +565,26 @@ void vtkF3DRenderer::ConfigureGridUsingCurrentActors()
...
+        right[0], right[1], right[2], 0, //
+        up[0], up[1], up[2], 0,          //
+        front[0], front[1], front[2], 0, //
+        0, 0, 0, 1,                      //
+      };
+      upMatrix->DeepCopy(m);
+    }
+    vtkNew<vtkMatrix4x4> upMatrixInv;
+    upMatrixInv->DeepCopy(upMatrix);
+    upMatrixInv->Transpose(); // matrix is orthonormal, no need to use `Invert()`
```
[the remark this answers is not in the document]
snoyer: I'm using both the matrix and it's inverse copy down the line. I guess I could hardcode the inverse one too, but is it worth it?

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -1548,6 +1640,109 @@ void vtkF3DRenderer::ResetCameraClippingRange()
...
+            if (!extendBoxArbitrary(prop3d, box))
+            {
+              const std::string repr = std::string(prop3d->GetClassName());
+              F3DLog::Print(F3DLog::Severity::Warning,
+                "Could not properly account for " + repr +
+                  " in non-axis-aligned bounds computation");
+              extendBoxAxisAligned(prop3d, box);
+            }
+          }
+        }
```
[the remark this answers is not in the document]
snoyer: I still think the bare minimum we can do is
```cpp
        if (prop3d) // we need an actual prop
        {
          if (isAxisAligned) // we want the fast code path if possible
          {
            extendBoxAxisAligned(prop3d, box);
          }
          else // otherwise we do the arbitrary computation
          {
            extendBoxArbitrary(prop3d, box); // but we don't check if it worked
          }
        }
```

but from there I don't think it costs much to handle the unlikely case of an unknown mapper and do
```cpp
        if (prop3d) // we need an actual prop
        {
          if (isAxisAligned) // we want the fast code path if possible
          {
            extendBoxAxisAligned(prop3d, box);
          }
          else if(!extendBoxArbitrary(prop3d, box)) // otherwise we try the arbitrary computation
          {
            extendBoxAxisAligned(prop3d, box); // the optimal arbitrary code didn't succeed so we fall back to the fast code that will work but give a sub-optimal (not tight) box
            // and we might as well add a log here I guess
          }
        }
```
which is basically the current version of the PR.

Am I missing something?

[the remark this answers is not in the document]
Meakk: Right

vtkext/private/module/vtkF3DRenderer.cxx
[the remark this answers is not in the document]
Meakk: Oh ok I missed you were using the initial matrix too.

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -500,10 +565,26 @@ void vtkF3DRenderer::ConfigureGridUsingCurrentActors()
...
-    double bounds[6];
-    this->ComputeVisiblePropBounds(bounds);
+    double* up = this->GetEnvironmentUp();
+    double* right = this->GetEnvironmentRight();
+    double front[3];
+    vtkMath::Cross(right, up, front);
 
-    vtkBoundingBox bbox(bounds);
+    vtkNew<vtkMatrix4x4> upMatrix;
+    {
```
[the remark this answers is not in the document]
Meakk: I don't think it's worth trying to free a variable manually by adding a scope in general, especially for stack variables or very small heap memory usage.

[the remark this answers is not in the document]
snoyer: TBH I didn't even thing about actually freeing it, it was only about the variable not being "visible" further down the function as it is only declared to be immediately passed to `DeepCopy()` :)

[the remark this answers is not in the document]
Meakk: I'd say bad style.

snoyer:
starting over with #2023

PR #1538, opened 2024-07-14
fix screenshot date test

snoyer:
Use pattern matching in logs rather that comparing filename on disk.
This avoids having to generate the expected date in the `cmake` file and having to run the test the same day.

### DIFF
--- a/application/testing/CMakeLists.txt
+++ b/application/testing/CMakeLists.txt
@@ -242,7 +242,7 @@ endfunction()
 function(f3d_ss_template_test)
   cmake_parse_arguments(F3D_SS_TEMPLATE_TEST "" "NAME;TEMPLATE;EXPECTED_REGEX" "ARGS" ${ARGN})
   f3d_test(NAME TestScreenshot${F3D_SS_TEMPLATE_TEST_NAME} DATA suzanne.ply ARGS --screenshot-filename=${F3D_SS_TEMPLATE_TEST_TEMPLATE} --dry-run --interaction-test-play=${F3D_SOURCE_DIR}/testing/recordings/TestScreenshot.log
-    REGEXP "saving screenshot to ${F3D_SS_TEMPLATE_TEST_EXPECTED_REGEX}" NO_BASELINE DEPENDS TestSetupScreenshots)
+    REGEXP "saving screenshot to .+[/\\]${F3D_SS_TEMPLATE_TEST_EXPECTED_REGEX}" NO_BASELINE DEPENDS TestSetupScreenshots)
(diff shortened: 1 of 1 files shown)


application/testing/CMakeLists.txt
```diff
@@ -253,9 +258,7 @@ f3d_ss_test(NAME Version TEMPLATE ${_screenshot_dir}/{app}_{version}_{version_fu
 f3d_ss_test(NAME Model TEMPLATE ${_screenshot_dir}/{model}_{model.ext}_{model_ext}.png EXPECTED ${_screenshot_dir}/suzanne_suzanne.ply_ply.png)
 f3d_ss_test(NAME ModelN1 TEMPLATE ${_screenshot_dir}/{model}_{n}_{n:2}.png EXPECTED ${_screenshot_dir}/suzanne_1_01.png)
 f3d_ss_test(NAME ModelN2 TEMPLATE ${_screenshot_dir}/{model}_{n}_{n:2}.png EXPECTED ${_screenshot_dir}/suzanne_2_02.png DEPENDS TestScreenshotModelN1)
-string(TIMESTAMP DATE_Y "%Y")
-string(TIMESTAMP DATE_Ymd "%Y%m%d")
-f3d_ss_test(NAME Date TEMPLATE ${_screenshot_dir}/{model}_{date}_{date:%Y}.png EXPECTED ${_screenshot_dir}/suzanne_${DATE_Ymd}_${DATE_Y}.png)
+f3d_ss_template_test(NAME Date TEMPLATE ${_screenshot_dir}/{model}_{date}_{date:%Y}.png EXPECTED_REGEX ${_screenshot_dir}/suzanne_[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]_[0-9][0-9][0-9][0-9]\.png)
```
mwestphal, line 261: since you use this only once I wonder if you really need a cmake function ? just declare the test using f3d_test ?

snoyer: I'd argue the function helps for readability. Maybe I can use it for some of the other tests to remove the "it's only used once" reservation (and insignificantly lighten the CI at the same time)?

mwestphal: that would be awesome if you see a partern for improving the regex tests.

snoyer: > that would be awesome if you see a partern for improving the regex tests.

I just meant replacing some of the `f3d_ss_test` with `f3d_ss_template_test`. I'm not a `ctest` wizard yet :)

mwestphal (approved):

PR #1561, opened 2024-08-12
Ignore .DS_Store file

Meakk:
Add `.DS_Store` to `.gitignore`. This file is used to store metadata of the file explorer on macos (Finder)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ dist
 __pycache__
 *.pyc
 _version.py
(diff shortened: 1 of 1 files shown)


mwestphal:
Why would such a file end up in f3d source dir ?

Meakk:
If you change a property (e.g. from default view to detailed view) while browsing F3D sources in Finder.

mwestphal (approved):

PR #1565, opened 2024-08-15
Cleanup image error, exception, testing and doc

mwestphal:
- Specifiy behavior when saving incompatible type image
 - Use correct exceptions
 - Update doc
 - Update tests

### DIFF
--- a/library/src/image.cxx
+++ b/library/src/image.cxx
@@ -38,9 +38,15 @@ namespace f3d
 {
 class image::internals
 {
-  inline static const std::string metadataKeyPrefix = "f3d:";
-
 public:
+  inline static const std::string metadataKeyPrefix = "f3d:";
+  inline static const std::map<SaveFormat, std::string> saveFormatString = {
+    { SaveFormat::PNG, "PNG" },
+    { SaveFormat::BMP, "BMP" },
+    { SaveFormat::JPG, "JPG" },
+    { SaveFormat::TIF, "TIF" },
+  };
+
   vtkSmartPointer<vtkImageData> Image;
   std::unordered_map<std::string, std::string> Metadata;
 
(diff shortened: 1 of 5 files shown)


library/src/image.cxx
```diff
@@ -93,6 +93,27 @@ class image::internals
       }
     }
   }
+
+  static bool checkSaveFormatCompatibility(const image& self, SaveFormat format)
```
snoyer, line 97: How about making this a `void` method that throws so the message can be more precise? e.g. "PNG only supports channel type BYTE and SHORT", "JPG only supports channel type BYTE" and so on.

library/src/image.cxx
```diff
@@ -36,11 +36,17 @@
...
+struct image::internals
 {
   inline static const std::string metadataKeyPrefix = "f3d:";
+  // clang-format off
+  inline static const std::map<SaveFormat, std::string> saveFormatString = { 
+    { SaveFormat::PNG, "PNG" },
+    { SaveFormat::BMP, "BMP" }, 
+    { SaveFormat::JPG, "JPG" }, 
+    { SaveFormat::TIF, "TIF" } };
+  // clang-format on
```
snoyer, line 49: Would it be enough to add a trailing comma instead of disabling the formatting?
```suggestion
  inline static const std::map<SaveFormat, std::string> saveFormatString = { 
    { SaveFormat::PNG, "PNG" },
    { SaveFormat::BMP, "BMP" },
    { SaveFormat::JPG, "JPG" },
    { SaveFormat::TIF, "TIF" },
  };
```

mwestphal: indeed!

Meakk (approved):

PR #1582, opened 2024-08-23
Add assimp plugin to wasm build and use npm in CI

Meakk:


webassembly/build.sh
```diff
@@ -3,10 +3,11 @@
 cmake -S /src -B /build \
     -DBUILD_SHARED_LIBS=OFF \
     -DF3D_STRICT_BUILD=ON \
+    -DF3D_PLUGIN_BUILD_ASSIMP=ON \
     -DF3D_PLUGIN_BUILD_EXODUS=OFF \
     -DF3D_WASM_COPY_APP=ON \
     -DCMAKE_BUILD_TYPE=$1 \
-    -DVTK_DIR:PATH=/depends/lib/cmake/vtk
+    -DCMAKE_FIND_ROOT_PATH:PATH=/depends/
```
mwestphal, line 10: CMAKE_PREFIX_PATH

Meakk: nop, it doesn't work because the emscripten toolchain is disabling it to avoid finding native packages.  
`CMAKE_FIND_ROOT_PATH` must be used when cross-compiling

mwestphal: TIL

mwestphal (approved):

PR #1287, opened 2024-02-12
Implements GLTF Draco importer

Meakk:

PR #1584, opened 2024-08-23
Fix draco decoder and factorize

Meakk:


Meakk:
@mwestphal review easier by ignoring white spaces: https://github.com/f3d-app/f3d/pull/1584/files?diff=unified&w=1

plugins/draco/module/vtkF3DGLTFDocumentLoader.cxx
mwestphal, line 1: Could you add a test of what you are fixing with a small data ?

Meakk: No it's not that easy. We have already discussed about it when I implemented Draco GLTF and decided to skip covering these edge cases.  
You can see the coverage of the previous PR, we had 13 lines missing: https://github.com/f3d-app/f3d/pull/1287#issuecomment-1940894832  
It's and improvement here with 4 lines missing only with the refactoring.

Meakk: Sorry, misunderstood your initial comment, let me check, but it won't be easy to find I think.

mwestphal: My question was more about testing than coverage, but fair enough.

mwestphal (approved):

PR #1585, opened 2024-08-23
add a test helper

snoyer:
Introduce a helper class to allow factoring away the current pattern of
```cpp
  if (!test1)
  {
    std::cerr << "test1 has failed" << std::endl;
    return EXIT_FAILURE;
  }
  if (!test2)
  {
    std::cerr << "test2 has failed" << std::endl;
    return EXIT_FAILURE;
  }
  if (!test3)
  {
    std::cerr << "test3 has failed" << std::endl;
    return EXIT_FAILURE;
  }
  return EXIT_SUCCESS;
```
into
```cpp
  Test test;
  test("test1", test1);
  test("test2", test2);
  test("test3", test3);
  return test.result();
```
and reduce noise in `TestSDK...` files.


library/testing/test.h
```diff
@@ -0,0 +1,139 @@
+#ifndef Test_h
+#define Test_h
+
+#include <functional>
+#include <iostream>
+#include <sstream>
+
+class Test
```
Meakk, line 8: Let's find a class and filename that match the style.  
Maybe we can merge it with `TestSDKHelpers` too?

library/testing/.clang-format
mwestphal, line 1: why do we need a custom conf ?

library/testing/test.h
mwestphal: agreed

mwestphal: some doc would go a long way to understand what this do btw

library/testing/.clang-format
snoyer: we don't _need_ it, but I expect with this boilerplate/helper there will be enough `[&](){ ... }` lambdas that it would be nice to use this config that formats as:
```cpp
    test("example test", [&]() {
      /* ... */
    });
```
instead of the project's one that does:
```cpp
    test("example test",
      [&]()
      {
         /* ... */
      });
```

mwestphal: indeed, look nicer, we may consider switching the whole project like this maybe ?

library/testing/test.h
snoyer: renamed, feel free to re-open if you have better naming ideas

library/testing/FakeUnitTest.h
```diff
@@ -0,0 +1,144 @@
...
+    return ss.str();
+  }
+
+private:
+  class Dummy : public std::exception
+  {
+  };
+};
+
+#endif
```
mwestphal, line 144: missing eol

library/testing/FakeUnitTest.h
```diff
@@ -0,0 +1,144 @@
+#ifndef Test_h
+#define Test_h
+
+#include <functional>
+#include <iostream>
+#include <sstream>
+
+class FakeUnitTest
```
mwestphal, line 8: add some doc

library/testing/FakeUnitTest.h
mwestphal, line 1: what make these tests fake in your opinion ?

snoyer: I think to do real unit testing each test would need to be properly independent (no side effects between tests, can be run out of order or in parallel). This class make the code (and the output feel) like unit tests but it's still a bunch of sequential checks in the same `ctest` `main()`.

Maybe "fake" wasn't the best choice of words... would `PseudoUnitTest` be better? or something else entirely?

mwestphal: I see what you mean. Pseudo is better indeed.

library/testing/PseudoUnitTest.h
```diff
@@ -0,0 +1,149 @@
...
+    this->record(success, label, this->comparisonMessage(actual, expected, success ? "~=" : "!="));
+  }
+
+  int result()
+  {
+    const bool success = this->failCount == 0;
+    return success ? EXIT_SUCCESS : EXIT_FAILURE;
+  }
+
+protected:
```
mwestphal, line 57: private ?

library/testing/.clang-format
```diff
@@ -0,0 +1,17 @@
...
+  AfterFunction: true
+  AfterNamespace: true
+  AfterObjCDeclaration: true
+  AfterStruct: true
+  AfterUnion: true
+  BeforeCatch: true
+  BeforeElse: true
+  IndentBraces: false
+  # ...except for lambdas
+  BeforeLambdaBody: false
```
mwestphal, line 17: why does this matter ?

mwestphal (commented):
small questions and comment, feel free to resolve

library/testing/PseudoUnitTest.h
snoyer: will do, I'll just keep `log()` `protected` because it's subclassed in `TestPseudoUnitTest.cxx`

library/testing/.clang-format
snoyer: the idea is to keep the global project style except for lambdas that get real noisy real fast, see https://github.com/f3d-app/f3d/pull/1585#discussion_r1739634772

mwestphal: thanks for the reminder

mwestphal (approved):

library/testing/PseudoUnitTest.h
```diff
@@ -0,0 +1,150 @@
+#ifndef Test_h
+#define Test_h
```
Meakk, line 2:
```suggestion
#ifndef PseudoUnitTest_h
#define PseudoUnitTest_h
```

library/testing/PseudoUnitTest.h
```diff
@@ -0,0 +1,150 @@
...
+   * - *fail* when anything else is caught
+   */
+  template<typename E>
+  void testFunction(
+    const std::string& label, std::function<void()> function, bool expectException = false)
+  {
+    try
+    {
+      function();
+      if (expectException)
```
Meakk, line 74: Is it making sense to drop this boolean and use `if constexpr (std::is_same_v<E, Dummy>)` here?

library/testing/PseudoUnitTest.h
```diff
@@ -0,0 +1,150 @@
...
+      this->record(true, label, e.what());
+    }
+    catch (const char* msg)
+    {
+      this->record(false, label, msg);
+    }
+    catch (const std::string& msg)
+    {
+      this->record(false, label, msg);
+    }
```
Meakk, line 94: do we throw string exceptions?

library/testing/PseudoUnitTest.h
```diff
@@ -0,0 +1,150 @@
...
+    std::stringstream ss;
+    ss << value;
+    return ss.str();
+  }
+
+protected:
+  virtual void log(const bool success, const std::string& label, const std::string& message)
+  {
+    const std::string line = message.empty() ? label : (label + ": " + message);
+    const std::string icon = success ? u8"\u2714" : u8"\u2718";
```
Meakk, line 140: Should we assume it works with any terminal?

library/testing/PseudoUnitTest.h
```diff
@@ -0,0 +1,150 @@
...
+
+  /** test the execution of a function */
+  void operator()(const std::string& label, std::function<void()> function)
+  {
+    this->testFunction<Dummy>(label, function);
+  }
+
+  /** test the execution of a function expecting a given exception */
+  template<typename E>
+  void expect(const std::string& label, std::function<void()> function)
```
Meakk, line 30: Not super important, but `std::function` is expensive, using template allow the compiler to optimize away the function and inline it. 
```suggestion
  template<typename E, typename F>
  void expect(const std::string& label, F function)
```

library/testing/PseudoUnitTest.h
```diff
@@ -0,0 +1,150 @@
...
+/** Helper to perform multiple checks within the same `ctest` test.
+ * Checks are performed using the various overloads of `operator()`
+ * and their results are logged and tracked so that the `result()` method
+ * will return `EXIT_FAILURE` if any of them failed (and `EXIT_SUCCESS` otherwise).
+ */
+class PseudoUnitTest
+{
+public:
+  /** test a boolean condition */
+  void operator()(const std::string& label, bool condition)
```
Meakk, line 17: I suppose the argument is always a string view rvalue, so I think this make more sense (here and everywhere):
```suggestion
  void operator()(std::string_view&& label, bool condition)
```

Meakk (commented):
classic nitpicking

library/testing/PseudoUnitTest.h
snoyer: it's used as a way to log failure details
```cpp
test("foo bar", [&]() { 
  if(not_good())
    throw "was not good";
});
```
will output
```
✘ foo bar: was not good
```
(that's throwing `char*` but it does the same with `std::string` so we can build the message dynamically if needed)

library/testing/PseudoUnitTest.h
snoyer: It works on Linux, Mac, Windows on the CI at least, I guess we can downgrade to non-utf8 later if we get cases where it doesn't work in the wild?

library/testing/PseudoUnitTest.h
snoyer: Way better, thanks!

library/testing/PseudoUnitTest.h
snoyer: leaving this one out because it's c++17 only

Meakk (approved):

PR #1588, opened 2024-08-24
Refactor libInfo struct

Meakk:


library/src/engine.cxx
```diff
@@ -290,29 +290,26 @@ engine::libInformation engine::getLibInfo()
...
-  tmp = "ON";
+  libInfo.Modules["OpenEXR"] = true;
 #else
-  tmp = "OFF";
+  libInfo.Modules["OpenEXR"] = false;
 #endif
-  libInfo.OpenEXRModule = tmp;
 
+  // First version of VTK including the version check (and the feature used)
+#if VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 0, 20200527)
```
mwestphal, line 312: how did that compiled before this change ?

mwestphal (commented):
a comment

library/src/engine.cxx
Meakk: that's a mistake on my side, you removed that with the drop of VTK 9.0
fixed

mwestphal: Ha right

mwestphal (approved):

Issue #1595, opened 2024-08-28
Parse mesh url from url-param search query to load https://f3d.app/web?mesh={mesh_url}


jo-chemla:
**Is your feature request related to a problem? Please describe.**
Long-time f3d user, however I just discovered the https://f3d.app/web WASM build which does look great so thanks a lot! It would be great if one could pass to the viewer a url-param which indicates a mesh file to load. 

**Describe the solution you'd like**
f3d could parse mesh url given search query like the following: https://f3d.app/web?mesh={mesh_url}
Regarding at how the [webassembly/app.html](https://github.com/f3d-app/f3d/blob/4f768e3f73b7f3a75c2f0ecaa6a0ce6e0623861b/webassembly/app.html#L129) loads supported meshes on file select, it seem that this could be hacked to initially load a model if the urlparam search query is detected. 

Some `os` urlparams seem to be parsed [here](https://github.com/f3d-app/f3d/blob/4f768e3f73b7f3a75c2f0ecaa6a0ce6e0623861b/assets/main.js#L56), but only to know which os to look for the wasm build.

**Describe alternatives you've considered**
3dviewer.net does it this way - although a weird hash # replaces the question-mark in the search-query: https://3dviewer.net/#model=${mesh_url}. 

**Additional context**
This would make it easier to inspect mesh files lying accessible on the web via a server endpoint using f3d. My interest is integrating f3d within STAC-browser (Spartio-Temporal Asset Catalogs) which is made to explore and standardize geospatial assets, among which meshes are an important part - see PR [here](https://github.com/radiantearth/stac-browser/pull/465) for context

**Contribution**
Would love to contribute, although I don't know mulch about the f3d codebase. Would need some pointers, the implementation probably has to happen in the app.html file, but building/bundling the wasm seem hard once implementation is finished.

snoyer:
This would be a nice improvement indeed!

> although a weird hash # replaces the question-mark in the search-query: [3dviewer.net#model=${mesh_url}](https://3dviewer.net/#model=$%7Bmesh_url%7D).

Doing it with a hash seems better in this case actually. If there was anything to do server side we'd need to pass the parameters in the `?` part of the URL, but in this case everything is local so we can use the `#` part that doesn't get send to the server and can be reacted to without reloading the page.

[...]

PR #1596, opened 2024-08-28
webassembly/app.html: Fetch mesh from search-query url-param

jo-chemla:
PR Following feature request thread here https://github.com/f3d-app/f3d/issues/1595

### DIFF
--- a/webassembly/app.html
+++ b/webassembly/app.html
@@ -222,27 +222,25 @@
         Module.engineInstance.getWindow().render();
         Module.engineInstance.getInteractor().start();
 
-        // Parse mesh from url-param search query hash
+        // Parse model from url-param search query hash via model url and extension
         function filename_for_model_url(model_url, extension_parsed, contentDisposition) {
           // Build filename given extension urlparam or response header content-disposition
           if (extension_parsed) {
             return `model_urlparam.${extension_parsed}`;
-          } else {
+          } else if (contentDisposition) {
             // If extension is not provided by user, try to get it auto from content-disposition header of url extension 
-            if (contentDisposition) {
-              return contentDisposition.split('filename=')[1].split(';')[0];
-            } else {
-              return model_url.split('/').pop();
-            }
+            return contentDisposition.split('filename=')[1].split(';')[0];
+          } else {
+            return model_url.split('/').pop();
           }
-          return filename;
+          throw new Error(`Could not parse filename/extension from either urlparam extension, response header content-disposition, nor filename present in url`);
         }
         function load_from_url(){
-          // Parse search-query mesh url-param or load default mesh file
+          // Parse search-query model url-param or load default model file
           // const params = new URLSearchParams(window.location.search);
           // Replace first hash with question mark to have real search query parsing and avoid leading # in first parsed urlparam
           const params = new URLSearchParams(window.location.hash.replace(/^#/, '?'));
-          const model_url_passed = params.get("mesh");
+          const model_url_passed = params.get("model");
           const extension_parsed = params.get("extension");
           if (model_url_passed) {
             const model_url = decodeURI(model_url_passed);
--- a/webassembly/app.html
+++ b/webassembly/app.html
@@ -225,18 +225,25 @@
         // Parse search-query mesh url-param or load default mesh file
         const params = new URLSearchParams(window.location.search);
         const mesh_url = params.get("mesh");
+        const extension = params.get("extension");
         if (mesh_url) {
           fetch(mesh_url)
           .then((response) => {
             if (!response.ok) {
               throw new Error(`HTTP error, status = ${response.status}`);
             }
-            return response.arrayBuffer();
-          })
-          .then((buffer) => {
+            // Build filename given extension urlparam or response header content-disposition
+            let filename 
+            if (extension) {
+              filename = 'mesh_urlparam' + extension
+            } else {
+              const contentDisposition = response.headers.get('content-disposition')
+              filename = contentDisposition.split('filename=')[1].split(';')[0];
+            }
             // Convert buffer to Uint8Array and openFile
-            Module.FS.writeFile('mesh_urlparam', new Uint8Array(buffer));
-            openFile('mesh_urlparam');
+            const buffer = response.arrayBuffer();
+            Module.FS.writeFile(filename, new Uint8Array(buffer));
+            openFile(filename);
           })
         } else {
           // load the file located in the virtual filesystem
--- a/webassembly/app.html
+++ b/webassembly/app.html
@@ -126,10 +126,12 @@
           const loader = Module.engineInstance.getLoader();
           if (loader.hasSceneReader(filePath))
           {
+            console.log('loader.hasSceneReader')
             loader.loadScene(filePath);
           }
           else if (loader.hasGeometryReader(filePath))
           {
+            console.log('loader.hasGeometryReader')
             loader.loadGeometry(filePath);
           }
           else
@@ -235,10 +237,20 @@
             // Build filename given extension urlparam or response header content-disposition
             let filename 
             if (extension) {
-              filename = 'mesh_urlparam' + extension
+              filename = `mesh_urlparam.${extension}`
             } else {
+              // If extension is not provided by user, try to get it auto from content-disposition header of url extension 
               const contentDisposition = response.headers.get('content-disposition')
-              filename = contentDisposition.split('filename=')[1].split(';')[0];
+              if (contentDisposition) {
+                filename = contentDisposition.split('filename=')[1].split(';')[0];
+              } else {
+                try {
+                  const url_extension = '.' + mesh_url.split('.').at(-1) 
+                  filename = `mesh_urlparam.${url_extension}`
+                } catch {
+                  throw new Error(`Could not infer file extension from content-disposition or url extension. Please enforce it via the extension urlparam`);
+                }
+              }
             }
             // Convert buffer to Uint8Array and openFile
             const buffer = response.arrayBuffer();
--- a/webassembly/app.html
+++ b/webassembly/app.html
@@ -212,8 +212,6 @@
             openFile(document.getElementById('file-name').innerHTML);
           });
 
-        // load the file located in the virtual filesystem
-        openFile('f3d.vtp');
 
         // setup the window size based on the canvas size
         const main = document.getElementById('main');
@@ -224,7 +222,7 @@
         Module.engineInstance.getWindow().render();
         Module.engineInstance.getInteractor().start();
 
-        // Parse url
+        // Parse search-query mesh url-param or load default mesh file
         const params = new URLSearchParams(window.location.search);
         const mesh_url = params.get("mesh");
         if (mesh_url) {
@@ -240,6 +238,9 @@
             Module.FS.writeFile('mesh_urlparam', new Uint8Array(buffer));
             openFile('mesh_urlparam');
           })
+        } else {
+          // load the file located in the virtual filesystem
+          openFile('f3d.vtp');
         }
       })
       .catch(error => console.error("Internal exception: " + error));
--- a/webassembly/app.html
+++ b/webassembly/app.html
@@ -126,12 +126,10 @@
           const loader = Module.engineInstance.getLoader();
           if (loader.hasSceneReader(filePath))
           {
-            console.log('loader.hasSceneReader')
             loader.loadScene(filePath);
           }
           else if (loader.hasGeometryReader(filePath))
           {
-            console.log('loader.hasGeometryReader')
             loader.loadGeometry(filePath);
           }
           else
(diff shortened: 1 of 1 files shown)


webassembly/app.html
```diff
@@ -223,6 +221,27 @@ <h1 class="title">F3D Web</h1>
...
+          .then((response) => {
+            if (!response.ok) {
+              throw new Error(`HTTP error, status = ${response.status}`);
+            }
+            return response.arrayBuffer();
+          })
+          .then((buffer) => {
+            // Convert buffer to Uint8Array and openFile
+            Module.FS.writeFile('mesh_urlparam', new Uint8Array(buffer));
+            openFile('mesh_urlparam');
```
snoyer, line 239:
```suggestion
            const name = "..."; // retrieve name from mesh_url
            Module.FS.writeFile(name, new Uint8Array(buffer));
            openFile(name);
```
`f3d` needs the file extension to know the format and open the file

jo-chemla: Ah that's right, is the extension parsed from the filename written to Module.FS? 
In that case, the mesh url has to end with the extension, and filename be retrieved from it, correct? 
Otherwise, we can as well add another url-param, like `loader` or `extension`, which if present will enforce loading with the corresponding loader, and if not present, will default to parsing extension from url. Would that make sense ?

snoyer: > Ah that's right, is the extension parsed from the filename written to Module.FS?

yes this filename can be made up but it need the proper extension

> Otherwise, we can as well add another url-param, like `loader` or `extension`, which if present will enforce loading with the corresponding loader, and if not present, will default to parsing extension from url. Would that make sense ?

It does make sense, if the model was a glTF file but there was no way to tell from the url alone  (e.g. `http://example.com/foo/bar`) then you'd need some way to pass the format so that you could write the file as `bar.gltf` in the filesystem

jo-chemla: Alright thanks for the precision that the file extension should be passed as filename. 
This extension can be enforced by user via url-param first, simply `extension` which would be appended to a placeholder name. 
At some point, one might try to extract it from the `Response Headers` as advised [here](https://github.com/kovacsv/Online3DViewer/issues/432), a bit trickier to parse, like `Content-Disposition: attachment; filename=odm_textured_model_geo.obj; filename*=UTF-8''odm_textured_model_geo.obj`

Meakk:
Just tried a local build and opening `http://192.168.0.12:3000/?mesh=https://github.com/KhronosGroup/glTF-Sample-Models/raw/main/2.0/DamagedHelmet/glTF-Binary/DamagedHelmet.glb` gives me the following error:
> Access to fetch at 'https://github.com/KhronosGroup/glTF-Sample-Models/raw/main/2.0/DamagedHelmet/glTF-Binary/DamagedHelmet.glb' from origin 'http://192.168.0.12:3000' has been blocked by CORS policy: The 'Access-Control-Allow-Origin' header contains the invalid value ''. Have the server send the header with a valid value, or, if an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.

Any idea?

jo-chemla:
True, CORS blocking is a standard error when fetching content that resides on a server, from an app that lives on another domain. You either need:
 - to use a url to a mesh file that has `Access-Control-Allow-Origin: *` headers, maybe [these](https://groups.csail.mit.edu/graphics/classes/6.837/F03/models/) should work like [teapot](https://groups.csail.mit.edu/graphics/classes/6.837/F03/models/teapot.obj)
 - or to use an Allow-CORS plugin (for example this plugin: [firefox](https://addons.mozilla.org/en-GB/firefox/addon/access-control-allow-origin/)/[chrome](https://chromewebstore.google.com/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf?pli=1))

Meakk:
Using `?mesh=https://groups.csail.mit.edu/graphics/classes/6.837/F03/models/teapot.obj`:
```
Uncaught (in promise) TypeError: Cannot read properties of null (reading 'split')
    at ?mesh=https://groups.csail.mit.edu/graphics/classes/6.837/F03/models/teapot.obj:241:45
```

Using `?mesh=https://groups.csail.mit.edu/graphics/classes/6.837/F03/models/teapot.obj&extension=.obj`:
No error is reported by the window is empty (a filename `meshurl_param.obj` is set in the open button)

image

snoyer:
How do we feel about using the hash part of the URL (`https://f3d.app/web/#http://example.com/model.glb`) instead of the search params?
Also (and independently of the hash vs search question) appending the extension as a hash to the model URL looks like it could be a simpler solution than adding a seperate parameter

```js
        const mesh_url = window.location.hash.substring(1);
        const j = mesh_url.lastIndexOf('/');
        const mesh_name = (j>0? mesh_url.substring(j+1):mesh_url);
        const sanitized_mesh_name = mesh_name.replace(/[^a-zA-Z0-9._-]/,'_');
        // ...
            Module.FS.writeFile(sanitized_mesh_name, new Uint8Array(buffer));
            openFile(sanitized_mesh_name);
```

- with `http://localhost:8000/app.html#http://localhost:8000/Buggy.glb`:
  - `mesh_url`: `http://localhost:8000/Buggy.glb`
  - `mesh_name`: `Buggy.glb`
  - `sanitized_mesh_name`: `Buggy.glb`

- with `http://localhost:8000/app.html#http://localhost:8000/cube#.glb`:
  - `mesh_url`: `http://localhost:8000/cube#.glb`
  - `mesh_name`: `cube#.glb`
  - `sanitized_mesh_name`: `cube_.glb`

jo-chemla:
Final edit: 🚀 Last commit replaces the search-query with a hash as requested by @snoyer URLs are parsed exactly like search queries, but with a `#` rather than `?` character. mesh URLs parsed are decodeURI'ed, so one can pass either a standard url, or an encoded one in case it includes special characters like `?, # etc`. https://f3d.app/web#mesh=${encodeURI(https://groups.csail.mit.edu/graphics/classes/6.837/F03/models/teapot.obj)}&extension=.obj I've also updated the above list of mesh files I could test it against, and all formats do work! From my point-of-view, it is now up to the f3d team whether or not to merge!

webassembly/app.html
```diff
@@ -126,10 +126,12 @@ <h1 class="title">F3D Web</h1>
           const loader = Module.engineInstance.getLoader();
           if (loader.hasSceneReader(filePath))
           {
+            console.log('loader.hasSceneReader')
```
Meakk, line 129: Please remove all `console.log` calls.

webassembly/app.html
```diff
@@ -223,6 +223,54 @@ <h1 class="title">F3D Web</h1>
...
+        const mesh_url = params.get("mesh");
+        const extension = params.get("extension");
+        if (mesh_url) {
+          fetch(mesh_url)
+          .then((response) => {
+            if (!response.ok) {
+              throw new Error(`HTTP error, status = ${response.status}`);
+            }
+            // Build filename given extension urlparam or response header content-disposition
+            let filename
```
Meakk, line 238: Sorry, I have a C++ background, so let's pretend JS semicolon insertion doesn't exist and add a semicolon after each statement 😬

webassembly/app.html
```diff
@@ -223,6 +223,54 @@ <h1 class="title">F3D Web</h1>
...
+                filename = contentDisposition.split('filename=')[1].split(';')[0];
+                console.log('filename constructed from contentDisposition header', filename)
+              } else {
+                try {
+                  filename = mesh_url.split('/').pop()
+                  console.log('filename constructed from mesh_url filename+extension', filename)
+                  const url_extension = '.' + mesh_url.split('.').at(-1) 
+                  if (url_extension.length > 10) {
+                    throw new Error(`Bad extension parsed from filename ${filename}`)
+                  }
```
Meakk, line 256: Is it necessary? If the extension is unknown, F3D will send an error message anyway.

webassembly/app.html
```diff
@@ -223,6 +223,54 @@ <h1 class="title">F3D Web</h1>
...
+                  // filename = `mesh_urlparam.${url_extension}`
```
Meakk, line 257: dead code to remove

Meakk (commented):
Thanks, I confirm it works, nice job :)  
Can you address the few cosmetic comments here, but it looks good to me.  
@snoyer can you also take a quick look at the changes?

jo-chemla:
Alright, just did include the suggested changes, thanks for the feedback! 
And I realized I forgot to push the `? -> #` change (search query to hash) suggested by snoyer, now it is included in this final commit.

webassembly/app.html
```diff
@@ -223,6 +221,49 @@ <h1 class="title">F3D Web</h1>
         // do a first render and start the interactor
         Module.engineInstance.getWindow().render();
         Module.engineInstance.getInteractor().start();
+
+        // Parse search-query mesh url-param or load default mesh file
+        // const params = new URLSearchParams(window.location.search);
+        // Replace first hash with question mark to have real search query parsing and avoid leading # in first parsed urlparam
+        const params = new URLSearchParams(window.location.hash.replace('#', '?'));
```
snoyer, line 228:
```suggestion
        const params = new URLSearchParams(window.location.hash.replace(/^#/, ''));
```
to remove only the `#` at the beginning

snoyer:
Nice feature! Here's a few improvements:

It'd be nice to clarify the order of priority used to get the filename between the download URL, response header, and `extension` parameter. I think it should be: from the header if available, otherwise from the url, and either way force the extension if the parameter is set. Maybe refactor to a `filename_for_model_url(model_url, format, header)` function so we can cover it with tests in the future.

Using the hash is nice and local, it doesn't get sent to the server and doesn't refresh the page when changed, but it therefore needs a listener to work on subsequent changes.

I'd say overall it would need to look something like:
```js
  function load_from_url(){
    const params = new URLSearchParams(window.location.hash.replace('#', '?'));
    // ...
    fetch(model_url).then((response) => {
         // ...
         const filename = filename_for_model_url(model_url, extension, response.headers.get("content-disposition")
         // ...
         openFile(filename);
    }
  }
  addEventListener("hashchange", (event) => {load_from_url()}); // do it whenever the hash changes
  load_from_url(); // do it the first time when page is loaded
```

jo-chemla:
Thanks, just refactored per your instructions, should be good to go! 

>  I think it should be: from the header if available, otherwise from the url, and either way force the extension if the parameter is set.

This is how it works at the moment - rephrased, if extension provided, enforce it, otherwise, get type from header if available, otherwise from url. 

```js
if (extension_parsed) {
  filename = `model.${extension_parsed}`;
} else {
  // If extension is not provided by user, try to get it auto from content-disposition header of url extension 
  if (contentDisposition) {
    filename = contentDisposition.split('filename=')[1].split(';')[0];
  } else {
    try {
      filename = model_url.split('/').pop();
```

webassembly/app.html
```diff
@@ -223,6 +221,57 @@ <h1 class="title">F3D Web</h1>
         // do a first render and start the interactor
         Module.engineInstance.getWindow().render();
         Module.engineInstance.getInteractor().start();
+
+        // Parse mesh from url-param search query hash
+        function filename_for_model_url(model_url, extension_parsed, contentDisposition) {
+          // Build filename given extension urlparam or response header content-disposition
+          let filename;
+          if (extension_parsed) {
+            filename = `model_urlparam.${extension_parsed}`;
```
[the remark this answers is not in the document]
jo-chemla: alright, done!

webassembly/app.html
```diff
@@ -223,6 +221,57 @@ <h1 class="title">F3D Web</h1>
...
+            } else {
+              try {
+                filename = model_url.split('/').pop();
+                const url_extension = '.' + model_url.split('.').at(-1) ;
+              } catch {
+                throw new Error(`Could not infer file extension from content-disposition or url extension. Please enforce it via the extension urlparam`);
+              }
+            }
+          }
+          return filename;
```
snoyer, line 242: oops, submitted the previous remark as a single comment instead of adding to a review, my bad!
```suggestion
          throw ...
```
You can throw here unconditionally, if we hit a good code path we would have returned already. That also ensure the function can never return an undefined `filename`

jo-chemla: Ah true, just removed the try/catch and throw blocks. Thanks again!

webassembly/app.html
```diff
@@ -223,6 +221,51 @@ <h1 class="title">F3D Web</h1>
...
+        }
+        function load_from_url(){
+          // Parse search-query mesh url-param or load default mesh file
+          // const params = new URLSearchParams(window.location.search);
+          // Replace first hash with question mark to have real search query parsing and avoid leading # in first parsed urlparam
+          const params = new URLSearchParams(window.location.hash.replace(/^#/, '?'));
+          const model_url_passed = params.get("mesh");
```
Meakk, line 245: files are not only meshes, they can also contain materials and lights.  
Can we rename the option to `model` instead?

webassembly/app.html
snoyer: `filename` doesn't exist anymore so you need to remove this last `return`

jo-chemla:
Thanks for the last suggestions, I included them and just pushed the final commit!

webassembly/app.html
```diff
@@ -223,6 +221,49 @@ <h1 class="title">F3D Web</h1>
         // do a first render and start the interactor
         Module.engineInstance.getWindow().render();
         Module.engineInstance.getInteractor().start();
+
```
CharlesGueunet, line 224: Test

Meakk (approved):

PR #1599, opened 2024-08-31
options: add a missing static

mwestphal:
Also remove an outdated todo


mwestphal:
does not work for cryptic python related reasons, closing.

PR #1610, opened 2024-09-08
Quake MDL format feature

Youva:
Adds class vtkQuakeMDLImporter that reads .MDL files.

The class : 
- Reads vertices and triangle data from the format to display a model.
- Reads a texture and texture coordinates from the format to display onto the model.
- Reads animations, stored as an array of vertices in the file, and saves them as a vector of polydata.
- Groups animations and displays them.
- Adds light sources to the scene.


mwestphal:
i think you already have a PR, why opening another ?

PR #1614, opened 2024-09-10
minor fixes to `TestSDKHelpers.h`

snoyer:
`TestSDKHelpers.h` is missing some includes, which isn't a problem as they are currently always already included before any use of the file, but still, let's have them in there.

Also `class TestSDKHelpers` only has static methods and holds no state so it can be a namespace instead

### DIFF
--- a/library/testing/TestSDKHelpers.h
+++ b/library/testing/TestSDKHelpers.h
@@ -1,6 +1,9 @@
 #ifndef TestSDKHelpers_h
 #define TestSDKHelpers_h
 
+#include <image.h>
+#include <window.h>
+
 #include <fstream>
 #include <iostream>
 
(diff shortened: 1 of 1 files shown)


library/testing/TestSDKHelpers.h
```diff
@@ -1,61 +1,65 @@
...
 #define TestSDKHelpers_h
 
+#include <image.h>
+#include <window.h>
+
 #include <fstream>
 #include <iostream>
 
-class TestSDKHelpers
+namespace TestSDKHelpers
```
mwestphal, line 10: I think you wanted to rename this file ?

snoyer: I thought it'd be nice if only the files containing actual tests were called `Test*` but I can't think of a better name for this one... it was probably not a very useful point to begin with anyway so unless it resonate with you we can just forget about it

mwestphal: fine by me :)

PR #1619, opened 2024-09-13
WSL: Fixes related to the errors caused by Windows Subsystem Linux's strict compilation rules.

Schwarzemann:
The first fix involves the `WriteEvent` method in `vtkF3DInteractorEventRecorder`. The mentioned method passed down an extra argument called `void* callData` was never declared in the original header file. This caused a compilation error in WSL. To fix it I overloaded the `WriteEvent` function to handle additional argument as expected. For the second fix I declared the method `Clear()`. This method was used in `vtkF3DInteractorEventRecorder` with no declarations. I also improved on `SetInteractor` a bit so it works better in strict compilation environments. This improvement involves checking for null pointers, cleanup and state management.


mwestphal:
Thanks for opening this PR ! Any specific reason why you closed it right away ?

Schwarzemann:
Certain CI tests were failing I will fix them and open a new PR.

mwestphal:
No need, you can keep it open and push the fix in tha same branch,the PR will update itself :)

PR #1625, opened 2024-09-15
WSL: Fixes related to the errors caused by Windows Subsystem Linux's strict compilation rules.

Schwarzemann:
The error that I have encountered was related to the mismatch between the number of arguments passed to the WriteEvent function. The `vtkInteractorEventRecorder::WriteEvent` expects 6 arguments but the actual code calls this function with 7 arguments. The arguments are listed as; 

1. `vtkCommand::GetStringFromEventId(event) -> const char*`
2. `rwi->GetEventPosition() -> int*`
3. `mod -> int`
4. `rwi->GetKeyCode() -> int`
5. `rwi->GetRepeatCount() -> int`
6. `rwi->GetKeySym() -> char*`
7. `callData` -> This is an extra argument that isn't defined in `WriteEvent` function signature.

I removed the mentioned argument and now it seems to be working just fine. I will implement and handle `callData` separately in my next commit. I also implemented the `Clear()` function which wasn't properly defined before and added a placeholder which I will also implement.

### DIFF
--- a/vtkext/private/module/vtkF3DInteractorEventRecorder.cxx
+++ b/vtkext/private/module/vtkF3DInteractorEventRecorder.cxx
@@ -34,6 +34,12 @@ void vtkF3DInteractorEventRecorder::SetInteractor(vtkRenderWindowInteractor* int
 }
 
 //------------------------------------------------------------------------------
+void vtkF3DInteractorEventRecorder::Clear()
+{
+  // Place holder
+}
+
+//------------------------------------------------------------------------------
 void vtkF3DInteractorEventRecorder::ProcessEvents(
   vtkObject* object, unsigned long event, void* clientData, void* callData)
 {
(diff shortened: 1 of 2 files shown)


mwestphal:
> callData -> This is an extra argument that isn't defined in WriteEvent function signature.

It is defined: https://vtk.org/doc/nightly/html/classvtkInteractorEventRecorder.html#a5b542ad9081fa30ac4ffe6a2f43de5a4

mwestphal:
> . I also implemented the Clear() function which wasn't properly defined before and added a placeholder which I will also implement.

Why would that be needed ?

Schwarzemann:
Oops my bad this fix seems to only work `VTK 9.1`. 

> > . I also implemented the Clear() function which wasn't properly defined before and added a placeholder which I will also implement.
> 
> Why would that be needed ?

It was failing for me but this might be related to the version as well so I'll check from my end. This can be closed I guess.

mwestphal:
Lets not close this and try to understand what is happening

mwestphal:
Ha right, F3D require VTK 9.2.6 now. this fix is not mergeable Im afraid.

PR #1630, opened 2024-09-21
Added libInformation and readerInformation to the python bindings

Nokse22:


mwestphal:
That looks nice, could you add tests for it ?

Please review @snoyer @Meakk

Nokse22:
> That looks nice, could you add tests for it ?

Yes, adding them.

python/F3DPythonBindings.cxx
```diff
@@ -334,7 +334,33 @@ PYBIND11_MODULE(pyf3d, module)
...
     .def_static(
       "autoload_plugins", &f3d::engine::autoloadPlugins, "Automatically load internal plugins")
-    .def_static("get_plugins_list", &f3d::engine::getPluginsList);
+    .def_static("get_plugins_list", &f3d::engine::getPluginsList)
+    .def_static("get_lib_info", &f3d::engine::getLibInfo, py::return_value_policy::reference)
+    .def_static("get_readers_info", &f3d::engine::getReadersInfo);
+
+  // libInformation
+  py::class_<f3d::engine::libInformation>(module, "LibInformation")
+    .def(py::init<>())
```
snoyer, line 343:
```suggestion
```
Should not be needed (same for `ReaderInformation`) unless we need to create this instances from python code for some reason?

Nokse22: Ok

Nokse22:
I just check the type in the tests, let me know if I need to check the actual content.

snoyer:
> I just check the type in the tests, let me know if I need to check the actual content.

I'd say it's not necessary to check for exact content but I'd check for non-emptiness in addition to type where applicable (`assert isinstance(xs, list) and xs` instead of just `assert isinstance(xs, list)` for lists and same for `dict` and `str` values)

Also you could rename `test_plugins.py` to `test_engine.py` (should have been done way back when we started using `pytest`, better late than never) and add your tests in there to limit the number of files.

mwestphal (approved):

mwestphal:
@snoyer please review

snoyer:
looks good

Issue #1638, opened 2024-09-27
`--camera-index` should override other camera options


mwestphal:
**Describe the bug**
F3D has many camera option but there is one that rely on camera provided by the file, however it should override other camera options but it does not

**To Reproduce**
 1. Open the file using `f3d --dry-run Cameras.gltf --camera-index=0 --camera-position=0,0,0`
 2. Camera should be positioned correctly but it is not

**Expected behavior**
Camera positioned as if camera-position was not provided

**Additional context**
In F3DStarter.cxx:1075, SetupCamera is always called. It should only be called if camera-index has no value, as its an std::optional

Also a test should be added.

mwestphal:
Of course, I just assigned it to you

mwestphal:
Fixed by #1639

PR #1639, opened 2024-09-28
Fix: --camera-index should override other camera options #1638

dk1242:
I have added an extra variable inside **F3DAppOptions** to track if `--camera-index` is passed through CLI arguments.

Fixes for: #1638

### DIFF
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -1101,7 +1101,7 @@ void F3DStarter::LoadFile(int index, bool relativeIndex)
         f3d::log::error("Could not load file: ", ex.what());
       }
 
-      if (!this->Internals->AppOptions.cameraIndexPassed && !this->Internals->AppOptions.NoRender)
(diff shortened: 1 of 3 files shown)


application/F3DStarter.cxx
```diff
@@ -1095,7 +1101,7 @@ void F3DStarter::LoadFile(int index, bool relativeIndex)
         f3d::log::error("Could not load file: ", ex.what());
       }
 
-      if (!this->Internals->AppOptions.NoRender)
+      if (!this->Internals->AppOptions.cameraIndexPassed && !this->Internals->AppOptions.NoRender)
```
mwestphal, line 1104: cameraIndexPassed is not needed, just check `scene.camera.index.has_value()`

testing/baselines/TestCameraIndexConfiguration.png
mwestphal, line 1: Hum, it looks like this behavior was tested so it may be intended.

mwestphal: nevermind, all good.

mwestphal (approved):

Issue #1636, opened 2024-09-27
`--help` argument priority


snoyer:
**Issue**
Running `f3d foo.bar --upp=y -h` exits after printing the suggestion to correct the option typo and does not show the help message.

The arguments (`--help`, `--version`, ...) that output a message and exit early instead of starting the F3D GUI should be checked and handled first.

**Suggested fix**

In F3DOptionTools.cxx, invert the order to check for unknow options and F3DExNoProcess options.

brown-eric:
Hey! I'm new to this, how could I help with this issue?

mwestphal:
Hey @brown-eric , could you come to our discord ? I'll be happy to mentor you :)

mwestphal:
https://discord.f3d.com

SlayerPower:
omg bro eres el mejor jaja, nose pe como que s chakali, o algo pe, pues "maanna taamaanngilanga"
 literalmente lo que dice alli, igual gracias por el trabjao

mwestphal:
> omg bro eres el mejor jaja, nose pe como que s chakali, o algo pe, pues "maanna taamaanngilanga" literalmente lo que dice alli, igual gracias por el trabjao

@SlayerPower this is an english speaking repository.

snoyer:
fixed by #1644, thanks @brown-eric!

PR #1644, opened 2024-10-01
Fix --help option priority

brown-eric:
Fix [https://github.com/f3d-app/f3d/issues/1636](url)
Switched the order of checking for unknown and checking for help/version arguments. This ensures that if a user types something like f3d --help --unknown, the help message will be shown rather than just the error about the unknown option.

### DIFF
--- a/application/testing/CMakeLists.txt
+++ b/application/testing/CMakeLists.txt
@@ -983,6 +983,12 @@ f3d_test(NAME TestBadRefNoOutput DATA cow.vtp ARGS --ref=${F3D_SOURCE_DIR}/testi
 # Test failure with a bad interaction play file, please do not create a dummy.log
 f3d_test(NAME TestPlayNoFile DATA cow.vtp ARGS --interaction-test-play=${CMAKE_BINARY_DIR}/Testing/Temporary/dummy.log WILL_FAIL)
 
+# Test that --help is displayed even when there is an unknown option
+f3d_test(NAME TestHelpPrecedenceWithUnknownOption ARGS --help --unknown REGEXP "F3D Command Line Help"  NO_BASELINE)
+
+# Test that --version is displayed even when there is an unknown option
+f3d_test(NAME TestVersionPrecedenceWithUnknownOption ARGS --version --unknown REGEXP "F3D Version" NO_BASELINE)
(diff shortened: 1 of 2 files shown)


mwestphal (changes requested):
- Please add a link to the issue in the description, eg: "Fix #1636 "
 - Please add a test for this behavior in application/testing/CMakeLists.txt, you can look at `TestHelp` for inspiration.

brown-eric:
Updated by adding testcases for --help or --version before unknown to make sure the code behaves as intended

application/testing/CMakeLists.txt
```diff
@@ -983,6 +983,12 @@ f3d_test(NAME TestBadRefNoOutput DATA cow.vtp ARGS --ref=${F3D_SOURCE_DIR}/testi
 # Test failure with a bad interaction play file, please do not create a dummy.log
 f3d_test(NAME TestPlayNoFile DATA cow.vtp ARGS --interaction-test-play=${CMAKE_BINARY_DIR}/Testing/Temporary/dummy.log WILL_FAIL)
 
+# Test that --help is displayed even when there is an unknown option
+f3d_test(NAME TestHelpPrecedenceWithUnknownOption ARGS --help --unknown REGEXP "F3D Command Line Help"  NO_BASELINE)
```
mwestphal, line 987: you may to check for "Usage:" instead, as the `TestHelp` does.

application/testing/CMakeLists.txt
```diff
@@ -983,6 +983,12 @@ f3d_test(NAME TestBadRefNoOutput DATA cow.vtp ARGS --ref=${F3D_SOURCE_DIR}/testi
...
+
+# Test that --version is displayed even when there is an unknown option
+f3d_test(NAME TestVersionPrecedenceWithUnknownOption ARGS --version --unknown REGEXP "F3D Version" NO_BASELINE)
```
mwestphal, line 990: you may want to check for `Version:` as the `TestVersion` does.

mwestphal (changes requested):
you can run your test locally to check them :)

mwestphal (approved):
@snoyer please review

Issue #1651, opened 2024-10-06
Volume Rendering require coloring array to be set to work correctly


mwestphal:
**Describe the bug**
F3D recently decoupled `--scalars` into `--scalar-coloring` and `--coloring-array` option, however this is an unepexted effect on `--volume` behavior when not using `--scalar-coloring`

**To Reproduce**
 1. Open the file using `f3d --dry-run ../src/testing/data/waveletArrays.vti -v --coloring-array=Result -b`
 2. Incorrect array is used

![a](https://github.com/user-attachments/assets/aabf8a1f-b630-420e-9ab4-0314aaf33c1e)

**Expected behavior**
Correct array is used for rendering

**F3D Information**
 Paste the content of `f3d --version`: master

**Additional context**

A workaround is to have `-s` in the command line.

Should not be fixed before https://github.com/f3d-app/f3d/pull/1609 is merged

This is caused by the `vtkF3DRendererWithColoring::SetColoring` with `enable` set to false which in turn ignore the array name.

mwestphal:
Fixing this is more work than I expected due to design choices in the vtkF3DRenderer.
I think creating a new class for handling the coloring cycle logic would be beneficial to both the vtkF3DMetaImporter and vtkF3DRenderer.

PR #1672, opened 2024-10-22
Move coloring info logic into a dedicated handler

mwestphal:
- Move ColoringInfo logic out of vtkF3DMetaImporter and vtkF3DRenderer into a dedicated F3DColoringInfoHandler
 - F3DColoringInfoHandler contains ColoringInfo maps as well as current coloring state
 - F3DColoringInfoHandler is owned by the meta importer and accessed by the renderer
 - Change the API in vtkF3DRenderer
 - vtkF3DRenderer now handles arrayName and EnableColoring as standard properties
 - Inform user when coloring is forced
 - Fix small issues in vtkF3DRenderer with log and coverage
 - Add/Update tests
 - Update docs
 - Fix https://github.com/f3d-app/f3d/issues/1651


vtkext/private/module/F3DColoringInfoHandler.cxx
```diff
@@ -0,0 +1,190 @@
...
+}
+
+//----------------------------------------------------------------------------
+void F3DColoringInfoHandler::CycleColoringArray(bool cycleToNonColoring)
+{
+  auto& data =
+    this->CurrentUsingCellData ? this->CellDataColoringInfo : this->PointDataColoringInfo;
+  if (!this->Coloring)
+  {
+    if(data.size() == 0)
```
snoyer, line 164:
```suggestion
    if(data.empty())
```

vtkext/private/module/F3DColoringInfoHandler.cxx
```diff
@@ -0,0 +1,190 @@
...
+
+  if (!enable)
+  {
+    // Not coloring
+    this->Coloring = false;
+  }
+  else if (nIndices == 0)
+  {
+    // Trying to color but no array available
+    F3DLog::Print(F3DLog::Severity::Debug, "No array to color with");
```
snoyer, line 119:
```suggestion
    if(!quiet) F3DLog::Print(F3DLog::Severity::Debug, "No array to color with");
```
feels like you need this test here

vtkext/private/module/F3DColoringInfoHandler.cxx
```diff
@@ -0,0 +1,193 @@
...
+  if (!this->CurrentColoringIter.has_value())
+  {
+    if(data.empty())
+    {
+      return;
+    }
+    else
+    {
+      this->CurrentColoringIter = data.begin();
+    }
```
snoyer, line 175: either
```suggestion
    if(!data.empty())
    {
      this->CurrentColoringIter = data.begin();
    }
```
or
```suggestion
    if(data.empty())
    {
      return;
    }
    this->CurrentColoringIter = data.begin();
```
to avoid the mix of complete branching and early return?

vtkext/private/module/F3DColoringInfoHandler.cxx
```diff
@@ -0,0 +1,193 @@
...
+  }
+  else
+  {
+    // XXX: Iter is always valid here
+    ++(this->CurrentColoringIter.value());
+    if (this->CurrentColoringIter.value() == data.end())
```
snoyer, line 181:
```suggestion
    if (++this->CurrentColoringIter.value() == data.end())
```

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -2739,30 +2627,39 @@ void vtkF3DRenderer::ConfigureRangeAndCTFForColoring(
...
+      // Cycle component if the current one is not valid
+      this->CycleComponentForColoring();
+    }
   }
   else
   {
-    // Cycle through arrays looping back to -1
-    // -1 0 1 2 -1 0 1 2 ...
-    this->ArrayIndexForColoring = (this->ArrayIndexForColoring + 2) % (nIndex + 1) - 1;
+    this->SetArrayNameForColoring(std::optional<std::string>());
```
snoyer, line 2662:
```suggestion
    this->SetArrayNameForColoring(std::nullopt);
```

vtkext/private/module/F3DColoringInfoHandler.cxx
```diff
@@ -0,0 +1,187 @@
...
+{
+  // XXX: This assumes importer do not import actors with an empty input
+  assert(dataset);
+
+  // Recover all possible names
+  std::set<std::string> arrayNames;
+
+  vtkDataSetAttributes* attr = useCellData
+    ? static_cast<vtkDataSetAttributes*>(dataset->GetCellData())
+    : static_cast<vtkDataSetAttributes*>(dataset->GetPointData());
```
Meakk, line 31: Do you really need to cast since it's a base class for both?

vtkext/private/module/F3DColoringInfoHandler.cxx
```diff
@@ -0,0 +1,187 @@
...
+            info.ComponentNames.emplace_back(compName ? compName : "");
+          }
+        }
+      }
+    }
+  }
+}
+
+//----------------------------------------------------------------------------
+std::optional<F3DColoringInfoHandler::ColoringInfo> F3DColoringInfoHandler::SetCurrentColoring(bool enable, bool useCellData, std::optional<std::string> arrayName, bool quiet)
```
Meakk, line 104: That's a lot of `bool` arguments, which will generate cryptic calls. Can we split this function in several calls like `SetEnableColoring`, `SetColoringUseCellData`, `SetColoringArrayName` or does it have to be in the same function?

Also, `arrayName` should be const ref.

vtkext/private/module/F3DColoringInfoHandler.cxx
```diff
@@ -0,0 +1,187 @@
...
+std::optional<F3DColoringInfoHandler::ColoringInfo> F3DColoringInfoHandler::GetCurrentColoringInfo() const
+{
+  if (this->CurrentColoringIter.has_value())
+  {
+    return this->CurrentColoringIter.value()->second;
+  }
+  else
+  {
+    return std::nullopt;
+  }
```
Meakk, line 158: useless `else`
```suggestion
  return std::nullopt;
```

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -338,7 +336,7 @@ void vtkF3DRenderer::InitializeUpVector(const std::string& upString)
   }
   else
   {
-    F3DLog::Print(F3DLog::Severity::Warning, upString + " is not a valid up direction");
+    F3DLog::Print(F3DLog::Severity::Warning, upString + " is not a valid up direction\n");
```
Meakk, line 339: Don't we already add a `\n` automatically?

mwestphal: Ill need to recheck on that.

vtkext/private/module/F3DColoringInfoHandler.cxx
mwestphal: It is possible but then I need to add a method that actually "triggers" the array lookup. That doesnt seems necessary to me as this method is very much internal to our vtkext private module and used only in vtkF3DRenderer.

Agreed on the const ref

vtkext/private/module/F3DColoringInfoHandler.cxx
mwestphal: I do, without it compiler complains that they are not the same type.

Meakk: Ok I remember I already encounter something like that before.

vtkext/private/module/vtkF3DRenderer.cxx
mwestphal: we do not, Ive fixed all instance, but I could put that in a dedicated MR.

Meakk: > I could put that in a dedicated MR.

Would be better yes

Meakk (approved):

Issue #228, opened 2022-02-15
possibility to setting number of color discretization


xyont:
dear Developers,

i did not know is this feature available yet or not, several times i read the documentation but i can not find for such purpose.

Paraview has capability to setting number of color discretization, hopefully is not complex things and F3D also capable to do the same. it's a common in engineering representation. 

best regards,
SN

![F3d_256discrettation](https://user-images.githubusercontent.com/72367311/153990831-08c6b19c-939f-458f-95a7-533fc12d4b28.jpg)
![Paravie_Number_Color_Discrtization_12](https://user-images.githubusercontent.com/72367311/153990832-33548f58-3088-4bcd-aaed-a4c927b33050.jpg)

mwestphal:
What about using Step color space ?

Meakk:
I think it's unrelated to step color space, just the number of computed colors in the lookup table.  
We could use `vtkDiscretizableColorTransferFunction` and add a new option to [set the number of colors](https://vtk.org/doc/nightly/html/classvtkDiscretizableColorTransferFunction.html#add05ecb2a72eeaec91588366fa7ea64d).

mwestphal:
Indeed, but being able to choose the step color space resolve the usecase in a different way and open the door to other color spaces.

mwestphal:
That being said, adding the number of colors should be quite an easy feature to add. Feel free to contribute if you can @xyont

xyont:
i'm new to graphics especially in VTK, not a really programmer also. however, it's interesting subject in which lead me to start learn something.

mwestphal:
Added context and suggested solution.

MarcosDanielTorres:
Hi! Maybe I can take this?

mwestphal:
Of course! Feel free to join our discord if you want :)

https://discord.f3d.app

mwestphal:
Unassigning as requested by @gapry

hant-hub:
Hello, I would like to work on this.

mwestphal:
Assigned ! :)

mwestphal:
Fixed by #2264, closing

PR #1678, opened 2024-10-25
Feature: Add an option to control the discretization of colormap

gapry:
## Related Issue:
This work addresses the Issue: https://github.com/f3d-app/f3d/issues/228

## Current Progress 
Over the past two weeks, I have made significant progress on https://github.com/f3d-app/f3d/issues/228, with my work available for review in the following commits:

- https://github.com/gapry/f3d/commit/6af934026d0578794ef06d30a51b50c3584bd441
- https://github.com/gapry/f3d/commit/9f9a2e9494aa7ea9f981121ad6ae94e28a9e1dfb

These commits introduce a new CLI option and extend the [vtkF3DRenderer](https://github.com/f3d-app/f3d/blob/master/vtkext/private/module/vtkF3DRenderer.h) class to handle this option.

Please note that the PR isn’t fully complete yet, as it does not fully resolve https://github.com/f3d-app/f3d/issues/228. I will continue to work on it to ensure that all aspects are addressed. 

Since I’ve already spent two weeks on this task, I wanted to report my progress, especially as we’re working remotely.

## Next Steps
I will review the VTK documentation to learn how to adjust color settings, specifically focusing on the [vtkDiscretizableColorTransferFunction](https://vtk.org/doc/nightly/html/classvtkDiscretizableColorTransferFunction.html#add05ecb2a72eeaec91588366fa7ea64d).

### DIFF
--- a/vtkext/private/module/vtkF3DRenderer.cxx
+++ b/vtkext/private/module/vtkF3DRenderer.cxx
@@ -2127,70 +2127,31 @@ void vtkF3DRenderer::SetColormap(const std::vector<double>& colormap)
   }
 }
 
+// ./build/bin/f3d assets/f3d-data/dragon.vtu --colormap=0,1,0,0,1,0,0,1 -xtgansb --coloring-component=0 --colormap-discretization=4
 void vtkF3DRenderer::SetColorDiscretization(const int discretization) 
 {
   std::string LogMsg1 = std::string("discretization value = ") + std::to_string(discretization);
   F3DLog::Print(F3DLog::Severity::Info, LogMsg1);
 
-  if (discretization > 0) 
+  if (discretization > 0)
   {
-    vtkSmartPointer<vtkDiscretizableColorTransferFunction> discretizableCTF = vtkSmartPointer<vtkDiscretizableColorTransferFunction>::New();
-
-    // this->ColorTransferFunction = vtkSmartPointer<vtkDiscretizableColorTransferFunction>::New();;
-    // vtkDiscretizableColorTransferFunction* const discretizableCTF = vtkDiscretizableColorTransferFunction::SafeDownCast(this->ColorTransferFunction);
-
-    if(discretizableCTF != nullptr) {
-      F3DLog::Print(F3DLog::Severity::Info, "discretizableCTF is not null");
+    this->ColorTransferFunction = vtkSmartPointer<vtkDiscretizableColorTransferFunction>::New();
+    vtkDiscretizableColorTransferFunction* const discretizableCTF = vtkDiscretizableColorTransferFunction::SafeDownCast(this->ColorTransferFunction);
     
-      discretizableCTF->SetNumberOfValues(discretization);
-      discretizableCTF->DiscretizeOn();
+    discretizableCTF->DiscretizeOn();
+    discretizableCTF->SetNumberOfValues(discretization);
 
-      std::vector<double> colormap = this->Colormap;
-      const int numCompents = colormap.size() / 4; 
-
-      std::string LogMsg2 = std::string("numCompents = ") + std::to_string(numCompents);
-      F3DLog::Print(F3DLog::Severity::Info, LogMsg2);
-
-      for (int i = 0; i < numCompents; i++) 
-      {
-        const double v = colormap[i * 4 + 0]; // Scalar
-        const double r = colormap[i * 4 + 1]; // Red
-        const double g = colormap[i * 4 + 2]; // Green
-        const double b = colormap[i * 4 + 3]; // Blue
-
-        std::string LogMsg4 = std::string("v, r, g, b = ") + std::to_string(v) + ", " 
-                                                           + std::to_string(r) + ", " 
-                                                           + std::to_string(g) + ", "
-                                                           + std::to_string(b);
-        F3DLog::Print(F3DLog::Severity::Info, LogMsg4);
-
-        discretizableCTF->AddRGBPoint(v, r, g, b);
-      }
-
-      // const int numCompents = discretizableCTF->GetSize();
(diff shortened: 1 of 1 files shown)


library/options.json
```diff
@@ -246,6 +246,10 @@
         "type": "double_vector",
         "default_value": "0.0, 0.0, 0.0, 0.0, 0.4, 0.9, 0.0, 0.0, 0.8, 0.9, 0.9, 0.0, 1.0, 1.0, 1.0, 1.0"
       },
+      "discretization" : {
+        "type": "bool",
```
mwestphal, line 250: This option definitely should not be a bool but an int I think.

mwestphal:
@gapry any news on this ?

mwestphal:
You need to install the SDK: 

`cmake --install . --component sdk`

from: https://f3d.app/doc/dev/BUILD.html


but you do not need really to build the libf3d examples to do this :)

mwestphal:
Hi @gapry 

Do you need any help moving forward with this ?

mwestphal:
Hi @gapry

Do you need any help moving forward with this ?

mwestphal:
> so, I don't know how to define the type. e.g. vector<RGB>, vector<RGBA>, ... , etc ???

What do you mean ? You do not need to define a new type. You can just change the discretization of the colormap in the vtkF3DRenderer in `ConfigureRangeAndCTFForColoring` by using the VTK method: vtkDiscretizableColorTransferFunction::SetNumberOfValues
https://vtk.org/doc/nightly/html/classvtkDiscretizableColorTransferFunction.html#add05ecb2a72eeaec91588366fa7ea64d

mwestphal:
From https://f3d.app/doc/user/OPTIONS.html

```
Set a custom colormap for the coloring.
This is a list of colors in the format val1,red1,green1,blue1,...,valN,redN,greenN,blueN
where all values are in the range (0,1).
Ignored if --colormap-file option is specified.
Use with the scalar option.
```

Opacity is never mentioned. There are four values because the first value is the scalar value.

VAL R G B

snoyer:
> Unfortunately, this confirms that my implementation is incorrect.

`--colormap=1.0,255.0,127.0,63.0` will give you a colormap with a single color regardless of your implementation tho. Maybe your implementation works but you just can't tell?
dragon_1


Try `--colormap=0,1,0,0,1,0,0,1` to map value `0` to red and `1` to blue
dragon_2

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -2472,6 +2489,12 @@ void vtkF3DRenderer::ConfigureRangeAndCTFForColoring(
     return;
   }
 
+  // Set Discretization
+  if(!this->DiscretizableColorTransferFunctionConfigured) {
+    this->DiscretizableColorTransferFunction->SetNumberOfValues(this->Discretization);
```
mwestphal, line 2494: Can you do that directly on the existing colorTransferFunction

mwestphal: I think you misunderstood, you can use SetNumberOfValues on the existing colorTransferFunction ?

snoyer:
> If my discretization of the colormap implementation is correct, the rendering result should look like this, right?
Yes you should see distinct bands instead of a smooth gradient

mwestphal:
Do you need any help moving forward @gapry ?

snoyer:
> Observed Behavior:
> 
>     * When using `--colormap=1,0,0,1,1,0,1,0,1,1,0,0`, the rendering only applies the last color in the list (red).
> 
>     * When using `--colormap=1,0,0,1,1,0,1,0`, the rendering applies only the last color (green).
> 
> 
> It seems that instead of applying the full colormap, f3d is using only the last specified color for rendering. I’m unsure why this happens. Any insights would be appreciated. Thanks 🙏

`--colormap` format is `t,r,g,b,t,r,g,b...` and both of your examples here map all their colors to the value `t=1`. see https://github.com/f3d-app/f3d/pull/1678#issuecomment-2508871384 for a 2 color example that should work for basic testing

snoyer:
> ```shell
> ./build/bin/f3d assets/f3d-data/dragon.vtu -xtganseb --colormap=0,1,0,0,1,0,0,1 --colormap-discretization=2

now you have a proper colormap but you're visualizing the normals magnitude and all the values are very close to `1` (as you can see on the colorbar in your screenshot). Try adding `--coloring-component=0` to visualize the X component which varies between `-1` and `+1`

snoyer:
@gapry this is where you should apply the disrcretization IMO (@mwestphal can hopefully confirm): https://github.com/f3d-app/f3d/blob/2ac4c0780716dda71fb7f36b94fd6ca6edbe6177/vtkext/private/module/vtkF3DRenderer.cxx#L2527

moving some of your code over there:
```diff
-  this->ColorTransferFunction = vtkSmartPointer<vtkColorTransferFunction>::New();
+  size_t xxx = 4;
+  if (xxx > 0)
+  {
+    this->ColorTransferFunction = vtkSmartPointer<vtkDiscretizableColorTransferFunction>::New();
+    vtkDiscretizableColorTransferFunction* const discretizableCTF =
+      vtkDiscretizableColorTransferFunction::SafeDownCast(this->ColorTransferFunction);
+    discretizableCTF->DiscretizeOn();
+    discretizableCTF->SetNumberOfValues(xxx);
+  }
+  else
+  {
+    this->ColorTransferFunction = vtkSmartPointer<vtkColorTransferFunction>::New();
+  }
```
I get:
image

so I guess that's the core concept, now how to connect everything properly is up to you ;)

mwestphal:
Yes that looks correct @snoyer

mwestphal:
Hi @gapry 

How is it going ? Do you need any help ?

mwestphal:
Hi @gapry

How is it going ? Do you need any help ?

Issue #595, opened 2023-02-01
Command line / Scripting interactions in a vim-like manner


mwestphal:
Let user use small command to control F3D, based on the libf3d::options with a unified api, eg:

 - `:render.effect.ssao toggle` 
 - `:render.effect.ssao on` 
 - `:render.effect.ssao off`
 - `:render.effect.ssao reset` 
 
 It could also include specific action, eg:

- `:camera.fit`

Of course, a libf3d API would need to be designed for this.

more discussions here: https://github.com/f3d-app/f3d/issues/535

First implementation in libf3d added in https://github.com/f3d-app/f3d/pull/1680

Autocompletion will be needed in the future and should be added on the command API using a dedicated callback

mwestphal:
So moving forward with this, here are my thoughts and a design proposal.

**Introducing f3dCommand**

f3dCommand will let user interact with f3d in a text based way and script certain f3d interactions.
It is centered on a few standard commands:

`set`, `inc`, `dec`, `toggle`, `print`, `alias`

And a few specific commands:

`exit`, `reset`.

Standard commands takes a `libf3d` options as arguments. specific commands do not support arguments at all.

**set**

Usage: `set option value`

 - `option` is any of the libf3d option
 - `value` is convertable-from-string value that will be converted to the correct type and passed to the option

Set the value of `option` to `value`. No effect if `value` is incompatible with `option`.

**inc/dec**

Usage: `inc option`

  - `option` is a libf3d option that supports inc/dec
  
 Increase/Decrease the value of `option`. The actual change depends of the option, which will be documented by option. May have no effect on certains options.
 
 Note: CycleScalars and such will be handled here.
 
 **toggle**
 
 Usage: `toggle option`

  - `option` is a libf3d option that supports toggle (boolean)

 Toggle the the value of a boolean option.
 
 Note: toggle animation will be handled here and added as an option instead/in supplement of being a dedicated C++ call.
 
**print**

Usage: `print [option]`

 - `option` is any of the libf3d option

Print the value of an option, without option, print info about the scene (current `?`  binding)

**alias**

Usage: `alias option value`

 - `option` is any of the libf3d option
 - `value` is a string alias to be used for the option

Add an alias for `option` named `value`. `value` can then be used anywhere instead of `option`.
All command-line long options will be provided as alias for the libf3d `option` counterpart by default in F3D application.

[...]

PR #1680, opened 2024-10-27
Add a simple command API

mwestphal:
- Add a interactor::command API
 - Add the following commands:
   - set
   - toggle
   - reset
   - unset
   - print
   -  many specific commands
 - Add dedicated documentation
 - Add python bindings and tests
 - Fix https://github.com/f3d-app/f3d/issues/1685
 - Change options::toggle behavior with optional bool
 - Add tests
 - Add triggerCommand binding for java and wasm
  
  Needed for: 
  - https://github.com/f3d-app/f3d/issues/595
  - https://github.com/f3d-app/f3d/issues/443
  
Other things to do:
 - Use in interaction API: https://github.com/f3d-app/f3d/pull/1687
 - Add a `--command-script`: https://github.com/f3d-app/f3d/issues/1694
 - autocompletion: Added mention in  #595 
 - add defaults API: Another PR, mentionned in https://github.com/f3d-app/f3d/pull/1687


library/src/interactor_impl.cxx
```diff
@@ -764,4 +766,87 @@ void interactor_impl::UpdateRendererAfterInteraction()
...
+  size_t nTokens = results.size();
+  if (nTokens > 3 || nTokens == 0)
+  {
+    log::error("Command: \"", commandString, "\" contains incorrect number of tokens, ignoring");
+    return *this;
+  }
+
+  std::string command = results[0];
+  std::string option = nTokens >= 1 ? results[1] : "";
+  std::string value = nTokens >= 2 ? results[2] : "";
```
snoyer, line 791:
```suggestion
  const std::regex commandRE("(\\S+)"          // command: non-spaces
                             "(?:\\s+(\\S+))?" // maybe option: non-spaces after some space(s)
                             "(?:\\s+(.+))?"   // maybe value: anything after some space(s)
  );
  std::smatch match;
  if (std::regex_match(commandString, match, commandRE))
  {
    const std::string command = match[1];
    const std::string option = match[2];
    const std::string value = match[3];
```
maybe regex is overkill but relying on raw `getline` feels a bit optimistic

mwestphal: reimplemented (without regex)

application/F3DStarter.cxx
```diff
@@ -798,45 +798,80 @@ int F3DStarter::Start(int argc, char** argv)
     this->Internals->ApplyPositionAndResolution();
 
     f3d::interactor& interactor = this->Internals->Engine->getInteractor();
+
+    interactor.addCommandCallback("load_previous_file_group",
+      [this](const std::string&) { return this->LoadRelativeFileGroup(-1); });
```
Meakk, line 803: should we add a warning if an argument is provided but not used?

application/F3DStarter.cxx
```diff
@@ -798,45 +798,80 @@ int F3DStarter::Start(int argc, char** argv)
...
+          return this->LoadRelativeFileGroup(0);
+        }
+        return true;
+      });
+
+    interactor.addCommandCallback("take_screenshot",
+      [this](const std::string& filename) -> bool
+      {
+        // TODO add test
+        std::string localFilename =
```
Meakk, line 829: what if the provided filename is invalid? Is it checked in `SaveScreenshot`?

application/F3DStarter.cxx
```diff
@@ -798,45 +798,80 @@ int F3DStarter::Start(int argc, char** argv)
...
+          filename.empty() ? this->Internals->AppOptions.ScreenshotFilename : filename;
+        this->SaveScreenshot(localFilename);
+        return true;
+      });
+
+    interactor.addCommandCallback("take_minimal_screenshot",
+      [this](const std::string& filename) -> bool
+      {
+        // TODO add test
```
Meakk, line 838: todo

application/F3DStarter.cxx
```diff
@@ -798,45 +798,80 @@ int F3DStarter::Start(int argc, char** argv)
...
+        return true;
+      });
+
     interactor.setKeyPressCallBack(
       [this](int, const std::string& keySym) -> bool
       {
         if (keySym == "Left")
         {
-          return this->LoadRelativeFileGroup(-1);
+          return this->Internals->Engine->getInteractor().triggerCommand(
```
Meakk, line 850: That'd be great to have a `map<string, string>` and just do something like:
```cpp
interactor.setKeyPressCallBack(
      [this](int, const std::string& keySym) -> bool
      {
        try
        {
          return this->Internals->Engine->getInteractor().triggerCommand(commandMap.at(keySym));
        }
        catch(const std::out_of_range&)
        {
          return false;
        }
      }
);
```

doc/libf3d/OPTIONS.md
```diff
@@ -141,6 +141,9 @@ Please note that when accessing optional options, special care must be used, eg:
     std::cout << "Line Width: unset" << std::endl;
   }
 ```
+
+It even more true with the few optional boolean options as std::optional has a boolean operator.
```
Meakk, line 145:
```suggestion
It's even more true with the few optional boolean options as std::optional has an implicit boolean cast operator.
```

doc/user/COMMANDS.md
```diff
@@ -0,0 +1,78 @@
+# Commands
+
+F3D will provides access to commands through interactions configuration and command line widget in the future.
```
Meakk, line 3:
```suggestion
F3D will provide access to commands through interactions configuration and command line widget in the future.
```

doc/user/COMMANDS.md
```diff
@@ -0,0 +1,78 @@
...
+Commands let you trigger specific behavior that may not be available otherwise.
+Please note commands are currently experimental and the behavior, names and arguments may change without deprecation.
+
+Standard commands have the following syntax:
+
+`action libf3d.option.name [values]`
+
+Specific commands have the following syntax:
+
+`specific_action [arg] [values]`
```
Meakk, line 13: I still disagree about differentiation of standard/specific commands.  
For me, it should be documented as `action [args]`.  
Then you can explain below that a few commands are related to options state modification (set/toggle/unset/reset/print) in which case the first argument correspond to the option name.

doc/user/COMMANDS.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Standard commands
+
+`set option.name values`: A command to set a libf3d option, eg: `set scene.up.direction +Z`.
+
+`toggle option.name`: A command to toggle a boolean libf3d option, eg: `toggle ui.scalar_bar`. 
+
+`unset option.name`: A command to unset an optional libf3d option, eg: `unset render.hdri.file`. 
+
+`reset option.name`: A command to reset a libf3d option to its default values, eg: `reset render.background.blur_coc`.
```
Meakk, line 29: What's the difference between `reset` and `unset`? If I'm using `reset` on an optional option, I'm expecting it to be unset, so it feels like it's redundant to me.

doc/user/COMMANDS.md
```diff
@@ -0,0 +1,78 @@
...
+
+`print option.name`: A command to print the value of an libf3d option, eg: `print scene.up.direction`.
+
+### Specific commands
+
+`cycle_animation`:  A specific command to cycle `scene.animation.index` option using model information, no arguments.
```
Meakk, line 35:
```suggestion
`cycle_animation`:  A specific command to cycle `scene.animation.index` option using model information, no argument.
```

Meakk: here and everywhere.

library/private/interactor_impl.h
```diff
@@ -38,6 +38,11 @@ class interactor_impl : public interactor
   interactor& setKeyPressCallBack(std::function<bool(int, std::string)> callBack) override;
   interactor& setDropFilesCallBack(std::function<bool(std::vector<std::string>)> callBack) override;
 
+  interactor& addCommandCallback(
```
Meakk, line 41: I'm wondering if we shouldn't have a `const std::vector<std::string>&` as argument of the callback.  
That way, we can handle splitting arguments at the framework level, and not reimplement it in every callback.  
I think parsing might become more complicated in the future (like handling escaped spaces, handling several consecutive spaces, quoted arguments, ...)

library/src/interactor_impl.cxx
```diff
@@ -172,228 +176,142 @@ class interactor_impl::internals
...
-
     // Available keycodes: None
     switch (keyCode)
     {
       case 'W':
-        self->AnimationManager->CycleAnimation();
-        self->Options.scene.animation.index = self->AnimationManager->GetAnimationIndex();
-        ren->SetAnimationnameInfo(self->AnimationManager->GetAnimationName());
-        render = true;
+        self->Interactor.triggerCommand("cycle_animation");
```
Meakk, line 183: same remark as for the app. A map instead of the switch/case would look better

library/src/interactor_impl.cxx
```diff
@@ -172,228 +176,142 @@ class interactor_impl::internals
...
-                           : ref < 10 ? 10
-                           :            25;
-        // clang-format on
-        /* new intensity in percents */
-        const int newIntensityPct = std::lround(intensity * 100) + (down ? -offsetPp : +offsetPp);
-
-        self->Options.render.light.intensity = std::max(newIntensityPct, 0) / 100.0;
-        render = true;
+        std::string command = "increase_light_intensity";
+        command += rwi->GetShiftKey() ? " negative" : "";
```
Meakk, line 257: Not super fan of this.  
Should we have `increase_light_intensity` and `decrease_light_intensity`?  
Or maybe `light_intensity increase/decrease`?

doc/user/COMMANDS.md
mwestphal: I agree that technically reset and unset is similar, so I could keep only reset.

application/F3DStarter.cxx
mwestphal: feels not necessary to me

application/F3DStarter.cxx
mwestphal: of course

application/F3DStarter.cxx
mwestphal: this will be reworked in https://github.com/f3d-app/f3d/pull/1687, not needed

doc/user/COMMANDS.md
mwestphal: Some specific commands are also related to option state modification, so that not really the right distinction

library/private/interactor_impl.h
mwestphal: > not reimplement it in every callback.

Only a few have it

The problem I have with this is that `set option.vector 1.3, 1.4, 1.5` will be split and then need to be recombined before being given to `options::set`.

library/src/interactor_impl.cxx
mwestphal: same answer

library/src/interactor_impl.cxx
mwestphal: >  Or maybe light_intensity increase/decrease?

That would break the `action` logic.

> increase_light_intensity and decrease_light_intensity

Ill do that.

application/F3DStarter.cxx
mwestphal: looking at the code, `fs::create_directories` is not checked . Should be fixed but unrelated to this PR.

library/private/interactor_impl.h
mwestphal: There is some benefit though, lets discuss pros/cons

doc/user/COMMANDS.md
Meakk: Ok, I agree it may not be the right distinction but I still think there is no reason to document a difference between the standard and specific commands.

doc/user/COMMANDS.md
Meakk: Yes I think `reset` only should be kept

library/private/interactor_impl.h
Meakk: I see you pushed this change. Should I consider we agree and close the discussion?

mwestphal: indeed

library/src/interactor_impl.cxx
```diff
@@ -558,16 +498,97 @@ class interactor_impl::internals
     this->Window.render();
   }
 
+  struct tokenize_exception : public exception
+  {
+    explicit tokenize_exception(const std::string& what = "")
+      : exception(what){}
+  };
+
+  static std::vector<std::string> Tokenize(std::string_view str)
```
Meakk, line 507:
```suggestion
  static std::vector<std::string> Tokenize(const std::string_view& str)
```

library/src/interactor_impl.cxx
```diff
@@ -558,16 +498,97 @@ class interactor_impl::internals
...
+  {
+    std::vector<std::string> tokens;
+    std::string token;
+    const auto accumulate = [&](const char &c) { token.push_back(c); };
+    const auto emit = [&]() {
+      if (!token.empty()) {
+        std::cout<<token<<std::endl;
```
Meakk, line 514: cout

doc/user/COMMANDS.md
mwestphal: The value I see is for documentation and ease of use for user. They would know that using a standard action require an option name.

application/F3DStarter.cxx
mwestphal: wth the new impl, I only check in callbacks that uses the args.

library/src/interactor_impl.cxx
```diff
@@ -558,16 +498,97 @@ class interactor_impl::internals
...
   scene_impl& Scene;
+  interactor_impl& Interactor;
   animationManager* AnimationManager;
 
   vtkSmartPointer<vtkRenderWindowInteractor> VTKInteractor;
   vtkNew<vtkF3DInteractorStyle> Style;
   vtkSmartPointer<vtkF3DInteractorEventRecorder> Recorder;
   std::map<unsigned long, std::pair<int, std::function<void()>>> TimerCallBacks;
 
+  std::map<std::string, std::function<bool(const std::vector<std::string>&)>> CommandCallbacks;
```
Meakk, line 590: I know that `std::unordered_map` needs a hash function but it's faster and we do not need sorting.

doc/user/COMMANDS.md
Meakk: And that should be documented at such in commands documentation. I just disagree about the specific command category.

library/src/interactor_impl.cxx
mwestphal: really does not matter for small maps imo.

library/src/interactor_impl.cxx
mwestphal: nope, thats a string_view

Meakk: and? I agree the copy is cheaper than a string but still

library/src/interactor_impl.cxx
Meakk: I agree but in practice we do not know the size of the map. That's not super important for now I guess anyway.

library/src/interactor_impl.cxx
```diff
@@ -603,6 +826,73 @@ interactor& interactor_impl::setDropFilesCallBack(
   return *this;
 }
 
+//----------------------------------------------------------------------------
+interactor& interactor_impl::addCommandCallback(
+  std::string_view action, std::function<bool(const std::vector<std::string>&)> callback)
+{
+  this->Internals->CommandCallbacks[std::string(action)] = callback;
```
Meakk, line 833: In the end you have a string in the map so why do you use a string_view?  

I think this would make more sense:
```cpp
interactor& interactor_impl::addCommandCallback(
  std::string action, std::function<bool(const std::vector<std::string>&)> callback)
{
  this->Internals->CommandCallbacks[std::move(action)] = callback;
  return *this;
}
```
or
```cpp
interactor& interactor_impl::addCommandCallback(
  std::string&& action, std::function<bool(const std::vector<std::string>&)> callback)
{
  this->Internals->CommandCallbacks[action] = callback;
  return *this;
}
```
but that works only with rvalues.

library/src/interactor_impl.cxx
```diff
@@ -603,6 +826,73 @@ interactor& interactor_impl::setDropFilesCallBack(
...
+  return *this;
+}
+
+//----------------------------------------------------------------------------
+interactor& interactor_impl::removeCommandCallback(std::string_view action)
+{
+  this->Internals->CommandCallbacks.erase(std::string(action));
+  return *this;
+}
```
Meakk, line 842: Just use a const ref string here, no reason to use a string_view
```suggestion
interactor& interactor_impl::removeCommandCallback(const std::string& action)
{
  this->Internals->CommandCallbacks.erase(action);
  return *this;
}
```

library/src/interactor_impl.cxx
```diff
@@ -603,6 +826,73 @@ interactor& interactor_impl::setDropFilesCallBack(
...
+  {
+    tokens = interactor_impl::Internals->Tokenize(command);
+  }
+  catch (const internals::tokenize_exception&)
+  {
+    log::error("Command: unable to tokenize command:\"", command, "\", ignoring");
+    return false;
+  }
+
+  const std::string action = tokens[0];
```
Meakk, line 858:
```suggestion
  const std::string& action = tokens[0];
```

library/src/interactor_impl.cxx
```diff
@@ -603,6 +826,73 @@ interactor& interactor_impl::setDropFilesCallBack(
...
+
+//----------------------------------------------------------------------------
+interactor& interactor_impl::removeCommandCallback(std::string_view action)
+{
+  this->Internals->CommandCallbacks.erase(std::string(action));
+  return *this;
+}
+
+//----------------------------------------------------------------------------
+bool interactor_impl::triggerCommand(std::string_view command)
```
Meakk, line 845: here I think it's justified because the argument doesn't need to be converted to a string.

library/src/interactor_impl.cxx
Meakk: Ok I just checked online and it seems like while const-ref it is possible, it looks like common practice is to just copy it.

library/src/interactor_impl.cxx
```diff
@@ -558,16 +498,92 @@ class interactor_impl::internals
...
 
+  struct tokenize_exception : public exception
+  {
+    explicit tokenize_exception(const std::string& what = "")
+      : exception(what)
+    {
+    }
+  };
+
+  static std::vector<std::string> Tokenize(std::string_view str)
```
Meakk, line 509: Maybe `Tokenize` could be moved in `f3d::utils::tokenize`? While it's not reused yet, maybe it will at some point?  
And I think it's a great candidate for fuzz testing to properly test it so we may need to access it from the test framework.

library/src/interactor_impl.cxx
```diff
@@ -558,16 +498,92 @@ class interactor_impl::internals
...
+    {
+      if (!token.empty())
+      {
+        tokens.push_back(token);
+        token.clear();
+      }
+    };
+    bool escaped = false;
+    char quoted = '\0';
+    for (const char& c : str)
```
Meakk, line 524: if const-ref a string_view is too much, this one too

library/src/interactor_impl.cxx
mwestphal: > why do you use a string_view?

Because wherever I would previously use a `const std::string&`, I should now use a `std::string_view` instead. This is how I understand string_view, I know there is more to it but I think its works in the general case.

>  std::string action

One copy too many

>  std::string&& action

Only rvalues


Whats wrong with string_view ? They are _designed to replace_  string const ref .

library/src/interactor_impl.cxx
mwestphal: this is C++11

library/src/interactor_impl.cxx
mwestphal: I agree to put into utils, but Ill not implement fuzz testing in this PR

library/src/interactor_impl.cxx
mwestphal: there is no char_view afaik ?

library/src/interactor_impl.cxx
mwestphal: the same argument can be made about almost any of the maps in the libf3d. Not worth it imo.

library/src/interactor_impl.cxx
mwestphal: string_view instead

library/src/interactor_impl.cxx
mwestphal: nevermind, you are right, this API should not use string_view I think.

library/src/interactor_impl.cxx
mwestphal: nevermind, you are right, this API should not use string_view I think.

python/F3DPythonBindings.cxx
```diff
@@ -223,6 +223,7 @@ PYBIND11_MODULE(pyf3d, module)
     .def("record_interaction", &f3d::interactor::recordInteraction, "Record an interaction file")
     .def("start", &f3d::interactor::start, "Start the interactor")
     .def("stop", &f3d::interactor::start, "Stop the interactor")
+    .def("trigger_command", &f3d::interactor::triggerCommand, "Trigger a command")
```
mwestphal, line 226: @Meakk do you know If I can add callback bindings ?

library/src/interactor_impl.cxx
```diff
@@ -115,11 +119,12 @@ class interactor_impl::internals
...
     VT_ISOMETRIC
   };
-  static void SetViewOrbit(ViewType view, internals* self)
+
+  void SetViewOrbit(ViewType view)
   {
     vtkNew<vtkMatrix3x3> transform;
-    self->ToEnvironmentSpace(transform);
-    camera& cam = self->Window.getCamera();
+    this->ToEnvironmentSpace(transform);
```
Meakk, line 126: Can you make this change in another PR?

library/src/interactor_impl.cxx
```diff
@@ -172,228 +199,142 @@ class interactor_impl::internals
...
-                           :            25;
-        // clang-format on
-        /* new intensity in percents */
-        const int newIntensityPct = std::lround(intensity * 100) + (down ? -offsetPp : +offsetPp);
-
-        self->Options.render.light.intensity = std::max(newIntensityPct, 0) / 100.0;
-        render = true;
+        std::string command =
+          rwi->GetShiftKey() ? "decrease_light_intensity" : "increase_light_intensity";
+        self->Interactor.triggerCommand(command);
```
Meakk, line 281: Save a copy:
```suggestion
        self->Interactor.triggerCommand(std::move(command));
```

python/F3DPythonBindings.cxx
Meakk: I think it should be possible. See https://pybind11.readthedocs.io/en/stable/advanced/cast/functional.html  
What happens if you just add `.def("add_callback", &f3d::interactor::addCallback, "Add a new command")`?

library/src/interactor_impl.cxx
mwestphal: sure

library/src/interactor_impl.cxx
Meakk: `for (char c : str)`

mwestphal: right

Meakk (approved):

PR #1682, opened 2024-10-29
Log: Add many missing EOL in F3DLog usages

mwestphal:


Meakk (commented):
Is it working as expected on Windows log window? (No additional new line)

Issue #1637, opened 2024-09-27
`--dry-run` option is not working


mwestphal:
**Describe the bug**
F3D usually uses a config file, but there is an option to disable this behavior, `--dry-run`. It seems to be non-functionnal currently

**To Reproduce**
Steps to reproduce the behavior, using a built f3d with BUILD_TESTING enabled:
 1. Open the file using `f3d --config=config_build --dry-run --verbose
 2. F3D appears with axis visible, it should not
 3. Check the output, a config file is read

**Expected behavior**
Config file should not be read

**Additional context**
in `F3DStarter.cxx:655`, although the comment says that dry-run is being checked, its actually not. Add the check here by checking the cli option and position the boolean accordingly. Also add a test for it in `application/testing/CMakeLists.txt`

snoyer:
Would this be the right time to rename this argument to `--no-config` or something?
`--dry-run` is [used in many CLI applications](https://www.google.com/search?q=%22--dry-run%22+site%3Alinux.die.net) to mean "run as normal but don't actually write/change anything" so it's a bit confusing (F3D's `--no-render` is closer to what `--dry-run` usually means)

mwestphal:
Good point @snoyer

mwestphal:
It is open!

mwestphal:
Hum, since you are working on another issue, I'd like to keep this one open for other developpers with no experience on the project if thats ok

Yogesh9000:
sure thats fine

rehanganapathy:
Hey there, this looks interesting, if no one else is working on it, can i pick it up? Thank you so much!
@mwestphal

mwestphal:
Sounds good! @rehanganapathy , go for it :)

mwestphal:
@rehanganapathy any news on this ?

mwestphal:
no feedback, unassigning it

t-h2o:
Hello!

With this patch, the config value is never set.

```patch
 application/F3DStarter.cxx | 6 +++++-
 1 file changed, 5 insertions(+), 1 deletion(-)

diff --git a/application/F3DStarter.cxx b/application/F3DStarter.cxx
index e1cc563a..6b00c5bc 100644
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -713,12 +713,16 @@ int F3DStarter::Start(int argc, char** argv)
   // but this duplicate the initialization value as it is present in
[...]

PR #1690, opened 2024-11-04
Renamed dry-run argument to no-config

rkoudsi:
for issue #1637
this will make it clearer as dry-run does not do what most people expect (not make changes).
This new argument will lead people to search for no-render if they actually intend to use the feature that most projects refer to as dry-run. 
no-config is a better name for this arg.


mwestphal:
Thats nice but you may also want to fix issue https://github.com/f3d-app/f3d/issues/1637. Afaik `--dry-run` is currently non fonctionnal ?

mwestphal:
Done in #1920

Issue #1159, opened 2024-01-11
Distinguish dropping 3D file and dropping a HDRI file using modifiers


mwestphal:
Currently, when dropping a file on F3D window, F3D will load it has a 3D model unless its a .hdr, then it will load it as a HDRI.

This is very practical for most users but break the logical paradigm of F3D where one interaction correspond to one action.

While we will keep this interaction, having a dedicated interaction (Ctrl+Drop and Shift+Drop) for dedicated actions would be beneficial and let user drop any kind of file as an HDRI if needed.

mwestphal:
@kidharb just comment on this issue, Ill then  assign it to you

kidharb:
I will work on this issue

mwestphal:
Need #443

PR #1701, opened 2024-11-10
Add interaction command for dropping files

mwestphal:
- Add a command callback "add_files" in libf3d to add files to a scene
  - Add an interaction command for "Drop" interaction to trigger "add_files"
  - Rework OnDropFiles interaction to trigger a "Drop" interaction with added "files" as arguments of the command
  - Add command callbacks in F3D application to "add_files" , "set_hdri", "add_files_or_set_hdri"
  - Add interactions commands for "Drop" -> "add_files_or_set_hdri", "CTRL+Drop"  -> add_files, "SHIFT+Drop" -> "set_hdri"
  - Remove user drop files callback logic
  - Rework tests to ensure multifile drop is tested
  - Add a modified drops test
  - Rework SDK callback test to test Drop interaction
  - Fix: https://github.com/f3d-app/f3d/issues/1159
  - Add libf3d doc
  - Add doc for the new commands
   
  To do later:

  Add defaults API: https://github.com/f3d-app/f3d/pull/1704
  Add config file control: Another PR
  Add complete interaction doc: Another PR
 Add automatic cheatsheet generation: Another PR


application/F3DStarter.cxx
```diff
@@ -843,23 +843,46 @@ int F3DStarter::Start(int argc, char** argv)
...
-      "Right", f3d::interactor::ModifierKeys::NONE, "load_next_file_group");
-    interactor.addInteractionCommand(
-      "Up", f3d::interactor::ModifierKeys::NONE, "reload_current_file_group");
-    interactor.addInteractionCommand(
-      "Down", f3d::interactor::ModifierKeys::NONE, "add_current_directories");
-    interactor.addInteractionCommand(
-      "F11", f3d::interactor::ModifierKeys::NONE, "take_minimal_screenshot");
-    interactor.addInteractionCommand("F12", f3d::interactor::ModifierKeys::NONE, "take_screenshot");
+    // This replace an existing command in libf3d
+    interactor.addCommandCallback("add_files",
```
Meakk, line 847: I'm wondering if we shouldn't have a way to specify we are replacing an existing command.
This way, the user would be able to catch a change in libf3d command name for example.  
Do you think we could add `replaceCommandCallback` and maybe throw if calling `addCommandCallback` is called with an existing command name?

library/src/interactor_impl.cxx
```diff
@@ -440,6 +360,76 @@ class interactor_impl::internals
   }
 
   //----------------------------------------------------------------------------
+  void TriggerInteractionCommands(const std::string& interaction, const std::string& argsString)
+  {
+    const auto modifiers = [&]()
```
Meakk, line 365: I'm not against functional programming, but I'm wondering if it's justified here.  
You could just in `ModifierKeys modifiers` local variable.

library/src/interactor_impl.cxx
```diff
@@ -440,6 +360,76 @@ class interactor_impl::internals
...
+      {
+        return ModifierKeys::SHIFT;
+      }
+      else
+      {
+        return ModifierKeys::NONE;
+      }
+    };
+
+    const auto to_string = [](const InteractionBind& bind)
```
Meakk, line 388: same here I guess. You can just store the result to a local `std::string`

application/F3DStarter.cxx
mwestphal: I'm not against it, but it does make the API a bit heavier. I do plan to add a `addDefaultCommandCallbacks` and a `removeAllCommandCallbacks` in a dedicated PR, maybe I should add your suggestion in the same PR ?

Meakk: fair enough

Meakk (approved):

PR #1710, opened 2024-11-15
Bump USD version

Meakk:
Bump to 24.11
- Compatible with oneTBB
- Boost no longer required


plugins/usd/module/CMakeLists.txt
```diff
@@ -8,13 +8,6 @@ vtk_module_add_module(f3d::vtkextUSD
 
 vtk_module_link(f3d::vtkextUSD PRIVATE ${PXR_LIBRARIES})
 
-# When built in monolithic mode, include must be specified manually here because
-# the "usd_ms" target does not propagate the directories automatically
-if(POLICY CMP0167)
-  cmake_policy(SET CMP0167 NEW)
-endif()
-find_package(Boost REQUIRED)
```
mwestphal, line 16: should we remove boost from CI then ?

Meakk: I think it's still used by openvdb

mwestphal: right

mwestphal (approved):

PR #1711, opened 2024-11-16
[draft] Interaction description

mwestphal:


library/src/interactor_impl.cxx
```diff
@@ -745,76 +753,186 @@ bool interactor_impl::triggerCommand(std::string_view command)
...
-  this->addBinding(F3D_EXIT_HOTKEY_SYM, ModifierKeys::NONE, "stop_interactor");
-  this->addBinding("Return", ModifierKeys::NONE, "reset_camera");
-  this->addBinding("Space", ModifierKeys::NONE, "toggle_animation");
-  this->addBinding("Drop", ModifierKeys::NONE, "add_files");
+  this->addBinding("W", ModifierKeys::NONE, "cycle_animation", std::bind(docStrVal, "Cycle animation", animStr));
+  this->addBinding("C", ModifierKeys::NONE, "cycle_coloring field", std::bind(docStrVal, "Cycle point/cell data coloring", pointCellStr));
+  this->addBinding("S", ModifierKeys::NONE, "cycle_coloring array", std::bind(docStrVal, "Cycle array to color with", arrayStr));
+  this->addBinding("Y", ModifierKeys::NONE, "cycle_coloring component", std::bind(docStrVal, "Cycle component to color with", compStr));
+  this->addBinding("B", ModifierKeys::NONE, "toggle ui.scalar_bar", std::bind(docStrVal, "Toggle the scalar bar display",
+    [&]() -> std::string { return docTgl(opts.ui.scalar_bar);}));
```
Meakk, line 832: why aren't you using std::bind here?

library/src/interactor_impl.cxx
```diff
@@ -830,6 +948,23 @@ interactor_impl::getBindingInteractions() const
...
+interactor_impl::getBindingsDocumentation() const
+{
+  std::vector<std::tuple<std::string, std::string, std::string>> docs;
+  for (internals::Bind bind : this->Internals->OrderedBinds)
+  {
+    auto docFunc = this->Internals->Bindings[bind].DocumentationCallback;
+    if (docFunc)
+    {
+      auto docPair = docFunc();
+      docs.emplace_back(bind.Format(), docPair.first, docPair.second);
```
Meakk, line 962:
```suggestion
      auto& [desc, val] = docFunc();
      docs.emplace_back(bind.Format(), desc, val);
```

testing/baselines/TestInteractionCheatsheetBlackBG.png
Meakk, line 1: I suppose we'll have to add a `binding group` to organize the bindings, but that can be later

application/F3DStarter.cxx
```diff
@@ -901,23 +905,48 @@ int F3DStarter::Start(int argc, char** argv)
...
   f3d::log::debug("Engine configured");
 
+  // Print bindings list and exits if needed
+  if (this->Internals->AppOptions.BindingsList)
+  {
+    f3d::log::info("Bindings:");
+    for (const auto& [key, desc, val] :
+      this->Internals->Engine->getInteractor().getBindingsDocumentation())
+    {
+      F3DOptionsTools::PrintHelpPair(key, desc, 12);
```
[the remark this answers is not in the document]
mwestphal: its the longest string our Binding have by default:  "ANY+Question"

application/F3DStarter.cxx
```diff
@@ -901,23 +905,48 @@ int F3DStarter::Start(int argc, char** argv)
...
 
-    interactor.addBinding("Left", f3d::interactor::ModifierKeys::NONE, "load_previous_file_group");
-    interactor.addBinding("Right", f3d::interactor::ModifierKeys::NONE, "load_next_file_group");
-    interactor.addBinding("Up", f3d::interactor::ModifierKeys::NONE, "reload_current_file_group");
-    interactor.addBinding("Down", f3d::interactor::ModifierKeys::NONE, "add_current_directories");
-    interactor.addBinding("F11", f3d::interactor::ModifierKeys::NONE, "take_minimal_screenshot");
-    interactor.addBinding("F12", f3d::interactor::ModifierKeys::NONE, "take_screenshot");
+    // "doc"
+    auto docString = [](const std::string& doc) -> std::pair<std::string, std::string>
+    { return std::make_pair(doc, ""); };
```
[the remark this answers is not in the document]
mwestphal: Why is it better ?

library/private/interactor_impl.h
```diff
@@ -44,11 +44,15 @@ class interactor_impl : public interactor
 
   interactor& initBindings() override;
   interactor& addBinding(const std::string& interaction, ModifierKeys modifiers,
-    std::vector<std::string> commands) override;
-  interactor& addBinding(
-    const std::string& interaction, ModifierKeys modifiers, std::string command) override;
+    std::vector<std::string> commands,
+    std::function<std::pair<std::string, std::string>()> documentationCallback = nullptr) override;
```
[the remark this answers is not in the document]
mwestphal: indeed

library/src/interactor_impl.cxx
```diff
@@ -745,76 +753,186 @@ bool interactor_impl::triggerCommand(std::string_view command)
...
+    return valStream.str();
+  };
+
+  // "ON/OFF"
+  auto docTgl = [](bool val) -> std::string
+  { return val ? "ON" : "OFF"; };
+
+  // "ON/OFF/NO SET"
+  auto docTglOpt = [](std::optional<bool> val) -> std::string
+  { return val.has_value() ? (val.value() ? "ON" : "OFF") : "NOT SET"; };
```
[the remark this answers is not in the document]
mwestphal: sure, but this was not introduced by this PR, this date back from the optional options

library/src/interactor_impl.cxx
mwestphal: Well, there is nothing to bindm there is no args to this method.

mwestphal: Wouldnt that bind that parameter to the current value of `opts.ui.scalar_bar` ? 

Maybe std::bind is confusing after all :)

application/F3DStarter.cxx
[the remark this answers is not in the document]
mwestphal: I added a a XXX to improve formating

application/F3DStarter.cxx
```diff
@@ -901,18 +905,32 @@ int F3DStarter::Start(int argc, char** argv)
...
+
+    using mod_t = f3d::interaction_bind_t::ModifierKeys;
+    interactor.addBinding({ mod_t::NONE, "Left" }, "load_previous_file_group", "Others",
```
Meakk, line 913: Maybe `Others` should just be just an empty string?

library/private/interactor_impl.h
```diff
@@ -42,13 +42,22 @@ class interactor_impl : public interactor
...
-  interactor& removeBinding(std::string interaction, ModifierKeys modifiers) override;
-  std::vector<std::pair<std::string, ModifierKeys>> getBindingInteractions() const override;
+  interactor& addBinding(const interaction_bind_t& bind, std::vector<std::string> commands,
+    std::string group = std::string(),
+    documentation_callback_t documentationCallback = nullptr) override;
+  interactor& addBinding(const interaction_bind_t& bind, std::string command,
+    std::string group = std::string(),
+    documentation_callback_t documentationCallback = nullptr) override;
+  interactor& removeBinding(const interaction_bind_t& bind) override;
+  /*  std::vector<std::pair<std::string, ModifierKeys>> getBindingInteractions() const override;
```
Meakk, line 54: commented code

library/public/interactor.h
```diff
@@ -65,77 +116,89 @@ class F3D_EXPORT interactor
...
+   * similar as `addBinding(bind, {command})`
    *
-   * Adding command for an existing combination of interaction and modifier will throw a
-   * interactor::already_exists_exception.
+   * Adding command for an existing bind will throw a interactor::already_exists_exception.
    */
-  virtual interactor& addBinding(
-    const std::string& interaction, ModifierKeys modifiers, std::string command) = 0;
+  virtual interactor& addBinding(const interaction_bind_t& bind, std::string command,
+    std::string group = std::string(),
```
Meakk, line 162:
```suggestion
    std::string group = {},
```
or
```suggestion
    std::string group = "",
```
to avoid repetition of string keyword?

library/public/interactor.h
```diff
@@ -65,77 +116,89 @@ class F3D_EXPORT interactor
...
    *
-   * Adding commands for an existing combination of interaction and modifier will throw a
-   * interactor::already_exists_exception.
+   * Adding commands for an existing bind will throw a interactor::already_exists_exception.
    */
-  virtual interactor& addBinding(
-    const std::string& interaction, ModifierKeys modifiers, std::vector<std::string> commands) = 0;
+  virtual interactor& addBinding(const interaction_bind_t& bind, std::vector<std::string> commands,
+    std::string group = std::string(),
+    std::function<std::pair<std::string, std::string>()> documentationCallback = nullptr) = 0;
```
Meakk, line 152: don't you have an alias for that to improve readability?

library/public/interactor.h
```diff
@@ -65,77 +116,89 @@ class F3D_EXPORT interactor
...
   /**
    * Remove binding corresponding to provided interaction and modifiers
    */
-  virtual interactor& removeBinding(std::string interaction, ModifierKeys modifiers) = 0;
+  virtual interactor& removeBinding(const interaction_bind_t& bind) = 0;
+
+  /**
+   * Return a vector of available bind groups, in order of addition
+   */
+  virtual std::vector<std::string> getBindGroups() const = 0;
```
Meakk, line 183: Since we have `addBinding`/`removeBinding`, I suppose it makes more sens to call this function
```suggestion
  virtual std::vector<std::string> getBindingGroups() const = 0;
```

library/public/interactor.h
```diff
@@ -65,77 +116,89 @@ class F3D_EXPORT interactor
...
 
   /**
-   * Return a string vector of all currently defined bind interactions
+   * Return a vector of bind for the specified group, in order of addition
    */
-  virtual std::vector<std::pair<std::string, ModifierKeys>> getBindingInteractions() const = 0;
+  virtual std::vector<interaction_bind_t> getBindsForGroup(std::string group) const = 0;
```
Meakk, line 188: Maybe we should have two function:
- `getBindings()` to query all bindings
- `getBindings(std::string group)` to query bindings of a specific group?

application/F3DStarter.cxx
mwestphal: That would be not great when using `--bindings-list` imo

library/public/interactor.h
mwestphal: Not in the public header, I dont think this would be a good idea.

application/F3DStarter.cxx
Meakk: Yes, but I'm thinking of the default argument of `addBinding` if the group is not specified. In this case it makes sense that the group name is empty, and then if needed convert the empty group to `Others` if needed when printing the list.

library/public/interactor.h
Meakk: why?

library/public/interactor.h
Meakk: Regardless of the naming, this comment still stands. The `ForGoup` suffix looks useless if the argument is called `group`. And having an API to retrieve all binds in all groups could be useful to simplify the user logic if group differentiation is not needed.

library/src/interactor_impl.cxx
```diff
@@ -745,89 +721,236 @@ bool interactor_impl::triggerCommand(std::string_view command)
...
-  this->addBinding("4", ModifierKeys::ANY, "roll_camera -90");
-  this->addBinding("5", ModifierKeys::ANY, "toggle scene.camera.orthographic");
-  this->addBinding("6", ModifierKeys::ANY, "roll_camera 90");
-  this->addBinding("7", ModifierKeys::ANY, "set_camera top");
-  this->addBinding("9", ModifierKeys::ANY, "set_camera isometric");
-  this->addBinding(F3D_EXIT_HOTKEY_SYM, ModifierKeys::NONE, "stop_interactor");
-  this->addBinding("Return", ModifierKeys::NONE, "reset_camera");
-  this->addBinding("Space", ModifierKeys::NONE, "toggle_animation");
-  this->addBinding("Drop", ModifierKeys::NONE, "add_files");
+  this->addBinding({mod_t::NONE, "W"}, "cycle_animation", "Scene", std::bind(docStrVal, "Cycle animation", animStr));
```
Meakk, line 796: Sorry but that still looks odd to me. I'm no fan of the lambda calling lambda factorized using std::bind
What about just:
```cpp
auto docAnim = [&](){ return std::pair("Cycle animation", this->Internals->AnimationManager->GetAnimationName()); }
this->addBinding({mod_t::NONE, "W"}, "cycle_animation", "Scene", docAnim);
```

library/src/interactor_impl.cxx
```diff
@@ -745,89 +721,236 @@ bool interactor_impl::triggerCommand(std::string_view command)
...
+  this->addBinding({mod_t::NONE, "C"}, "cycle_coloring field", "Scene", std::bind(docStrVal, "Cycle point/cell data coloring", pointCellStr));
+  this->addBinding({mod_t::NONE, "S"}, "cycle_coloring array", "Scene", std::bind(docStrVal, "Cycle array to color with", arrayStr));
+  this->addBinding({mod_t::NONE, "Y"}, "cycle_coloring component", "Scene", std::bind(docStrVal, "Cycle component to color with", compStr));
```
Meakk, line 799: similarly for these 3 ones

library/src/interactor_impl.cxx
```diff
@@ -745,89 +721,236 @@ bool interactor_impl::triggerCommand(std::string_view command)
...
+  this->addBinding({mod_t::NONE, "B"}, "toggle ui.scalar_bar", "Scene", std::bind(docStrVal, "Toggle the scalar bar display",
```
Meakk, line 800: What about simplifying by removing a level of lambda:
```cpp
auto docTgl = [](const std::string& doc, const bool& opt)
{
  return std::pair(doc, val ? "ON" : "OFF";);
};

this->addBinding({mod_t::NONE, "B"}, "toggle ui.scalar_bar", "Scene", std::bind(docTgl, "Toggle the scalar bar display", std::cref(opts.ui.scalar_bar));
```
I've never used std::cref but that looks appropriate here

library/public/interactor.h
mwestphal: because then the name  would need to be f3d_documentation_callback_t or something.

Meakk: Idk, since it's just an alias for convenience, I don't see it as a true type. We have an example here: https://github.com/f3d-app/f3d/blob/d8ea37a730ba8ca7f5373c8337111ad463917874/library/public/context.h#L26

application/F3DStarter.cxx
Meakk: ok fair enough

library/public/interactor.h
```diff
@@ -13,6 +13,57 @@
...
+    SHIFT = 0x2,     // 00000010
+    CTRL_SHIFT = 0x3 // 00000011
+  };
+
+  ModifierKeys mod;
+  std::string inter;
+
+  bool operator<(const interaction_bind_t& bind) const
+  {
+    return this->inter < bind.inter || (this->inter == bind.inter && this->mod < bind.mod);
```
Meakk, line 36: Maybe it's better to check the modifier first since the comparison is cheaper?

library/public/interactor.h
```diff
@@ -13,6 +13,57 @@
...
+  }
+
+  bool operator==(const interaction_bind_t& bind) const
+  {
+    return this->inter == bind.inter && this->mod == bind.mod;
```
Meakk, line 41: same here

library/public/interactor.h
mwestphal: You are not afraid of conflicts with other headers ?

Meakk: not if the alias is in the class scope

mwestphal: 🤦

Meakk (approved):

PR #1718, opened 2024-11-16
Remove build step from static analysis

Meakk:
F3DIcon.h is now generated at configure-time. The build step should not be necessary.


.github/actions/static-analysis-ci/action.yml
```diff
@@ -71,12 +71,6 @@ runs:
         -DF3D_STRICT_BUILD=ON
         -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
 
-    # It's necessary to run the build step to generate F3DIcon.h
```
mwestphal, line 74: looks like this is still true, just not for F3DIcon

PR #1725, opened 2024-11-20
escape quotes in dropped filenames

snoyer:
`OnDropFiles()` needs to escape quotes in filenames before adding quotes around them.
Also `TriggerBinding()` should add a space between commands and first argument instead of expecting said space to be there already.


mwestphal:
Ideally a test using a file with a name withs spaces should be added/modified into, but I can take care of it if needed.

mwestphal:
superseeded by  https://github.com/f3d-app/f3d/pull/1728

PR #1728, opened 2024-11-23
OnDropFiles: Add support for spaces, quotes and backslahes in filename

mwestphal:
`OnDropFiles()` needs to escape quotes and backslashes in filenames before adding quotes around them.
Also `TriggerBinding()` should add a space between commands and first argument instead of expecting said space to be there already.

Test for spaces in filename is not supported by the interaction testing mechanism.
Test for quotes in filename is supported but github actions do not like quotes in filenames, so testing is limited.
Test for backslashes in filename is supported on Windows

Superseed https://github.com/f3d-app/f3d/pull/1725

### DIFF
--- a/library/src/interactor_impl.cxx
+++ b/library/src/interactor_impl.cxx
@@ -30,6 +30,7 @@
 #include <chrono>
 #include <cmath>
 #include <map>
+#include <regex>
 #include <vector>
 
 #include "camera.h"
@@ -244,10 +245,19 @@ public:
   {
     internals* self = static_cast<internals*>(clientData);
     vtkStringArray* filesArr = static_cast<vtkStringArray*>(callData);
+    const std::regex charsToEscape("([\"])");
     std::string filesString;
     for (int i = 0; i < filesArr->GetNumberOfTuples(); i++)
     {
-      filesString.append(" \"" + filesArr->GetValue(i) + "\" ");
+      const vtkStdString& filename = filesArr->GetValue(i);
+      const std::string escapedFilename = std::regex_replace(filename, charsToEscape, "\\$1");
+      if (i > 0)
+      {
+        filesString.push_back(' ');
+      }
+      filesString.push_back('"');
+      filesString.append(escapedFilename);
+      filesString.push_back('"');
     }
 
     self->TriggerBinding("Drop", filesString);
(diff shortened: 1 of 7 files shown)


mwestphal:
@snoyer @Meakk please review

library/src/interactor_impl.cxx
```diff
@@ -244,10 +245,19 @@ class interactor_impl::internals
   {
     internals* self = static_cast<internals*>(clientData);
     vtkStringArray* filesArr = static_cast<vtkStringArray*>(callData);
+    const std::regex charsToEscape(R"(([\"\\]))");
```
snoyer, line 248:
```suggestion
    const std::regex charsToEscape(R"((["\\]))");
```
shouldn't you also drop the first backslash when going from `"([\"\\\\])"` to an `R"` string?

mwestphal: absolutely !

library/src/utils.cxx
```diff
@@ -31,7 +31,15 @@ std::vector<std::string> utils::tokenize(std::string_view str)
...
-        escaped = true;
+        if (escaped)
+        {
+          accumulate(c);
+          escaped = false;
+        }
+        else
+        {
+          escaped = true;
+        }
```
Meakk, line 42:
```suggestion
        if (escaped)
        {
          accumulate(c);
        }
        escaped = !escaped;
```

Meakk (approved):
classic nitpicking comment, but otherwise lgtm

library/src/interactor_impl.cxx
```diff
@@ -424,15 +434,21 @@ class interactor_impl::internals
...
           // XXX: Ignore the boolean return of triggerCommand,
           //  error is already logged by triggerCommand
-          this->Interactor.triggerCommand(command + argsString);
+          this->Interactor.triggerCommand(commandWithArgs);
         }
         catch (const f3d::interactor::command_runtime_exception& ex)
         {
-          log::error("Interaction: error running command:\"", command + argsString, "\":");
+          log::error("Interaction: error running command:\"", commandWithArgs, "\":");
           log::error(ex.what());
```
snoyer, line 452:
```suggestion
          log::error("Interaction: error running command: \"" + commandWithArgs + "\": " + ex.what());
```
untested suggestion to improve spacing and line breaks, feel free to double check and/or tweak

PR #1729, opened 2024-11-24
Tests: Add more interaction test on Windows and macOS

mwestphal:
Do not put all interaction tests in LONG_TIMEOUT and only disable LONG_TIMEOUT on macos.
Exclude a few tests that are failing because of macOS openGL impl.

The results is more testing on Windows and macOS


Meakk (approved):
LGTM

Issue #1604, opened 2024-09-05
Improvements to f3d web model url parsing feature 


jo-chemla:
Following PR https://github.com/f3d-app/f3d/pull/1596 and issue https://github.com/f3d-app/f3d/issues/1595
Here is a thread to discuss possible improvements to the model url parsing implementation of f3d web, after some discussion on the f3d discord

Quick notes parsing through the discussion 
- [x] parse more url-params like up-vector, widgets, rendering, lighting
- [x] Visual feedback for download (can be done via fetch, more recent browser api than xhr) 
- [x] Document feature 
- [x] Detect cors and inform user, or any 404 

Sidenotes, 
- Search query vs hash can be parsed and not necessarily sent to server, that's more standard than hashes 
- Bundling, ts module loaded in browser without need to build

jo-chemla:
PR has been updated, mergeable with newest master branch from my point-of-view!

mwestphal:
Lets wait for the discussion to settle on the PR before closing this :)

mwestphal:
Works perfectly!

PR #1738, opened 2024-11-27
f3d-web updates: parse other ui params, loading visual feedback, alert user on error, documentation

jo-chemla:
Solves improvements of web url-param parsing listed in https://github.com/f3d-app/f3d/issues/1604 : 
 - parse more url-params: `axis, grid, fxaa, tone, ssao, ambient, up`
 - Visual feedback for download via progress bar from bulma
 - Alert user when provided model url yields error
 - Document in web readme


Meakk:
Awesome! I'll try that locally as soon as I can.

Meakk:
Here's my feedback:
It seems to work fine when loading the URL with all options set :+1:, however when options are edited several times in the URL input, it seems broken. Maybe a render() is missing? Or the browser cache is causing issues?
Also, I cannot see the progress bar.

webassembly/app.html
```diff
@@ -255,10 +278,43 @@ <h1 class="title">F3D Web</h1>
...
+
+          // Parse other options like #axis=false&grid=false&fxaa=false&tone=false&ssao=false&ambient=false
+          const options = Module.engineInstance.getOptions()
+          for (let [id, option] of idOptionMappings) {
+            // Set f3d option parameter values
+            const parsed_param_value = params.get(id);
+            if (parsed_param_value && parsed_param_value.toLowerCase() === 'false') {
+              // we can use toggle, set_string, set_integer, set_color
+              // sad thing is toggle requires knowing the param state beforehand, rather than setting it explicitly
+              options.toggle(option)
```
snoyer, line 298: As you already figured out, the JS bindings are not complete enough yet to achieve what you're attempting here.
https://github.com/f3d-app/f3d/blob/f3d7627047a8d584d0d2cd469d1bc616ffaca539/webassembly/F3DEmscriptenBindings.cxx#L80-L84

You could use the value of the UI switches to guess the current values and decide whether to toggle but that'd be sketchy as there is no proper guarantee that they'll be in sync.

At the very least `set_boolean` would need to be added, but ideally the bindings should expose the newer `setAsString` method that would let you do `options.set_as_string(option, parsed_param_value)` directly (passing a `"false"`/`"true"` string and having the parsing done internally, as opposed to converting from string to bool in the JS and calling `.set_boolean`).

jo-chemla:
Thanks for the feedback! 
 - Progress-bar only appears when you load a heavy model via the File Loader, on my side a 50MB local file will show progress bar for 2s. Will need to check if one can add the same progress bar for file loaded from url.
 - on my side options do work even with no model url passed. Indeed sometimes I have to hard-refresh the page with `SHIFT + CTRL + R` for options to be taken into account - I assumed this was because we decided to pass the options via hash rather than search query.

webassembly/app.html
jo-chemla: The current implementation takes for granted that: 
 - every DOM `input` is `checked` when page loads, see [this code block](https://github.com/f3d-app/f3d/blob/f3d7627047a8d584d0d2cd469d1bc616ffaca539/webassembly/app.html#L54-L76) 
 - the corresponding f3d `settings.options` are toggled on, see this [code block](https://github.com/f3d-app/f3d/blob/f3d7627047a8d584d0d2cd469d1bc616ffaca539/webassembly/app.html#L102-L110)
 - so I can only watch for when these `axis, grid, fxaa, tone, ssao, ambient` url-params are set to false, in which case I toggle them a second time before calling a final `Module.engineInstance.getWindow().render();` just before returnging from method `load_from_url()`

The current implementation does work under these assumptions from my testing. 
I think this PR can be merged, and the code could then be updated to account for a new `set_boolean` function + binding, or even higher-level `set_as_string` binding, once either is available.

snoyer:
>  Indeed sometimes I have to hard-refresh the page with `SHIFT + CTRL + R` for options to be taken into account - I assumed this was because we decided to pass the options via hash rather than search query.

Changing the hash in the browser URL box indeed should not produce a full page reload, but a reload shouldn't be needed because the script is already listening for the hash change event (`addEventListener("hashchange", (event) => {load_from_url()});`). You may want to add a `console.log` inside `load_from_url()` to investigate whether it is triggered correctly or if something is missing with the hash handling.

jo-chemla:
Just pushed a quick PR to add a fetchWithProgress function so progress-bar can be displayed both for FileReader and model url parsed. 
Regarding changing hash in the browser url bar, I did retry logging bits and it always listens to hash changes during my testing. Indeed, when hash does not change is when the reload needs to be triggered.

webassembly/app.html
jo-chemla: Alright, this is correct, I don't usually change hash after page was loaded but you're right. I'll include your modifications.

jo-chemla:
Pushed 2 more commits addressing the suggested modifications, should be good to go!

webassembly/app.html
```diff
@@ -83,6 +83,10 @@ <h1 class="title">F3D Web</h1>
         </div>
       </div>
     </div>
+    <div class="event-log" style="position: absolute; z-index: 10; bottom: 50px; width: 100%; display: none;">
+      <p style="text-align: center;""> Loading 3D object...</p>
```
snoyer, line 87:
```suggestion
      <p style="text-align: center;"> Loading 3D object...</p>
```

mwestphal:
FYI the wasm bindings have been improved to support set_as_string/get_as_string

jo-chemla:
Alright just committed a minor update, hich should be mergeable - don't hesitate to edit the PR in case you'd want a better approach.

jo-chemla:
Ready to be reviewed by f3d team from my point of view

jo-chemla:
It seems that conflicts were introduced after this PR was finalized and submitted. I'm trying to resolve them to merge master back into this PR branch before this one can get merged.

jo-chemla:
Just solved the last conflicts and re-tested the whole thing, from my point of view everything does work as expected!
Can be tested locally with this url: https://f3d.app/web#model=https://groups.csail.mit.edu/graphics/classes/6.837/F03/models/teapot.obj&up=%2BY&grid=false&fxaa=false&tone=false&ssao=false&ambient=false&axis=false Note that model requires a CORS plugin, note added in the web readme.

jo-chemla:
Just ran prettier style formatting and final commits.

webassembly/app.html
```diff
@@ -321,28 +410,73 @@ <h1 class="title">F3D Web</h1>
...
+              options.set_string("scene.up_direction", parsed_up_value);
+              if (parsed_up_value == "+Y") {
+                document.getElementById("z-up").classList.remove("is-active");
+                document.getElementById("y-up").classList.add("is-active");
+                openFile(document.getElementById("file-name").innerHTML);
+              } else {
+                document.getElementById("z-up").classList.add("is-active");
+                document.getElementById("y-up").classList.remove("is-active");
+                openFile(document.getElementById("file-name").innerHTML);
+              }
```
snoyer, line 475: Same open file in both branches of the `if`/`else` could be outside.

Also isn't `openFile()` the function that's called on with the "internal" filename on the virtual filesystem? do we know that `document.getElementById("file-name").innerHTML` is actually the correct virtual filename?

Also also what happens when `parsed_up_value` is neither `+Y` nor `+Z`?

jo-chemla: - Thanks committed the fix. 
 - These openFile calls should execute after that element has been set, only if a up value has been parsed. 
 - If the value is neither +Y nor +Z, then the default will be +Z, since only the first condition strictly evaluates against +Y string.

snoyer: >     * These openFile calls should execute after that element has been set, only if a up value has been parsed.

Sure but do we _know_ that `document.getElementById("file-name").innerHTML` contains the correct filename on the internal filesystem so that `openFile` opens the correct file?
If what you're intending is to reload the current model wouldn't you need to store the actual filename of the previous `openFile()` call so you can reuse that in some kind of `reloadCurrentFile()` function?


>     * If the value is neither +Y nor +Z, then the default will be +Z, since only the first condition strictly evaluates against +Y string.

yeah that one was rhetorical ;) if you pass `+X` through the hash the model will display X-up but the UI will say Z-up. What about removing the `is-active` class from both the Y and Z widgets and adding it back by testing respectively for each one (that way none of them would be active if `parsed_up_value` is `+x`, `-z`, ...).

jo-chemla: alright, did not know the api would accept +x, -z etc, I was only testing for +z vs +y. Pushed one last commit to accomodate this. 
Regarding the first comment, my point-of-view is that that dom-element innerHTML _does_ store the current filename, so I don't see a need for such `reloadCurrentFile` function. I can add it if you do think I'm wrong here, which is totally possible.

Meakk:
Running `https://f3d.app/web#model=https://groups.csail.mit.edu/graphics/classes/6.837/F03/models/teapot.obj` fails with the following error:
```
Internal exception: 7423688 [web:352:35](https://f3d.app/web/#model=https://groups.csail.mit.edu/graphics/classes/6.837/F03/models/teapot.obj)
    <anonymous> https://f3d.app/web/#model=https://groups.csail.mit.edu/graphics/classes/6.837/F03/models/teapot.obj:352
    (Async: promise callback)
    <anonymous> https://f3d.app/web/#model=https://groups.csail.mit.edu/graphics/classes/6.837/F03/models/teapot.obj:352
```
I've tried on Firefox with `CORS everywhere` and `Allow CORS` plugins

jo-chemla:
That's weird, were the plugins running (icon turned to green or plain color rather than red or transparent)?

It's also possible that this particular url does not work because it might in fact be served via plain http without ssl certificate, which induces mixed content loading (https for the f3d webapp and files but http for the file request) which is prevented by the browser. You could try with another url of a file that is served over real https.

jo-chemla:
I've re-tested locally with all the following urls, and it does work - loads the correct mesh file, with the correct set o parameters. 
local URL looks like the following: 

```url
http://127.0.0.1:8081/app.html#model=https://groups.csail.mit.edu/graphics/classes/6.837/F03/models/teapot.obj&up=%2BY&grid=false&fxaa=false&tone=false&ssao=false&ambient=false&axis=true
```

With link
http://127.0.0.1:8081/app.html#model=https://github.com/KhronosGroup/glTF-Sample-Models/raw/main/2.0/DamagedHelmet/glTF-Binary/DamagedHelmet.glb&up=%2BY&grid=false&fxaa=false&tone=false&ssao=false&ambient=false&axis=true

Tested with the below mesh urls: 

- https://github.com/KhronosGroup/glTF-Sample-Models/raw/main/2.0/DamagedHelmet/glTF-Binary/DamagedHelmet.glb
- https://groups.csail.mit.edu/graphics/classes/6.837/F03/models/teapot.obj
- https://hub.dronedb.app/orgs/pierotofy/ds/brighton-beach/download?path=3d-model%2Fodm_textured_model_geo.obj

Meakk:
Thanks @jo-chemla  
Sorry I was quite busy these past few weeks. I'll try to clarify why it's not working on my side and will try to take a closer look soon.

jo-chemla:
No worries, thanks for getting back!

Meakk (approved):
Finally had time to check again, and I confirm it was an issue on my side.  
I tried locally and it seems to work quite well. The progress bar isn't very smooth but it's better than nothing.
Thank you @jo-chemla !

Meakk:
@mwestphal do you want to take a look or should I merge it?

mwestphal:
> @mwestphal do you want to take a look or should I merge it?

go merge

mwestphal:
I just tried it on the live website and it fails with:

```
Error occurred while fetching the model at provided url (passed as url-param). 

This might be a CORS issue (you can avoid it by using a Allow-CORS plugin) because server does not allow cross-origin requests, or the url to the file is wrong. 

Complete error message: NetworkError when attempting to fetch resource.
```
Same on Firefox and Chromium.
I did install allow-cors plugin extension to no avail.

Am I missing something ?

@jo-chemla @Meakk

jo-chemla:
Thanks for merging the PR! AFK this week end, but probable that on top of adding the allow-cors plugins on either browser you also have to activate them - usually by opening them from the plug-in bar, and clicking their icon so it becomes colored or full.

Otherwise, you could try with server meshes endpoints that do allow cors natively, I think some of the urls provided in this PR thread other than the teapot allow that.

mwestphal:
That was it! Thanks!

PR #1744, opened 2024-12-02
refactor string templating to `f3d::utils`

snoyer:
Move the core of the variable substitution functionality used to generate screenshot filenames into `f3d::utils`.
This will allow it to be tested independently and possibly reused in other parts of the code.


library/public/utils.h
```diff
@@ -57,6 +59,28 @@ class F3D_EXPORT utils
...
   };
+
+  /** String template allowing substitution of variables enclosed in curly braces.
+    ```
+    StringTemplate("{greeting} {name}!")
+      .substitute({ { "greeting", "hello" }, { "name", "World" } })
+      .str() == "hello World!"
+    ```
+   */
+  class StringTemplate
```
mwestphal, line 70: maybe you need a F3D_EXPORT here ?

snoyer: thanks :)

Windows CI now has a compilation error instead of a link error https://github.com/f3d-app/f3d/actions/runs/12134916053/job/33833013791?pr=1744#step:3:1074
I assume that's progress but I have no idea what to do about it :/

library/public/utils.h
```diff
@@ -57,6 +61,94 @@ class F3D_EXPORT utils
...
```
mwestphal, line 72: All our public class and types (not enums) are snake case. This should be named `string_template` I believe

library/public/utils.h
```diff
@@ -57,6 +61,94 @@ class F3D_EXPORT utils
...
+    {
+      std::vector<std::string> variables;
+      for (auto [fragment, isVariable] : this->fragments)
+        if (isVariable)
+          variables.emplace_back(fragment);
+      return variables;
+    }
+  };
 };
 }
```
mwestphal, line 153: can you declare all the methods inline outside of the class declaration (but still in the header) ?

library/public/utils.h
```diff
@@ -4,6 +4,10 @@
 #include "exception.h"
 #include "export.h"
 
+#include <functional>
```
Meakk, line 7: not needed anymore

library/public/utils.h
```diff
@@ -57,7 +61,112 @@ class F3D_EXPORT utils
...
+    else
+      ss << fragment;
+  return ss.str();
+}
+
+//------------------------------------------------------------------------------
+inline std::vector<std::string> utils::string_template::variables() const
+{
+  std::vector<std::string> variables;
+  for (auto [fragment, isVariable] : this->fragments)
```
Meakk, line 165:
```suggestion
  for (const auto& [fragment, isVariable] : this->fragments)
```

library/public/utils.h
```diff
@@ -57,7 +61,112 @@ class F3D_EXPORT utils
...
+  const std::map<std::string, std::string>& lookup)
+{
+  return this->substitute([&](const std::string& key) { return lookup.at(key); });
+}
+
+//------------------------------------------------------------------------------
+inline std::string utils::string_template::str() const
+{
+  std::ostringstream ss;
+  for (auto [fragment, isVariable] : this->fragments)
```
Meakk, line 153:
```suggestion
  for (const auto& [fragment, isVariable] : this->fragments)
```

Meakk (approved):
LGTM

application/F3DStarter.cxx
```diff
@@ -325,112 +326,70 @@ class F3DStarter::F3DInternals
...
+      return { stringTemplate.str() };
     }
 
-    /* apply numbering in the non-substituted fragments and join */
-    const auto applyNumbering = [&](const size_t i)
+    const auto numberingLookup = [&](const size_t i)
     {
-      std::stringstream joined;
-      for (const auto& [fragment, processed] : fragments)
+      return [&numberingRe, i](const std::string& var)
```
mwestphal, line 364: please avoid single letter var

application/F3DStarter.cxx
```diff
@@ -325,112 +326,70 @@ class F3DStarter::F3DInternals
...
-          joined << std::regex_replace(fragment, numberingRe, formattedNumber.str());
+          return std::regex_replace(var, numberingRe, formattedNumber.str());
         }
-        else
-        {
-          joined << fragment;
-        }
-      }
-      return joined.str();
+        throw std::out_of_range(var);
```
mwestphal, line 384: throw f3d specific exception instead

mwestphal: Wait, this is in app, do we really need an exception here ?

library/public/utils.h
```diff
@@ -57,7 +60,114 @@ class F3D_EXPORT utils
...
+{
+  std::ostringstream ss;
+  // cppcheck-suppress unassignedVariable
+  // (false positive, fixed in cppcheck 2.8)
+  for (const auto& [fragment, isVariable] : this->fragments)
+    if (isVariable)
+      ss << "{" << fragment << "}";
+    else
+      ss << fragment;
+  return ss.str();
```
mwestphal, line 159: use braces

library/public/utils.h
```diff
@@ -57,7 +60,114 @@ class F3D_EXPORT utils
...
+}
+
+//------------------------------------------------------------------------------
+inline std::vector<std::string> utils::string_template::variables() const
+{
+  std::vector<std::string> variables;
+  for (const auto& [fragment, isVariable] : this->fragments)
+    if (isVariable)
+      variables.emplace_back(fragment);
+  return variables;
```
mwestphal, line 169: braces

library/testing/TestSDKUtils.cxx
```diff
@@ -60,5 +60,38 @@ int TestSDKUtils(int argc, char* argv[])
   test.expect<f3d::utils::tokenize_exception>("tokenize_exception with unfinishied escape",
     [&]() { f3d::utils::tokenize(R"(set render.hdri.file file path back\)"); });
 
+  //
```
mwestphal, line 63: empty comment ?

snoyer: just for spacing

application/F3DStarter.cxx
snoyer: that's the variable lookup function saying "not found" (same way `std::map::at` does) if it doesn't have a value for that key

mwestphal: ha right, this is caught in utils.cxx and I see it is documented in utils.h. However, using a std exception means that it is easy to catch an exception we did not intend to catch if at any point the std throw this for some reason (in not yet written code). Using a utils::string_template::exception ensure this will never happen.

mwestphal (changes requested):
one small change about execption

application/F3DStarter.cxx
snoyer: I thought `std::out_of_range` was more specific than it actually is, oh well custom one it is then

mwestphal (approved):
Nice!

PR #1761, opened 2024-12-06
Add support for keeping the camera state when switching files

jia-kai:
- Add a new option `keep-camera-between-file` for the global default behavior
- Add `_keeping_camera` variants to the commands `load_{previous,next}_file_group`
- Add default key bindings to the above new commands
- Add tests and docs


application/F3DOptionsTools.cxx
```diff
@@ -74,7 +74,8 @@ static inline const std::array<CLIGroup, 8> CLIOptions = {{
       { "watch", "", "Watch current file and automatically reload it whenever it is modified on disk", "<bool>", "1" },
       { "load-plugins", "", "List of plugins to load separated with a comma", "<paths or names>", "" },
       { "scan-plugins", "", "Scan standard directories for plugins and display available plugins (result can be incomplete)", "", "" },
-      { "screenshot-filename", "", "Screenshot filename", "<filename>", "" } } },
+      { "screenshot-filename", "", "Screenshot filename", "<filename>", "" } ,
+      { "keep-camera-between-file", "", "Keep the camera state when loading a new file group.", "<bool>", "1" } } },
   { "General",
```
mwestphal, line 79: lets not add a CLI option for this

doc/user/COMMANDS.md
```diff
@@ -54,9 +54,13 @@ eg: `add_files /path/to/dragon.vtu`.
...
 
-`load_previous_file_group`: A specific command to load the previous file or file group. No argument.
+`load_previous_file_group`: A specific command to load the previous file or file group. No argument. Whether the camera state is kept or not depends on the `keep-camera-between-file` option.
 
-`load_next_file_group`: A specific command to load the next file or file group. No argument.
+`load_previous_file_group_keeping_camera`: A specific command to load the previous file or file group and keep the camera state. No argument.
+
+`load_next_file_group`: A specific command to load the next file or file group. No argument. Whether the camera state is kept or not depends on the `keep-camera-between-file` option.
+
+`load_next_file_group_keeping_camera`: A specific command to load the next file or file group and keep the camera state. No argument.
```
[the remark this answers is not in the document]
jia-kai: > lets make this an arg of load_next/load_previous ?
> 
> Wdyt @Meakk

I've thought about this. It seems like a more elegant solution to me, but that would require `addBinding` to support a nondefault arg, which seems not implemented yet.

application/F3DOptionsTools.cxx
jia-kai: > lets not add a CLI option for this

Do you think it's worth keeping the it as a json config option? If so, should the entry in `DefaultAppOptions` be kept? (The comment says it is for ` The "App" CLI options`; I am wondering if there is a list of configurable non-CLI options)

doc/user/COMMANDS.md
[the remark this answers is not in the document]
mwestphal: it does support it, see roll_camera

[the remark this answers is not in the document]
jia-kai: > it does support it, see roll_camera

Oh I see! I misread some code. It is handled by `this->Interactor.triggerCommand(commandWithArgs);`

I would refactor the code after coming back :)

application/F3DOptionsTools.cxx
jia-kai: > > lets not add a CLI option for this
> 
> Do you think it's worth keeping it as a json config option? If so, should the entry in `DefaultAppOptions` be kept? (The comment says it is for ` The "App" CLI options`; I am wondering if there is a list of configurable non-CLI options)

Upon a second thought, I don't think we should add a json config option either; if the user wants keeping camera state as the default behavior, they can remap the Left/Right keys in their configuration.

mwestphal: its not an option at all, just two different bindings

application/F3DStarter.cxx
```diff
@@ -1513,11 +1515,34 @@ void F3DStarter::AddCommands()
...
+    [](const std::vector<std::string>& args, std::string_view commandName, bool defaultValue)
+  {
+    if (args.empty())
+    {
+      return defaultValue;
+    }
+    if (args.size() != 1)
+    {
+      throw std::invalid_argument{ std::string("Command: ") + std::string(commandName) +
+        " takes at most 1 argument, got " + std::to_string(args.size()) + " arguments instead." };
```
mwestphal, line 1528: you can add a dedicated script test to cover this code.

mwestphal (changes requested):
One small last change needed

jia-kai:
@mwestphal Do you have any idea on how to fix `TestInteractionCheatsheetAnimationNameRaytracing` on macos 9.3.1/9.2.6? The error is `rendering difference with reference image: 0.0609307 is higher than the threshold of 0.05`. I guess it is some font rendering problem, but not sure how to reproduce/fix it as I don't have the development environment.

mwestphal:
> @mwestphal Do you have any idea on how to fix `TestInteractionCheatsheetAnimationNameRaytracing` on macos 9.3.1/9.2.6? The error is `rendering difference with reference image: 0.0609307 is higher than the threshold of 0.05`. I guess it is some font rendering problem, but not sure how to reproduce/fix it as I don't have the development environment.

This failure is expected because you are adding an interaction, which is shown in the spreadsheet.
If you do not have the raytrading setup locally, you can easily grab the new baselines from the CI, right here:
https://github.com/f3d-app/f3d/actions/runs/12216735579

Let me know if you need further help :)

You can also ask such question on discord for faster feedback

application/F3DStarter.cxx
```diff
@@ -657,6 +657,8 @@ class F3DStarter::F3DInternals
       // clang-format off
       interactor.addBinding({ mod_t::NONE, "Left" }, "load_previous_file_group", "Others", std::bind(docString, "Load previous file group"));
       interactor.addBinding({ mod_t::NONE, "Right" }, "load_next_file_group", "Others", std::bind(docString, "Load next file group"));
+      interactor.addBinding({ mod_t::CTRL, "Left" }, "load_previous_file_group true", "Others", std::bind(docString, "Load previous file group while keeping the camera state"));
+      interactor.addBinding({ mod_t::CTRL, "Right" }, "load_next_file_group true", "Others", std::bind(docString, "Load next file group while keeping the camera state"));
```
mwestphal, line 661: Can the doc be shorter ? This is increasing the width of the cheatsheet greatly.

Maybe:

`Load previous file group, keep camera`

Wdyt ?

mwestphal (changes requested):
one small change

mwestphal:
Please also rebase on master, some change were just merged and they conflich with some baselines updates.

Issue #1502, opened 2024-06-30
Improve libf3d API in regards to time management


mwestphal:
**Describe the bug**
Improve libf3d API in regards to time management

There is currently no way in libf3d to:
 - extract the timerange
 - select a timestep and ONLY call LoadAtTime, not a complete loadScene

Nokse22:
For Exhibit to be able to show animations I need to know the time range and if any animation is preset so #1628 would be needed. Then using `scene.animation.time` I can play animations controlling them from the app.

snoyer:
> select a timestep and ONLY call LoadAtTime, not a complete loadScene

For this one, as discussed a while back on Discord, there are 2 ways to go about it:

1. expose `LoadAtTime()` in the API somehow
2. have `scene.animation.time` _not_ be a "load time" option, and internally call `LoadAtTime()` if the value of `scene.animation.time` has changed between calls to `render()`

The second one would be more convenient for users and doesn't require expanding the API, but there's a concern about silently triggering IO as some readers may require reloading the whole file.

Issue #20, opened 2021-11-12
Better scalar range handling with animation


mwestphal:
**Context**
F3D is able to show data using a colormap. Such a colormap has a range. This range is initialized using the min and max value of the data at the first time step. This works in most cases but in same cases, this would lead to unreadable results.

A workaround to this is of course to specify the colormap manually, however, F3D could have a better behavior.

**Suggested Solution**
F3D should **expand** the colormap on each timestep so that when the animation has been played fully, the colormap will not change anymore and further play through the animation will not make the color map evolve.

To implement this, one would need to ensure that changing the timestep in `animationManager` turn a flag on in `vtkF3DRendererWithColoring` so that it knows that the colormap should be expanded, and take that information into account in `vtkF3DRendererWithColoring::ConfigureRangeAndCTFForColoring` .

PR #1764, opened 2024-12-07
Animation API improvements

mwestphal:
- Add animation API to scene: Fix https://github.com/f3d-app/f3d/issues/1502
 - Remove scene.animation.time option
 - Improve setupCamera logic with bounds reset
 - Cleanup related code
 - Add support for extending color range: Fix https://github.com/f3d-app/f3d/issues/20
 - Fix an issue with interaction play that could trigger a segfault
 - Fix an issue with animation and multifile mode
 - Add python bindings for the new animation API


vtkext/private/module/vtkF3DMetaImporter.cxx
```diff
@@ -470,48 +497,52 @@ bool vtkF3DMetaImporter::UpdateAtTimeValue(double timeValue)
...
+        vtkPolyDataMapper* pdMapper = vtkPolyDataMapper::SafeDownCast(actor->GetMapper());
+        assert(pdMapper);
+
+        // Update coloring vectors, with a dedicated logic for generic importer
+        vtkDataSet* datasetForColoring = pdMapper->GetInput();
+        vtkF3DGenericImporter* genericImporter = vtkF3DGenericImporter::SafeDownCast(importerPair.Importer);
+        if (genericImporter)
         {
-          datasetForColoring = genericImporter->GetImportedPoints();
+          // TODO This will be improved with proper composite support
```
Meakk, line 529: Avoid the `TODO` keyword if you don't plan to do it in this PR.

doc/user/ANIMATIONS.md
```diff
@@ -34,10 +34,17 @@ F3D animation behavior can be fully controlled from the command line using the f
...
 - Press <kbd>Space</kbd> to play/pause animation
 
 ## Time Units
-- When F3D plays an animation, it assumes the time unit is in seconds to show accurate speed of animation.
+
+When F3D plays an animation, it assumes the time unit is in seconds to show accurate speed of animation.
+
+## Coloring range
+
+When coloring while loading a time value or playing the animation, the coloring range will be automatically expanded from previously loaded time value,
```
Meakk, line 49: Shouldn't we just compute the range of all timesteps when the coloring is requested?

mwestphal: This is not the right approach imo, as it will be 0(nTimeSteps). eg the following call would take a long time:

`f3d file_with_10000_ts.ex2 --animation-time=2 -sb`

vtkext/private/module/vtkF3DMetaImporter.cxx
mwestphal: I'm just moving this comment around, but I can write XXX instead if you prefer.

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -2426,18 +2442,32 @@ void vtkF3DRenderer::ConfigureRangeAndCTFForColoring(
     }
   }
 
-  if (autoRange)
+  if (this->UsingExpandingRange)
```
Meakk, line 2445: let's add a `f3d::log::info` 😜

Meakk (changes requested):
few changes needed

application/F3DStarter.cxx
```diff
@@ -141,9 +142,12 @@ class F3DStarter::F3DInternals
       cam.setPosition(pos);
       reset = true;
     }
-    if (camConf.CameraPosition.size() != 3 && camConf.CameraZoomFactor > 0)
+    if (camConf.CameraPosition.size() != 3)
     {
-      zoomFactor = camConf.CameraZoomFactor;
+      if (camConf.CameraZoomFactor > 0)
```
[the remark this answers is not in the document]
mwestphal: It is related and was caused by the order of loading/reseting bounds.

I do list the change in the MR descr though:

> Improve setupCamera logic with bounds reset

library/src/interactor_impl.cxx
```diff
@@ -1049,7 +1049,6 @@ bool interactor_impl::playInteraction(const std::string& file)
 
     std::string cleanFile = vtksys::SystemTools::CollapseFullPath(file);
     this->Internals->Recorder->SetFileName(cleanFile.c_str());
-    this->Internals->Window.UpdateDynamicOptions();
```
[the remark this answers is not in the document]
mwestphal: TBH I dont know why this was ever needed at all, it dates back to my original implementation of the options.

In any case, options are dynamically updated on render, and with this change I fall into a segfault in VTK.
I fixed the segfault in VTK https://gitlab.kitware.com/vtk/vtk/-/merge_requests/11740, but since this code was not needed at all, I just removed it instead of adding version checks

application/F3DStarter.cxx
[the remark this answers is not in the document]
Meakk: Ok I missed that in the description, thanks.

library/src/interactor_impl.cxx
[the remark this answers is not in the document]
Meakk: Ok I don't remember the details but I think this was also introducing some duplicated warnings in some cases.
Anyway, good to see useless code removed.

Meakk (approved):

PR #1778, opened 2024-12-11
Add ImGui in external libs

Meakk:


vtkext/private/module/CMakeLists.txt
```diff
@@ -106,8 +106,13 @@ if(F3D_MODULE_EXR)
   list(APPEND classes vtkF3DEXRReader)
 endif()
 
-if(F3D_MODULE_UI)
-  find_package(imgui REQUIRED)
+if(F3D_MODULE_UI AND NOT F3D_USE_EXTERNAL_IMGUI)
+  list(APPEND sources
+    "${F3D_SOURCE_DIR}/external/imgui/imgui.cpp"
+    "${F3D_SOURCE_DIR}/external/imgui/imgui_demo.cpp"
```
Meakk, line 112: I decided to keep `imgui_demo.cpp` because it's useful while developing the UI.  
It's also recommended to do so:
```
//---------------------------------------------------
// PLEASE DO NOT REMOVE THIS FILE FROM YOUR PROJECT!
//---------------------------------------------------
// Message to the person tempted to delete this file when integrating Dear ImGui into their codebase:
// Think again! It is the most useful reference code that you and other coders will want to refer to and call.
// Have the ImGui::ShowDemoWindow() function wired in an always-available debug menu of your game/app!
// Also include Metrics! ItemPicker! DebugLog! and other debug features.
// Removing this file from your project is hindering access to documentation for everyone in your team,
// likely leading you to poorer usage of the library.
// Everything in this file will be stripped out by the linker if you don't call ImGui::ShowDemoWindow().
// If you want to link core Dear ImGui in your shipped builds but want a thorough guarantee that the demo will not be
// linked, you can setup your imconfig.h with #define IMGUI_DISABLE_DEMO_WINDOWS and those functions will be empty.
// In another situation, whenever you have Dear ImGui available you probably want this to be available for reference.
// Thank you,
// -Your beloved friend, imgui_demo.cpp (which you won't delete)
```

mwestphal: I disagree, I really dont see the point of compiling it.

Keeping the file around can make sense though, at least it simpler to update.

Meakk: What do you suggest when a developer has to use the dev tools? They have to add the file `"${F3D_SOURCE_DIR}/external/imgui/imgui_demo.cpp"` manually everytime?

Meakk: > // Everything in this file will be stripped out by the linker if you don't call ImGui::ShowDemoWindow().

That's what convinced me to keep it. Since it doesn't hurt, what's the problem?

mwestphal: You lost me there, what does imgui_demo provides ?

Meakk: Many debug tools, please read the block in my first message. TLDR:
> // Also include Metrics! ItemPicker! DebugLog! and other debug features.

You'll get the importance of it when you'll start working with ImGui.

mwestphal: Alright, trusting you here.

mwestphal (approved):

PR #1797, opened 2024-12-17
Animation: Fix a small issue with animation and improve animation tests

mwestphal:


library/src/animationManager.cxx
```diff
@@ -209,8 +209,13 @@ void animationManager::Tick()
...
     if (this->CurrentTime < this->TimeRange[0] || this->CurrentTime > this->TimeRange[1])
     {
+      auto modulo = [](double val, double mod)
+      {
+        const double remainder = fmod(val, mod);
+        return remainder < 0 ? remainder + mod : remainder;
+      };
       this->CurrentTime = this->TimeRange[0] +
-        std::fmod(this->CurrentTime - this->TimeRange[0], this->TimeRange[1] - this->TimeRange[0]);
+        modulo(this->CurrentTime - this->TimeRange[0], this->TimeRange[1] - this->TimeRange[0]);
```
mwestphal, line 218: @Meakk I know remember why fmod did not work, nor do std::remainder

mwestphal: the modulo as to loop around correctly, even with negative values

PR #1802, opened 2024-12-18
wasm: Fix/Improve bindings

mwestphal:


Meakk (approved):
checked locally, it works with the pushed fix

PR #1804, opened 2024-12-18
prepare for `pybind11-stubgen`

snoyer:
It seems like https://github.com/sizmailov/pybind11-stubgen is now mature enough to generate the `f3d` bindings stubs automatically.

In order to properly generate the stubs we need to:
- reorder some bindings declaration so that everything is known before it is referenced
- use `----ignore-unresolved-names` to ignore `f3d::point3_t` and `f3d::vector3_t` as we don't actually map them but let them auto-convert to and from `tuple[float, float, float]`
- use `--enum-class-locations SaveFormat:Image` to allow proper identification of the default value for the `format` parameters of `Image.save()` and `Image.save_buffer()`
- do some post-processing to clean up a few thing

With this PR's changes we can run the new `generate_stubs.py` script to generate the `.pyi` files

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "scikit_build_core.build"
 
 [project]
 name = "f3d"
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 dynamic = ["version"]
 description = "F3D, a fast and minimalist 3D viewer"
 readme = "README.md"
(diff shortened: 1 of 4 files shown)


snoyer:
Adding the type aliases for `point3_t` and `vector3_t` in `__init__.py` does not work properly because the following gets generated in `__init__.pyi`
```py
from builtins import tuple as point3_t
from builtins import tuple as vector3_t
```
this is technically correct but means that variables anotated as `f3d.point3_t` or `f3d.vector3_t` are seen as being of type `tuple[Any, ...]` that is tuples alright but of unknown length and with elements of any type.

the following more precise definition should be added after the imports in the generated `pyf3d.pyi` _somehow_
```py
point3_t = tuple[float, float, float]
vector3_t = tuple[float, float, float]
```

snoyer:
taken over in #1818

PR #1806, opened 2024-12-18
Updateoptions

void-inject:
Updated logic in `UpdateOptions` to detect "default" as a special value


void-inject:
CodeCov report is because of my ide's auto formater, i think

Issue #972, opened 2023-09-05
[feature] Add stubs for the python bindings


melMass:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when autocomplete doesn't work

**Describe the solution you'd like**
Whatever fit best the maintainer's workflow, a few ideas:
- [pybind11-stubgen](https://github.com/sizmailov/pybind11-stubgen)
- mypy's stubgen (used in the POC)

**Describe alternatives you've considered**
For a working POC I tried `mypy`'s stubgen:

```sh
pip install mypy
stubgen -m f3d
stubgen -m f3d.pyf3d
```

Then copy the generated pyi files to `site-packages/f3d` 

**Additional context**
The POC output
[f3d-stubs.zip](https://github.com/f3d-app/f3d/files/12523577/f3d-stubs.zip)

**Contribution**
Probably easier if it's done by maintainers but if I'm pointed to the process I can dig into it yes!

mwestphal:
@snoyer @Meakk

Meakk:
Let's come back to this after https://github.com/f3d-app/f3d/pull/962 is merged

mwestphal:
It is now merged, what about this issue then ? :)

Meakk:
I think it's important to add the stub files so let's target 2.3.  
Additionally, we should improve the bindings to name all arguments and document each function.

Meakk:
Added by https://github.com/f3d-app/f3d/pull/1818

PR #1818, opened 2024-12-21
Generate Python stubs

Meakk:
Add a script used to generate Python stubs for our current Python bindings. It will generate `.pyi` files containing all the available Python API used by IDE to do static code verification.

Fix https://github.com/f3d-app/f3d/issues/972


python/CMakeLists.txt
```diff
@@ -83,6 +86,15 @@ endif()
 configure_file("${CMAKE_CURRENT_SOURCE_DIR}/__init__.py.in"
   "${CMAKE_CURRENT_BINARY_DIR}/__init__.py-install" @ONLY)
 
+# Stubs
+if (F3D_BINDINGS_PYTHON_GENERATE_STUBS)
+  add_custom_command(
+    TARGET pyf3d
+    POST_BUILD
+    COMMAND python ${CMAKE_CURRENT_SOURCE_DIR}/generate_stubs.py --into=${f3d_binary_dir}
```
snoyer, line 94:
```suggestion
    COMMAND PYTHONPATH="" python ${CMAKE_CURRENT_SOURCE_DIR}/generate_stubs.py --into=${f3d_binary_dir}
```
for some reason that works for me locally (linux)

snoyer:
some changes got lost in a merge
```
pybind11_stubgen - [  ERROR] In f3d.pyf3d.Camera.state. : Invalid expression 'f3d::camera_state_t'
Warning: stubgen - [WARNING] Raw C++ types/values were found in signatures extracted from docstrings.
```
`py::class_<f3d::camera_state_t>` needs to be moved above `py::class_<f3d::camera, ...>` again, see #1804

pyproject.toml
```diff
@@ -30,6 +30,10 @@ classifiers = [
   "Topic :: Scientific/Engineering :: Visualization",
 ]
 
+dependencies = [
+  "pybind11_stubgen"
+]
```
snoyer, line 35: this will download and install `pybind11_stubgen` when doing `pip install f3d` which isn't necessary for normal use

there is an extra/optional dependency mechanism that could be used so that `pybind11_stubgen` is only installed on `pip install f3d[dev]`
```suggestion
[project.optional-dependencies]
dev = [
  "pybind11_stubgen"
]
```
I'm not sure what the exact workflow is tho (build wheel vs build bindings from cmake vs install from source), maybe it's not necessary at all because it has to be installed externally first anyway?

Meakk: Yeah I'm not sure. It's a build dependency, so it's needed if you type `pip install .` from f3d repository but not needed if installed using wheels.  
I'll try to find the correct way to do it, but maybe it's `optional-dependencies` as you said.

snoyer: there's also
```
[build-system]
requires = ["scikit-build-core", "setuptools-scm>=8.0"]
```
maybe it should go in there?

Meakk: yeah let's try that.

Meakk: indeed it works, thanks :)

.github/actions/python-ci/action.yml
```diff
@@ -53,6 +53,7 @@ runs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install pytest==8.0.0
+        python -m pip install pybind11_stubgen
```
Meakk, line 56: probably not needed

mwestphal:
please write a description explaining what this do, idk what is a stub

python/F3DPythonBindings.cxx
```diff
@@ -407,44 +430,22 @@ PYBIND11_MODULE(pyf3d, module)
...
-
   py::enum_<f3d::log::VerboseLevel>(log, "VerboseLevel")
     .value("DEBUG", f3d::log::VerboseLevel::DEBUG)
     .value("INFO", f3d::log::VerboseLevel::INFO)
     .value("WARN", f3d::log::VerboseLevel::WARN)
     .value("ERROR", f3d::log::VerboseLevel::ERROR)
     .value("QUIET", f3d::log::VerboseLevel::QUIET)
     .export_values();
+
+  log //
```
mwestphal, line 444: // ?

snoyer: it prevents `clang-format` from bringing the following `.def()` on the same line

Meakk: not related to this PR anyway, so let's not block the review for that.

mwestphal (approved):

snoyer (approved):

Issue #1697, opened 2024-11-09
Provide info about available rendering backends


mwestphal:
**Is your feature request related to a problem? Please describe.**
F3D can use specific rendering backends using the `--rendering-backend` cli option, however there is no way to ask F3D what rendering backend are currently available and will work when asked.

**Describe the solution you'd like**
 - Add a CLI option to list available and ready rendering backends : `--list-rendering-backends`

PR #1831, opened 2024-12-23
Option to list rendering backends

Meakk:
Fix https://github.com/f3d-app/f3d/issues/1697

Add information about available rendering backends

### DIFF
--- a/application/F3DOptionsTools.cxx
+++ b/application/F3DOptionsTools.cxx
@@ -70,6 +70,7 @@ static inline const std::array<CLIGroup, 8> CLIOptions = {{
       { "no-config", "", "Do not read the configuration file", "<bool>", "1" },
       { "no-render", "", "Do not render anything and quit right after loading the first file, use with --verbose to recover information about a file.", "<bool>", "1" },
       { "rendering-backend", "", "Backend to use when rendering (auto|glx|wgl|egl|osmesa)", "<string>", "" },
+      { "rendering-backend-list", "", "List of rendering backends available on this system", "", "" },
       { "max-size", "", "Maximum size in Mib of a file to load, leave empty for unlimited", "<size in Mib>", "" },
 #if F3D_MODULE_DMON
       { "watch", "", "Watch current file and automatically reload it whenever it is modified on disk", "<bool>", "1" },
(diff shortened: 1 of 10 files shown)


snoyer:
would need to use `--rendering-backends-list` (plural) to match `--readers-list` and `--bindings-list`
However I believe the singular is grammatically correct so the other 2 would need to change instead/eventually?

application/F3DOptionsTools.cxx
```diff
@@ -176,7 +176,7 @@ static inline const std::array<CLIGroup, 8> CLIOptions = {{
  * True boolean options need to be filtered out in ParseCLIOptions
  * This is the easiest, compile time way to do it
  */
-constexpr std::array<std::string_view, 4> CLIBooleans = {"version", "help", "readers-list", "scan-plugins"};
+constexpr std::array CLIBooleans = {"version", "help", "readers-list", "scan-plugins", "renderin-backend-list"};
```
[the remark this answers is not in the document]
Meakk: yes since C++14

doc/user/CONFIGURATION_FILE.md
```diff
@@ -65,7 +65,7 @@ The third block specifies raytracing usage for .gltf and .glb files.
 The last block specifies that volume rendering should be used with .mhd files.
 
 The following options <b> cannot </b> be set via config file:
-`help`, `version`, `readers-list`, `config`, `dry-run` and `input`.
+`help`, `version`, `list-readers`, `config`, `dry-run` and `input`.
```
mwestphal, line 68: I think `list-rendering-backends` and `scan-plugins` should be added here

doc/user/OPTIONS.md
```diff
@@ -13,7 +13,7 @@ Options|Default|Description
 \-\-no-background||Use with \-\-output to output a png file with a transparent background.
 -h, \-\-help||Print *help* and exit. Ignore `--verbose`.
 \-\-version||Show *version* information and exit. Ignore `--verbose`.
-\-\-readers-list||List available *readers* and exit. Ignore `--verbose`.
+\-\-list-readers||List available *readers* and exit. Ignore `--verbose`.
```
mwestphal, line 16: I'd expect list-bindings and list-rendering-backends to appear in this file

mwestphal (changes requested):
small changes needed

doc/user/OPTIONS.md
```diff
@@ -13,7 +13,9 @@ Options|Default|Description
...
+\-\-list-bindings||List available *bindings* and exit. Ignore `--verbose`.
```
mwestphal, line 17: not sure that it ignore `--verbose` though

mwestphal (changes requested):
small question

doc/user/OPTIONS.md
Meakk: Indeed. It's copied from other list commands.

Meakk: Just checked, it's correct. The verbosity level is not taken into account yet

mwestphal (approved):

PR #1841, opened 2024-12-27
engine: Update API for C++17

mwestphal:
- Rewrite getPluginsList
 - Update API for C++17
 - Add/Update test for better testing and coverage
 - Improve some exceptions doc and logic


library/src/engine.cxx
```diff
@@ -41,16 +41,16 @@ engine::engine(
 
   // build default cache path
 #if defined(_WIN32)
-  std::string cachePath = vtksys::SystemTools::GetEnv("LOCALAPPDATA");
-  cachePath = cachePath + "/f3d";
+  fs::path cachePath(std::getenv("LOCALAPPDATA"));
```
Meakk, line 44: Creating a `fs::path` with nullptr will crash the app. You need to check the result of `std::getenv`

library/src/engine.cxx
```diff
@@ -213,7 +213,7 @@ interactor& engine::getInteractor()
 }
 
 //----------------------------------------------------------------------------
-void engine::loadPlugin(const std::string& pathOrName, const std::vector<std::string>& searchPaths)
+void engine::loadPlugin(std::string_view pathOrName, const std::vector<fs::path>& searchPaths)
```
Meakk, line 216: `pathOrName` is used many times to create a string, I think keeping it const ref might be better?

library/src/engine.cxx
```diff
@@ -364,6 +364,9 @@ std::vector<std::string> engine::getPluginsList(const std::string& pluginPath)
       }
     }
   }
+  catch (const std::filesystem::filesystem_error&)
```
Meakk, line 367:
```suggestion
  catch (const fs::filesystem_error&)
```

library/src/engine.cxx
```diff
@@ -442,7 +445,7 @@ std::vector<engine::readerInformation> engine::getReadersInfo()
 }
 
 //----------------------------------------------------------------------------
-engine& engine::setCachePath(const std::string& cachePath)
+engine& engine::setCachePath(const std::filesystem::path& cachePath)
```
Meakk, line 448:
```suggestion
engine& engine::setCachePath(const fs::path& cachePath)
```

library/src/window_impl.cxx
```diff
@@ -549,9 +549,9 @@ void window_impl::SetImporter(vtkF3DMetaImporter* importer)
 }
 
 //----------------------------------------------------------------------------
-void window_impl::SetCachePath(const std::string& cachePath)
+void window_impl::SetCachePath(const std::filesystem::path& cachePath)
 {
-  this->Internals->CachePath = cachePath;
+  this->Internals->CachePath = cachePath.string();
```
Meakk, line 554: Maybe `CachePath` should be a fs::path too?

library/testing/TestSDKEngine.cxx
```diff
@@ -71,5 +72,9 @@ int TestSDKEngine(int argc, char* argv[])
     return EXIT_FAILURE;
   }
 
+  // coverage for getPluginList
+  std::ignore = f3d::engine::getPluginsList(std::string(argv[1]) + "configs");
+  std::ignore = f3d::engine::getPluginsList("innexistent");
```
Meakk, line 77:
```suggestion
  std::ignore = f3d::engine::getPluginsList("inexistent");
```

library/testing/TestSDKEngineExceptions.cxx
```diff
@@ -1,99 +1,82 @@
...
+      "olderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfol"
+      "derfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolde"
+      "rfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderf"
+      "olderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfol"
+      "derfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolde"
+      "rfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderf"
+      "olderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfol"
+      "derfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolde"
+      "rfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderf"
+      "olderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolderfolder/file.ext");
```
Meakk, line 78: Very verbose, what about this?
```suggestion
    f3d::engine::loadPlugin("/" + std::string(1000, 'x') + "/file.ext");
```

Meakk (changes requested):

application/F3DPluginsTools.cxx
```diff
@@ -6,21 +6,27 @@
...
 {
 //----------------------------------------------------------------------------
-std::vector<std::string> GetPluginSearchPaths()
+std::vector<fs::path> GetPluginSearchPaths()
 {
+  std::vector<fs::path> searchPaths;
+
   // Recover F3D_PLUGINS_PATH first
-  auto searchPaths = F3DSystemTools::GetVectorEnvironnementVariable("F3D_PLUGINS_PATH");
+  std::vector<std::string> stringPaths =
```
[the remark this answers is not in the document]
mwestphal: definitely not, this function can be used for any env var, it is not path specific.

library/src/engine.cxx
mwestphal: good catch!

library/src/window_impl.cxx
mwestphal: indeed

library/src/engine.cxx
```diff
@@ -39,18 +39,29 @@ engine::engine(
...
+
+#if defined(_WIN32)
+  cachePath /= "f3d";
 #elif defined(__APPLE__)
-  std::string cachePath = vtksys::SystemTools::GetEnv("HOME");
-  cachePath += "/Library/Caches/f3d";
+  cachePath = cachePath / "Library" / "Caches" / "f3d";
 #elif defined(__ANDROID__)
-  std::string cachePath = ""; // no default
+//  fs::path cachePath;
```
mwestphal, line 62: no default does not makes much sense, we cannot force libf3d user to set a cachePath. Is there no reasonnable cache path we could use by default ?

mwestphal: @Meakk

application/F3DPluginsTools.cxx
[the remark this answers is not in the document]
Meakk: I meant the `GetPluginSearchPaths()` function but it seems like you changed it.

library/src/window_impl.cxx
```diff
@@ -128,38 +121,39 @@ window_impl::window_impl(const options& options, const std::optional<Type>& type
   {
 #if defined(VTK_OPENGL_HAS_EGL)
     this->Internals->RenWin = vtkSmartPointer<vtkF3DEGLRenderWindow>::New();
-#else
-    throw engine::no_window_exception("Window type is EGL but VTK EGL support is not enabled");
 #endif
+    // No #else needed, it would be unreachable
```
Meakk, line 125: Why isn't it unreachable? `VTK_OPENGL_HAS_EGL` can be OFF

Meakk: Good idea, I think `#error` will help us in the future if we change something in `context` class.

mwestphal: obv #error wont work as this is still compiled, I've added asserts

Meakk (approved):

Issue #1795, opened 2024-12-16
UI related segfault when creating multiple engines with the libf3d


mwestphal:
**Describe the bug**
When creating multiple engine, there seems to be some memory related issue

**To Reproduce**
 1. Run this python script
 
 ```
import f3d 

def stop_fn(inter):
    inter.stop()

def print_fn():
    print("test")

engine = f3d.Engine.create(False)
inter = engine.interactor
inter.start(1 / 30, partial(stop_fn, inter))
del engine

engine2 = f3d.Engine.create(False)
inter2 = engine2.interactor
inter2.start(1 / 30, print_fn)
```
 2. Press Escape to show console
 3. Press Escape to hide console
 4. Segfault

```
../src/python/testing/test_interactor_start.py 
Program received signal SIGSEGV, Segmentation fault.
f3d::detail::interactor_impl::internals::OnConsoleEvent (event=1202, clientData=0x555555bf5e80, data=0x0) at /home/glow/dev/f3d/f3d/src/library/src/interactor_impl.cxx:234
234           self->Options.ui.console = (event == vtkF3DConsoleOutputWindow::ShowEvent);
(gdb) bt
#0  f3d::detail::interactor_impl::internals::OnConsoleEvent (event=1202, clientData=0x555555bf5e80, data=0x0) at /home/glow/dev/f3d/f3d/src/library/src/interactor_impl.cxx:234
#1  0x00007fffedd23a8f in vtkCallbackCommand::Execute (this=0x555555be3020, caller=0x555555a6a8f0, event=1202, callData=0x0) at /home/glow/dev/vtk/vtk1/src/Common/Core/vtkCallbackCommand.cxx:30
#2  0x00007fffede5fd1c in vtkSubjectHelper::InvokeEvent (this=0x555555be3d00, event=1202, callData=0x0, self=0x555555a6a8f0) at /home/glow/dev/vtk/vtk1/src/Common/Core/vtkObject.cxx:638
#3  0x00007fffede602df in vtkObject::InvokeEvent (this=0x555555a6a8f0, event=1202, callData=0x0) at /home/glow/dev/vtk/vtk1/src/Common/Core/vtkObject.cxx:807
#4  0x00007ffff5f02cfa in vtkObject::InvokeEvent (this=0x555555a6a8f0, event=1202) at /home/glow/dev/vtk/vtk1/src/Common/Core/vtkObject.h:223
[...]

PR #1847, opened 2024-12-28
Bug fix for segfault iss #1795

hant-hub:
This fixes a UI related segfault due to a stale observer reference in VTK. This was first found in issue #1795 where a python application segfaults if you delete and recreate the f3d engine.


hant-hub:
I'm unsure how to generate synthetic vtkWindowEvents for testing, I guess I could make a c++ test for this?

hant-hub:
For the test there is a thing where I believe due to memory layout issues the segfault is inconsistent. I could only get it to reliably detect the segfault by repeating the test multiple times. Currently I set that to 5 since that almost always catches the bug in my testing, but we might want to change that or install some kind of other memory validation.

hant-hub:
not really sure why the last two tests are failing, any advice would be appreciated.

library/testing/CMakeLists.txt
```diff
@@ -62,6 +63,8 @@ configure_file("${F3D_SOURCE_DIR}/testing/recordings/TestSDKInteractorCallBack.l
                "${CMAKE_BINARY_DIR}/TestSDKInteractorCallBack.log") # Dragon.vtu; S
 configure_file("${F3D_SOURCE_DIR}/testing/recordings/TestSDKInteractorDropFullScene.log.in"
                "${CMAKE_BINARY_DIR}/TestSDKInteractorDropFullScene.log") # world.obj; S
+configure_file("${F3D_SOURCE_DIR}/testing/recordings/TestSDKEngineRecreation.log.in"
```
Meakk, line 66: I don't think you need a configure_file, just use a .log file directly

mwestphal:
> Currently I set that to 5 since that almost always catches the bug in my testing, but we might want to change that or install some kind of other memory validation.

Thats fine, memory issues are like that :)

Meakk:
> not really sure why the last two tests are failing, any advice would be appreciated.

Not sure about the static analysis error, it seems unrelated.  
About the crash with VTK 9.2.6, I can try to reproduce locally next week.

mwestphal:
> not really sure why the last two tests are failing, any advice would be appreciated.

VTK v9.2.6: Probably due to an already fixed bug in VTK, just not add the test with VTK 9.2.6 in the CMake logic

static_analysis: I dont get why this would be wrong and definitely not caused by ytour changes , is your branch based on master ?

library/testing/TestSDKEngineRecreation.cxx
```diff
@@ -0,0 +1,24 @@
+#include <engine.h>
+#include <interactor.h>
+#include <log.h>
+#include <options.h>
+#include <scene.h>
+#include <window.h>
```
mwestphal, line 6: not all includes needed

mwestphal (changes requested):
nice fix! the main change looks good, small changes needed.

snoyer:
The static analysis error is unrelated but it looks valid, the real question is why is it showing up only now?
```
library/src/interactor_impl.cxx:1049:66: style: Unused variable: doc [unusedVariable]
  std::vector<std::tuple<std::string, std::string, std::string>> doc;
```
https://github.com/f3d-app/f3d/blob/27d1e8a117b978bf830cc7d3a8601827f18309ed/library/src/interactor_impl.cxx#L1041-L1053

mwestphal:
> The static analysis error is unrelated but it looks valid, the real question is why is it showing up only now?
> 
> ```
> library/src/interactor_impl.cxx:1049:66: style: Unused variable: doc [unusedVariable]
>   std::vector<std::tuple<std::string, std::string, std::string>> doc;
> ```
> 
> https://github.com/f3d-app/f3d/blob/27d1e8a117b978bf830cc7d3a8601827f18309ed/library/src/interactor_impl.cxx#L1041-L1053

Good point, moving this discussion to: https://github.com/f3d-app/f3d/pull/1848

mwestphal:
> The static analysis error is unrelated but it looks valid, the real question is why is it showing up only now?
> 
> ```
> library/src/interactor_impl.cxx:1049:66: style: Unused variable: doc [unusedVariable]
>   std::vector<std::tuple<std::string, std::string, std::string>> doc;
> ```
> 
> https://github.com/f3d-app/f3d/blob/27d1e8a117b978bf830cc7d3a8601827f18309ed/library/src/interactor_impl.cxx#L1041-L1053

Good point, moving this discussion here:

mwestphal:
(my bad, I did not intend to do that)

mwestphal:
Anyway, static_analysis CI is fixed in master @hant-hub

mwestphal (approved):

mwestphal:
Nice fix @hant-hub

mwestphal:
@Meakk wanna review ?

library/testing/TestSDKEngineRecreation.cxx
```diff
@@ -0,0 +1,20 @@
...
+  std::string interactionFilePath = std::string(argv[1]) + "/recordings/" + filename + ".log";
+  f3d::engine* eng = new f3d::engine(f3d::engine::create(false));
+  eng->getInteractor().start(1 / 30, [eng]() { eng->getInteractor().stop(); });
+  delete eng;
+  for (int i = 0; i < 5; i++)
+  {
+    f3d::engine* eng2 = new f3d::engine(f3d::engine::create(false));
+    eng2->getInteractor().playInteraction(interactionFilePath);
+    delete eng2;
+  }
```
Meakk, line 18: let's modernize the memory management using unique_ptr's
```suggestion
  auto eng = std::make_unique<f3d::engine>(f3d::engine::create(false));
  eng->getInteractor().start(1 / 30, [eng]() { eng->getInteractor().stop(); });
  eng = nullptr;
  for (int i = 0; i < 5; i++)
  {
    auto eng2 = std::make_unique<f3d::engine>(f3d::engine::create(false));
    eng2->getInteractor().playInteraction(interactionFilePath);
  }
```

Meakk (changes requested):
small comment regarding code style.  
Otherwise, it looks good to me

library/testing/TestSDKEngineRecreation.cxx
snoyer: can you scope the first engine too instead of doing `= nullptr`?

Meakk: it's possible too. I'm fine with both

CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.21)
 include(CMakeDependentOption)
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS true)
```
Meakk, line 4: Why is it needed? That's an option we usually set in command line when configuring (`-DCMAKE_EXPORT_COMPILE_COMMANDS=ON`) for static analysis tools (clang-tidy/cppcheck)

hant-hub: Sorry had that there since it made my workflow a bit easier. I can get rid of that.

Meakk (approved):
Thank you!

mwestphal:
@hant-hub thanks for your contribution!

PR #1850, opened 2024-12-30
image: Modernize API for C++17

mwestphal:


library/src/image.cxx
```diff
@@ -175,36 +177,42 @@ image::image(unsigned int width, unsigned int height, unsigned int channelCount,
...
 
-  std::string fullPath = vtksys::SystemTools::CollapseFullPath(path);
-  if (!vtksys::SystemTools::FileExists(fullPath))
+  try
   {
-    throw read_exception("Cannot open file " + path);
-  }
+    if (!fs::exists(filePath))
+    {
+      throw read_exception("Cannot open file " + filePath.string());
```
Meakk, line 189: Can we cover it?

library/src/image.cxx
```diff
@@ -492,14 +500,25 @@ const image& image::save(const std::string& path, SaveFormat format) const
...
+    fs::create_directories(filePath.parent_path());
+
+    writer->SetFileName(filePath.string().c_str());
+    writer->SetInputData(this->Internals->Image);
+    writer->Write();
 
-  if (writer->GetErrorCode() != 0)
+    if (writer->GetErrorCode() != 0)
+    {
+      throw write_exception("Cannot write " + filePath.string());
```
Meakk, line 514: Coverage

Meakk (changes requested):

library/public/image.h
```diff
@@ -45,13 +46,13 @@ class F3D_EXPORT image
   };
 
   /**
-   * Create an image from file, the following formats are supported:
+   * Create an image from provided file path, used as is, the following formats are supported:
```
snoyer, line 49:
```suggestion
   * Read an image from provided file path, used as is, the following formats are supported:
```
"Create" kind of implies making a new/blank one to me? "Read [from]" (or "Load [from]") matches better with the "Save [to]" used later on.

Although I guess it depends whether "image" is supposed to be read as a regular noun or a c++ type because a constructor does indeed _create_ an instance
```suggestion
   * Create an `image` from provided file path, used as is, the following formats are supported:
```

library/src/image.cxx
mwestphal: its supposed to be already covered, this is what I see locally at least.

library/public/image.h
mwestphal: reworded

library/src/image.cxx
snoyer, line 514:
```suggestion
      throw write_exception("Cannot write image " + filePath.string());
```
or
```suggestion
      throw write_exception("Cannot write file " + filePath.string());
```
not sure which one, there's both a `throw read_exception("Cannot open file " + filePath.string());` and a `throw read_exception("Cannot read image " + filePath.string());` on the reading side

Meakk (approved):

PR #1858, opened 2025-01-02
options: Modernize API for C++17

mwestphal:


library/src/options.cxx
```diff
@@ -117,12 +117,12 @@ std::vector<std::string> options::getNames() const
 }
 
 //----------------------------------------------------------------------------
-std::pair<std::string, unsigned int> options::getClosestOption(const std::string& option) const
+std::pair<std::string, unsigned int> options::getClosestOption(std::string_view option) const
```
Meakk, line 120: `utils::textDistance` is currently using strings, let's use string_view too

Meakk (changes requested):

library/src/options.cxx
mwestphal: it is not using strings, it is already using string_view

https://github.com/f3d-app/f3d/pull/1855

Meakk: right, my local branch wasn't up to date

Meakk (approved):

PR #1859, opened 2025-01-02
utils: Adding collapsePath

mwestphal:


library/src/utils.cxx
```diff
@@ -97,6 +99,14 @@ std::vector<std::string> utils::tokenize(std::string_view str)
   return tokens;
 }
 
+//----------------------------------------------------------------------------
+std::string utils::collapsePath(const std::string& path, const std::string& baseDirectory)
+{
+  return path.empty()       ? path
```
Meakk, line 105: weird formatting. clang-format is happy with that?

library/public/utils.h
```diff
@@ -55,6 +55,18 @@ class F3D_EXPORT utils
...
+  /**
+   * Collapse a string filesystem path by:
+   * - Expanding tilda `~` into home dir in a cross-platform way
+   * - Transform relative path into an absolute path based on basedDirectory if provided, or the
+   * current directory if not
+   * - Remove any `..` if any
+   * Rely on vtksys::SystemTools::CollapseFullPath but return empty string if the provided
+   * string is empty.
+   */
+  [[nodiscard]] static std::string collapsePath(
```
Meakk, line 67: should we use `std::filesystem::path` as arguments?

library/src/utils.cxx
mwestphal: its clang-format doing that, not me ^^

library/public/utils.h
mwestphal: nope, because vtksys wants std::string. I dont see the point of converting back and forth

Meakk: I agree but it's semantically more correct. And where do you plan to use this function? I suppose you will have a `fs::path` when calling `collapsePath` so you are converting to a string anyway.

mwestphal: Actually I dont convert to string because I use this method on inputted string by the user, but as you see fit.

library/testing/TestSDKUtils.cxx
```diff
@@ -101,5 +105,17 @@ int TestSDKUtils(int argc, char* argv[])
       .str(),
     "hello {foo}!");
 
+  //
+
+  test("collapsePath: empty", f3d::utils::collapsePath("").empty());
+  test("collapsePath: relative to absolute",
+    fs::path(f3d::utils::collapsePath("folder/file.ext")).is_absolute());
```
Meakk, line 112:
```suggestion
    f3d::utils::collapsePath("folder/file.ext").is_absolute());
```

library/testing/TestSDKUtils.cxx
```diff
@@ -101,5 +105,17 @@ int TestSDKUtils(int argc, char* argv[])
...
+  test("collapsePath: relative with folder",
+    fs::path(f3d::utils::collapsePath("folder/file.ext", "/")) == fs::path("/folder/file.ext"));
```
Meakk, line 114:
```suggestion
    f3d::utils::collapsePath("folder/file.ext", "/") == "/folder/file.ext");
```

library/testing/TestSDKUtils.cxx
```diff
@@ -101,5 +105,17 @@ int TestSDKUtils(int argc, char* argv[])
...
+  test(
+    "collapsePath: remove dotdot", f3d::utils::collapsePath("/folder/../file.ext") == "/file.ext");
+  test("collapsePath: expand home",
+    fs::path(f3d::utils::collapsePath("~/folder/file.ext")).is_absolute());
```
Meakk, line 118:
```suggestion
    f3d::utils::collapsePath("~/folder/file.ext").is_absolute());
```

Meakk (changes requested):

Meakk (approved):

Issue #1741, opened 2024-11-29
Rolling camera and then interacting with the view provide incorrect interactions


mwestphal:
**Describe the bug**
Rolling the camera (by pressing `6`) and then interacting with the mouse is incorrect.
The camera "jerks" back to the original position.

**To Reproduce**
 1. Open the file using `f3d --dry-run dragon.vtu`
 2. press 6
 3. interact with view

**Expected behavior**
No jerking

mwestphal:
Caused by 

```
    // orthogonalize up vector based on focal direction
    vtkMath::MultiplyScalar(dir, dot);
    vtkMath::Subtract(up, dir, dir);
    vtkMath::Normalize(dir);
    camera->SetViewUp(dir);
```

in /vtkF3DInteractorStyle.cxx

Since rolling camera does not impact view-up, im not sure of the right way to fix this.

TopViewCamera as the same issue

samoncrief:
I'm looking into this now, feel free to assign it to me

mwestphal:
Sure, assigned.

A possible implementation is to store a temporary view-up when rolling

mwestphal:
Fixed by #1879

PR #1879, opened 2025-01-05
#1741 Rolling camera and then interacting with the view provide incorrect interactions

samoncrief:
A fix for the following bug: https://github.com/f3d-app/f3d/issues/1741


mwestphal (commented):
Nice! Please add a test. A simple interaction test should do the trick

application/testing/CMakeLists.txt
```diff
@@ -825,6 +825,7 @@ f3d_test(NAME TestInteractionOrthographicProjection DATA cow.vtp INTERACTION) #5
 f3d_test(NAME TestInteractionZoomToggleOrthographicProjection DATA cow.vtp INTERACTION) #MouseWheel;5;Mousewheelx6;5
 f3d_test(NAME TestInteractionRotateCameraMinus90 DATA f3d.glb INTERACTION)
 f3d_test(NAME TestInteractionRotateCamera90 DATA f3d.glb INTERACTION)
+f3d_test(NAME TestInteractionRollCameraRotation DATA f3d.glb INTERACTION)
```
mwestphal, line 828: please add a comment after the test detailing what does the interaction file do

library/src/interactor_impl.cxx
```diff
@@ -187,6 +187,7 @@ class interactor_impl::internals
     /* set camera coordinates back */
     cam.setPosition(newPos);
     cam.setViewUp(up);
+    Style->EndTemporaryUp();
```
Meakk, line 190: Where is `Style` coming from? Add `this->` for clarity if it's a class member

Meakk (commented):
I'm trying to understand the behavior with this PR. When the camera is rolled, the temporary up is used, but when do we restore the old up value? Only when we reset the camera?  
I think we should slowing converge back the temporary up direction to the initial up direction by doing a slerp

mwestphal:
> I'm trying to understand the behavior with this PR. When the camera is rolled, the temporary up is used, but when do we restore the old up value? Only when we reset the camera? I think we should slowing converge back the temporary up direction to the initial up direction by doing a slerp

Thats not what we want here I think.

samoncrief:
> I'm trying to understand the behavior with this PR. When the camera is rolled, the temporary up is used, but when do we restore the old up value? Only when we reset the camera? I think we should slowing converge back the temporary up direction to the initial up direction by doing a slerp

The bug was that rotating the camera around the object while rolled causes the camera to jerk back to its original orientation. This PR does not change up in the renderer, it gives the interactor a definition to use after rolling with `roll_camera` that reflects the view up of the camera. When the camera is reset, the camera is oriented in a way where the renderer's up value is correct again, so at that point the flag is set to false.

application/testing/CMakeLists.txt
samoncrief: comment added using notation from other tests

library/src/interactor_impl.cxx
samoncrief: Added

library/src/interactor_impl.cxx
```diff
@@ -187,6 +187,7 @@ class interactor_impl::internals
     /* set camera coordinates back */
     cam.setPosition(newPos);
     cam.setViewUp(up);
+    this->Style->EndTemporaryUp();
```
mwestphal, line 190: can this line be put somewhere else ? It doesnt seem to need to be in between camera lines.

samoncrief: Fair point, moving to the `set_camera` definition to be similar to the other use cases.

vtkext/private/module/vtkF3DInteractorStyle.cxx
```diff
@@ -153,6 +153,8 @@ void vtkF3DInteractorStyle::Rotate()
   double dir[3];
   camera->GetDirectionOfProjection(dir);
   double* up = ren->GetUpVector();
+  InterpolateTemporaryUp(0.1, up);
```
mwestphal, line 156:
```suggestion
  this->InterpolateTemporaryUp(0.1, up);
```

vtkext/private/module/vtkF3DInteractorStyle.h
```diff
@@ -106,13 +106,26 @@ class vtkF3DInteractorStyle : public vtkInteractorStyleTrackballCamera
...
    * Overridden to support being disabled
    */
   void Dolly(double factor) override;
 
   bool CameraMovementDisabled = false;
+
+  /**
+   * Temporary up vector to support rolled camera interaction
+   */
+  double TemporaryUp[3] = { 0, 0, 0 };
```
mwestphal, line 128: does this default value has an impact in normal usage, when not "rolled" ?

samoncrief: Originally, this was meant as a way to have a value on initialization. With the interpolate step, now if it is interpolated before being set it would normalize into the target vector. If a default null pointer is preferred I'll adjust.

Meakk: I think it's fine since we set the value at initialization anyway

testing/baselines/TestInteractionRollCameraRotation.png
Meakk, line 1: It's hard to know just by looking at the baseline if the camera was properly restored. Can you add the grid in the test by adding `ARGS --grid` in the `f3d_test` call?

samoncrief: Added

Meakk: Why is it flipped in the baseline? The interpolation to the initial up isn't completed yet? (it's fine if it's the case, I'm just asking)

samoncrief: Yes, this baseline is to test that the original bug (snapping back to up) is not occurring. Getting a value between 0 and 90 degrees is intentional.

Meakk: alright, sounds good then

Meakk (approved):

vtkext/private/module/vtkF3DInteractorStyle.h
```diff
@@ -106,13 +106,26 @@ class vtkF3DInteractorStyle : public vtkInteractorStyleTrackballCamera
...
   void FindPokedRenderer(int vtkNotUsed(x), int vtkNotUsed(y));
 
+  /**
+   * Temporary up vector to support rolled camera interaction
+   */
+  void ResetTemporaryUp();
+  void SetTemporaryUp(const double *tempUp);
+  void InterpolateTemporaryUp(const double factor, const double* input);
+
+
```
mwestphal, line 116: you can delete this line

mwestphal (changes requested):
small style change then gtg

mwestphal (approved):

mwestphal:
Thanks for your contribution @samoncrief !

PR #1886, opened 2025-01-05
Improve fps counter 

GabrielNakamoto:
> Accumulate it over a few frames to give an average instead of showing the last frame performance only

- In vtkF3DUIActor I changed `SetFpsValue()` to `UpdateFpsValue()` which will automatically accumulate fps values before setting the display value to the average. The number of frames it waits to average is decided by the static member variable `vtkF3DUIActor::FramesToAverage`

> Updating it only when the whole scene is rendered

- In the `vtkF3DRenderer::Render()` function I updated it to check if its information has `vtkF3DRenderPass::RENDER_UI_ONLY()` set to 1 If it does it calculates the fps value and calls the UI Actor's `UpdateFpsValue()`

Let me know what you think the number of frames to average should be (currently set to 5).

From building it and using some of the example models it seems that the render UI only is working however I may be wrong.

Seeing this is my first time contributing I'm starting a pull request to get some advice.

### DIFF
--- a/vtkext/private/module/vtkF3DRenderer.cxx
+++ b/vtkext/private/module/vtkF3DRenderer.cxx
@@ -1545,21 +1545,21 @@ void vtkF3DRenderer::Render()
 
   bool uiOnly = info->Get(vtkF3DRenderPass::RENDER_UI_ONLY());
 
-  if (! uiOnly)
+  if (!uiOnly)
   {
 	  // Get CPU frame time
-	  double elapsedTime = std::chrono::duration_cast<std::chrono::microseconds>(cpuElapsed).count() * 1e-6;
+    double elapsedTime = std::chrono::duration_cast<std::chrono::microseconds>(cpuElapsed).count() * 1e-6;
 
 #if !defined(__ANDROID__) && !defined(__EMSCRIPTEN__)
-	  glEndQuery(GL_TIME_ELAPSED);
-	  GLint elapsed;
-	  glGetQueryObjectiv(this->Timer, GL_QUERY_RESULT, &elapsed);
+	glEndQuery(GL_TIME_ELAPSED);
+	GLint elapsed;
+	glGetQueryObjectiv(this->Timer, GL_QUERY_RESULT, &elapsed);
 
-	  // Get min between CPU frame time and GPU frame time
-	  elapsedTime = std::min(elapsedTime, elapsed * 1e-9);
+	// Get min between CPU frame time and GPU frame time
+	elapsedTime = std::min(elapsedTime, elapsed * 1e-9);
 #endif
 
-	  this->UIActor->UpdateFpsValue(elapsedTime);
+	this->UIActor->UpdateFpsValue(elapsedTime);
   }
 }
 
--- a/vtkext/private/module/vtkF3DRenderer.cxx
+++ b/vtkext/private/module/vtkF3DRenderer.cxx
@@ -1554,7 +1554,7 @@ void vtkF3DRenderer::Render()
   fps = std::min(fps, static_cast<int>(std::round(1.0 / (elapsed * 1e-9))));
 #endif
 
-  this->UIActor->SetFpsValue(fps);
+  this->UIActor->UpdateFpsValue(fps);
 }
 
 //----------------------------------------------------------------------------
--- a/vtkext/private/module/vtkF3DRenderer.cxx
+++ b/vtkext/private/module/vtkF3DRenderer.cxx
@@ -1547,19 +1547,19 @@ void vtkF3DRenderer::Render()
 
   if (!uiOnly)
   {
-	  // Get CPU frame time
-    double elapsedTime = std::chrono::duration_cast<std::chrono::microseconds>(cpuElapsed).count() * 1e-6;
+   // Get CPU frame time
+   double elapsedTime = std::chrono::duration_cast<std::chrono::microseconds>(cpuElapsed).count() * 1e-6;
 
 #if !defined(__ANDROID__) && !defined(__EMSCRIPTEN__)
-	glEndQuery(GL_TIME_ELAPSED);
-	GLint elapsed;
-	glGetQueryObjectiv(this->Timer, GL_QUERY_RESULT, &elapsed);
+   glEndQuery(GL_TIME_ELAPSED);
+   GLint elapsed;
+   glGetQueryObjectiv(this->Timer, GL_QUERY_RESULT, &elapsed);
 
-	// Get min between CPU frame time and GPU frame time
-	elapsedTime = std::min(elapsedTime, elapsed * 1e-9);
+   // Get min between CPU frame time and GPU frame time
+   elapsedTime = std::min(elapsedTime, elapsed * 1e-9);
 #endif
 
-	this->UIActor->UpdateFpsValue(elapsedTime);
+   this->UIActor->UpdateFpsValue(elapsedTime);
   }
 }
 
--- a/vtkext/private/module/vtkF3DRenderer.cxx
+++ b/vtkext/private/module/vtkF3DRenderer.cxx
@@ -1547,20 +1547,19 @@ void vtkF3DRenderer::Render()
 
   if (! uiOnly)
   {
-  // Get CPU frame per seconds
-  int fps = static_cast<int>(std::round(
-    1.0 / (std::chrono::duration_cast<std::chrono::microseconds>(cpuElapsed).count() * 1e-6)));
+	  // Get CPU frame time
+	  double elapsedTime = std::chrono::duration_cast<std::chrono::microseconds>(cpuElapsed).count() * 1e-6;
 
 #if !defined(__ANDROID__) && !defined(__EMSCRIPTEN__)
-  glEndQuery(GL_TIME_ELAPSED);
-  GLint elapsed;
-  glGetQueryObjectiv(this->Timer, GL_QUERY_RESULT, &elapsed);
+	  glEndQuery(GL_TIME_ELAPSED);
+	  GLint elapsed;
+	  glGetQueryObjectiv(this->Timer, GL_QUERY_RESULT, &elapsed);
 
-  // Get min between CPU frame per seconds and GPU frame per seconds
-  fps = std::min(fps, static_cast<int>(std::round(1.0 / (elapsed * 1e-9))));
+	  // Get min between CPU frame time and GPU frame time
+	  elapsedTime = std::min(elapsedTime, elapsed * 1e-9);
 #endif
 
-  this->UIActor->UpdateFpsValue(fps);
+	  this->UIActor->UpdateFpsValue(elapsedTime);
   }
 }
 
--- a/vtkext/private/module/vtkF3DRenderer.cxx
+++ b/vtkext/private/module/vtkF3DRenderer.cxx
@@ -1541,6 +1541,12 @@ void vtkF3DRenderer::Render()
 
   auto cpuElapsed = std::chrono::high_resolution_clock::now() - cpuStart;
 
+  vtkInformation *info = this->GetInformation();
+
+  bool uiOnly = info->Get(vtkF3DRenderPass::RENDER_UI_ONLY());
+
+  if (! uiOnly)
+  {
   // Get CPU frame per seconds
   int fps = static_cast<int>(std::round(
     1.0 / (std::chrono::duration_cast<std::chrono::microseconds>(cpuElapsed).count() * 1e-6)));
@@ -1555,6 +1561,7 @@ void vtkF3DRenderer::Render()
 #endif
 
   this->UIActor->UpdateFpsValue(fps);
+  }
 }
 
 //----------------------------------------------------------------------------
--- a/vtkext/private/module/vtkF3DUIActor.cxx
+++ b/vtkext/private/module/vtkF3DUIActor.cxx
@@ -72,7 +72,7 @@ void vtkF3DUIActor::UpdateFpsValue(int fps)
   this->AccumulatedFpsValue += fps;
   this->FramesAccumulated++;
 
-  if (this->FramesAccumulated == vtkF3DUI::FramesToAverage)
+  if (this->FramesAccumulated == vtkF3DUIActor::FramesToAverage)
   {
 	  this->FpsValue = static_cast<int>(this->AccumulatedFpsValue / this->FramesAccumulated);
 
--- a/vtkext/private/module/vtkF3DUIActor.cxx
+++ b/vtkext/private/module/vtkF3DUIActor.cxx
@@ -71,16 +71,17 @@ void vtkF3DUIActor::SetFpsCounterVisibility(bool show)
 //----------------------------------------------------------------------------
 void vtkF3DUIActor::UpdateFpsValue(const double elapsedFrameTime)
 {
+	this->TotalFrameTimes += elapsedFrameTime;
 	this->FrameTimes.push_back(elapsedFrameTime);
 
-	// add window size check here
-	if (this->FrameTimes.size() == 6)
+	while (this->TotalFrameTimes > 1.0)
 	{
+		double t = this->FrameTimes.front();
 		this->FrameTimes.pop_front();
+		this->TotalFrameTimes -= t;
 	}
 
-	double averageFrameTime = std::accumulate(this->FrameTimes.begin(), this->FrameTimes.end(), 0.0) / this->FrameTimes.size();
-
+	double averageFrameTime = this->TotalFrameTimes / this->FrameTimes.size();
 	this->FpsValue = static_cast<int>(std::round(1.0 / averageFrameTime));
 }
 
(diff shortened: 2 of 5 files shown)


vtkext/private/module/vtkF3DUIActor.cxx
```diff
@@ -67,9 +67,18 @@ void vtkF3DUIActor::SetFpsCounterVisibility(bool show)
...
-void vtkF3DUIActor::SetFpsValue(int fps)
+void vtkF3DUIActor::UpdateFpsValue(int fps)
 {
-  this->FpsValue = fps;
+  this->AccumulatedFpsValue += fps;
+  this->FramesAccumulated++;
+
+  if (this->FramesAccumulated == vtkF3DUIActor::FramesToAverage)
+  {
+	  this->FpsValue = static_cast<int>(this->AccumulatedFpsValue / this->FramesAccumulated);
```
mwestphal, line 77: IMO it should be a slided average, wdyt ?

mwestphal (commented):
This is looking good, wdyt @Meakk ?

vtkext/private/module/vtkF3DUIActor.cxx
GabrielNakamoto: Yes that looks better. I can add that once I'm home from school.

Meakk: I think the average must be done in frame time, rather than fps.  
So maybe you could pass the elapsed time in this function, average the elapsed time, and only inverse it at the end to convert to fps.  
If you don't do that, the average will be biased with highest fps values.

Also, I don't like the idea of the fps value at 0 for the first 5 frames.

About the slided average, I'm afraid it will still produce large variation of fps and will still make the fps value unreadable. But maybe if we increase the window size to something larger than 5, this effect could be mitigated.

Here's a suggestion:
- Add the elapsed time in a container (`std::queue` but maybe another container is more adapted)
- Remove the oldest value if the number of elements is above the window size
- Average all the elements
- Inverse the value to get the fps value

Let's see if the value can be read and if changing the window size have an effect

GabrielNakamoto: Sounds good will try to implement this.

GabrielNakamoto: > * Remove the oldest value if the number of elements is above the window size

How can the renderer get the window size (in it's information maybe?). And why should that affect the fps average?

GabrielNakamoto:
Don't review the latest commit yet I have to fix some things sorry

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -1541,20 +1541,26 @@ void vtkF3DRenderer::Render()
...
+	  glGetQueryObjectiv(this->Timer, GL_QUERY_RESULT, &elapsed);
 
-  // Get min between CPU frame per seconds and GPU frame per seconds
-  fps = std::min(fps, static_cast<int>(std::round(1.0 / (elapsed * 1e-9))));
+	  // Get min between CPU frame time and GPU frame time
+	  elapsedTime = std::min(elapsedTime, elapsed * 1e-9);
 #endif
 
-  this->UIActor->SetFpsValue(fps);
+	  this->UIActor->UpdateFpsValue(elapsedTime);
```
mwestphal, line 1562: would it make sense to only compute the FPS when we need it (aka when the fps counter is displayed) ?

vtkext/private/module/vtkF3DUIActor.cxx
Meakk: > How can the renderer get the window size (in it's information maybe?). And why should that affect the fps average?

Sorry for the confusion, what I call "window size" is the number of elements we want in the list (6 in your case).

vtkext/private/module/vtkF3DRenderer.cxx
Meakk: It's cheap, I think it's fine to update it in the background so we can have a good value directly when the user make it visible.

vtkext/private/module/vtkF3DUIActor.cxx
```diff
@@ -67,9 +69,19 @@ void vtkF3DUIActor::SetFpsCounterVisibility(bool show)
...
 
 //----------------------------------------------------------------------------
-void vtkF3DUIActor::SetFpsValue(int fps)
+void vtkF3DUIActor::UpdateFpsValue(const double elapsedFrameTime)
 {
-  this->FpsValue = fps;
+	this->FrameTimes.push_back(elapsedFrameTime);
+
+	// add window size check here
+	if (this->FrameTimes.size() == 6)
```
Meakk, line 77: Actually, 6 is too small I think. I tried locally and with +700fps, it's still barely readable.  
Obviously, a large number might be a problem because it will increase the cost of the `std::accumulate`.  
I suggest we keep track of the total elapsed time instead of computing it everytime in a new variable.  
Then, if the elapsed time exceeds 1 second, let `pop_front()` and subtract the removed elements until the total goes below 1 second.

```cpp
this->TotalFrameTimes += elapsedFrameTime;
this->FrameTimes.push_back(elapsedFrameTime);

while (this->TotalFrameTimes > 1.0)
{
  double t = this->FrameTimes.pop_front();
  this->TotalFrameTimes -= t;
}

double averageFrameTime = std::accumulate(this->FrameTimes.begin(), this->FrameTimes.end(), 0.0) / this->FrameTimes.size();
this->FpsValue = static_cast<int>(std::round(1.0 / averageFrameTime));
```

What do you think?

GabrielNakamoto: Yeah so the 6 was just a placeholder because I didn't know how to check the window size, I'll add your changes in and see how it does!

Meakk: Cool! I haven't tested the code above, so let me know if it works.

GabrielNakamoto: Just pushed the change.

I still don't think it is behaving properly, at least on my computer. I get way too high fps values most of the time... I don't know why though?

Meakk: I'll take a look

vtkext/private/module/vtkF3DUIActor.cxx
```diff
@@ -4,6 +4,8 @@
 #include <vtkOpenGLRenderWindow.h>
 #include <vtkViewport.h>
 
+#include <numeric>
```
Meakk, line 7: You probably don't need this include anymore

vtkext/private/module/vtkF3DUIActor.cxx
Meakk: I think it works well!

https://github.com/user-attachments/assets/df2b82db-612f-44dc-a171-c5cde981858d

vtkext/private/module/vtkF3DUIActor.cxx
```diff
@@ -67,9 +69,20 @@ void vtkF3DUIActor::SetFpsCounterVisibility(bool show)
...
 {
-  this->FpsValue = fps;
+	this->TotalFrameTimes += elapsedFrameTime;
+	this->FrameTimes.push_back(elapsedFrameTime);
+
+	while (this->TotalFrameTimes > 1.0)
+	{
+		double t = this->FrameTimes.front();
+		this->FrameTimes.pop_front();
+		this->TotalFrameTimes -= t;
```
Meakk, line 81: In order to cover these lines I think you have no choice, you need a VTK test.  
You can do something like that:
- Add a file `vtkext\private\module\Testing\TestF3DFpsCounter.cxx` containing:
```cpp
#include <vtkNew.h>
#include <vtkObjectFactory.h>

#include "vtkF3DUIActor.h"

#include <iostream>

// subclass to retrieve internal protected values
class vtkF3DTestUIActor : public vtkF3DUIActor
{
public:
  static vtkF3DTestUIActor* New();
  vtkTypeMacro(vtkF3DTestUIActor, vtkF3DUIActor);

  size_t GetNumberOfFrameTimes() { return this->FrameTimes.size(); }
  double GetTotalFrameTimes() { return this->TotalFrameTimes; }
  int GetFpsValue() { return this->FpsValue; }
};

vtkObjectFactoryNewMacro(vtkF3DTestUIActor);

int TestF3DFpsCounter(int argc, char* argv[])
{
  vtkNew<vtkF3DTestUIActor> uiActor;

  // add 1000 frames at 0.01s
  for (int i = 0; i < 1000; i++)
  {
    uiActor->UpdateFpsValue(0.01);
  }

  // make sure only the 100 last records are kept
  if (uiActor->GetNumberOfFrameTimes() > 100)
  {
    std::cerr << "Number of frame times must be at most 100" << std::endl;
    return EXIT_FAILURE;
  }

  // make sure only the total time kept doesn't exceed 1 second
  if (uiActor->GetTotalFrameTimes() > 1.0)
  {
    std::cerr << "Number of total frame times must be at most 1.0" << std::endl;
    return EXIT_FAILURE;
  }

  // make sure the FPS value is exactly 100
  if (uiActor->GetFpsValue() != 100)
  {
    std::cerr << "Number of FPS value must be exactly 100" << std::endl;
    return EXIT_FAILURE;
  }

  return EXIT_SUCCESS;
}
```
- Make sure to register this file in `vtkext\private\module\Testing\CMakeLists.txt` in the `test_sources` list

GabrielNakamoto: Okay sounds good, thanks for all the help!

Meakk (approved):
Great! Thank you for your contribution!

vtkext/private/module/vtkF3DUIActor.cxx
```diff
@@ -67,9 +67,20 @@ void vtkF3DUIActor::SetFpsCounterVisibility(bool show)
...
-void vtkF3DUIActor::SetFpsValue(int fps)
+void vtkF3DUIActor::UpdateFpsValue(const double elapsedFrameTime)
 {
-  this->FpsValue = fps;
+	this->TotalFrameTimes += elapsedFrameTime;
+	this->FrameTimes.push_back(elapsedFrameTime);
+
+	while (this->TotalFrameTimes > 1.0)
+	{
+		double t = this->FrameTimes.front();
```
mwestphal, line 77: please avoid single letter var

mwestphal (changes requested):
please convert all tabs to space

Meakk:
> please convert all tabs to space

Why is clang-format not detecting that?

mwestphal:
> > please convert all tabs to space
> 
> Why is clang-format not detecting that?

no idea, to improve definitely

GabrielNakamoto:
I pushed the style changes

vtkext/private/module/vtkF3DUIActor.cxx
```diff
@@ -67,9 +67,21 @@ void vtkF3DUIActor::SetFpsCounterVisibility(bool show)
...
-  this->FpsValue = fps;
+  this->TotalFrameTimes += elapsedFrameTime;
+  this->FrameTimes.push_back(elapsedFrameTime);
+
+  while (this->TotalFrameTimes > 1.0)
+  {
+    double oldestFrameTime = this->FrameTimes.front();
+
+    this->FrameTimes.pop_front();
+	this->TotalFrameTimes -= oldestFrameTime;
```
mwestphal, line 80:
```suggestion
    this->TotalFrameTimes -= oldestFrameTime;
```

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -1541,20 +1541,26 @@ void vtkF3DRenderer::Render()
...
-  GLint elapsed;
-  glGetQueryObjectiv(this->Timer, GL_QUERY_RESULT, &elapsed);
+	glEndQuery(GL_TIME_ELAPSED);
+	GLint elapsed;
+	glGetQueryObjectiv(this->Timer, GL_QUERY_RESULT, &elapsed);
 
-  // Get min between CPU frame per seconds and GPU frame per seconds
-  fps = std::min(fps, static_cast<int>(std::round(1.0 / (elapsed * 1e-9))));
+	// Get min between CPU frame time and GPU frame time
+	elapsedTime = std::min(elapsedTime, elapsed * 1e-9);
```
mwestphal, line 1559: still a bunch of tabs here

GabrielNakamoto: Sorry about that

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -1541,20 +1541,26 @@ void vtkF3DRenderer::Render()
...
+   // Get CPU frame time
+   double elapsedTime = std::chrono::duration_cast<std::chrono::microseconds>(cpuElapsed).count() * 1e-6;
 
 #if !defined(__ANDROID__) && !defined(__EMSCRIPTEN__)
-  glEndQuery(GL_TIME_ELAPSED);
-  GLint elapsed;
-  glGetQueryObjectiv(this->Timer, GL_QUERY_RESULT, &elapsed);
+   glEndQuery(GL_TIME_ELAPSED);
+   GLint elapsed;
+   glGetQueryObjectiv(this->Timer, GL_QUERY_RESULT, &elapsed);
```
mwestphal, line 1556:
```suggestion
  glEndQuery(GL_TIME_ELAPSED);
  GLint elapsed;
  glGetQueryObjectiv(this->Timer, GL_QUERY_RESULT, &elapsed);
```

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -1541,20 +1541,26 @@ void vtkF3DRenderer::Render()
...
 
-  // Get min between CPU frame per seconds and GPU frame per seconds
-  fps = std::min(fps, static_cast<int>(std::round(1.0 / (elapsed * 1e-9))));
+   // Get min between CPU frame time and GPU frame time
+   elapsedTime = std::min(elapsedTime, elapsed * 1e-9);
```
mwestphal, line 1559:
```suggestion
  // Get min between CPU frame time and GPU frame time
  elapsedTime = std::min(elapsedTime, elapsed * 1e-9);
```

vtkext/private/module/vtkF3DRenderer.cxx
mwestphal, line 1556:
```suggestion
    glEndQuery(GL_TIME_ELAPSED);
    GLint elapsed;
    glGetQueryObjectiv(this->Timer, GL_QUERY_RESULT, &elapsed);
```

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -1541,20 +1541,26 @@ void vtkF3DRenderer::Render()
...
-  int fps = static_cast<int>(std::round(
-    1.0 / (std::chrono::duration_cast<std::chrono::microseconds>(cpuElapsed).count() * 1e-6)));
+  vtkInformation *info = this->GetInformation();
+
+  bool uiOnly = info->Get(vtkF3DRenderPass::RENDER_UI_ONLY());
+
+  if (!uiOnly)
+  {
+   // Get CPU frame time
+   double elapsedTime = std::chrono::duration_cast<std::chrono::microseconds>(cpuElapsed).count() * 1e-6;
```
mwestphal, line 1551:
```suggestion
    // Get CPU frame time
    double elapsedTime = std::chrono::duration_cast<std::chrono::microseconds>(cpuElapsed).count() * 1e-6;
```

vtkext/private/module/vtkF3DRenderer.cxx
mwestphal, line 1559:
```suggestion
    // Get min between CPU frame time and GPU frame time
    elapsedTime = std::min(elapsedTime, elapsed * 1e-9);
```

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -1541,20 +1541,26 @@ void vtkF3DRenderer::Render()
...
+   glGetQueryObjectiv(this->Timer, GL_QUERY_RESULT, &elapsed);
 
-  // Get min between CPU frame per seconds and GPU frame per seconds
-  fps = std::min(fps, static_cast<int>(std::round(1.0 / (elapsed * 1e-9))));
+   // Get min between CPU frame time and GPU frame time
+   elapsedTime = std::min(elapsedTime, elapsed * 1e-9);
 #endif
 
-  this->UIActor->SetFpsValue(fps);
+   this->UIActor->UpdateFpsValue(elapsedTime);
```
mwestphal, line 1562:
```suggestion
    this->UIActor->UpdateFpsValue(elapsedTime);
```

mwestphal (changes requested):
still incorrect indenting, sorrt about that

Meakk:
Let's merge it. Formatting will be handle in https://github.com/f3d-app/f3d/pull/1912

Meakk:
Thanks @GabrielNakamoto!

PR #1918, opened 2025-01-16
Skip frame when we are late for animation

mwestphal:


library/src/animationManager.cxx
```diff
@@ -204,6 +204,17 @@ void animationManager::Tick()
 {
   if (this->Playing)
   {
+    std::chrono::steady_clock::time_point currentTick = std::chrono::steady_clock::now();
+    if (!vtksys::SystemTools::HasEnv("CTEST_F3D_PREVENT_SKIP_FRAME") &&
```
Meakk, line 208: I remember that querying an environment variable is actually quite expensive.  
Can we move it to `animationManager` constructor and save it to a class member?

Meakk:
You don't need to wait for the fix in VTK and update the VTK sha?

library/src/animationManager.cxx
mwestphal: good point.

mwestphal:
> You don't need to wait for the fix in VTK and update the VTK sha?

We need both to fix animation on Windows with the usd file, but in essence these are separated features.

Meakk (approved):

PR #1938, opened 2025-01-21
Improve ci graph

mwestphal:
Add a cache_dependencies job for each platform so that all dependencies but VTK is cached.

Job where a dependency (OCCT 7_8_0) is rebuilt:
 - cache_dependencies: 29min14: https://github.com/f3d-app/f3d/actions/runs/12891637495/job/35944038822
 - any windows_ci: 15min: 
   - https://github.com/f3d-app/f3d/actions/runs/12891637495/job/35946831929
   - https://github.com/f3d-app/f3d/actions/runs/12891637495/job/35946832815
 
 Job where no dependency is rebuilt:
  - cache_dependencies: 59s: https://github.com/f3d-app/f3d/actions/runs/12890727504/job/35941005286
  - any windows_ci: 15min: https://github.com/f3d-app/f3d/actions/runs/12890727504/job/35941110620
  
  Fix: https://github.com/f3d-app/f3d/issues/1939

### DIFF
--- a/.github/actions/coverage-ci/action.yml
+++ b/.github/actions/coverage-ci/action.yml
@@ -25,28 +25,16 @@ runs:
         lfs_sha: ${{inputs.lfs_sha}}
         cache_postfix: cache-0 
 
-    - name: Dependencies Dir
-      shell: bash
-      working-directory: ${{github.workspace}}
-      run: |
-        mkdir dependencies
-        cd dependencies
-        mkdir install
-
-    - name: Install VTK dependencies
-      uses: ./source/.github/actions/vtk-dependencies
-
-    - name: Install Raytracing Dependencies
-      uses: ./source/.github/actions/ospray-sb-install-dep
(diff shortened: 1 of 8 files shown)


Meakk:
Is the cache properly saved even if the CI fails?

mwestphal:
> Is the cache properly saved even if the CI fails?

The cache_dependencies job will indeed save the cache if another job fail. However, if we the cache_dependencies job fails for any reason, the individual caches will not be saved, even if the step for this specific cache succesful.

Alternatively I could create individual job for each dependency, but it would require to have a way to handle dependencies depency, which is basically package management. Since this must be at job/workflow level so that cache is being saved.

It seems to be possible by using reusable workflow, I can definitely try it, let me know what you think.

Meakk (approved):

Meakk:
It's ok, let's stick with that for now.

PR #1946, opened 2025-01-24
merge release into master

mwestphal:
No patch release planned, merging release into master.

### DIFF
--- a/application/F3DColorMapTools.cxx
+++ b/application/F3DColorMapTools.cxx
@@ -26,14 +26,17 @@ fs::path Find(const std::string& str)
       }
     }
 
-    std::vector<fs::path> dirsToCheck{ F3DSystemTools::GetUserConfigFileDirectory() / "colormaps",
+    std::vector<fs::path> dirsToCheck
+    {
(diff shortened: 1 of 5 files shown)


Meakk:
Why is the formatting failing here?

mwestphal:
> Why is the formatting failing here?

Because the workflow is using the one from master but this commit HEAD doesnt contains the needed source change.

Meakk:
> > Why is the formatting failing here?
> 
> Because the workflow is using the one from master but this commit HEAD doesnt contains the needed source change.

Should we fix the formatting in this PR then?

mwestphal:
> > > Why is the formatting failing here?
> > 
> > 
> > Because the workflow is using the one from master but this commit HEAD doesnt contains the needed source change.
> 
> Should we fix the formatting in this PR then?

Already fixed in master, nothing to fix here.

We could fix it in release, but there is no point.

Meakk:
I don't get it. Aren't you breaking the formatting in master with this PR?

mwestphal:
> I don't get it. Aren't you breaking the formatting in master with this PR?

Sorry I should have checked the diff first, Something is wrong here.

PR #1948, opened 2025-01-27
CI: Improve versions mgt logic

mwestphal:
- Use save/restore logic for dependencies cache to avoid rebuild dependencies when not needed
 - Improve dependents logic to invalidate cache when another dependencies changes
 - Add support for not restoring cache of not needed dependencies when version is empty
 - Add support for not build OpenVDB and building VTK without OpenVDB

### DIFF
--- a/.github/actions/alembic-install-dep/action.yml
+++ b/.github/actions/alembic-install-dep/action.yml
@@ -20,7 +20,7 @@ runs:
 
     - name: Cache Alembic
       id: cache-alembic
-      uses: actions/cache@v4
+      uses: actions/cache/restore@v4
       with:
         path: dependencies/alembic_install
         key: alembic-${{inputs.version}}-${{runner.os}}-${{inputs.cpu}}-0
@@ -67,3 +67,10 @@ runs:
       working-directory: ${{github.workspace}}/dependencies/alembic_install
       shell: bash
       run: cp -r ./* ../install/
+
+    - name: Save cache
+      if: steps.cache-alembic.outputs.cache-hit != 'true'
+      uses: actions/cache/save@v4
+      with:
+        key: ${{ steps.cache-alembic.outputs.cache-primary-key }}
+        path: dependencies/alembic_install
--- a/.github/actions/f3d-dependencies/action.yml
+++ b/.github/actions/f3d-dependencies/action.yml
@@ -7,43 +7,46 @@ inputs:
     default: 'x86_64'
   alembic_version:
     description: 'Version of alembic to build'
-    required: true
+    required: false
   assimp_version:
     description: 'Version of assimp to build'
-    required: true
+    required: false
   draco_version:
     description: 'Version of draco to build'
-    required: true
+    required: false
   occt_version:
     description: 'Version of occt to build'
-    required: true
+    required: false
   openexr_version:
     description: 'Version of openexr to build'
-    required: true
+    required: false
   pybind11_version:
     description: 'Version of pybind11 to build'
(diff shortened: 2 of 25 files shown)


.github/actions/alembic-install-dep/action.yml
```diff
@@ -20,7 +20,7 @@ runs:
 
     - name: Cache Alembic
       id: cache-alembic
-      uses: actions/cache@v4
+      uses: actions/cache/restore@v4
```
Meakk, line 23: Is it disabling automatic cache save? So using `cache/restore` and `cache/save` is more flexible?

.github/actions/f3d-dependencies/action.yml
```diff
@@ -7,43 +7,46 @@ inputs:
...
     description: 'Version of usd to build'
-    required: true
+    required: false
 
 runs:
   using: "composite"
   steps:
 
     - name: Install OCCT dependency
+      if: inputs.occt_version != ''
```
Meakk, line 35: it means that no version specified is equivalent to no dependency build?

.github/actions/occt-install-dep/action.yml
```diff
@@ -54,7 +54,7 @@ runs:
       if: steps.cache-occt.outputs.cache-hit != 'true'
       working-directory: ${{github.workspace}}/dependencies
       shell: bash
-      run:
+      run: |
```
Meakk, line 57: is it a fix?

.github/actions/usd-install-dep/action.yml
```diff
@@ -8,17 +8,26 @@ inputs:
...
+        [[ "${{ inputs.openexr_version }}" ]] || { echo "openexr_version input is empty" ; exit 1; }
+
     - name: Cache USD
       id: cache-usd
-      uses: actions/cache@v4
+      uses: actions/cache/restore@v4
       with:
         path: dependencies/usd_install
-        key: usd-${{inputs.version}}-${{runner.os}}-${{inputs.cpu}}-3
+        key: usd-${{inputs.version}}-${{inputs.openexr_version}}-${{runner.os}}-${{inputs.cpu}}-3
```
Meakk, line 30: Reset key increment to 0

```suggestion
        key: usd-${{inputs.version}}-${{inputs.openexr_version}}-${{runner.os}}-${{inputs.cpu}}-0
```

.github/actions/alembic-install-dep/action.yml
mwestphal: Exactly

.github/actions/f3d-dependencies/action.yml
mwestphal: Yes, that's the point

.github/actions/occt-install-dep/action.yml
mwestphal: It seems to work without it but it's more consistent with it

Meakk: I think `mkdir occt_install` was probably ignored, but it's not a big deal since the directory is created automatically by CMake during the installation step.  
Ideally, `mkdir occt_build` shouldn't be necessary if we use `cmake -B x -S x` so we can remove this job step entirely.  
Not related to this PR tho, we can do it later.

Meakk (approved):

.github/actions/occt-install-dep/action.yml
mwestphal: good point!

PR #1949, opened 2025-01-28
CI: Set default versions in a dedicated action

mwestphal:


.github/workflows/nightly_vtk_master.yml
```diff
@@ -35,6 +35,47 @@ jobs:
       shell: bash
       run: echo "vtk_sha=$(git log -n 1 --pretty=format:%H)" >> $GITHUB_OUTPUT
 
+#----------------------------------------------------------------------------
+# Default versions: Set default version for all dependencies
+#----------------------------------------------------------------------------
+  default_versions:
+    runs-on: ubuntu-22.04
```
Meakk, line 42: Looks odd to me to run an action, just to write something to `$GITHUB_OUTPUT`  
Can't we just add an external file containing versions (raw text, json or even cmake file) and read it directly?

mwestphal: I mean, I can do that but the logic of parsing the file will then be put into an action or be duplicated between workflows.

Meakk: fait enough

Meakk (approved):

PR #1950, opened 2025-01-28
Rework Android/Wasm CI

Meakk:
- Rework the CI related to Android in order to use Docker images with prebuilt dependencies instead
- Enable more plugins to the Android CI
- Move wasm to the CI workflow and uses the new VTK-versioned Docker images


mwestphal:
How do we change vtk version then ?

Meakk:
> How do we change vtk version then ?

In the Docker image directly, just like we do with wasm

mwestphal:
I don't like the wasm approach TBH, let's discuss this.

.github/actions/android-ci/action.yml
```diff
@@ -4,49 +4,11 @@ inputs:
...
-    required: true
 
 runs:
   using: "composite"
   steps:
 
-    - name: Check required inputs
-      shell: bash
-      run: |
-        [[ "${{inputs.arch}}" ]] || { echo "arch input is empty" ; exit 1; }
```
mwestphal, line 26: this seems still needed

.github/workflows/ci.yml
```diff
@@ -646,7 +646,7 @@ jobs:
         arch: [armeabi-v7a, arm64-v8a, x86, x86_64]
 
     runs-on: ubuntu-22.04
-    container: ghcr.io/f3d-app/f3d-ci
+    container: ghcr.io/f3d-app/f3d-android-${{ matrix.arch }}:${{needs.default_versions.outputs.vtk_commit_sha}}
```
mwestphal, line 649: what happens if the container does not exists ?

mwestphal (changes requested):
some questions

.github/workflows/ci.yml
Meakk: It will fail until the Docker images are available, so it needs a PR in `f3d-docker-images`.  
I know it's not ideal, I'm planning to improve it, but at least it ensures VTK sync.

.github/actions/android-ci/action.yml
Meakk: nice catch

.github/workflows/ci.yml
mwestphal: afaiu f3d-docker-images builds the image name based on f3d master, so that cant work ?

Meakk: nop: https://github.com/f3d-app/f3d-docker-images/blob/237b9cfc157f99550d056fd95706666afe4db400/.github/workflows/build_docker_image.yml#L12

Meakk: Yes that's what I had in mind.

mwestphal (approved):

Issue #1926, opened 2025-01-18
ImGui console is rendered below scalar bar and axis


mwestphal:
**Describe the bug**
The ImGui console is rendered below scalar bar and axis, which can causes issue when inputing commands.

**To Reproduce**
 1. Open the file using `f3d dragon.vtu -s`
 2. Press Esc

![Image](https://github.com/user-attachments/assets/1ad5b212-1b38-4b64-95fe-f8e8368fd3b7)

**Expected behavior**
Console should be above

**F3D Information**
Paste the content of `f3d --version`: 3.0

mwestphal:
Sure, go for it !

Yogesh9000:
Hii @mwestphal I was able to fix the issue with scalar bar rendering above imgui console thanks to help from @Meakk.
This pr has those changes: https://github.com/f3d-app/f3d/pull/1951.
I was not able to solve the issue with axis rendering above imgui console, and I do not have any idea how to solve it, so maybe you can fix that.

mwestphal:
Well, the scalar bar is the one that is causing issue so its great already!

PR #1951, opened 2025-01-28
Imgui console is rendered below scalar bar and axis #1926

Yogesh9000:
whats done:
  - Fix the issue with scalar bar rendering above imgui console
whats to be done:
  - Fix axis rendering above imgui console


mwestphal:
Thanks for the contribution @Yogesh9000 !

PR #1971, opened 2025-02-09
fix zsh completion

jia-kai:
Completion on zsh was broken, with the error message `_arguments:comparguments:327: invalid argument:                                 -y, means direct scalars (default: -1)`.

The error was caused by improper parsing of the following help message:
```
  -y, --coloring-component [=<comp_index>(=-2)]
                                Component from the array to color with. -1
                                means magnitude, -2 or the short option,
                                -y, means direct scalars (default: -1)
```

This PR fixes this issue by requiring `--` to follow `, ` for short options.

### DIFF
--- a/resources/completion.zsh
+++ b/resources/completion.zsh
@@ -7,7 +7,7 @@ local f3dhelp
 
 f3dhelp=$(f3d --help 2>&1 | sed '1,/Examples/!d' | sed 's/\[.*\] *//')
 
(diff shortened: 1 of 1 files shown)


mwestphal:
@Meakk should we add a step in release to check for completion ?

Meakk:
That would be great. Not sure how to do it though.

Meakk (approved):

mwestphal:
> That would be great. Not sure how to do it though.

I mean a manual test :)

mwestphal:
Thanks @jia-kai !

Issue #1562, opened 2024-08-13
Add a binding to control opacity


githubkusi:
**Context**
F3D support `--opacity` to control the mesh opacity. While this is a great feature, it is not possible to modify the opacity interactively. Add a bind to do so would make F3D more reactive.

**Improve suggestion**
 1. Add a dedicated command to increase/decrease opacity (see light intensity command for inspiration)
 2. Add a dedicated bind to call that new command. `Ctrl + T`  to decrase and `Shift + T` to increase comes to mind
 3. Add dedicated tests

Meakk:
Sounds like a good idea to me.
I think it will probably be addressed in F3D 3.0 with the interactor rework or at least the interactive console. Right @mwestphal ?

mwestphal:
definitely!

mwestphal:
need: https://github.com/f3d-app/f3d/issues/443

mwestphal:
This can now be adressed!

mwestphal:
I've improved the issue so it can be contributed more easily.

19jrushlow:
First-time contributor here, I would like to work on this if possible.

mwestphal:
Great, I've just assigned it to you.

Please join our discord if you can, more easy to discuss and help you when needed :)
https://discord.f3d.app

19jrushlow:
@mwestphal should this issue be marked as resolved now?

mwestphal:
Yes!

PR #1972, opened 2025-02-09
opacity binding

19jrushlow:
The binding works, but the value displayed in the cheat sheet does not update until some other binding has been called by the user. I suspect the issue lies in docDblOpt, but I cannot figure it out and would like someone to review the problem.


library/src/interactor_impl.cxx
```diff
@@ -938,6 +962,16 @@ interactor& interactor_impl::initBindings()
...
     return std::pair(doc, valStream.str());
   };
+  
+  // "doc", "value (default to 1)"
+  auto docDblOpt = [](const std::string& doc, const std::optional<double>& val)
+  {
+	std::stringstream valStream;
+	valStream.precision(2);
+    valStream << std::fixed;
+	val.has_value() ? (valStream << val.value()) : (valStream << 1.0);
```
snoyer, line 972: ternary operator as a replacement for `if else` without assigning the value looks confusing, would be clearer as:
```suggestion
	valStream << (val.has_value() ? val.value() : 1.0);
```
and then `std::optional` already has a shortcut for this pattern
```suggestion
	valStream << val.value_or(1.0);
```

mwestphal:
> it doesn't look like it is (yet) when changing the opacity, see 

@snoyer is right, please add `     this->CheatSheetConfigured = false; ` in the method he pointed.

mwestphal:
>  should clang-format go just before all the this->addBindings instead?

Agreed

mwestphal:
> should the bindings added by this PR go on the P key instead of T to match the existing "Toggle Translucency" binding? 
 
 Good point
 
 > And should increasing/decreasing opacity turn translucency support on if it was off?
 
 I think that make sense but I would like @Meakk input on this

library/src/interactor_impl.cxx
19jrushlow: fixed

19jrushlow:
The bug appears to be fixed after doing the cheat sheet invalidation thing, and I have switched the binding to P. It seems like all that should be left is:

1. Add unit tests
2. Possibly turn on translucency support when calling the binding, pending a decision from Meakk

Am I missing anything? Here is the original [issue](https://github.com/f3d-app/f3d/issues/1562)

library/src/interactor_impl.cxx
19jrushlow: Actually, I am just now realizing that it may not be best to default to 1 automatically, in case docDblOpt needs to be used for a different binding in the future. Should I pass the default value as a parameter to the lambda instead?

library/src/interactor_impl.cxx
```diff
@@ -938,6 +962,16 @@ interactor& interactor_impl::initBindings()
...
     return std::pair(doc, valStream.str());
   };
+  
+  // "doc", "value (default to 1)"
+  auto docDblOpt = [](const std::string& doc, const std::optional<double>& val)
+  {
+    std::stringstream valStream;
+    valStream.precision(2);
+    valStream << std::fixed;
+    valStream << val.value_or(1.0);
```
mwestphal, line 972: Hum, I dont think thats correct, it should default to "N/A" to be coherent with `docTglOpt` imo.

19jrushlow: Is that not misleading? The user may think that the opacity binding is incompatible with their model if they see that, and you said it can be forced to 1 if it is not set. In fact, I think using "N/A" for docTglOpt might also be a bit misleading, since when I initially saw the values displayed as "N/A" I assumed that I would be unable to change those values even though that is not the case.

mwestphal: "1" is incorrect, but maybe N/A is not the right choice. The idea to convey is that it is not defined. I dont have a better idea than N/A.

In any case, lets be coherent at least :)

mwestphal: Lets discuss that on discord

mwestphal: so this should be valur_or("Unset") ? :)

mwestphal:
LGTM. just that one test to fix.

mwestphal (changes requested):
You also need to update the doc, doc/user/INTERACTIONS.md

19jrushlow:
should just need to write the unit tests, assuming all the checks pass

library/src/interactor_impl.cxx
```diff
@@ -938,6 +962,23 @@ interactor& interactor_impl::initBindings()
...
+  
+  // "doc", "value/Unset"
+  auto docDblOpt = [](const std::string& doc, const std::optional<double>& val)
+  {
+    std::stringstream valStream;
+    valStream.precision(2);
+    valStream << std::fixed;
+    if (val.has_value())
+    {
+      valStream << val.value();
```
mwestphal, line 974: Ideally you may also want to add/modify a test that show the cheatsheet with the opacity modified in order to cover  this line.

Adding a "TestInteractionCheatsheetOpacity" comes to mind. the log only need to press "H" to show the cheatsheet, the rest can be command line args.

mwestphal (approved):

PR #1978, opened 2025-02-11
add command remove_file_groups

yankadeshkovski:
Adds the command "remove_file_groups"

Checks for animation and stops if so. 
Clears Scene.
Then removes all files groups.
Reloads window title.


yankadeshkovski:
After running the command the dropzone_info does not show up in the body of the application. If clicking the left or right arrow key the dropzone_info does load in. Also an error is thrown if the dropzone_info is not reloaded and a file is dragged into F3D. This may mean we need to loadnextfilegroup before the remove_file_groups command is finished.

yankadeshkovski:
Actually, may require some help understanding this problematic behavior too. After importing a file that was previously removed via the "remove_file_groups" command the command line opens as the image is imported... which isn't an issue since you can just close it, but definitely not a wanted or pleasant bug.

application/F3DStarter.cxx
```diff
@@ -1618,6 +1618,21 @@ void F3DStarter::AddCommands()
...
+    {
+      if (!this->Internals->AppOptions.NoRender)
+      {
+        this->Internals->Engine->getInteractor().stopAnimation();
+      }
+      f3d::scene& scene = this->Internals->Engine->getScene();
+      scene.clear();
+      this->Internals->FilesGroups.clear();
+      f3d::window& window = this->Internals->Engine->getWindow();
+      window.setWindowName(F3D::AppTitle).setIcon(F3DIcon, sizeof(F3DIcon));
```
mwestphal, line 1632: lets put that in a `ResetWindowName` method.

application/F3DStarter.cxx
```diff
@@ -1618,6 +1618,21 @@ void F3DStarter::AddCommands()
...
+      this->LoadRelativeFileGroup(0, true, true);
+    });
```
mwestphal, line 1634: you should be able to set the options like this:

```
  f3d::options& options = this->Internals->Engine->getOptions();
  options.ui.dropzone = true;
  options.ui.filename_info = "";
```

application/F3DStarter.cxx
```diff
@@ -1512,7 +1512,7 @@ int F3DStarter::AddFile(const fs::path& path, bool quiet)
       bool found = false;
       std::vector<std::vector<fs::path>>::iterator it;
       for (it = this->Internals->FilesGroups.begin(); it != this->Internals->FilesGroups.end();
-           it++)
+        it++)
```
mwestphal, line 1515: for some reason your local clang-format is slightly different than our CI, lets not worry about it for now.

yankadeshkovski:
Awesome, will make sure to make these changes soon! Unfortunately heading to sleep now, so I will deal with this tomorrow morning 👍

application/F3DStarter.cxx
yankadeshkovski: This should be added now!

application/F3DStarter.cxx
yankadeshkovski: This should be fixed now.

application/F3DStarter.h
```diff
@@ -82,6 +82,11 @@ class F3DStarter
    */
   void EventLoop();
 
+  /**
+   * Resets the window name for f3d app
+   */
+  void ResetWindowName();
```
mwestphal, line 88: nice, lets use this method in `F3DStarter::Start` as well then :)

yankadeshkovski: Ah, we can defo do that. I'll implement this as well and commit.

yankadeshkovski: This should be done in the new commit.

yankadeshkovski:
Test for remove_file_groups command is added. This test runs to completion and seems to work properly. However, there is a bug that is overlooked. After calling the command (remove_file_groups), re-adding any files that were inside of the FilesGroups previously will load but the window will not show them (blank screen and only a window name change to the file name). If the files that are added afterwards were not part of the removed FilesGroups then everything works properly and they are displayed.

application/F3DStarter.cxx
```diff
@@ -1618,6 +1624,22 @@ void F3DStarter::AddCommands()
...
+  interactor.addCommand("remove_file_groups",
+    [this](const std::vector<std::string>&)
+    {
+      if (!this->Internals->AppOptions.NoRender)
+      {
+        this->Internals->Engine->getInteractor().stopAnimation();
+      }
+      f3d::scene& scene = this->Internals->Engine->getScene();
+      scene.clear();
+      this->Internals->FilesGroups.clear();
```
mwestphal, line 1636: you need to add this

`this->Internals->LoadedFiles.clear();`

application/testing/CMakeLists.txt
```diff
@@ -923,6 +923,7 @@ f3d_test(NAME TestCommandScriptPrintColoring DATA dragon.vtu SCRIPT TestCommandS
 f3d_test(NAME TestCommandScriptPrintMesh DATA dragon.vtu SCRIPT TestCommandScriptPrintMesh.txt REGEXP "Number of points: 13268" NO_BASELINE) # print_mesh_info
 f3d_test(NAME TestCommandScriptPrintOptions DATA dragon.vtu SCRIPT TestCommandScriptPrintOptions.txt REGEXP "interactor.invert_zoom: false" NO_BASELINE) # print_options_info
 f3d_test(NAME TestParseOptionalBoolExtraArg DATA dragon.vtu SCRIPT TestParseOptionalBoolExtraArg.txt REGEXP "Command: load_previous_file_group takes at most 1 argument, got 2 arguments instead." NO_BASELINE)
+f3d_test(NAME TestRemoveFileGroups DATA dragon.vtu SCRIPT TestRemoveFileGroups.txt)
```
mwestphal, line 926:
```suggestion
f3d_test(NAME TestRemoveFileGroups DATA dragon.vtu SCRIPT TestRemoveFileGroups.txt NO_DATA_FORCE_RENDER)
```

application/F3DStarter.cxx
yankadeshkovski: Ahhhhh... I never realized we had a LoadedFiles attribute, but that behavior now makes sense. Will fix and push soon.

application/testing/CMakeLists.txt
yankadeshkovski: Will fix this as well.

yankadeshkovski:
Wait, still failing the format. Will fix again :/

mwestphal (approved):

mwestphal:
LGTM, do you want to review @Meakk ?

mwestphal (changes requested):
I forgot, you need to update the doc /doc/user/COMMANDS.md

Meakk (approved):

mwestphal (approved):

mwestphal:
Thanks for your contribution @yankadeshkovski !

PR #1981, opened 2025-02-12
Unset optional display value

19jrushlow:
std::optional variables are currently displayed as "N/A" on the cheatsheet. This is misleading, so it is being changed to "Unset."
Note: not all of the baselines have been updated, I need to retrieve some from the CI


mwestphal (approved):

mwestphal:
Thanks for your contribution @19jrushlow !

Issue #1500, opened 2024-06-29
Final shader should not be applied on overlay


Meakk:
**Describe the bug**
Overlay widgets like filename or cheatsheet should be drawn after the final shader

mwestphal:
Awesome, I just assigned it to you.

Meakk:
@samoncrief this one is challenging. Currently, there is a main pass taking care of everything (background, scene and overlay), then the userpass is postprocessing the result.  
Basically, the overlay step in the main pass should be extracted to its own pass and be applied after the user pass.
Don't hesitate to ping me on Discord if you need to discuss

mwestphal:
Done in #1984

PR #1984, opened 2025-02-13
Final shader applied on overlay

samoncrief:
@Meakk Here's the current version with linear blending applied. The code is all handled in `vtkF3DOverlayRenderPass`.

Fix for https://github.com/f3d-app/f3d/issues/1500


mwestphal:
Sorry I canceled your run, working on a CI issue right now

mwestphal:
I restarted CI :)

mwestphal:
LGTM but its more a @Meakk style of code.

vtkext/private/module/vtkF3DOverlayRenderPass.cxx
```diff
@@ -0,0 +1,229 @@
...
+#include <vtkProp.h>
+#include <vtkRenderState.h>
+#include <vtkRenderer.h>
+#include <vtkShaderProgram.h>
+#include <vtkSkybox.h>
+#include <vtkTextureObject.h>
+
+vtkStandardNewMacro(vtkF3DOverlayRenderPass);
+
+void vtkF3DOverlayRenderPass::Render(const vtkRenderState* s)
```
Meakk, line 22: Missing `-------` separator above each function implementation (see other files)

vtkext/private/module/vtkF3DOverlayRenderPass.cxx
```diff
@@ -0,0 +1,229 @@
...
+void vtkF3DOverlayRenderPass::Initialize(const vtkRenderState* s)
+{
+  this->OverlayProps.clear();
+
+  // assign props to the overlay pass
+  vtkProp** props = s->GetPropArray();
+  for (int i = 0; i < s->GetPropArrayCount(); i++)
+  {
+    vtkProp* prop = props[i];
+    if (!vtkSkybox::SafeDownCast(prop) && !vtkProp3D::SafeDownCast(prop))
```
Meakk, line 77: I know this code was there already, but `vtkSkybox` derives from `vtkProp3D`, so the first check can be removed.

vtkext/private/module/vtkF3DOverlayRenderPass.cxx
```diff
@@ -0,0 +1,229 @@
...
+  ostate->vtkglViewport(pos[0], pos[1], size[0], size[1]);
+  ostate->vtkglScissor(pos[0], pos[1], size[0], size[1]);
+
+  this->QuadHelper->Render();
+
+  this->OverlayPass->GetColorTexture()->Deactivate();
+  this->ColorTexture->Deactivate();
+
+  vtkOpenGLCheckErrorMacro("failed after Render");
+}
```
Meakk, line 229: missing empty line at the end of the file

vtkext/private/module/vtkF3DOverlayRenderPass.h
```diff
@@ -0,0 +1,64 @@
...
+  /**
+   * Release graphics resources and ask components to release their own resources.
+   */
+  void ReleaseGraphicsResources(vtkWindow* w) override;
+
+  /**
+   * Forbidden copies.
+   */
+  vtkF3DOverlayRenderPass(const vtkF3DOverlayRenderPass&) = delete;
+  void operator=(const vtkF3DOverlayRenderPass&) = delete;
```
mwestphal, line 41: both should be private

mwestphal: nevermind

mwestphal (commented):
LGTM appart the small codepath that is not reached by coverage

vtkext/private/module/vtkF3DOverlayRenderPass.cxx
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  renWin->GetState()->PushFramebufferBindings();
+  this->RenderDelegate(
+    s, size[0], size[1], size[0], size[1], this->FrameBufferObject, this->ColorTexture);
+  renWin->GetState()->PopFramebufferBindings();
+
+  if (this->QuadHelper && this->QuadHelper->ShaderChangeValue < this->GetMTime())
+  {
+    this->QuadHelper = nullptr;
```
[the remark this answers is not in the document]
samoncrief: Codecov says not, this was just a holdover from the format of the user pass code. I can take it out if we don't need to test for it.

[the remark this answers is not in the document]
mwestphal: Unless I'm mistaken this can only happen if we render multiple times and the "ShaderChangeValue" changes between render. Can this happen ?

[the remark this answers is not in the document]
mwestphal: I guess not since you removed it :)

mwestphal (approved):
LGTM but this will need @Meakk review as well

.github/baselines/install_output.png
Meakk, line 1: I think there are many other tests having filename+FXAA, so I'm surprised we do not have more tests failing there. E.g. `TestDefaultConfigFileVTU`  
Maybe there's a threshold issue @mwestphal?

Meakk:
Amazing work @samoncrief!  
Can you add a new test with a final shader (e.g. negative shader) and a filename to properly test that the filename widget isn't affected?

.github/baselines/install_output.png
mwestphal: no idea, one need to check the diff value locally

testing/baselines/TestHDRIRaytracingAmbientOnlyNoBackground.png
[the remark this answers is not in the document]
samoncrief: The overlay pass takes the render (after shaders/FXAA) as a source that isn't alpha premultiplied (so we run `toLinear(result.rgb)` without dividing) to composite the image with the overlay, then convert it back to sRGB and fill in the remaining space (where `1.0 - result.a > 0.0`) with the original rgb. This applies even when the overlay has no UI to show. I wasn't aware of this being an issue, so this may require a separate fix in the future if the overlay pass isn't available to convert the raytracing images.

samoncrief:
> Amazing work @samoncrief! Can you add a new test with a final shader (e.g. negative shader) and a filename to properly test that the filename widget isn't affected?

Sure. Do you want it for a shader where alpha equals 1.0 like in the examples, or with the alpha values preserved? The example code treats the background as a solid object because of the alpha, so it may not give the intended results.

.github/baselines/install_output.png
mwestphal: > Image comparison success with an error difference of: 0.0259376

The image are notably different but its not enough to reach the 0.04 threshold.
We could try defaulting to 0.02 but we may have false positive.

I dont think we should update the baselines though.

samoncrief: The tests in question have the file name panel rendered at a darker color than in other baselines. I can roll back the baselines, but unless there is something specific to those tests, (or something platform-specific that the baselines are taken from) the lighter panel and sharper text over the scalar bar would be intended behavior.

Meakk: It's intended behavior. Ok let's update them later if we reach the 0.04 threshold.

Meakk:
> Do you want it for a shader where alpha equals 1.0 like in the examples, or with the alpha values preserved?

It's up to you :)

Meakk (approved):

PR #2001, opened 2025-02-18
Tests: Improve cheatsheet testing and limit baselines update

mwestphal:
Stop duplicating baselines and avoid future baseline updates by only showing the interesting part of the cheatsheet.

Rely on mouse wheel scroll down.

This is not perfect, baseline update will still be needed sometimes, but I expect it to be less frequent.


mwestphal:
@Meakk @19jrushlow please review

Meakk (approved):

Issue #2003, opened 2025-02-18
Add proper parsing for color_t


mwestphal:
**Context**
color_t option type has recently been added in libf3d, however parsing is actually the same as for double_vector.
Lets implement proper parsing of colors, such as RGB and other syntax.

**Suggested solution**
 - Implement proper parsing in options_tools::parse for this type
 - Check that formating is correct in options_tools::format and improve it if needed
  -Add testing in TestSDKOptions.cxx and TestSDKOptionsIO.cxx

**Syntax to Support**
 - HDR double vector: `val, val, val`
 - RGB hexa: `#RRGGBB` (case independent)
 - named colors: https://vtk.org/doc/nightly/html/classvtkNamedColors.html#a66191aacb2a0e4eafd13bbdd3decb5b3
 - `hsl()`: https://www.w3schools.com/colors/colors_hsl.asp
 - `rgb()`: https://www.w3schools.com/colors/colors_rgb.asp
 - `hwb()`: https://www.w3schools.com/colors/colors_hwb.asp (optional)
 - `cmyk()`: https://www.w3schools.com/colors/colors_cmyk.asp (optional)

VittorioSanchez:
Hello, can I work on this issue?

mwestphal:
Of course!

mwestphal:
Being adressed in https://github.com/f3d-app/f3d/pull/2015

PR #2015, opened 2025-02-24
Add parsing color_t

VittorioSanchez:
Added parsing for: 
- #RRGGBB
- rgb(R, G, B)
- hsl(H, S%, L%)
- hsv(H, S%, V%)
- hwb(H, W%, B%)
- cmyk(C%, M%, Y%, K%)
- CSS color names

These are case insensitive and whitespace insensitive. % symbol is also optional.
Limits are checked and throws error if out of range.
Updated doc and added tests

Fix: https://github.com/f3d-app/f3d/issues/2003


library/private/options_tools.h
```diff
@@ -191,6 +194,46 @@ ratio_t parse(const std::string& str)
...
+  const std::regex rgbRegex("^rgb\\((\\s*\\d{1,3})\\s*,\\s*(\\d{1,3})\\s*,\\s*(\\d{1,3})\\s*\\)$", std::regex_constants::syntax_option_type::icase);
+  std::smatch rgbMatch;
+  if (std::regex_search(s, rgbMatch, rgbRegex))
+  {
+    std::vector<double> vec;
+    for (size_t i = 1; i < rgbMatch.size(); i++)
+    {
+      vec.push_back(std::stod(rgbMatch[i]) / 255.0);
+    }
+    return color_t(vec);
```
snoyer, line 225:
```suggestion
    return color_t(                   //
      std::stod(rgbMatch[1]) / 255.0, //
      std::stod(rgbMatch[2]) / 255.0, //
      std::stod(rgbMatch[3]) / 255.0  //
    );
```
you can skip the loop and the `vector` here

library/private/options_tools.h
```diff
@@ -191,6 +194,46 @@ ratio_t parse(const std::string& str)
...
+  {
+    std::vector<double> vec;
+    const std::string hexStr = hexMatch[1];
+    for (size_t i = 0; i < hexStr.length(); i += 2)
+    {
+      unsigned int byte = std::stoul(hexStr.substr(i, 2), nullptr, 16);
+      vec.push_back(byte / 255.0);
+    }
+    return color_t(vec);
+  }
```
snoyer, line 213:
```suggestion
  const std::regex hexRegex(
    "^#([0-9a-f]{2})([0-9a-f]{2})([0-9a-f]{2})$", std::regex_constants::icase);
  std::smatch hexMatch;
  if (std::regex_search(s, hexMatch, hexRegex))
  {
    return color_t(                                 //
      std::stoul(hexMatch[1], nullptr, 16) / 255.0, //
      std::stoul(hexMatch[2], nullptr, 16) / 255.0, //
      std::stoul(hexMatch[3], nullptr, 16) / 255.0  //
    );
  }
```
using a group for each component is maybe more straightforward? also using `icase` can avoid having to specify `a-fA-F`

library/private/options_tools.h
```diff
@@ -191,6 +194,46 @@ ratio_t parse(const std::string& str)
 template<>
 color_t parse(const std::string& str)
 {
+  std::string s = options_tools::trim(str);
+  std::transform(s.begin(), s.end(), s.begin(), [](unsigned char c) { return std::tolower(c); });
```
snoyer, line 198:
```suggestion
```
do we need to lowercase the whole string if we're going to regex match with `icase` anyways?

library/private/options_tools.h
```diff
@@ -191,6 +194,46 @@ ratio_t parse(const std::string& str)
 template<>
 color_t parse(const std::string& str)
 {
+  std::string s = options_tools::trim(str);
```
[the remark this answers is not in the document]
VittorioSanchez: let's do that, it does simplifies the regexes too

library/private/options_tools.h
```diff
@@ -184,13 +188,114 @@ ratio_t parse(const std::string& str)
...
+    if (rgb[0] > 1.0 || rgb[1] > 1.0 || rgb[2] > 1.0)
+    {
+      throw options::parsing_exception("Cannot parse " + str + " into a color_t");
+    }
+    return color_t(rgb[0], rgb[1], rgb[2]);
+  }
+
+  /* Hue-based format search: hsl, hsv, hwb */
+  const std::regex hueRegex(
+    "^([a-z]{3})\\((\\d{1,3}),(\\d{1,3})%{0,1},(\\d{1,3})%{0,1}\\)$", std::regex_constants::icase);
```
snoyer, line 231:
```suggestion
    "^(hsl|hsv|hwb)\\((\\d{1,3}),(\\d{1,3})%?,(\\d{1,3})%?\\)$", std::regex_constants::icase);
```
you can restrict the known keywords in the regex, and `?` is more idiomatic for "zero or one"

VittorioSanchez: will fix that

library/private/options_tools.h
```diff
@@ -186,13 +189,112 @@ ratio_t parse(const std::string& str)
...
+    return color_t(rgb[0], rgb[1], rgb[2]);
+  }
+
+  /* Hue-based format search: hsl, hsv, hwb */
+  const std::regex hueRegex(
+    "^(hsl|hsv|hwb)\\((\\d{1,3}),(\\d{1,3})%?,(\\d{1,3})%?\\)$", std::regex_constants::icase);
+  std::smatch hueMatch;
+  if (std::regex_search(strCompact, hueMatch, hueRegex))
+  {
+    double h = std::stod(hueMatch[2]) / 360.0;
```
snoyer, line 236:
```suggestion
    const double h = std::stod(hueMatch[2]) / 360.0;
```

library/private/options_tools.h
```diff
@@ -186,13 +189,112 @@ ratio_t parse(const std::string& str)
...
+    {
+      throw options::parsing_exception("Cannot parse " + str + " into a color_t");
+    }
+
+    std::string hueFormat = hueMatch[1].str();
+    std::transform(hueFormat.begin(), hueFormat.end(), hueFormat.begin(),
+      [](unsigned char c) { return std::tolower(c); });
+    if (hueFormat == "hsl")
+    {
+      double l = std::stod(hueMatch[4]) / 100.0;
```
snoyer, line 249:
```suggestion
      double l = v;
```

library/private/options_tools.h
```diff
@@ -186,13 +189,112 @@ ratio_t parse(const std::string& str)
...
+  /* CMYK format search */
+  const std::regex cmykRegex(
+    "^cmyk\\((\\d{1,3})%?,(\\d{1,3})%?,(\\d{1,3})%?,(\\d{1,3})%?\\)$", std::regex_constants::icase);
+  std::smatch cmykMatch;
+  if (std::regex_search(strCompact, cmykMatch, cmykRegex))
+  {
+    double c = std::stod(cmykMatch[1]) / 100.0;
+    double m = std::stod(cmykMatch[2]) / 100.0;
+    double y = std::stod(cmykMatch[3]) / 100.0;
+    double k = std::stod(cmykMatch[4]) / 100.0;
```
snoyer, line 278: can all be const

snoyer (commented):
some nitpicks as per tradition but look good :)

snoyer:
@mwestphal does the helptext need to be changed from `--background-color <R,G,B>` to `--background-color <color>` or something?

mwestphal:
Good point, probably yes!

library/private/options_tools.h
VittorioSanchez: fixed

VittorioSanchez:
> @mwestphal does the helptext need to be changed from `--background-color <R,G,B>` to `--background-color <color>` or something?

I changed it

snoyer:
Looks good.
The one thing I can see however is that `rgb/cmyk/hsv/...()` only work with integer params and not double (matching with `\d{1,3}`). I don't know if @mwestphal and @Meakk want them to work with reals too (I do know that matching real numbers with regex not the most fun tho).
Maybe it could be merged as is and improved in a follow-up PR?

mwestphal:
> Looks good. The one thing I can see however is that `rgb/cmyk/hsv/...()` only work with integer params and not double (matching with `\d{1,3}`). I don't know if @mwestphal and @Meakk want them to work with reals too (I do know that matching real numbers with regex not the most fun tho). Maybe it could be merged as is and improved in a follow-up PR?

w3schools uses integer, so I'd say no, especially as the classic vector of double syntax supports doubles anyway

mwestphal:
@VittorioSanchez Is this ready for review ? :)

VittorioSanchez:
> @VittorioSanchez Is this ready for review ? :)

Yes it is

doc/user/PARSING.md
```diff
@@ -67,7 +67,17 @@ When formatting a vector into a string, individual token are formatted according
...
-Color are parsed and formatted as a vector of double.
+The following formats are supported when parsing a color, case insensitive:
+
+- R,G,B where R, G, B are doubles between 0 and 1
+- #RRGGBB where RR, GG, BB are hexadecimal values
+- hsl(H, S%, L%) where H \<= 360, S and L in percent
+- hsv(H, S%, V%) where H \<= 360, S and V in percent
+- hwb(H, W%, B%) where H \<= 360, W and B in percent
+- cmyk(C%, M%, Y%, K%) where C, M, Y, K in percent
+- color name, see [HTML color names](https://en.wikipedia.org/wiki/Web_colors#HTML_color_names)
```
mwestphal, line 78: should this link be used ?

https://htmlpreview.github.io/?https://github.com/Kitware/vtk-examples/blob/gh-pages/VTKNamedColorPatches.html

library/private/options_tools.h
```diff
@@ -8,6 +8,9 @@
 #include <cassert>
 #include <regex>
 #include <sstream>
+#include <vtkMath.h>
+#include <vtkNamedColors.h>
+#include <vtkSmartPointer.h>
```
mwestphal, line 13: put vtk includes before system includes

library/private/options_tools.h
```diff
@@ -186,13 +189,112 @@ ratio_t parse(const std::string& str)
...
- * Supported formats: "R,G,B"
- * rely on parse<std::vector<double>>(str)
+ * Supported formats: "R,G,B", "#RRGGBB", "rgb(R,G,B)", "hsl(H,S%,L%)", "hsv(H,S%,V%)",
+ *                    "hwb(H,W%,B%)", cmyk(C%,M%,Y%,K%), "CSS3 color name"
  * Can throw options::parsing_exception in case of failure to parse
  */
 template<>
 color_t parse(const std::string& str)
 {
+  const std::string strCompact = std::regex_replace(str, std::regex("\\s"), "");
```
mwestphal, line 199: what does this do ?

doc/user/PARSING.md
```diff
@@ -67,7 +67,17 @@ When formatting a vector into a string, individual token are formatted according
...
 
-Color are parsed and formatted as a vector of double.
+The following formats are supported when parsing a color, case insensitive:
+
+- R,G,B where R, G, B are doubles between 0 and 1
+- #RRGGBB where RR, GG, BB are hexadecimal values
+- hsl(H, S%, L%) where H \<= 360, S and L in percent
+- hsv(H, S%, V%) where H \<= 360, S and V in percent
+- hwb(H, W%, B%) where H \<= 360, W and B in percent
+- cmyk(C%, M%, Y%, K%) where C, M, Y, K in percent
```
mwestphal, line 77: add links to the w3schools stuff ?

doc/user/PARSING.md
```diff
@@ -67,7 +67,17 @@ When formatting a vector into a string, individual token are formatted according
...
+- color name, see [HTML color names](https://en.wikipedia.org/wiki/Web_colors#HTML_color_names)
+
+When formatting a color into a string, it is formatted as `R,G,B` where R, G, B are doubles between 0 and 1.
```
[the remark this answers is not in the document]
snoyer: yeah format as `#rrggbb` if all components are `<=1` and multiples of `1./255` makes sense to me, otherwise format as double vector

library/private/options_tools.h
```diff
@@ -186,13 +189,112 @@ ratio_t parse(const std::string& str)
...
+  /* CMYK format search */
+  const std::regex cmykRegex(
+    "^cmyk\\((\\d{1,3})%?,(\\d{1,3})%?,(\\d{1,3})%?,(\\d{1,3})%?\\)$", std::regex_constants::icase);
+  std::smatch cmykMatch;
+  if (std::regex_search(strCompact, cmykMatch, cmykRegex))
+  {
+    const double c = std::stod(cmykMatch[1]) / 100.0;
+    const double m = std::stod(cmykMatch[2]) / 100.0;
+    const double y = std::stod(cmykMatch[3]) / 100.0;
+    const double k = std::stod(cmykMatch[4]) / 100.0;
```
mwestphal, line 278: you need to catch exceptions from stod

mwestphal (changes requested):
nice work, a few questions though

doc/user/PARSING.md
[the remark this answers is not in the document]
mwestphal: @VittorioSanchez do you think you could do that in this PR or in another PR ?

library/private/options_tools.h
VittorioSanchez: removes all spaces in input string. It's to simplify the regexes and avoid to put \s* in between every character. And to allow the user to put spaces of course

doc/user/PARSING.md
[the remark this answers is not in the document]
VittorioSanchez: So output in #rrggbb for hex and rgb() and double vector for others right ?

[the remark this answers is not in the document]
VittorioSanchez: If so I think I can do it in this PR

doc/user/PARSING.md
VittorioSanchez: only hsv is not in w3school others ok

mwestphal: We are discussing it on discord right now

doc/user/PARSING.md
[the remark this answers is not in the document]
mwestphal: nice

library/private/options_tools.h
mwestphal: right!

doc/user/PARSING.md
mwestphal: https://www.w3.org/TR/css-color-3/#rgb-color this ones looks better

mwestphal:
Please resolve discussions you have adressed and ping me and @snoyer for a review when ready @VittorioSanchez :)

doc/user/PARSING.md
```diff
@@ -67,7 +67,20 @@ When formatting a vector into a string, individual token are formatted according
 
 ## Color
 
-Color are parsed and formatted as a vector of double.
+The following formats are supported when parsing a color, case insensitive:
+
+- R,G,B where R, G, B are doubles between 0 and 1
+- #RRGGBB where RR, GG, BB are hexadecimal values
+- rgb(R, G, B) where R, G and B \<= 255
```
mwestphal, line 74:
```suggestion
- rgb(R, G, B) where R, G and B are integer [0, 255]
```

Use the same phrasing everywhere below

doc/user/PARSING.md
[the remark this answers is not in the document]
mwestphal: so this needs updating :)

library/private/options_tools.h
```diff
@@ -186,21 +189,128 @@ ratio_t parse(const std::string& str)
...
   }
+  catch (std::invalid_argument const&)
+  {
+    throw options::parsing_exception("Cannot parse " + str + " into a color_t");
+  }
+  catch (std::out_of_range const&)
+  {
+    throw options::parsing_exception(
+      "Cannot parse " + str + " into a color_t as it would go out of range");
+  }
```
mwestphal, line 313: please add a test covering this.

library/private/options_tools.h
```diff
@@ -186,21 +189,128 @@ ratio_t parse(const std::string& str)
...
+    }
     return color_t(options_tools::parse<std::vector<double>>(str));
   }
   catch (const f3d::type_construction_exception& ex)
   {
     throw options::parsing_exception("Cannot parse " + str + " into a color_t: " + ex.what());
   }
+  catch (std::invalid_argument const&)
+  {
+    throw options::parsing_exception("Cannot parse " + str + " into a color_t");
```
mwestphal, line 307: please add a test covering this.

library/private/options_tools.h
```diff
@@ -327,8 +437,21 @@ std::string format(const std::vector<T>& var)
  */
 std::string format(color_t var)
 {
-  // TODO generate a proper color string
-  return options_tools::format(static_cast<std::vector<double>>(var));
+  if (std::fmod(var.r() * 255., 1) == 0. && //
+    std::fmod(var.g() * 255., 1) == 0. &&   //
+    std::fmod(var.b() * 255., 1) == 0.)     //
```
snoyer, line 442: you'll want to check that all values are between `0` and `1`.
Also are the `... == 0.` reliable or do we need some kind of `abs(...) < tiny_tolerance` here?

library/private/options_tools.h
snoyer: assuming we're expecting the `std::invalid_argument` from `stod` it may prove hard to cover as the input would (or at least should) always be a 1 to 3 digit integer as matched by the regex

mwestphal: Then this is unreachable code and can be safely removed :). Adding a comment why we do not catch the exception would be nice though

mwestphal: Hum, i think std::stoul should let us reach the catch with input such as `#HHHHHH`

snoyer: the `stoul` calls are also behind a regex (`[0-9a-f]{2}`)

mwestphal: of course, I missed that.

mwestphal: the out of range may also be unreachable ?

snoyer:
the `regex_search("^...$")` calls can be replaced with `regex_match("...")`

doc/user/PARSING.md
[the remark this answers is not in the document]
VittorioSanchez: right ^^'

library/private/options_tools.h
VittorioSanchez: yes at the end the out of range is only possible in the case we have a vector double and it is already catched in the **type_construction_exception**. I will remove it and the **invalid_argument** as it is covered by the regexes

library/private/options_tools.h
VittorioSanchez: see previous comment

library/private/options_tools.h
VittorioSanchez: @snoyer is it necessary to add the check of the values as it is checked in the parsing function ?

library/private/options_tools.h
```diff
@@ -186,21 +189,128 @@ ratio_t parse(const std::string& str)
...
+
+    /* Named colors search */
+    vtkSmartPointer<vtkNamedColors> color = vtkSmartPointer<vtkNamedColors>::New();
+    if (color->ColorExists(strCompact))
+    {
+      double rgba[4];
+      color->GetColor(strCompact, rgba);
+      return color_t(rgba[0], rgba[1], rgba[2]);
+    }
     return color_t(options_tools::parse<std::vector<double>>(str));
```
VittorioSanchez, line 299: just thought about it but it lacks a check on negative values here. Others are covered by regexes, I'll add it.

library/private/options_tools.h
snoyer: The hex, rgb, ... parsed inputs are indeed guaranteed to produce only components within `[0,1]` but the raw double vector form isn't.
The values could also have been set "by hand" when using the library, bypassing the parsing entirely.
And I think that while we should never have negative values, having values greater than `1` is legit for HDR colors, maybe @Meakk can expand on the hows and whens that comes into play.
Either way, formatting to `#rrggbb` should only be done when all the conditions are met and anything else would go to the double vector format, valid or not

doc/user/PARSING.md
```diff
@@ -67,7 +67,20 @@ When formatting a vector into a string, individual token are formatted according
...
+- rgb(R, G, B) where R, G and B are integer [0, 255]
+- hsl(H, S%, L%) where H is integer [0, 360], S and L are integer [0, 100]
+- hsv(H, S%, V%) where H is integer [0, 360], S and V are integer [0, 100]
+- hwb(H, W%, B%) where H is integer [0, 360], W and B are integer [0, 100]
+- cmyk(C%, M%, Y%, K%) where C, M, Y, K are integer [0, 100]
+- [color name](https://htmlpreview.github.io/?https://github.com/Kitware/vtk-examples/blob/gh-pages/VTKNamedColorPatches.html)
+
+See [W3C](https://www.w3.org/TR/css-color-3/#rgb-color) doc for more details on these formats.
+
+When formatting a color into a string, it is formatted as `#RRGGBB` if values are multiple of 255. Otherwise, it is formatted as `R,G,B` where R, G, B are doubles between 0 and 1.
```
mwestphal, line 83:
```suggestion
When formatting a color into a string, it is formatted as `#RRGGBB` if values are multiple of 255. Otherwise, it is formatted as vector of doubles.
```

doc/user/PARSING.md
```diff
@@ -67,7 +67,20 @@ When formatting a vector into a string, individual token are formatted according
 
 ## Color
 
-Color are parsed and formatted as a vector of double.
+The following formats are supported when parsing a color, case insensitive:
+
+- R,G,B where R, G, B are doubles [0, 1]
```
mwestphal, line 72:
```suggestion
- R,G,B where R, G, B are doubles >= 0
```

library/private/options_tools.h
mwestphal: not adressed yet

library/private/options_tools.h
```diff
@@ -186,17 +189,123 @@ ratio_t parse(const std::string& str)
...
+        throw options::parsing_exception("Cannot parse " + str + " into a color_t");
+      }
+      return color_t(           //
+        (1.0 - c) * (1.0 - k),  //
+        (1.0 - m) * (1.0 - k),  //
+        (1.0 - y) * (1.0 - k)); //
+    }
+
+    /* Named colors search */
+    vtkSmartPointer<vtkNamedColors> color = vtkSmartPointer<vtkNamedColors>::New();
```
mwestphal, line 292:
```suggestion
    vtkNew<vtkNamedColors> color;
```

library/private/options_tools.h
```diff
@@ -186,17 +189,123 @@ ratio_t parse(const std::string& str)
...
+    if (color->ColorExists(strCompact))
+    {
+      double rgba[4];
+      color->GetColor(strCompact, rgba);
+      return color_t(rgba[0], rgba[1], rgba[2]);
+    }
+
+    /* Vector double format */
+    std::vector<double> vecColor = options_tools::parse<std::vector<double>>(str);
+    if (std::any_of(vecColor.begin(), vecColor.end(), [](double value) { return !(value >= 0.0); }))
```
mwestphal, line 302: I wonder if we should remove this check. Anyone using the struct API can directly put negative values in anyway so its not like we can then skip checks later on, and it restrict a bit what the user can do.

If it makes sense to someone to input negative values, it may be simpler to let them (and catch the error later when we use the value, which we already do)

library/private/options_tools.h
```diff
@@ -186,17 +189,123 @@ ratio_t parse(const std::string& str)
...
+    {
+      throw options::parsing_exception("Cannot parse " + str + " into a color_t");
+    }
+    return color_t(vecColor);
   }
+  /* We do not catch std::invalid_argument exception from stod as it is covered by the regex */
```
mwestphal, line 308: nor std::out_of_range

library/private/options_tools.h
```diff
@@ -323,12 +432,26 @@ std::string format(const std::vector<T>& var)
...
+ * Formats in hex color string "#RRGGBB" if values are multiple of 1/255.
+ * Otherwise rely on format(std::vector<double>&)
  */
 std::string format(color_t var)
 {
-  // TODO generate a proper color string
-  return options_tools::format(static_cast<std::vector<double>>(var));
+  const std::vector<double> colors = { var.r(), var.g(), var.b() };
+  if (std::all_of(colors.begin(), colors.end(),
+        [](double val) { return (val >= 0 && val <= 1 && std::fmod(val * 255., 1) < 1e-9); }))
```
mwestphal, line 442: does using std::epsilon makes sense here ?

library/testing/TestSDKOptionsIO.cxx
```diff
@@ -110,8 +110,39 @@ int TestSDKOptionsIO(int argc, char* argv[])
 
   test.parse<f3d::color_t>("color_t", "0.1,0.2,0.3", { 0.1, 0.2, 0.3 });
   test.parse<f3d::color_t>("color_t", "  0.1,  0.2 , 0.3 ", { 0.1, 0.2, 0.3 });
+  test.parse<f3d::color_t>("color_t", "#FFFFFF", { 1.0, 1.0, 1.0 });
```
mwestphal, line 113: looking at w3c, it looks like they also list #FFF as supported format. Does that work ? If not is it easy to add ?

library/testing/TestSDKOptionsIO.cxx
```diff
@@ -110,8 +110,39 @@ int TestSDKOptionsIO(int argc, char* argv[])
...
+    "invalid rgb() value color_t", "rgb(300,255,255)");
+  test.parse_expect<f3d::color_t, parsing_exception>(
+    "invalid hsl() value color_t", "hsl(361,120,255)");
+  test.parse_expect<f3d::color_t, parsing_exception>("invalid hsv() color_t", "hsv(100,120%,0)");
+  test.parse_expect<f3d::color_t, parsing_exception>("invalid format color_t", "hxb(240,0%,0%)");
+  test.parse_expect<f3d::color_t, parsing_exception>("invalid color_t", "cmyk(200%,12%,34%,56%)");
   test.parse_expect<f3d::color_t, parsing_exception>("incorrect size color_t", "0.1,0.2,0.3,0.4");
+  test.parse_expect<f3d::color_t, parsing_exception>("invalid color_t", "-0.1,-0.2,-0.3");
+  test.parse_expect<f3d::color_t, parsing_exception>(
+    "out of range color_t", std::string(outOfRangeDoubleStr + ",0.2,0.3"));
```
mwestphal, line 140: Im not following this test ?

mwestphal (changes requested):
some questions and remarks

Meakk:
I made a quick review and it looks super great! Thank you @VittorioSanchez! 👍  
I'll let @mwestphal make the final approval when the small remarks are fixed.

mwestphal:
Hi @VittorioSanchez 

Do you need help moving forward ? :)

library/testing/TestSDKOptionsIO.cxx
VittorioSanchez: it was to check in case we got negative values and values out of range

library/testing/TestSDKOptionsIO.cxx
VittorioSanchez: I didn't implement it. It shouldn't be too difficult, I'll add it

doc/user/PARSING.md
```diff
@@ -67,7 +67,20 @@ When formatting a vector into a string, individual token are formatted according
 
 ## Color
 
-Color are parsed and formatted as a vector of double.
+The following formats are supported when parsing a color, case insensitive:
+
+- R,G,B where R, G, B are doubles >= 0
+- #RRGGBB where RR, GG, BB are hexadecimal values
```
mwestphal, line 73: missing `#RGB`

application/F3DOptionsTools.cxx
```diff
@@ -89,7 +89,7 @@ static inline const std::array<CLIGroup, 8> CLIOptions = {{
       { "grid-absolute", "", "Position grid at the absolute origin instead of below the model", "<bool>", "1" },
       { "grid-unit", "", "Size of grid unit square, automatically computed by default", "<value>", "" },
       { "grid-subdivisions", "", "Number of grid subdivisions", "<value>", "" },
-      { "grid-color", "", "Color of main grid lines", "<R,G,B>", "" },
+      { "grid-color", "", "Color of main grid lines", "<color>", "" },
```
mwestphal, line 92: both `OPTIONS.md` should be improved in the same way

library/private/options_tools.h
mwestphal: still not adressed yet

library/private/options_tools.h
```diff
@@ -186,17 +189,130 @@ ratio_t parse(const std::string& str)
 //----------------------------------------------------------------------------
 /**
  * Parse provided string into a color_t.
- * Supported formats: "R,G,B"
- * rely on parse<std::vector<double>>(str)
+ * Supported formats: "R,G,B", "#RRGGBB", "rgb(R,G,B)", "hsl(H,S%,L%)", "hsv(H,S%,V%)",
```
mwestphal, line 192: missing `#RGB`

library/private/options_tools.h
```diff
@@ -186,17 +189,130 @@ ratio_t parse(const std::string& str)
...
+    /* Short hex format search */
+    const std::regex shortHexRegex("#([0-9a-f])([0-9a-f])([0-9a-f])", std::regex_constants::icase);
+    std::smatch shortHexMatch;
+    if (std::regex_match(strCompact, shortHexMatch, shortHexRegex))
+    {
+      return color_t(
+        std::stoul(shortHexMatch[1].str() + shortHexMatch[1].str(), nullptr, 16) / 255.0,
+        std::stoul(shortHexMatch[2].str() + shortHexMatch[2].str(), nullptr, 16) / 255.0,
+        std::stoul(shortHexMatch[3].str() + shortHexMatch[3].str(), nullptr, 16) / 255.0);
+    }
```
mwestphal, line 225: I would put this one before the long hex ?

library/private/options_tools.h
```diff
@@ -323,12 +439,30 @@ std::string format(const std::vector<T>& var)
 //----------------------------------------------------------------------------
 /**
  * Format provided var into a string from provided color_t
- * rely on format(std::vector<double>&)
+ * Formats in hex color string "#RRGGBB" if values are multiple of 1/255.
```
mwestphal, line 442: should short hex be used when possible ?

mwestphal (changes requested):
small changes needed

application/F3DOptionsTools.cxx
VittorioSanchez: done already

library/private/options_tools.h
mwestphal: true, lets keep as is.

library/private/options_tools.h
```diff
@@ -8,6 +8,9 @@
 #include <cassert>
 #include <regex>
 #include <sstream>
+#include <vtkMath.h>
+#include <vtkNamedColors.h>
```
mwestphal, line 12: put vtk includes before system includes

Meakk (approved):
LGTM

mwestphal:
Thanks for your contribution @VittorioSanchez !

PR #2023, opened 2025-02-28
handle arbitrary up vector

### LINKED ISSUE #1533: Generalize "UP" vector to be any direction instead of axis aligned

snoyer:
would fix #1533


vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -288,61 +291,64 @@ void vtkF3DRenderer::Initialize()
...
+
+  std::array<double, 3> up = { upVec[0], upVec[1], upVec[2] };
+  std::array<double, 3> right = { 1, 0, 0 };
 
-  // TODO: Remove this when https://github.com/f3d-app/f3d/pull/1536/ is merged
-  int upIndex = -1;
-  bool error = true;
-  for (int i = 0; i < 3; i++)
+  /* if `up` is `(0,0,0)` make it `(0,1,0)` */
+  if (isNullVector(up))
```
mwestphal, line 304: warn maybe ?

testing/baselines/TestGridX.png
snoyer, line 1: drawing order of the axes has changed (blue over green vs the other way around)

mwestphal: I guess thats fine, but I dont see why it would related to VTK versions though

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -655,6 +687,116 @@ void vtkF3DRenderer::ConfigureGridUsingCurrentActors()
...
+          }
+          else
+          {
+            if (!extendBoxArbitrary(prop3d, box))
+            {
+              const std::string classname = std::string(prop3d->GetClassName());
+              F3DLog::Print(F3DLog::Severity::Warning,
+                "Could not properly account for " + classname +
+                  " in non-axis-aligned bounds computation");
+              extendBoxAxisAligned(prop3d, box);
```
snoyer, line 789: I believe this block is needed as it _could_ be reach if we had some unanticipated type of actors. Obviously we'd want to anticipate as much as possible which makes pretty hard to cover with a test.
But at the same time it feels wrong to drop the block altogether because such unanticipated actors could come from plugins.

testing/baselines/TestGridX.png
snoyer: possibly due to minor differences in the image comparison algorithm between VTK versions

vtkext/private/module/vtkF3DRenderer.cxx
mwestphal: what do you mean by unanticipated type of actors ?

mwestphal: ha got it, just put an `assert(actor)` in extendBoxArbitrary and return void.

snoyer: I removed the fallback code as discussed on the assumption that we only deal with `vtkPolyDataMapper`s

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -655,6 +687,110 @@ void vtkF3DRenderer::ConfigureGridUsingCurrentActors()
...
+    prop3d->PokeMatrix(tmpMatrix);
+
+    box.AddBounds(prop3d->GetBounds());
+
+    prop3d->PokeMatrix(nullptr);
+  };
+
+  /* Use custom logic to extend box.
+   * Should give the tightest bounds even when non-axis-aligned.
+   * Only implemented for `vtkPolyDataMapper`-based actors. */
```
mwestphal, line 733: should that computation be added to VTK ? It looks useful.

mwestphal:
doc to update

mwestphal (approved):
nice, needs a @Meakk review though :)

testing/baselines/TestConfigFileMultiFileSTL.png
Meakk, line 1: looks like a grid line position has changed. Is it expected?

vtkext/private/module/vtkF3DOpenGLGridMapper.cxx
```diff
@@ -52,8 +48,8 @@ void vtkF3DOpenGLGridMapper::ReplaceShaderValues(
   );
   vtkShaderProgram::Substitute(VSSource, "//VTK::PositionVC::Impl",
     "gridCoord = vertexMC.xy * fadeDist;\n"
-    "gridOffset = originOffset." + axes2d + ";\n"
-    "gl_Position = MCDCMatrix * vec4(vertexMC." + axes3d + " * fadeDist, 1.0);\n"
+    "gridOffset = originOffset.xz;\n"
+    "gl_Position = MCDCMatrix * vec4(vertexMC.xzy * fadeDist, 1.0);\n"
```
Meakk, line 52: I'm not entirely understanding the logic here, but swapping Y and Z makes me think a rotation matrix could be missing in the model matrix.

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -290,59 +293,65 @@ void vtkF3DRenderer::InitializeUpVector(const std::vector<double>& upVec)
...
+  if (isNullVector(up))
+  {
+    up[1] = 1.0;
+    F3DLog::Print(F3DLog::Severity::Warning, "null up vector, using (0,0,1) instead");
   }
+  vtkMath::Normalize(up.data());
 
-  if (!error)
+  /* make sure `right` is not colinear with `up` */
+  for (size_t i = 0; ::abs(vtkMath::Dot(right, up)) > 0.999 && i < 3; ++i)
```
Meakk, line 314: It feels like the for loop is not required. What about?
```cpp
if (::abs(vtkMath::Dot(right, up)) > 0.999)
{
  right = { 0.0, 1.0, 0.0 };
}
```

vtkext/private/module/vtkF3DOpenGLGridMapper.cxx
Meakk: Also, `originOffset` is used as a 2D vector with the Y value ignored. Could we change the uniform to a 2D offset? And maybe use it in the fragment shader directly?

testing/baselines/TestConfigFileMultiFileSTL.png
snoyer: not really expected but the orientation/position of the grid used to be handled in the shader and now it is set on the actor so maybe that could explain some minor differences?

vtkext/private/module/vtkF3DRenderer.cxx
snoyer: true, I reused the code from a previous version of the function where `right` was a param, it's now hardcoded so we can simplify

vtkext/private/module/vtkF3DOpenGLGridMapper.cxx
snoyer: > I'm not entirely understanding the logic here, but swapping Y and Z makes me think a rotation matrix could be missing in the model matrix.

not sure either I just reused one of the axis-aligned cases that was already there :/



> Also, `originOffset` is used as a 2D vector with the Y value ignored. Could we change the uniform to a 2D offset? And maybe use it in the fragment shader directly?

you're right, I'll drop `originOffset` and make `gridOffset` a uniform instead

Meakk: > not sure either I just reused one of the axis-aligned cases that was already there :/

Ok, it's probably related to the fact `vertexMC` is 2D in the grid plane (XZ). So using `vertexMC.xzy` is equivalent to `vec3(vertexMC.x, 0, vertexMC.y)`.  
Anyway, that's fine.

Meakk (approved):

vtkext/private/module/vtkF3DRenderer.cxx
snoyer: Not a VTK expert myself so I'll let you ponder that one ;p

mwestphal:
Thanks for the contribution @snoyer :)

PR #2029, opened 2025-03-01
Cache: Add documentation and add XDG support

mwestphal:


doc/user/OPTIONS.md
```diff
@@ -182,3 +182,13 @@ For example the screenshot filename is configured as `{app}/{model}_{n}.png` by
...
+
+## HDRI Caches
+
+When using HDRI related options, F3D will create and use a cache directory to store related data in order to speed up rendering.
+These cache files can be safely removed at the cost of recomputing them on next use.
+
+The cache directory location is as follows, in order, using the first defined environnement variables:
+ - Windows: `%LOCALAPPDATA%\f3d`
+ - Linux: `${XDG_CACHE_HOME}/f3d`,`${HOME}/.cache/f3d`
+ - macOS: `${XDG_CACHE_HOME}/f3d`,`${HOME}/Library/Caches/f3d`
```
mwestphal, line 194: @Meakk is that the expected behavior on macOS ?

Meakk: I need to double check but I don't think `XDG_CACHE_HOME` is used tbh

mwestphal: my reference is this so post: https://stackoverflow.com/questions/3373948/equivalents-of-xdg-config-home-and-xdg-data-home-on-mac-os-x but its not clear to me if these variables are set or not on macOS.

Meakk: Just checked, they are not

library/src/engine.cxx
```diff
@@ -52,32 +52,43 @@ engine::engine(
   // Ensure all lib initialization is done (once)
   detail::init::initialize();
 
+  fs::path cachePath;
 #if defined(_WIN32)
-  static constexpr const char* CACHE_ENV_VAR = "LOCALAPPDATA";
+  const char* appData = std::getenv("APPDATA");
```
Meakk, line 57: Why is it `APPDATA` instead of `LOCALAPPDATA` now?

mwestphal: mistake on my part, it should be LOCALAPPDATA afaiu

mwestphal, line 57:
```suggestion
  const char* appData = std::getenv("LOCALAPPDATA");
```

Meakk (approved):

PR #2031, opened 2025-03-02
Fix wasm option string issue

Meakk:


webassembly/F3DEmscriptenBindings.cxx
```diff
@@ -43,7 +43,7 @@ std::string get_as_string(f3d::options& o, const std::string& name)
 }
 f3d::options& set_string(f3d::options& o, const std::string& name, const std::string& value)
 {
-  return o.set(name, value);
+  return o.setAsString(name, value);
```
mwestphal, line 46: I dont think this is correct. you should change that in the app code, not the bindings code.

mwestphal: my change was actually not retrocompatible as it changed the underlying type of up_direction.

mwestphal (commented):
changes needed

mwestphal (changes requested):
changes needed

PR #2038, opened 2025-03-03
parse mixed `xyz` directions

### LINKED ISSUE #2021: Add proper parsing for direction_t
**Context**
direction_t option type has recently been added in libf3d, however parsing is actually the same as for double_vector and simple [+/-][X/Y/Z]
Lets implement proper parsing of colors, such as -x+y

**Suggested solution**
 - [x] Implement proper parsing in options_tools::parse for this type
 - [x] Improve formating 

**Syntax to Support**
 - [-/+][z/y/z]  (case agnostic)
 - repeated version of the above, eg: "-x+y+z"
 - contracted version: "-xy"

snoyer:
would fix #2021

### DIFF
--- a/library/private/options_tools.h
+++ b/library/private/options_tools.h
@@ -247,7 +247,7 @@ direction_t parse(const std::string& str)
       {
         return direction_t(options_tools::parse<std::vector<double>>(str));
       }
-      catch (const options::parsing_exception& ex)
+      catch (const options::parsing_exception&)
       {
         throw options::parsing_exception("Cannot parse " + str + " into a direction_t");
       }
--- a/library/private/options_tools.h
+++ b/library/private/options_tools.h
@@ -243,7 +243,14 @@ direction_t parse(const std::string& str)
     }
     else
     {
-      return direction_t(options_tools::parse<std::vector<double>>(str));
+      try
+      {
+        return direction_t(options_tools::parse<std::vector<double>>(str));
+      }
+      catch (const options::parsing_exception& ex)
+      {
+        throw options::parsing_exception("Cannot parse " + str + " into a direction_t");
+      }
     }
   }
   catch (const f3d::type_construction_exception& ex)
--- a/library/private/options_tools.h
+++ b/library/private/options_tools.h
@@ -206,7 +206,7 @@ color_t parse(const std::string& str)
 //----------------------------------------------------------------------------
 /**
  * Parse provided string into a direction_t.
- * Supported formats: "X,Y,Z", "[+|-][X|Y|Z]"
+ * Supported formats: "X,Y,Z", "[[+|-]X][[+|-]Y][[+|-]Z]"
  * rely on parse<std::vector<double>>(str)
  * Can throw options::parsing_exception in case of failure to parse
  */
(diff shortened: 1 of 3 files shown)


library/private/options_tools.h
```diff
@@ -206,7 +206,7 @@ color_t parse(const std::string& str)
 //----------------------------------------------------------------------------
 /**
  * Parse provided string into a direction_t.
- * Supported formats: "X,Y,Z", "[+|-][X|Y|Z]"
+ * Supported formats: "X,Y,Z", "[[+|-]X][[+|-]Y][[+|-]Z]"
```
mwestphal, line 209: please precise that order should be respected and case does not matter ?

library/private/options_tools.h
```diff
@@ -215,21 +215,42 @@ direction_t parse(const std::string& str)
...
     else
     {
-      return direction_t(options_tools::parse<std::vector<double>>(str));
+      try
+      {
+        return direction_t(options_tools::parse<std::vector<double>>(str));
+      }
+      catch (const options::parsing_exception&)
+      {
+        throw options::parsing_exception("Cannot parse " + str + " into a direction_t");
```
mwestphal, line 252: is that a bug fix ?

library/private/options_tools.h
snoyer: order is already be implied by this syntax, no?

library/private/options_tools.h
snoyer: just improves the messages by not raising the "could not parse into a double" that may not be relevant.
For example the input `"foobar"` will not match the regex and be tried as a double vector and fail but "could not parse foobar into a double" is not really useful in that case

mwestphal: fair.

library/private/options_tools.h
mwestphal: true.

library/private/options_tools.h
```diff
@@ -206,7 +206,7 @@ color_t parse(const std::string& str)
 //----------------------------------------------------------------------------
 /**
  * Parse provided string into a direction_t.
```
mwestphal, line 208: do you think we should improve format as well ?

snoyer: you mean detect if arbitrary x,y,z values can be formatted as `[±X][±Y][±Z]`?

mwestphal: Yes, this is what I mean

snoyer: should be doable, you want it in this PR?

mwestphal: as you prefer

library/private/options_tools.h
```diff
@@ -206,7 +206,7 @@ color_t parse(const std::string& str)
...
- * Supported formats: "X,Y,Z", "[+|-][X|Y|Z]"
+ * Supported formats: "X,Y,Z", "[[+|-]X][[+|-]Y][[+|-]Z]" (case insensitive)
```
mwestphal, line 209: please update `doc/user/PARSING.md` accordingly

mwestphal (changes requested):
doc update needed

mwestphal (approved):

mwestphal:
Thanks for the contribution @snoyer !

PR #2040, opened 2025-03-03
Add support for imperative options in configuration files

### LINKED ISSUE #1601: Add imperative config option support
With the option rework the behavior of CLI VS config file is clarified, with CLI always overriding.
However it would be nice to have "imperative" config options that would override CLI and dynamic options

When parsing config file option, detect `!` in front of the option name to know its an imperative and implement the rest

**Additional context**
https://github.com/f3d-app/f3d/pull/1580

mwestphal:
- Add support for imperative option in config files using the `!` syntax
 - Add tests and docs
 - Fix #1601


application/F3DConfigFileTools.cxx
```diff
@@ -85,7 +85,8 @@ std::vector<fs::path> GetConfigPaths(const std::string& configSearch)
 }
 
 //----------------------------------------------------------------------------
-std::pair<F3DOptionsTools::OptionsEntries, F3DConfigFileTools::BindingsEntries>
+std::tuple<F3DOptionsTools::OptionsEntries, F3DOptionsTools::OptionsEntries,
```
Meakk, line 88: It's not clear at first why we do have two options entries. It's probably a good idea to create a struct to name the two lists.

application/F3DConfigFileTools.cxx
```diff
@@ -217,26 +219,46 @@ F3DConfigFileTools::ReadConfigFiles(const std::string& userConfig)
...
           for (const auto& item : optionsBlock.items())
           {
+            // Individual item can be imperative, store in the right dict
+            std::string key = item.key();
+            F3DOptionsTools::OptionsDict* localEntry = &entry;
+            if (!key.empty() && key[0] == '!')
+            {
+              localEntry = &imperativeEntry;
+              key = key.substr(1);
+            }
```
Meakk, line 232: Pointers are so much C 😛 
```suggestion
            std::reference_wrapper<F3DOptionsTools::OptionsDict> localEntry = std::ref(entry);
            if (!key.empty() && key[0] == '!')
            {
              localEntry = std::ref(imperativeEntry);
              key = key.substr(1);
            }
```

application/F3DConfigFileTools.cxx
```diff
@@ -217,26 +219,46 @@ F3DConfigFileTools::ReadConfigFiles(const std::string& userConfig)
...
+
             if (item.value().is_number() || item.value().is_boolean())
             {
-              entry[item.key()] = nlohmann::to_string(item.value());
+              (*localEntry)[key] = nlohmann::to_string(item.value());
```
Meakk, line 236: If using a reference wrapper:
```suggestion
              localEntry.get()[key] = nlohmann::to_string(item.value());
```

application/F3DStarter.cxx
```diff
@@ -842,7 +841,7 @@ int F3DStarter::Start(int argc, char** argv)
     F3DOptionsTools::ParseCLIOptions(argc, argv, inputFiles);
 
   // Store in a option entries for easier processing
-  this->Internals->CLIOptionsEntries.emplace_back(cliOptionsDict, fs::path(), "CLI options");
+  this->Internals->CLIOptionsEntries.emplace_back(std::tie(cliOptionsDict, "", "CLI options"));
```
Meakk, line 844: Is this change necessary?

Meakk (changes requested):

application/F3DConfigFileTools.cxx
mwestphal: when you dont know how to write something, just write it wrong and ask for a review :p

doc/user/CONFIGURATION_FILE.md
```diff
@@ -70,6 +70,29 @@ The following options <b>are only taken on the first load</b>:
 Boolean options that have been turned on in the configuration file can be turned
 off on the command line if needed, eg: `--point-sprites=false`.
 
+### Imperative Options
+
+Command line options and options that are changed interactively overrides options that are set in configuration files.
+This is not always a wanted behavior, so to force an option to always be taken into account even if set in command line are changed interactively, it is possible to use imperative options, by adding a `!` in front of the option name, eg:
```
Meakk, line 76:
```suggestion
This is not always a desired behavior, so in order to force an option to always be taken into account even if set in command line or changed interactively, it is possible to use imperative options, by adding a `!` in front of the option name, eg:
```

Meakk (approved):

Meakk:
Small angliche suggestion, but otherwise it looks good to me

PR #2041, opened 2025-03-04
format `direction_t` as XYZ when possible

snoyer:
addresses formatting part of #2021

### DIFF
--- a/library/private/options_tools.h
+++ b/library/private/options_tools.h
@@ -385,7 +385,7 @@ std::string format(direction_t var)
           str += new_sign;
           sign = new_sign;
         }
-        str += 'X' + i;
+        str += static_cast<char>('X' + i);
       }
     }
     if (str.empty())
--- a/library/private/options_tools.h
+++ b/library/private/options_tools.h
@@ -359,33 +359,33 @@ std::string format(color_t var)
  */
 std::string format(direction_t var)
 {
-  const auto is_zero = [](double a) { return std::abs(a) < 1e-12; };
-  const auto abs_diff = [](double a, double b) { return std::abs(a) - std::abs(b); };
+  const auto isZero = [](double a) { return std::abs(a) < 1e-12; };
+  const auto absDiff = [](double a, double b) { return std::abs(a) - std::abs(b); };
   const auto formatAsXYZ = [&]()
   {
     std::string str = "";
-    double first_nonzero = 0; // not non-zero until first
-    char sign = ' ';          // initially not `+`/`-` to force first sign
-    for (size_t i = 0; i < 3; ++i)
+    double firstNonZero = 0; // not non-zero until first
+    char sign = '\0';        // initially not `+`/`-` to force first sign
+    for (size_t componentIndex = 0; componentIndex < 3; ++componentIndex)
     {
-      const double v = var[i];
-      if (!is_zero(v))
+      const double componentValue = var[componentIndex];
+      if (!isZero(componentValue))
       {
-        if (is_zero(first_nonzero))
+        if (isZero(firstNonZero))
         {
-          first_nonzero = v;
+          firstNonZero = componentValue;
         }
-        else if (!is_zero(abs_diff(v, first_nonzero)))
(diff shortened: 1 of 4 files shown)


library/private/options_tools.h
```diff
@@ -354,13 +354,55 @@ std::string format(color_t var)
...
 std::string format(direction_t var)
 {
-  // TODO generate a proper direction string
-  return options_tools::format(static_cast<std::vector<double>>(var));
+  const auto is_zero = [](double a) { return std::abs(a) < 1e-12; };
+  const auto abs_diff = [](double a, double b) { return std::abs(a) - std::abs(b); };
+  const auto formatAsXYZ = [&]()
+  {
+    std::string str = "";
+    double first_nonzero = 0; // not non-zero until first
```
mwestphal, line 367:
```suggestion
    double firstNonZero = 0; // not non-zero until first
```

library/private/options_tools.h
```diff
@@ -354,13 +354,55 @@ std::string format(color_t var)
...
+    char sign = ' ';          // initially not `+`/`-` to force first sign
```
mwestphal, line 368: init to `\0` may be more explicit ?

library/private/options_tools.h
```diff
@@ -354,13 +354,55 @@ std::string format(color_t var)
...
+      {
+        if (is_zero(first_nonzero))
+        {
+          first_nonzero = v;
+        }
+        else if (!is_zero(abs_diff(v, first_nonzero)))
+        {
+          throw std::invalid_argument("not all same");
+        }
+        const char new_sign = v < 0 ? '-' : '+';
```
mwestphal, line 382:
```suggestion
        const char newSign = v < 0 ? '-' : '+';
```

library/testing/TestSDKOptions.cxx
```diff
@@ -142,7 +142,7 @@ int TestSDKOptions(int argc, char* argv[])
 
   std::stringstream ssDir;
   ssDir << f3d::direction_t(0, 0, -1.0);
-  test("direction to string", ssDir.str() == "0,0,-1");
+  test("direction to string", ssDir.str(), "-Z");
```
mwestphal, line 145: maybe add a test above checking the code path that format to the vector format ?

mwestphal: all good, you did it in OptionIO (which is the right place)

library/private/options_tools.h
```diff
@@ -354,13 +354,55 @@ std::string format(color_t var)
...
+  const auto is_zero = [](double a) { return std::abs(a) < 1e-12; };
+  const auto abs_diff = [](double a, double b) { return std::abs(a) - std::abs(b); };
+  const auto formatAsXYZ = [&]()
+  {
+    std::string str = "";
+    double first_nonzero = 0; // not non-zero until first
+    char sign = ' ';          // initially not `+`/`-` to force first sign
+    for (size_t i = 0; i < 3; ++i)
+    {
+      const double v = var[i];
```
mwestphal, line 371: lets avoid single letter var

library/private/options_tools.h
```diff
@@ -354,13 +354,55 @@ std::string format(color_t var)
...
+      if (!is_zero(v))
+      {
+        if (is_zero(first_nonzero))
+        {
+          first_nonzero = v;
+        }
+        else if (!is_zero(abs_diff(v, first_nonzero)))
+        {
+          throw std::invalid_argument("not all same");
```
mwestphal, line 380: i dont get this code path, what does this correspond to ?

mwestphal (changes requested):
also need a doc update

library/private/options_tools.h
snoyer: we can only write as `±XYZ` if there are one or more non-zero components and they all have the same absolute value (`1,0,0`, `2,2,0`, `3,-3,3`, ...) if either condition doesn't hold then we bail an fall back to vector format

mwestphal (approved):

PR #2045, opened 2025-03-05
options: Add path_t and use it for hdri

mwestphal:


Meakk:
Any reason to not use `std::filesystem::path`?

mwestphal:
> Any reason to not use `std::filesystem::path`?

Im using std::filesystem::path already, inside of path_t. Its required to be done like this as std::filesystem::path is not part of the variant API. Adding a new type to the variant API is a possibility but I'd rather not.

mwestphal:
nevermind you are right @Meakk , I can use std::filesystem::path directly

mwestphal:
actually a valid approach  compared to https://github.com/f3d-app/f3d/pull/2046

Issue #1570, opened 2024-08-20
Add more option types


mwestphal:
**Is your feature request related to a problem? Please describe.**
Options generation from .json file was added in https://github.com/f3d-app/f3d/pull/1471, but only a few types were added:

- bool
- int
- double
- ratio
- double_vector
- string

Lets add more types.
 - [x] direction (specialisation of double_vector): https://github.com/f3d-app/f3d/issues/1645
 - [x] color (specialisation of double_vector): https://github.com/f3d-app/f3d/issues/1762
 - [x] path (specialisation of string)
 - [x] rgba_colormap (specialisation of double_vector)
 
  To add a type it is needed to:
 - Add it to types.h with the correct API
 - Add a naive parse/format methods to options_tools.h.in
 - Use it for at least one option in options.json
 - Add testing in TestSDKOptions*

Yogesh9000:
Is this issue still open? If it is I would like to work on it.

mwestphal:
Yes it is open! Lets discuss this one discord though :)

Yogesh9000:
sure lets discuss on discord

amannages:
Good afternoon, I was wondering if this issue was still open? If so, is it possible I could work on it?

mwestphal:
It is still open but @Yogesh9000 is working on it in https://github.com/f3d-app/f3d/pull/1648

mwestphal:
fixed by the addition of colormap: https://github.com/f3d-app/f3d/pull/2061

PR #2046, opened 2025-03-05
options: Add support for path using std::filesystem::path

mwestphal:
- Add a `_option_variant_convert` logic in f3dOptions.cmake for conversion from an option type to a variant type
- Add `path` aka `std::filesystem::path` has a valid option type
- Add parse/format for path
- Use path for hdri.file, font_file, all texture options
- Add functionnal, unit and coverage testing
- Add doc
- Added deprecation code for collapse behavior

related to https://github.com/f3d-app/f3d/issues/1570


doc/user/PARSING.md
```diff
@@ -57,7 +58,15 @@ When formatting a ratio into a string, it is formatted as a double.
...
 ## String
 
-String are parsed and formatted as is.
+String are trimmed of leading and trailing space when parsed.
+String are formatted as is.
+
+## String
+
+Path are formatted by creating a [std::filesystem::path](https://en.cppreference.com/w/cpp/filesystem/path/path)
+and then collapsed using [f3d::utils::collapsePath](https://f3d.app/doc/libf3d/doxygen/api/classf3d_1_1utils.html#_CPPv4N3f3d5utils12collapsePathERKNSt10filesystem4pathERKNSt10filesystem4pathE).
```
mwestphal, line 67: Im not convinced by my doc here but I also do not know if I should be super specific.

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -87,6 +87,26 @@
 
 namespace
 {
+std::string DeprecatedCollapsePath(const fs::path& path)
+{
+  std::string collapsed;
+  std::string origin = path.string();
+
+  // Handle retro-compatibility but warn for deprecation
+  // Foe easier removal when removing deprecation: F3D_DEPRECATED
```
Meakk, line 96:
```suggestion
  // For easier removal when removing deprecation: F3D_DEPRECATED
```

Meakk (approved):
Small typo, but otherwise it looks good.  
I still would like to clarify if we still need the variant, or if we can get rid of variant and actual types, but it's out of scope of this PR

mwestphal:
>  I still would like to clarify if we still need the variant, or if we can get rid of variant and actual types, but it's out of scope of this PR

We still need it, but we could deprecate it and remove the API on the next release. Ill open an issue.

mwestphal:
https://github.com/f3d-app/f3d/issues/2054

Issue #1492, opened 2024-06-28
F3D segfault with a specific FBX file


mwestphal:
**Describe the bug**
F3D segfault with a specific FBX file:  https://sketchfab.com/3d-models/wolf-e319f122c2474d3186b841b019fb0edd

[wolf.zip](https://github.com/user-attachments/files/16025588/wolf.zip)

**To Reproduce**
 1. Open the file using `./bin/f3d ~/data/tmp/source/Wolf_test_animation_.fbx`
 2. segfault

**Expected behavior**
No segfault

**System Information:**
 - OS: ArchLinux
 - GPU and GPU driver: Unrelevant

**F3D Information**
Paste the content of `f3d --version`:

```
F3D 2.4.0

F3D - A fast and minimalist 3D viewer
Version: 2.4.0-87-gbb439bdb.
Build date: 2024-06-28 06:54:00.
Build system: Linux 64-bits.
Compiler: GNU 14.1.1.
External rendering module: OFF.
Raytracing module: OFF.
VTK version: 9.3.0-3549-g2e9a7c71a2 (date: 20240627).
Copyright (C) 2019-2021 Kitware SAS.
Copyright (C) 2021-2024 Michael Migliore, Mathieu Westphal.
License BSD-3-Clause.
By Michael Migliore, Mathieu Westphal and Joachim Pouderoux.

```

**Additional context**

```
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[New Thread 0x7fffe6bff6c0 (LWP 103344)]
Configuration file for "config" could not be found

Thread 1 "f3d" received signal SIGSEGV, Segmentation fault.
0x00007ffff2de466b in Assimp::LimitBoneWeightsProcess::ProcessMesh(aiMesh*) () from /home/glow/dev/f3d/deps/assimp/install/lib/libassimp.so.5
(gdb) bt
#0  0x00007ffff2de466b in Assimp::LimitBoneWeightsProcess::ProcessMesh(aiMesh*) () from /home/glow/dev/f3d/deps/assimp/install/lib/libassimp.so.5
#1  0x00007ffff2de3ff4 in Assimp::LimitBoneWeightsProcess::Execute(aiScene*) () from /home/glow/dev/f3d/deps/assimp/install/lib/libassimp.so.5
[...]

PR #2051, opened 2025-03-08
Add validation flag for assimp

exbluesbreaker:
Passing this flag makes assimp perform validation of file before post-processing which can give a chance for assimp to report an error of some kind.
In particular case when issue was detected FBX file was problematic to assimp and without check it could lead to crash.

Issue was discovered in https://github.com/f3d-app/f3d/issues/1492


exbluesbreaker:
@Meakk @mwestphal

mwestphal:
Can you add a test for this ?

You can add an invalid.fbx in testing/data and check it doe not crash for example ?

exbluesbreaker:
> Can you add a test for this ?
> 
> You can add an invalid.fbx in testing/data and check it doe not crash for example ?

Let me figure out how to do it

exbluesbreaker:
I added test case but I am not sure it is the best way to do. I don't know how to check that run just doesn't crash, so I added checking regex about no rendering being done, but on the other hand if assimp support for such fbx file will start working this test case could start failing. Maybe test should be implemented in a different way, I would appreciate help.

exbluesbreaker:
I am trying to build 9.3.1 VTK to see what could be wrong with running this test with it. I had problems while attempting to build 9.2.6 which I could not solve. If there are some Dockerfile for container where I can build f3d with it I would appreciate it.

mwestphal:
> I am trying to build 9.3.1 VTK to see what could be wrong with running this test with it. I had problems while attempting to build 9.2.6 which I could not solve. If there are some Dockerfile for container where I can build f3d with it I would appreciate it.

Its expected, detecting the filaure to load a file was not possible before VTK 9.4. You can put your test next to `TestInvalidFile` which check the VTK version beforehand.

mwestphal:
> I added test case but I am not sure it is the best way to do. I don't know how to check that run just doesn't crash, so I added checking regex about no rendering being done, but on the other hand if assimp support for such fbx file will start working this test case could start failing. Maybe test should be implemented in a different way, I would appreciate help.

Your test is fine, assimp is not going to update itself. If we update assimp and it fixes the issue, we will update the test at that time.

mwestphal:
However this is two issues with the data.

1. Its too big <1Mo would be much better.
2. There is no license for it and we have no idea what license it would be.

So unless assimp contains a file we can reuse to test that flag, Im not sure we can test this.

Can you quickly check in assimp data if they have something that fit ? If not, lets drop the test.

exbluesbreaker:
> There is no license for it and we have no idea what license it would be.

I was thinkg it could be a problem, I will try to check assimp.  But since their viewer reports an error for such file, they probably do not yet consider supporting such stuff.

In general it should not be a problem who familiar with to make fbx file with bone name repeated(though I do not know how common is it).

mwestphal:
>  I will try to check assimp. But since their viewer reports an error for such file, they probably do not yet consider supporting such stuff.

Of course, thats why they may have a dataset to check that error case, I hope.

> In general it should not be a problem who familiar with to make fbx file with bone name repeated(though I do not know how common is it).

I suppose, however I'm not a 3D artist and I dont want to bother the few in the community that could do it for such an edges case :).

exbluesbreaker:
I checked fbx files in assimp repo subfolders, and they do not reproduce this problem.  I also tried to browse some models online but ended up only finding original wolf. License is specified on Sketchfab. 

image

mwestphal:
CC-NC is actually ok and 2Mo size is acceptable, lets keep the test in then :)

Please add a mention in DATA_LICENSES.md

testing/data/DATA_LICENSES.md
```diff
@@ -58,8 +58,9 @@
 - world\*: VTK Data: BSD-3-Clause
 - 16bits.\*: @bisechen: [CC0](https://creativecommons.org/publicdomain/zero/1.0/)
 - (ノಠ益ಠ )ノ.vtp: VTK Data: BSD-3-Clause
+- Wolf.fbx: OLEKSO: [CC-NC 4.0](http://creativecommons.org/licenses/by-nc/4.0/)
 
-All other datasets are licensed under the BSD-3-Clause F3D license.
+All other datasets are licensed under the BSD-3-Clause F3D license.`
```
mwestphal, line 63:
```suggestion
All other datasets are licensed under the BSD-3-Clause F3D license.
```

mwestphal (changes requested):
small change needed

plugins/assimp/module/vtkF3DAssimpImporter.cxx
```diff
@@ -592,7 +592,8 @@ class vtkF3DAssimpImporter::vtkInternals
     {
       // Work around for https://github.com/assimp/assimp/issues/4620
       this->Importer.SetPropertyBool(AI_CONFIG_IMPORT_FBX_PRESERVE_PIVOTS, false);
-      this->Scene = this->Importer.ReadFile(filePath, aiProcess_LimitBoneWeights);
+      this->Scene = this->Importer.ReadFile(
+        filePath, aiProcess_LimitBoneWeights | aiProcess_ValidateDataStructure);
```
Meakk, line 596: A few things I can read in the documentation of this flag:

> This post-processing step is not time-consuming. Its use is not compulsory, but recommended.

So I think it's a good idea to enable it.

> It is recommended that you capture Assimp's log output if you use this flag

That sounds like something we should do. We should transfer Assimp logs to F3D log so that all messages from Assimp are sent to F3D console.

exbluesbreaker:
Please take a look. I haven't worked extensively with exception handling. Maybe instead of throwing, I should return an error directly.

In general, the DeadlyImportError occurs within Assimp and is caught internally, which results in a log message being recorded in the Assimp log. As a result, the exception never propagates outside of Assimp (probably). Given this, I'm unsure whether the try-catch block here is necessary, as DeadlyImportError might never actually escape from Assimp.

Here’s the comment from Assimp regarding this exception::

```
/** FOR IMPORTER PLUGINS ONLY: Simple exception class to be thrown if an
 *  unrecoverable error occurs while importing. Loading APIs return
 *  nullptr instead of a valid aiScene then.  */
```

exbluesbreaker:
I removed error here, as it actually checked later on.  I just didn't see it without "--verbose". In general maybe errors from `vtkErrorWithObjectMacro` should not be hidden unless it is like `--quite` mode of running

Meakk:
It seems like you should use `Assimp::LogStream` for that purpose. See https://assimp-docs.readthedocs.io/en/latest/usage/use_the_lib.html#logging  
You need to subclass `LogStream`, and call `vtk[Error|Warning|Debug]MacroWithObject` depending on the severity.  
We are doing something similar for USD here: https://github.com/f3d-app/f3d/blob/9d5bc108cb3baded7f433974f4ef10750a7dd9fa/plugins/usd/module/vtkF3DUSDImporter.cxx#L1177

exbluesbreaker:
> It seems like you should use `Assimp::LogStream` for that purpose. See https://assimp-docs.readthedocs.io/en/latest/usage/use_the_lib.html#logging You need to subclass `LogStream`, and call `vtk[Error|Warning|Debug]MacroWithObject` depending on the severity. We are doing something similar for USD here:
> 
> https://github.com/f3d-app/f3d/blob/9d5bc108cb3baded7f433974f4ef10750a7dd9fa/plugins/usd/module/vtkF3DUSDImporter.cxx#L1177

As I understood assimp sets error in its log which could be retrieved by `this->Importer.GetErrorString()` which is done if `Scene` after import is `nullptr` later in `ReadScene`

```
...
     vtkErrorWithObjectMacro(this->Parent, "Assimp failed to load: " << filePath);

      auto errorDescription = this->Importer.GetErrorString();
      vtkErrorWithObjectMacro(this->Parent, "Assimp error: " << errorDescription);// Message is forwarded to vtk log
      return false;
...
```

You need to use `--verbose` to see it in console but I guess it is true for other `vtkErrorWithObjectMacro` errors.
So I guess current code is ok. This is console messages with `--verbose`:

image

Meakk:
Isn't it possible to get a valid scene but still have warnings registered in assimp?

exbluesbreaker:
> Isn't it possible to get a valid scene but still have warnings registered in assimp?

As I understood, situation happening with this fbx could not be handled by assimp. It is not a warning level issue but an error. Assimp validation tep reports it as an error and if validation is skipped it would lead to crash.

Issue itself (the same name for different bones) is not contradicting FBX format itself, but it probably couldn't be handled properly by assimp yet.

exbluesbreaker:
Sorry I forgot about this requested change. But I do not understand why build fails

mwestphal:
No idea, I've rerun the job

mwestphal:
Any reason why you closed this ? I think it was almost ready for merge, was it not ?

exbluesbreaker:
> Any reason why you closed this ? I think it was almost ready for merge, was it not ?

Oh sorry, I will redo it, I had a local copy.  Or if it is possible to recover somehow, I would appreciate it. I was just trying to update my master in fork repositoty but it created a merge commit, so I deleted fork and created a new one. I started working with pull request through forks in github only recently. I didn't expect it will happen like this.

exbluesbreaker:
I reopening it

Issue #1945, opened 2025-01-24
Assimp format (fbx, dae, ...) provide empty metadata


mwestphal:
**Describe the bug**
F3D support many formats through the assimp plugin, however the meta data provided for these formats is empty

**To Reproduce**
 1. Open the file using `f3d --no-config -m 16bit.fbx`

![Image](https://github.com/user-attachments/assets/3b9c2bb3-3367-4105-b12b-a039c3e275fe)

**Expected behavior**
Meta data should contain information

**System Information:**
 - OS: Any
 - GPU and GPU driver: Ant

**F3D Information**
Paste the content of `f3d --version`: 3.0.0

**Additional context**
F3D uses ` vtkF3DMetaImporter::GetMetaDataDescription()` to recover that information, something must be going wrong there.

mwestphal:
Point sprites also do not work with assimp, I think this is related.

exbluesbreaker:
I will work on this issue :)

exbluesbreaker:
I haven't found a fix yet, but I believe I've identified the root cause.

To gather all necessary information, F3D first requires actors, which it retrieves using:

```
vtkActorCollection* actorCollection = importer->GetImportedActors();
```
in `bool vtkF3DMetaImporter::Update()`.

The issue is that `vtkF3DAssimpImporter` does not update the field accessed by `GetImportedActors`. Instead, it performs the following operation:

```
void vtkF3DAssimpImporter::ImportActors(vtkRenderer* renderer)
{
  this->Internals->ImportRoot(renderer);
}
```
This means that it updates `Internals`, which maintains its own structures for tracking actors. As a result, `this->Internals` is aware of the actors, but `this` itself is not, leading to "invisible" actors. 

Probably `vtkF3DAssimpImporter::ImportActors` should also somehow update its own `ActorCollection` together with `Internals`  but I do not have an implementation yet.

mwestphal:
Yes, you are absolutely right. The importer needs to be reworked to populate the actor collection.

exbluesbreaker:
I tried adding copying of data from internals into importer itself.  Pull request is https://github.com/f3d-app/f3d/pull/2058

Now you can see this with 16bit.fbx, hope it is correct:

![Image](https://github.com/user-attachments/assets/f20dc24b-683e-431b-9abb-c8728637aa2e)

mwestphal:
Fixed by #2058

PR #2058, opened 2025-03-11
assimp: Record imported actors from internals

exbluesbreaker:
Importer itself did't have actor stored which prevented metadata retrieval for display

Issues with missing metadata was reported in https://github.com/f3d-app/f3d/issues/1945

### DIFF
--- a/application/testing/CMakeLists.txt
+++ b/application/testing/CMakeLists.txt
@@ -650,7 +650,7 @@ if(F3D_PLUGIN_BUILD_ASSIMP)
   endif()
 
   if(VTK_VERSION VERSION_GREATER_EQUAL 9.3.20240707)
-    f3d_test(NAME TestAssimpMetaData DATA 16bit.fbx ARGS --verbose --load-plugins=assimp REGEXP "Number of actors: 1" NO_BASELINE)
+    f3d_test(NAME TestAssimpMetaData DATA duck.fbx ARGS --verbose --load-plugins=assimp REGEXP "Number of points: 12636" NO_BASELINE)
   endif()
 
 endif()
--- a/application/testing/CMakeLists.txt
+++ b/application/testing/CMakeLists.txt
@@ -649,6 +649,10 @@ if(F3D_PLUGIN_BUILD_ASSIMP)
     f3d_test(NAME TestThumbnailConfigFileAssimpDAE DATA duck.dae CONFIG thumbnail_build LONG_TIMEOUT TONE_MAPPING)
   endif()
 
+  if(VTK_VERSION VERSION_GREATER_EQUAL 9.3.20240707)
+    f3d_test(NAME TestAssimpMetaData DATA 16bit.fbx ARGS --verbose --load-plugins=assimp REGEXP "Number of actors: 1" NO_BASELINE)
+  endif()
+
 endif()
 
 if(F3D_PLUGIN_BUILD_DRACO)
--- a/application/testing/CMakeLists.txt
+++ b/application/testing/CMakeLists.txt
@@ -649,7 +649,11 @@ if(F3D_PLUGIN_BUILD_ASSIMP)
     f3d_test(NAME TestThumbnailConfigFileAssimpDAE DATA duck.dae CONFIG thumbnail_build LONG_TIMEOUT TONE_MAPPING)
   endif()
 
-f3d_test(NAME TestAssimpVerbose DATA duck.fbx ARGS --verbose --load-plugins=assimp REGEXP "Number of points: 12636" NO_BASELINE)
+  f3d_test(NAME TestAssimpVerbose DATA duck.fbx ARGS --verbose --load-plugins=assimp REGEXP "Number of points: 12636" NO_BASELINE)
+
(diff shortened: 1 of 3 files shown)


exbluesbreaker:
I am not exactly sure if on older version of vtk any fix is needed. I added guard for older vtk. f3d gets metadata in a different way in this case, but I do not have such vtk to test with, maybe some other fix is needed, but I cannot test it.

mwestphal:
> I am not exactly sure if on older version of vtk any fix is needed. I added guard for older vtk. f3d gets metadata in a different way in this case, but I do not have such vtk to test with, maybe some other fix is needed, but I cannot test it.

That looks correct! Can you add a test for it though ? A verbose + REGEXP test should do the trick

exbluesbreaker:
> > I am not exactly sure if on older version of vtk any fix is needed. I added guard for older vtk. f3d gets metadata in a different way in this case, but I do not have such vtk to test with, maybe some other fix is needed, but I cannot test it.
> 
> That looks correct! Can you add a test for it though ? A verbose + REGEXP test should do the trick

I'll do it tomorrow

exbluesbreaker:
So I added a simple test for number of actors, it is not super excisting as it is usually 1, but I didn't find a way to get number of cells or points

application/testing/CMakeLists.txt
```diff
@@ -649,6 +649,10 @@ if(F3D_PLUGIN_BUILD_ASSIMP)
     f3d_test(NAME TestThumbnailConfigFileAssimpDAE DATA duck.dae CONFIG thumbnail_build LONG_TIMEOUT TONE_MAPPING)
   endif()
 
+  if(VTK_VERSION VERSION_GREATER_EQUAL 9.3.20240707)
+    f3d_test(NAME TestAssimpMetaData DATA 16bit.fbx ARGS --verbose --load-plugins=assimp REGEXP "Number of actors: 1" NO_BASELINE)
```
mwestphal, line 653: See this test:

```
f3d_test(NAME TestVerbose DATA dragon.vtu ARGS --verbose REGEXP "Number of points: 13268\nNumber of cells: 26532" NO_BASELINE)
```

You cant do that same for the assimp data ?

exbluesbreaker: > You cant do that same for the assimp data ?

Let me have a look, probably something else is then needed for supporting metadata output to console, output in GUI seems working. In console I only saw actors and files after current implementation of fix.

exbluesbreaker:
I updated `Internals->Description` with points and cells data to see in on console output. I also updated test case for more points model so number is more "specific"

application/testing/CMakeLists.txt
mwestphal: indeed, you are right

mwestphal:
Ha right I forgot that this API is a bit all over the place. In any case its a good fix.

application/testing/CMakeLists.txt
```diff
@@ -649,6 +649,10 @@ if(F3D_PLUGIN_BUILD_ASSIMP)
     f3d_test(NAME TestThumbnailConfigFileAssimpDAE DATA duck.dae CONFIG thumbnail_build LONG_TIMEOUT TONE_MAPPING)
   endif()
 
+  if(VTK_VERSION VERSION_GREATER_EQUAL 9.3.20240707)
+    f3d_test(NAME TestAssimpMetaData DATA duck.fbx ARGS --verbose --load-plugins=assimp REGEXP "Number of points: 12636" NO_BASELINE)
+  endif()
```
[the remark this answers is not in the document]
exbluesbreaker: > Then you are not testing the metadata but the verbose mode. So I guess you dont need the VTK_VERSION check and the test should be named TestAssimpVerbose.
> 
> Lets not add a specific test for assimp metadata using the UI, that doesnt seems worth it.

Good point. I guess there should be some way to interact with GUI and retrieve text from it, but I didn't have much experience with advanced GUI programming.

exbluesbreaker:
Also branch name is a bit misleading I noticed it only recently, it should be assimp not just fbx fix

mwestphal:
> Also branch name is a bit misleading I noticed it only recently, it should be assimp not just fbx fix

Dont worry about the branch name :)

application/testing/CMakeLists.txt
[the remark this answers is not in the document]
exbluesbreaker: > > I guess there should be some way to interact with GUI and retrieve text from it, but I didn't have much experience with advanced GUI programming.
> 
> Actually if you want to add a GUI test, its pretty simple, basically we just capture the image and compare it against a baseline :)
> 
> See
> 
> ```
> f3d_test(NAME TestMetaDataImporter DATA BoxAnimated.gltf ARGS -m UI)
> ```

Understood, will have a look

exbluesbreaker:
Thanks for the tip about GUI test, I added it, now it seems complete

mwestphal (approved):

mwestphal:
Nice! lets wait fox @Meakk review before merging though.

Meakk (approved):

mwestphal:
Thanks for the contribution @exbluesbreaker !

PR #2061, opened 2025-03-13
[draft] Adding trivial colormap_t

mwestphal:


library/private/options_tools.h
```diff
@@ -428,6 +448,17 @@ std::string format(direction_t var)
   }
 }
 
+//----------------------------------------------------------------------------
+/**
+ * Format provided var into a string from provided colormap_t.
+ * Rely on `format(std::vector<double>&)`
+ * TODO add proper formatting
+ */
+std::string format(colormap_t var)
```
Meakk, line 457: Any reason it's passed by copy?
```suggestion
std::string format(const colormap_t& var)
```

Meakk (approved):
Small comment, but otherwise, that looks good to me

library/private/options_tools.h
mwestphal: no reason ^^

Issue #1727, opened 2024-11-23
`context.h` exceptions should be documented  and caught


mwestphal:
**Describe the bug**
`context.h` can throw exceptions, however they are not documented and not caugth, relying on main.cxx catch all to not segfault, this should be fixed.

**To Reproduce**
 1. comment main.cxx catch all
 2. `ctest -R WGL`
 3. segfault

saneck2150:
Hi, i would like to work on this issue, if is still relevant

mwestphal:
It is! Do you know guidance ?

mwestphal:
> where can I get documentation? 

https://f3d.app

https://discord.f3d.app

> I thought it is a good issue to get started with

It is! But I wrote it a bit fast, I can add more context

mwestphal:
Do you need any help moving forward @saneck2150 ?

saneck2150:
Hello, yes, I had problems with running the program, I asked tech support to help with solving it in your discord, but in the end nothing worked out. So I can't work with this project, sorry :(

mwestphal:
No worries @saneck2150 , thanks for the feedback.

exbluesbreaker:
I would like to try to work on this issue.

mwestphal:
Sure!

mwestphal:
>  I see there are loading_exception and symbol_exception thrown by context itself. Does it mean we need to rework those 2, or task is about handling other possible causes of exception in context.cxx - loading/getting symbols for different OpenGL wrappers (egl,glx etc.)?

Actually, not at all.

The point of this task is to catch loading_exception and symbol_exception in the application in a proper way in order to handle erroring out cleanly.

The catch all in main.cxx is a last resort and not suposed to be reachable.

The modification should be in the application.

eg in F3DStarter.cxx there are calls for `createGLX()`. These create* methods are documented as:

```
   * Throws a context::loading_exception if a needed graphic library cannot be loaded
   * Throws a context::symbol_exception if a needed symbol cannot be found in graphic library
   * Throws a engine::no_window_exception if the window cannot be created for another reason
   * Throws a engine::cache_exception if the default cache directory cannot be used
```

But in F3DStarter we never catch these exceptions. They should be cached, logged and handled properly.

exbluesbreaker:
I have initial version of commit prepared, but I need to find out how to test last 3 exceptions.

exbluesbreaker:
[...]

PR #2064, opened 2025-03-14
context: Add exception handling

exbluesbreaker:
Related to https://github.com/f3d-app/f3d/issues/1727

### DIFF
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -936,29 +936,52 @@ int F3DStarter::Start(int argc, char** argv)
     bool offscreen = !this->Internals->AppOptions.Reference.empty() ||
       !this->Internals->AppOptions.Output.empty() || this->Internals->AppOptions.BindingsList;
 
-    if (this->Internals->AppOptions.RenderingBackend == "egl")
+    try
     {
-      this->Internals->Engine = std::make_unique<f3d::engine>(f3d::engine::createEGL());
-    }
-    else if (this->Internals->AppOptions.RenderingBackend == "osmesa")
-    {
-      this->Internals->Engine = std::make_unique<f3d::engine>(f3d::engine::createOSMesa());
-    }
-    else if (this->Internals->AppOptions.RenderingBackend == "glx")
-    {
-      this->Internals->Engine = std::make_unique<f3d::engine>(f3d::engine::createGLX(offscreen));
-    }
-    else if (this->Internals->AppOptions.RenderingBackend == "wgl")
-    {
-      this->Internals->Engine = std::make_unique<f3d::engine>(f3d::engine::createWGL(offscreen));
-    }
-    else
-    {
-      if (this->Internals->AppOptions.RenderingBackend != "auto")
+      if (this->Internals->AppOptions.RenderingBackend == "egl")
       {
-        f3d::log::warn("--rendering-backend value is invalid, falling back to \"auto\"");
+        this->Internals->Engine = std::make_unique<f3d::engine>(f3d::engine::createEGL());
       }
-      this->Internals->Engine = std::make_unique<f3d::engine>(f3d::engine::create(offscreen));
+      else if (this->Internals->AppOptions.RenderingBackend == "osmesa")
+      {
+        this->Internals->Engine = std::make_unique<f3d::engine>(f3d::engine::createOSMesa());
+      }
+      else if (this->Internals->AppOptions.RenderingBackend == "glx")
+      {
+        this->Internals->Engine = std::make_unique<f3d::engine>(f3d::engine::createGLX(offscreen));
+      }
+      else if (this->Internals->AppOptions.RenderingBackend == "wgl")
+      {
+        this->Internals->Engine = std::make_unique<f3d::engine>(f3d::engine::createWGL(offscreen));
+      }
+      else
+      {
+        if (this->Internals->AppOptions.RenderingBackend != "auto")
+        {
+          f3d::log::warn("--rendering-backend value is invalid, falling back to \"auto\"");
+        }
+        this->Internals->Engine = std::make_unique<f3d::engine>(f3d::engine::create(offscreen));
+      }
+    }
+    catch (const f3d::context::loading_exception& ex)
+    {
+      f3d::log::error("Could not load graphic library: ", ex.what());
+      return EXIT_FAILURE;
+    }
+    catch (const f3d::context::symbol_exception& ex)
+    {
+      f3d::log::error("Could not find needed symbol in graphic library: ", ex.what());
+      return EXIT_FAILURE;
+    }
+    catch (const f3d::engine::no_window_exception& ex)
+    {
+      f3d::log::error("Could not create the window: ", ex.what());
+      return EXIT_FAILURE;
+    }
+    catch (const f3d::engine::cache_exception& ex)
+    {
+      f3d::log::error("Could not use default cache directory: ", ex.what());
+      return EXIT_FAILURE;
     }
 
     this->ResetWindowName();
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -968,6 +968,11 @@ int F3DStarter::Start(int argc, char** argv)
       f3d::log::error("Could not load graphic library: ", ex.what());
       return EXIT_FAILURE;
     }
+    catch (const f3d::context::symbol_exception& ex)
+    {
+      f3d::log::error("Could not find needed symbol in graphic library: ", ex.what());
+      return EXIT_FAILURE;
+    }
     catch (const f3d::engine::no_window_exception& ex)
     {
       f3d::log::error("Could not create the window: ", ex.what());
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -968,11 +968,6 @@ int F3DStarter::Start(int argc, char** argv)
       f3d::log::error("Could not load graphic library: ", ex.what());
       return EXIT_FAILURE;
     }
-    catch (const f3d::context::symbol_exception& ex)
-    {
-      f3d::log::error("Could not find needed symbol in graphic library: ", ex.what());
-      return EXIT_FAILURE;
-    }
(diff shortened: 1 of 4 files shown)


exbluesbreaker:
I do not know yet how to test all of the exceptions, I know there are already existing test case for not supported GL wrappers, but I am not sure how to test other exception being returned in f3d application. I saw some fabricated code in unit tests, only.

mwestphal:
Some of these are already tested:

 - TestRenderingBackendWGLFailure
 - TestRenderingBackendGLXNoDisplay

Lets see what coverage says and we will see what we need to add

application/F3DStarter.cxx
```diff
@@ -936,29 +936,52 @@ int F3DStarter::Start(int argc, char** argv)
...
-    else
+    catch (const f3d::engine::cache_exception& ex)
     {
-      if (this->Internals->AppOptions.RenderingBackend != "auto")
-      {
-        f3d::log::warn("--rendering-backend value is invalid, falling back to \"auto\"");
-      }
-      this->Internals->Engine = std::make_unique<f3d::engine>(f3d::engine::create(offscreen));
+      f3d::log::error("Could not use default cache directory: ", ex.what());
+      return EXIT_FAILURE;
```
mwestphal, line 984: I think a test like this would cover it:

`f3d_test(NAME TestInvalidCache ENV "HOME=" REGEX "Could not use default cache directory" NO_BASELINE)`

application/F3DStarter.cxx
```diff
@@ -936,29 +936,52 @@ int F3DStarter::Start(int argc, char** argv)
...
-      this->Internals->Engine = std::make_unique<f3d::engine>(f3d::engine::createOSMesa());
+      f3d::log::error("Could not load graphic library: ", ex.what());
+      return EXIT_FAILURE;
     }
-    else if (this->Internals->AppOptions.RenderingBackend == "glx")
+    catch (const f3d::context::symbol_exception& ex)
     {
-      this->Internals->Engine = std::make_unique<f3d::engine>(f3d::engine::createGLX(offscreen));
+      f3d::log::error("Could not find needed symbol in graphic library: ", ex.what());
+      return EXIT_FAILURE;
```
mwestphal, line 974: I wonder which test cover this one ? Any ideas @exbluesbreaker ?

application/F3DStarter.cxx
exbluesbreaker: > I think a test like this would cover it:
> 
> `f3d_test(NAME TestInvalidCache ENV "HOME=" REGEX "Could not use default cache directory" NO_BASELINE)`

I actually catch a memory leak with this test. This is asan traces:

```
555: Could not use default cache directory: Could not setup cache, please set the appropriate environment variable
555: 
555: =================================================================
555: ==21824==ERROR: LeakSanitizer: detected memory leaks
555: 
555: Direct leak of 32 byte(s) in 1 object(s) allocated from:
555:     #0 0x7ffff78fe548 in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:95
555:     #1 0x7ffff6ed1d47 in f3d::engine::engine(std::optional<f3d::window::Type> const&, bool, std::function<void (*(char const*))()> const&) /home/bluesbreaker/Dev/contrib_repos/f3d/library/src/engine.cxx:50
555:     #2 0x7ffff6ed29c3 in f3d::engine::create(bool) /home/bluesbreaker/Dev/contrib_repos/f3d/library/src/engine.cxx:113
555:     #3 0x55555578403e in F3DStarter::Start(int, char**) /home/bluesbreaker/Dev/contrib_repos/f3d/application/F3DStarter.cxx:963
555:     #4 0x5555557ef7dc in main /home/bluesbreaker/Dev/contrib_repos/f3d/application/main.cxx:13
555:     #5 0x7ffff642a1c9 in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
555:     #6 0x7ffff642a28a in __libc_start_main_impl ../csu/libc-start.c:360
555:     #7 0x55555560ebc4 in _start (/home/bluesbreaker/Dev/contrib_repos/f3d/build/bin/f3d+0xbabc4) (BuildId: 499edc7daeecb48f608acea2f6a45ca584e14aa5)
555: 
555: SUMMARY: AddressSanitizer: 32 byte(s) leaked in 1 allocation(s).
```

As I understood it is related to this really weird to passing new in constructor initializer:

```
engine::engine(
  const std::optional<window::Type>& windowType, bool offscreen, const context::function& loader)
  : Internals(new engine::internals)
```
I guess the right thing without reworking a lot is to do delete on  throwing exception, though it is ugly and smart_pointer-like approach is better.

application/F3DStarter.cxx
exbluesbreaker: I'm not entirely sure, but based on my understanding and the context, it seems that it essentially loads the symbol for the basic "findProc" symbol in the library. This is nearly impossible to retrieve without missing the whole library, unless you create a fake library specifically for testing purposes. 

In this particular case, loading and symbol exceptions appear to be closely linked, as it checks for the most fundamental symbol in the library.

application/F3DStarter.cxx
mwestphal: > I guess the right thing without reworking a lot is to do delete on throwing exception, though it is ugly and smart_pointer-like approach is better.

Indedd. Both solutions are okay but a std::unique_ptr on the Internals should fix this easily I think

application/F3DStarter.cxx
mwestphal: Yes I agree, however codecov report this line to be covered, and I wonder how is that possible.

exbluesbreaker: I guess I should remove this exception handling because in this place only loading of library could fail, missing symbol in library could be simulated only by "fabricated" library, it can be raised at some other path working with symbol load when it tries to load some speicfic functions which could be missing by some reason.

I will try to fix it, add memory leak fix and also add test case as you proposed.

exbluesbreaker:
> You are modifying libf3d public API! ⚠️Please update bindings accordingly⚠️! You can find them in their respective directories: `python`, `java`, `webassembly`.

Oops

exbluesbreaker:
Can somebody have a look about bindings. I changed `Internals` field in engine from raw to unique_ptr, it doesn't seem it is directly provided in bindings

exbluesbreaker:
Ok, Windows also complains about exporting now, I guess I will just went back to raw pointer and fix allocation.

application/F3DStarter.cxx
mwestphal: I'm not following why you should not catch this. Indeed, it may not be testable but if someones as a faulty lib this definitely can happen.

mwestphal: Here is how I was able to reach this:

```
 sudo mv /usr/lib/libEGL.so.1.1.0 /usr/lib/libEGL.so.1.1.0.old
 sudo cp /usr/lib/libtwhatever.so /usr/lib/libEGL.so.1.1.0
 ./bin/f3d --rendering-backend=egl ../src/testing/data/dragon.vtu --output=a.png
 ```
 
```
terminate called after throwing an instance of 'f3d::context::symbol_exception'
  what():  Cannot find eglGetProcAddress symbol
```

Of course we cannot test like this in CI but codecov do report this catch as reached and I dont understand why.

Comment (exbluesbreaker) [reply application/F3DStarter.cxx]: Yes, it definitely can cause exception in such situation, I could add catching this exception back, but then we need to somehow deal with having coverage.

Comment (mwestphal) [reply application/F3DStarter.cxx]: its ok to not cover something if we cant. Coverage is only a metric :)

Please indeed add it back.

but the problem is that it is reported as covered by codecov and we need to understand why.

File: application/testing/CMakeLists.txt [hunk 3]
```diff
@@ -1136,6 +1136,9 @@ f3d_test(NAME TestBindingsList ARGS --list-bindings REGEXP "Any.5        Toggle
 # Test rendering backends
 # For some reason the sanitizer detects leaks because of EGL and OSMesa
 f3d_test(NAME TestRenderingBackendAuto DATA cow.vtp RENDERING_BACKEND auto)
+if(NOT WIN32)
+  f3d_test(NAME TestInvalidCache ENV "HOME=" REGEXP "Could not use default cache directory" NO_BASELINE)
```
mwestphal, line 1140: this test should not be put here, lets put it with other linux specific tests, eg before `TestHOMEOutput`

application/testing/CMakeLists.txt
```diff
@@ -1190,6 +1187,7 @@ set_tests_properties(f3d::TestNoFileFileNameTemplate PROPERTIES PASS_REGULAR_EXP
 set_tests_properties(f3d::TestNoFileFileNameTemplate PROPERTIES ENVIRONMENT "CTEST_F3D_NO_DATA_FORCE_RENDER=1")
 
 if(NOT WIN32)
+  f3d_test(NAME TestInvalidCache ENV "HOME=" REGEXP "Could not use default cache directory" NO_BASELINE)
```
mwestphal, line 1190: lets add an empty line after the test, clearer this way

application/F3DStarter.cxx
mwestphal: nevermind, it is not reported as covered, its just ignored by codecov for some reason. no worries.

application/F3DStarter.cxx
mwestphal: Somehow codecov still reports it as not covered but we can be sure it is. codecov bug I guess ? Lets ignore.

mwestphal (approved):

mwestphal (approved):
Approved but lets give some time for @Meakk to review

application/F3DStarter.cxx
exbluesbreaker: > Somehow codecov still reports it as not covered but we can be sure it is. codecov bug I guess ? Lets ignore.

Yeah, test case for cache exception seems passing.

mwestphal:
Thanks for the contribution @exbluesbreaker !

PR #2065, opened 2025-03-15
Improve command exception handling

mwestphal:
- Fix interactor::triggerCommand to actually return true on success
- Expose interactor::invalid_args_exception publicly and update usages
- Use it in application instead of using std::runtime_error
- Catch command_runtime_exception in appropriate places
- Issues found by running tests without the catch all in main.cxx


library/src/interactor_impl.cxx
```diff
@@ -538,7 +538,17 @@ class interactor_impl::internals
...
+      try
+      {
+        // XXX: Ignore the boolean return of triggerCommand,
+        // error is already logged by triggerCommand
+        this->Interactor.triggerCommand(this->CommandBuffer.value());
+      }
+      catch (const f3d::interactor::command_runtime_exception& ex)
+      {
+        log::error(
+          "Interaction: error running command: \"" + this->CommandBuffer.value() + "\": " + ex.what());
```
mwestphal, line 550: technically reachable in a SDK test but a pain to implement ^^

library/src/interactor_impl.cxx
```diff
@@ -538,7 +538,17 @@ class interactor_impl::internals
...
-      this->Interactor.triggerCommand(this->CommandBuffer.value());
+      try
+      {
+        // XXX: Ignore the boolean return of triggerCommand,
+        // error is already logged by triggerCommand
+        this->Interactor.triggerCommand(this->CommandBuffer.value());
+      }
+      catch (const f3d::interactor::command_runtime_exception& ex)
+      {
+        log::error("Interaction: error running command: \"" + this->CommandBuffer.value() +
```
Meakk, line 549: Looks like those lines are not covered

library/src/interactor_impl.cxx
```diff
@@ -622,8 +632,8 @@ interactor& interactor_impl::initCommands()
   {
     if (args.size() != expectedSize)
     {
-      throw interactor_impl::invalid_args_exception(std::string("Command: ") +
-        std::string(actionName) + " is expecting " + std::to_string(expectedSize) + " arguments");
+      throw interactor::invalid_args_exception(std::string("Command: ") + std::string(actionName) +
```
Meakk, line 635: Do we even need to add the scope of the exception here (and everywhere)?

mwestphal: Not technically needed but this is more understandable this way. I've been doing that for exceptions everywhere in the code.

library/testing/TestSDKInteractorCallBack.cxx
```diff
@@ -173,5 +173,16 @@ int TestSDKInteractorCallBack(int argc, char* argv[])
...
+  // Esc;"exception";Enter;Esc
+  if (!inter.playInteraction(
+        std::string(argv[1]) + "recordings/TestSDKInteractorCallBackConsoleException.log"))
+  {
+  return EXIT_FAILURE;
+}
+return EXIT_SUCCESS;
+    return EXIT_FAILURE;
+  }
+
```
mwestphal, line 186: lol wtf

Meakk (approved):

PR #2066, opened 2025-03-15
Colormap parse format

mwestphal:
- Add proper colormap parsing and formating : `val, color, val, color`  is now supported
 - Add documentation
 - Add testing
 - Fix/Improve related doc


Meakk (approved):
Awesome! LGTM

PR #2074, opened 2025-03-15
Updating dmon to 1.3.5

mwestphal:


mwestphal:
Cannot update because of https://github.com/septag/dmon/issues/37

PR #2083, opened 2025-03-22
Fix: Add validation for invalid "reset-" option keys in config files

Denys619:
### Summary

This PR improves the handling of invalid config entries like `"reset-"` with no option name.
Such entries previously resulted in confusing logs like:

`'' option ... does not exists , did you mean 'up'?`

Now, a clear and informative warning is logged, and the entry is safely skipped:

`Invalid option: 'reset-' must be followed by a valid option name, ignoring entry`

### Changes

- Added a simple validation to detect invalid `reset-` keys
- Logs a clear warning message only once, even if config is applied multiple times

### Why

Better UX for users writing config files and prevents misleading log messages.

### Checklist

- [x] PR from fork
- [x] Feature branch created
- [x] CI passed
- [x] Coverage added

### DIFF
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -505,13 +505,7 @@ public:
                 }
                 else
                 {
-                  static bool alreaedyWarned = false;
-                  if (!alreaedyWarned)
-                  {
-                    f3d::log::warn("Invalid option: 'reset-' must be followed by a valid option "
-                                   "name, ignoring entry");
-                    alreaedyWarned = true;
-                  }
+                  f3d::log::warn("Invalid option: 'reset-' must be followed by a valid option name, ignoring entry");
                   continue;
                 }
               }
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -505,7 +505,7 @@ public:
                 }
                 else
                 {
-                  f3d::log::warn("Invalid option: 'reset-' must be followed by a valid option "
+                  f3d::log::warn("Invalid option: 'reset' must be followed by a valid option "
                                  "name, ignoring entry");
                   continue;
                 }
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -496,10 +496,24 @@ public:
               // XXX: Use starts_with once C++20 is supported
               if (libf3dOptionName.rfind("reset-", 0) == 0)
               {
-                reset = true;
-                libf3dOptionName = libf3dOptionName.substr(6);
-                keyForLog = libf3dOptionName;
-                libf3dOptionValue = "reset";
+                if (libf3dOptionName.size() > 6) 
+                {
+                  reset = true;
+                  libf3dOptionName = libf3dOptionName.substr(6);
+                  keyForLog = libf3dOptionName;
+                  libf3dOptionValue = "reset";
+                }
+                else
+                {
+                  static bool alreaedyWarned = false;
+                  if (!alreaedyWarned)
+                  {
+                    f3d::log::warn(
+                        "Invalid option: 'reset-' must be followed by a valid option name, ignoring entry");
+                    alreaedyWarned = true;
+                  }
+                  continue;
+                }
               }
 
               // Handle reader options
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -505,7 +505,8 @@ public:
                 }
                 else
                 {
-                  f3d::log::warn("Invalid option: 'reset-' must be followed by a valid option name, ignoring entry");
+                  f3d::log::warn("Invalid option: 'reset-' must be followed by a valid option "
+                                 "name, ignoring entry");
                   continue;
                 }
               }
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -508,8 +508,8 @@ public:
                   static bool alreaedyWarned = false;
                   if (!alreaedyWarned)
                   {
-                    f3d::log::warn(
-                        "Invalid option: 'reset-' must be followed by a valid option name, ignoring entry");
+                    f3d::log::warn("Invalid option: 'reset-' must be followed by a valid option "
+                                   "name, ignoring entry");
                     alreaedyWarned = true;
                   }
                   continue;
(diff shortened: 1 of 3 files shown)


application/F3DStarter.cxx
```diff
@@ -496,10 +496,24 @@ class F3DStarter::F3DInternals
...
+                if (libf3dOptionName.size() > 6)
+                {
+                  reset = true;
+                  libf3dOptionName = libf3dOptionName.substr(6);
+                  keyForLog = libf3dOptionName;
+                  libf3dOptionValue = "reset";
+                }
+                else
+                {
+                  static bool alreaedyWarned = false;
```
mwestphal, line 508: lets not use static.

mwestphal: tbh I dont think you need the alreadyWarned logic at all.

mwestphal (changes requested):
changes needed

application/F3DStarter.cxx
Denys619: Hi @mwestphal, thanks the feedback!

You're right — using `static` is not ideal.

Just to clarify: without this logic, the warning is printed. I added the flag mainly to avoid that duplicate message.

image

That said, if duplicate warnings are not a problem from a UX standpoint, I'm happy to remove the logic altogether — no issue there.

If we do want to keep the "log once" behavior but avoid `static`, I can move the flag to a more appropriate scope, for example:
- using a local `std::unordered_set<std::string>` in `UpdateOptions()` to track warned keys for this session,
- or storing such flags in a field inside `F3DStarter::Internals`, depending on what's preferred.

Let me know how you'd like to proceed, and I’ll adjust accordingly!

mwestphal: Its ok to have the warning two times.

application/testing/CMakeLists.txt
```diff
@@ -1097,6 +1097,10 @@ f3d_test(NAME TestNonExistentConfigFileStem DATA cow.vtp CONFIG "dummy" REGEXP "
 # Test invalid config file
 f3d_test(NAME TestInvalidConfigFile DATA cow.vtp CONFIG ${F3D_SOURCE_DIR}/testing/configs/invalid.json REGEXP "Unable to parse the configuration file" NO_BASELINE)
 
+# Test invalid reset key in config file
+f3d_test(NAME TestInvalidResetKey DATA cow.vtp CONFIG ${F3D_SOURCE_DIR}/testing/configs/invalid_reset.json REGEXP "Invalid option: 'reset-' must be followed by a valid option name, ignoring entry" NO_BASELINE)
+
+
```
mwestphal, line 1103: you can remove this line

application/testing/CMakeLists.txt
```diff
@@ -1097,6 +1097,10 @@ f3d_test(NAME TestNonExistentConfigFileStem DATA cow.vtp CONFIG "dummy" REGEXP "
 # Test invalid config file
 f3d_test(NAME TestInvalidConfigFile DATA cow.vtp CONFIG ${F3D_SOURCE_DIR}/testing/configs/invalid.json REGEXP "Unable to parse the configuration file" NO_BASELINE)
 
+# Test invalid reset key in config file
+f3d_test(NAME TestInvalidResetKey DATA cow.vtp CONFIG ${F3D_SOURCE_DIR}/testing/configs/invalid_reset.json REGEXP "Invalid option: 'reset-' must be followed by a valid option name, ignoring entry" NO_BASELINE)
```
mwestphal, line 1101: instead of relying on a config file, it would be better to test with `--reset-` in the command line.

```suggestion
f3d_test(NAME TestInvalidResetOptions DATA cow.vtp ARGS `--reset-` REGEXP "Invalid option: 'reset-' must be followed by a valid option name, ignoring entry" NO_BASELINE)
```

This way you can drop the config file.

Denys619: I reworked the test to use --reset- as suggested, but it doesn’t reach UpdateOptions() — it's rejected earlier by the CLI parser. So either we drop the test or keep it config-based. Let me know what works best for you!

application/testing/CMakeLists.txt
```diff
@@ -1097,6 +1097,9 @@ f3d_test(NAME TestNonExistentConfigFileStem DATA cow.vtp CONFIG "dummy" REGEXP "
 # Test invalid config file
 f3d_test(NAME TestInvalidConfigFile DATA cow.vtp CONFIG ${F3D_SOURCE_DIR}/testing/configs/invalid.json REGEXP "Unable to parse the configuration file" NO_BASELINE)
 
+# Test invalid reset key in config file
+f3d_test(NAME TestInvalidResetOptions DATA cow.vtp ARGS `--reset-` REGEXP "Invalid option: 'reset-' must be followed by a valid option name, ignoring entry" NO_BASELINE)
```
mwestphal, line 1101:
```suggestion
f3d_test(NAME TestInvalidResetOptions DATA cow.vtp ARGS `--reset=` REGEXP "Invalid option: 'reset-' must be followed by a valid option name, ignoring entry" NO_BASELINE)
```

Denys619: The same result

mwestphal: not the same. You can use `-VV` to look at the command line and figure out how to fix it :)

application/testing/CMakeLists.txt
```diff
@@ -1097,6 +1097,9 @@ f3d_test(NAME TestNonExistentConfigFileStem DATA cow.vtp CONFIG "dummy" REGEXP "
 # Test invalid config file
 f3d_test(NAME TestInvalidConfigFile DATA cow.vtp CONFIG ${F3D_SOURCE_DIR}/testing/configs/invalid.json REGEXP "Unable to parse the configuration file" NO_BASELINE)
 
+# Test invalid reset key in config file
+f3d_test(NAME TestInvalidResetOptions DATA cow.vtp ARGS `--reset=` REGEXP "Invalid option: 'reset' must be followed by a valid option name, ignoring entry" NO_BASELINE)
```
mwestphal, line 1101:
```suggestion
f3d_test(NAME TestInvalidResetOptions DATA cow.vtp ARGS --reset= REGEXP "Invalid option: 'reset' must be followed by a valid option name, ignoring entry" NO_BASELINE)
```

mwestphal (approved):

mwestphal:
Thanks for your contribution @Denys619 !

Denys619:
@mwestphal I’m still new to open source and C++, so I truly appreciate all the guidance and the time you took to help me through this.

Looking forward to learning more and contributing again!

PR #2094, opened 2025-03-25
Fix JSON key quotes in CONFIGURATION_FILE.md

### LINKED ISSUE #2087: JSON files keys must be quoted
**Describe the bug**
In the documentation, the configuration file example doesn't have quoted keys, see https://f3d.app/doc/user/CONFIGURATION_FILE.html
While some JSON parser might support it, the JSON standard requires keys to be quoted. See https://stackoverflow.com/questions/949449/do-the-json-keys-have-to-be-surrounded-by-quotes

**Expected fix**
The file `doc/user/CONFIGURATION_FILE.md` must be fixed, and the F3D source code must be scanned to make sure that's the only location where we do have this issue.

Denys619:
### Description

This PR updates the JSON examples in `doc/user/CONFIGURATION_FILE.md` to use double quotes around all keys, as required by the official JSON specification (RFC 8259).

Prettier does not modify JSON in Markdown files properly.
All examples were validated using [jsonlint.com](https://jsonlint.com/) to ensure correctness.

### Checklist

- [x] Verified that similar issues do not occur in other documentation files.
- [x] This change affects documentation only.

resolves #2087

### DIFF
--- a/doc/user/CONFIGURATION_FILE.md
+++ b/doc/user/CONFIGURATION_FILE.md
@@ -20,40 +20,40 @@ A typical config file with options may look like this:
 ```javascript
 [
   {
-    options: {
+    "options": {
       "background-color": "0.7,0.7,0.7",
-      color: "0.5,0.1,0.1",
+      "color": "0.5,0.1,0.1",
       "anti-aliasing": true,
-      timer: true,
-      progress: true,
-      axis: true,
-      bar: true,
-      roughness: 0.2,
-      grid: true,
-      "scalar-coloring": true,
-    },
+      "timer": true,
+      "progress": true,
+      "axis": true,
+      "bar": true,
+      "roughness": 0.2,
+      "grid": true,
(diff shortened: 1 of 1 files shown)


mwestphal:
please configure prettier to use the correct formating:

https://stackoverflow.com/questions/55430906/prettier-single-quote-for-javascript-and-json-double-quote-for-html-sass-and-c

Denys619:
Oh God) The issue was caused by the code block being marked as `javascript` instead of `json`, which prevented Prettier from applying the correct formatting rules. I’ve updated the code block language, and Prettier now formats it properly without additional configuration.

Denys619:
@mwestphal please check)

mwestphal:
Nice find!

mwestphal (approved):

mwestphal:
Thanks for this fast fix @Denys619 !

PR #2116, opened 2025-04-02
doc: Initial commit to rework docs

exbluesbreaker:
Initial change for potential rework.

`commands_doc.h` file is added. It contains:

- `struct InteractiveCommand` which is new alternative to previously only `std::function`, now it is `std::function` to do the command and documentation (strings describing function and arguments).

- `struct InteractiveCommandDoc` templated by `N` (number of arguments), it basically stores all constant fields of documentation for every function (basically all `std::string` data associated). Probably some way to handle variable number should be added somehow later on (not by template but by naming argument in doc in some specific way). 

- List of constexpr objects of `struct InteractiveCommandDoc` for every function (except for one defined in tests, which are defined on the spot).
- templated `makeCommand` function to create `struct InteractiveCommand`  given struct InteractiveCommandDoc`  and `std::function` callback for the command

And I basically replaced storing callbacks only in map of commands into storing  `struct InteractiveCommand`  objects (doc+callback).

I used chatgpt for some code genaration (to make boilerplate alternatives to old addCommand, needs to be checked)

### DIFF
--- a/library/src/interactor_impl.cxx
+++ b/library/src/interactor_impl.cxx
@@ -906,8 +906,7 @@ interactor& interactor_impl::initCommands()
 //----------------------------------------------------------------------------
 interactor& interactor_impl::addCommand(InteractiveCommand command)
 {
-  const auto [it, success] =
-    this->Internals->Commands.insert({ command.Name, std::move(command) });
+  const auto [it, success] = this->Internals->Commands.insert({ command.Name, std::move(command) });
   if (!success)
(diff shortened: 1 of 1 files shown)


exbluesbreaker:
Let's discuss in general is it good/bad approach before actually going for tests/other checks which will fail

mwestphal:
superseeded by https://github.com/f3d-app/f3d/pull/2434

PR #2123, opened 2025-04-07
libf3d: Fix a crash with ui.axis with external window

mwestphal:


Meakk (approved):

library/src/window_impl.cxx
```diff
@@ -378,9 +381,13 @@ void window_impl::UpdateDynamicOptions()
   renderer->UpdateLights();
 
   const options& opt = this->Internals->Options;
-  renderer->ShowAxis(opt.ui.axis);
-  renderer->SetUseTrackball(opt.interactor.trackball);
-  renderer->SetInvertZoom(opt.interactor.invert_zoom);
+
+  if (this->Internals->Interactor)
+  {
+    renderer->ShowAxis(opt.ui.axis);
```
Meakk, line 387: should we add a warning if `opt.ui.axis` is true but there is no interactor?

mwestphal: We definitely can, but then isnt there many options with "no effect" where we should warn as well. eg: `ui.filename` without `ui.filename_info` set.

Meakk: ok fair enough

Issue #2125, opened 2025-04-08
F3D triggers memory leak and freezes system when window is resized to minimum


Denys619:
**Description**

When launching F3D without a model and manually resizing the application window down to a minimal size (as small as 1x1 pixel), the window becomes unresponsive. Once I try to maximize it again:

- The content of the window stays completely black.
- The window cannot be closed.
- The system becomes increasingly unresponsive.
- Eventually, the session crashes, and the system restarts the graphical environment (on Linux).

Using htop, I monitored system resource usage while reproducing the issue. Upon shrinking and then maximizing the window:

- Memory usage starts growing rapidly.
- Eventually, F3D consumes 100% of the system RAM.
- This leads to a system-wide hang and session crash, consistent with a memory leak.

https://github.com/user-attachments/assets/83cfe614-ef4a-4eb8-8195-a9c6b7ff1ea9

**To Reproduce**

1. Launch F3D without a model.
2. Manually resize the window by dragging its edges to the minimum possible size.
3. Maximize the window again.
4. Observe black screen, unresponsiveness, and eventual system hang.

**Expected behavior**

- The window should have a minimum size limit to prevent this.
- Resizing should not result in black screen or freeze.

**System Information:**

- OS: Ubuntu 22.04
- GPU: RX 5700XT

**F3D Information**

F3D 3.0.0

Version: 3.0.0-167-g93a117f8.
Build date: 2025-03-31 11:53:20.
Build system: Linux 64-bits.
Compiler: GNU 13.3.0.
Module ImGui: ON.
Module OpenEXR: OFF.
Module Raytracing: OFF.
VTK version: 9.4.1.

mwestphal:
Its in the DropZone code: 

```
Thread 1 "f3d" received signal SIGINT, Interrupt.
0x00007ffff777a8a7 in ?? () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff777a8a7 in ?? () from /usr/lib/libc.so.6
#1  0x00007ffff7d9c6d8 in ImDrawList::PrimReserve(int, int) () from /home/glow/dev/f3d/f3d/build/lib/libf3d.so.3
#2  0x00007ffff7d9ccee in ImDrawList::AddPolyline(ImVec2 const*, int, unsigned int, int, float) ()
   from /home/glow/dev/f3d/f3d/build/lib/libf3d.so.3
[...]

PR #2112, opened 2025-03-31
Bug Fix: Visual glitch in drop zone border (missing corner pixel)

Denys619:
This PR fixes a visual artifact in the drop zone rendering, where the bottom-left corner appeared to have a missing pixel due to an imperfect alignment between the horizontal and vertical border lines.

Although all lines were technically rendered, the tick alignment and drawing logic caused a gap to appear at the intersection of the left and bottom lines. To resolve this, i slightly adjusted the tick spacing and line start offset to ensure better visual continuity and proper corner stitching.

related #2102

PR #2126, opened 2025-04-08
Add viewport size guard to drop zone rendering

Denys619:
**Bug**

When the application window is resized to a very small size (e.g. 1x1), 
the drop zone tries to render using invalid dimensions. This causes ImDrawList
functions (like `AddLine`) to behave incorrectly, potentially causing high memory usage,
UI freeze, black screen, or even full system lock-up due to memory exhaustion.

**Fix**

This patch adds a guard in `vtkF3DImguiActor::RenderDropZone()` to skip rendering 
when the viewport size is too small (< 50x50 pixels). This prevents dangerous 
rendering operations under invalid dimensions.

**Result**

- Prevents memory leak and system crash when minimizing window
- Keeps the UI responsive
- Zero impact on normal-sized viewports

---

Let me know if you'd prefer a different threshold size or an alternative approach.

Related #2125


Denys619:
Sorry, I accidentally added other (wrong) commits, I'll fix it now

vtkext/private/module/vtkF3DImguiActor.cxx
```diff
@@ -295,6 +295,12 @@ void vtkF3DImguiActor::RenderDropZone()
   if (this->DropZoneVisible)
   {
     ImGuiViewport* viewport = ImGui::GetMainViewport();
+
+    if (viewport->WorkSize.x < 50 || viewport->WorkSize.y < 50)
```
mwestphal, line 299: I think the guard should be much smaller, like 10x10. I know that the text is already not fitting at that size but I can image someone willing to see the dropzone with smaller window.

vtkext/private/module/vtkF3DImguiActor.cxx
```diff
@@ -295,6 +295,12 @@ void vtkF3DImguiActor::RenderDropZone()
...
+    {
+        return;
```
mwestphal, line 301: looks like you are using tabs here, please use spaces

vtkext/private/module/vtkF3DImguiActor.cxx
```diff
@@ -295,6 +295,12 @@ void vtkF3DImguiActor::RenderDropZone()
   if (this->DropZoneVisible)
   {
     ImGuiViewport* viewport = ImGui::GetMainViewport();
+
```
mwestphal, line 298: Lets add some doc in libf3d/OPTIONS.md

mwestphal (changes requested):
small changes needed

mwestphal:
> Should this be covered by a test?

Yes please, it should be pretty simple to add :)

mwestphal:
Looks like you made some changes to the line drawing logic, could you add some explainer in the PR description about the change ?

Denys619:
Oh i got it) You are about this changes

image

I made these changes back when I was fixing the [lost pixel](https://github.com/f3d-app/f3d/pull/2112). I don't know why these changes are still here))

mwestphal (approved):

mwestphal:
Thanks for the contribution @Denys619 !

PR #2129, opened 2025-04-10
Fix animationManager for dynamic range and single frame

mwestphal:
- Fix animationManager to handle dynamic time ranges in multiple animations
- Properly support single frame animation
- Add tests
- Update single frame animation test baseline where console badge disappears
- Follow up issue: https://github.com/f3d-app/f3d/issues/2130

### DIFF
--- a/library/private/animationManager.h
+++ b/library/private/animationManager.h
@@ -104,10 +104,10 @@ public:
    */
   std::pair<double, double> GetTimeRange();
 
-private:
   animationManager(animationManager const&) = delete;
   void operator=(animationManager const&) = delete;
 
+private:
   /**
    * Update members for animation index from options
    */
--- a/library/private/animationManager.h
+++ b/library/private/animationManager.h
@@ -29,7 +29,7 @@ class window_impl;
 class animationManager
 {
 public:
-  animationManager(const options& options, window_impl& window);
+  animationManager(options& options, window_impl& window);
   ~animationManager() = default;
 
   /**
@@ -60,13 +60,9 @@ public:
 
   /**
    * Cycle onto and play the next available animation
+   * This modifies the scene.animation.index option
    */
   void CycleAnimation();
-
-  /**
-   * Enable only the current animation
-   */
-  void EnableOnlyCurrentAnimation();
 
   /**
    * Get the current animation index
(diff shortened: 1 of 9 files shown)


library/src/animationManager.cxx
```diff
@@ -304,38 +263,92 @@ int animationManager::GetAnimationIndex()
...
+      // Accumulate time ranges
+      this->TimeRange[0] = std::min(timeRange[0], this->TimeRange[0]);
+      this->TimeRange[1] = std::max(timeRange[1], this->TimeRange[1]);
+    }
+  }
+  if (this->TimeRange[0] > this->TimeRange[1])
+  {
+    log::warn("Animation(s) time range delta is invalid: [", this->TimeRange[0], ", ",
+      this->TimeRange[1], "]. Swapping range.");
+    std::swap(this->TimeRange[0], this->TimeRange[1]);
```
mwestphal, line 338: There is no way to cover this at the moment, at least without editing some binary exodus files, which I will not do.

library/src/animationManager.cxx
```diff
@@ -98,50 +97,10 @@ void animationManager::Initialize()
...
-    return;
-  }
-  else
-  {
-    log::debug("Animation(s) time range is: [", this->TimeRange[0], ", ", this->TimeRange[1], "].");
-  }
-  log::debug("");
+  // Reset animation index to an invalid value before updating
+  // TODO: Rework animation index to be a vector of int
+  this->AnimationIndex = -2;
```
mwestphal, line 102: I'm not a big fan of this but I will rework this post release:

mwestphal: https://github.com/f3d-app/f3d/issues/2130

testing/baselines/TestMetaData.png
Meakk, line 1: why has it changed? There's not animation in this file

mwestphal: TestMetaData opens pdiag.vtu, which contains a single frame of animation.

Single frame animation used to cause a warning, something like: "Animation time range incorrect, disabling animation", but it is now a valid time range and single frame animation are properly supported, that means that the console badge does not show up anymore, changing the baseline.

Same for other baselines I have yet to update.

library/private/animationManager.h
```diff
@@ -103,27 +99,31 @@ class animationManager
    */
   bool LoadAtTime(double timeValue);
 
-  animationManager(animationManager const&) = delete;
```
Meakk, line 106: Clang-tidy is suppose to warning about this.  
Read https://discourse.llvm.org/t/why-does-clang-tidy-recommend-deleted-member-functions-should-be-public/46728

Meakk (approved):

library/private/animationManager.h
mwestphal: I've fixed it but no idea no warnings

PR #2131, opened 2025-04-12
Application: Add a f3d-console application on Windows

mwestphal:


application/CMakeLists.txt
```diff
@@ -31,169 +191,16 @@ endif()
...
-  target_link_libraries(f3d PRIVATE "-framework Cocoa")
-endif()
-
-# Windows executable without console
-if(WIN32)
-  if(F3D_WINDOWS_GUI)
-    set_target_properties(f3d PROPERTIES WIN32_EXECUTABLE TRUE)
-  else()
-    # force usage of wWinMainCRTStartup in console mode for Unicode support
-    list(APPEND f3d_link_options_public "/ENTRY:wWinMainCRTStartup")
```
mwestphal, line 190: I removed this, it makes the CMake logic more complex and it doesnt matter for the f3d GUI usecase.

.github/actions/python-ci/action.yml
```diff
@@ -119,7 +119,7 @@ runs:
         -DF3D_MACOS_BUNDLE=OFF;
         -DF3D_MODULE_EXR=ON;
         -DF3D_MODULE_UI=ON;
-        -DF3D_WINDOWS_GUI=OFF" >> $GITHUB_ENV
+        -DF3D_BUILD_APPLICATION=OFF" >> $GITHUB_ENV
```
Meakk, line 122: That was definitely incorrect before

application/CMakeLists.txt
Meakk: Why? It's still useful for displaying unicode characters for `f3d-console.exe`.

mwestphal: Is that ok if I define it all the time ? I'd rather not use option in the cmake function.

.github/actions/python-ci/action.yml
mwestphal: Yeah, I have no idea what the original intent was, looks better like this anyuway

Meakk (approved):
LGTM but I think it would be great to test unicode strings during manual RC testing. This is confusing, and the official `/SUBSYSTEM` documentation doesn't really help.

PR #2133, opened 2025-04-12
Application: Catch parsing exception with app options

mwestphal:


Meakk (approved):
Looks good to me (when styling is fixed)

Issue #2000, opened 2025-02-18
File with multiple actors only show first actor when coloring


mwestphal:
**Describe the bug**
Opening a file providing multiple actors is not working well with coloring

**To Reproduce**

1.  Open the file using `f3d --no-config InterpolationTest.glb `
2.  Press S

![Image](https://github.com/user-attachments/assets/467a4866-f012-4c88-8136-5e9a34afa00f)

**Expected behavior**
All actors should be visible, instead of the first one

exbluesbreaker:
I would like to work on this issue

mwestphal:
Assigned! Its probably an issue in vtkF3DMetaImporter:245

mwestphal:
Hum, I'd have expected this to be more straitgorward, I can take a look if you want.

exbluesbreaker:
I got a bit sidetracked focusing on visibility, but the real issue was that all actors were **being rendered in the same position**—not just one being visible. You were right about the location in the code; it was in the constructor called at this line. I'll create a merge request soon.

exbluesbreaker:
Adding screenshot with fix from pull request

![Image](https://github.com/user-attachments/assets/baf97624-44ef-4afe-a0cf-853bb9012732)

mwestphal:
> I got a bit sidetracked focusing on visibility, but the real issue was that all actors were **being rendered in the same position**—not just one being visible. You were right about the location in the code; it was in the constructor called at this line. I'll create a merge request soon.

Damn, I should have thought of that!

Issue #1901, opened 2025-01-11
Morping and skining doest not work with coloring


mwestphal:
**Describe the bug**
F3D supports coloring on any data, even data is is skinned and morphed, however, in that case, skinning and morphing is ignored

**To Reproduce**
 1. Open the file using `f3d dota/scene.gltf --armature`
 2. Press s

![a](https://github.com/user-attachments/assets/f525c91f-a085-478f-b7a5-67a14740f3a0)

**Expected behavior**
Coloring should not prevent morphing and skining and the model should line up correctly (and support animation)

**System Information:**
 - OS: Any
 - GPU and GPU driver: Any

**F3D Information**
Paste the content of `f3d --version`: 3.0.0-RC2

**Additional context**

Related to https://github.com/f3d-app/f3d/issues/1010

exbluesbreaker:
It seems it is indeed related to https://github.com/f3d-app/f3d/pull/2151. Though fully correct implementation is still under question. I tried ShallowCopy, with it I noticed skeleton starts being always with spherical joints. Not sure if everything is right.

How it looked without ShallowCopy:

![Image](https://github.com/user-attachments/assets/37588e24-ad2d-48d3-8ea9-99a65dc4ede3)

One of the many instances of the scene after I iterated using "s" with ShallowCopy of original actor:

![Image](https://github.com/user-attachments/assets/52941b0f-477f-4281-8241-12facee6286a)

PR #2151, opened 2025-04-15
MetaImporter: copy user matrix from original actor

exbluesbreaker:
Related to https://github.com/f3d-app/f3d/issues/2000.
It needs to be checked if something else about copy from original actor is missing there as well.

### DIFF
--- a/vtkext/private/module/vtkF3DMetaImporter.h
+++ b/vtkext/private/module/vtkF3DMetaImporter.h
@@ -68,6 +68,7 @@ public:
       this->Actor->GetProperty()->SetRoughness(0.3);
       this->Actor->GetProperty()->SetInterpolationToPBR();
       this->Actor->SetMapper(this->Mapper);
+      this->Actor->SetUserMatrix(originalActor->GetUserMatrix());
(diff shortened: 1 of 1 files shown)


exbluesbreaker:
@mwestphal It seems now it looks like it should be. But maybe something else is missing in regard of original actor state.

mwestphal:
Thats interesting! Does this relates to https://github.com/f3d-app/f3d/issues/1901 as well maybe ?

mwestphal:
Maybe you could just plainly ShallowCopy the original actor into the actor ?

exbluesbreaker:
> Maybe you could just plainly ShallowCopy the original actor into the actor ?
> 
> And does animation works as expected ?

It seems not, I will work on it.

exbluesbreaker:
> Maybe you could just plainly ShallowCopy the original actor into the actor ?
> 
> And does animation works as expected ?

I tried to do ShallowCopy at the beginning of consturctor, but now it is all red and animation still doesn't work in color

image

mwestphal:
Superseeded by https://github.com/f3d-app/f3d/pull/2172

PR #2166, opened 2025-04-19
CI: Update version timestamp and add warning on change

mwestphal:
Docker run: https://github.com/f3d-app/f3d-docker-images/actions/runs/14546311818


.github/workflows/version-timestamp-warn.yml
```diff
@@ -0,0 +1,16 @@
+name: Public API Warning
```
Meakk, line 1: Change name

Meakk (approved):

PR #2169, opened 2025-04-20
QuakeMDL: Set Normals array properly

### LINKED ISSUE #2153: F3DWeb is unable to open .mdl files
**Describe the bug**
With the addition of QuakeMDL in F3D, F3DWeb should be able to open .mdl file, but it came up empty

**To Reproduce**

1.  Navigate to https://f3d.app/web/
2.  Open a file -> All files -> any .mdl file
3.  F3DWeb is empty

**Expected behavior**
F3DWeb should be able to open a mdl file this way

**F3D Information**
master

Lgt2x:
In QuakeMDLImporter, the normals array was allocated, but this did not change its actual size. We now set the array size directly, which allocates memory just like it did before. Having a normals array without data confused the web renderer which did not display anything (#2153)

Tests have been updated with corrected normals.

Closes #2153

### DIFF
--- a/plugins/native/module/vtkF3DQuakeMDLImporter.cxx
+++ b/plugins/native/module/vtkF3DQuakeMDLImporter.cxx
@@ -488,7 +488,7 @@ vtkIdType vtkF3DQuakeMDLImporter::GetNumberOfAnimations()
 //----------------------------------------------------------------------------
 std::string vtkF3DQuakeMDLImporter::GetAnimationName(vtkIdType animationIndex)
 {
-  assert(animationIndex < this->Internals->AnimationNames.size());
+  assert(animationIndex < static_cast<vtkIdType>(this->Internals->AnimationNames.size()));
   assert(animationIndex >= 0);
   return this->Internals->AnimationNames[animationIndex];
 }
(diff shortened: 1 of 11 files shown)


Lgt2x:
Failing tests `TestDefaultConfigFileQuakeMDL` and `TestThumbnailConfigFileQuakeMDL` are disabled on my local build, I'm not sure how to enable them

mwestphal:
> Failing tests `TestDefaultConfigFileQuakeMDL` and `TestThumbnailConfigFileQuakeMDL` are disabled on my local build, I'm not sure how to enable them

Set `F3D_TESTING_ENABLE_LONG_TIMEOUT_TESTS` to ON, as documented here: https://f3d.app/doc/dev/TESTING.html

mwestphal (changes requested):
Looks good but missing some new baselines.

Lgt2x:
Thanks, added missing baselines.

mwestphal:
With correct normals it looks quite dark. I wonder if we should increase the default lights intensity for mdl, what do you think ?

Lgt2x:
> With correct normals it looks quite dark. I wonder if we should increase the default lights intensity for mdl, what do you think ?

I think it does make sense, Quake models seem to generally be on the darker side

Meakk (changes requested):
I think the dark issue is coming from the `F3DMDLNormalVectors` table being already stored as floating values between 0 and 1, but it's divided by `255` in the code. I'd try to remove that division to see if it's better.

Meakk:
Also, if we do have smooth shading now with normals, maybe it would be good to bring back the specular part by commenting out the `SetBaseIOR()` call.

Lgt2x:
> I think the dark issue is coming from the `F3DMDLNormalVectors` table being already stored as floating values between 0 and 1, but it's divided by `255` in the code. I'd try to remove that division to see if it's better.

I'm afraid that does not change anything visually

> Also, if we do have smooth shading now with normals, maybe it would be good to bring back the specular part by commenting out the `SetBaseIOR()` call.

That makes a slight difference (enough to make the test fail, barely), it's visually a little lighter.

BaseIOR default
TestQuakeMDL

BaseIOR = 1
TestQuakeMDL

Meakk:
Superseded by https://github.com/f3d-app/f3d/pull/2183

PR #2170, opened 2025-04-20
Add support for VTKHDF

Lgt2x:
### Describe your changes

[VTKHDF](https://docs.vtk.org/en/latest/design_documents/VTKFileFormats.html#vtkhdf-file-format) is the new-generation HDF5-based file format for VTK. F3D can read both legacy and XML VTK format, so it would make sense to also add support for VTKHDF.

VTKHDF is still under development, so all VTK 9.X versions may not have full support.

Test data generated from ParaView 5.13.3.

- [x] I have added [documentation](https://f3d.app/doc/dev/TESTING.html) for new features
- [x] If it is a modifying the default versions, I have updated timestamp

### Continuous integration

- [x] Style checks
- [x] Fast CI
- [x] Coverage cached CI
- [x] Analysis cached CI
- [x] WASM CI
- [x] Android CI
- [x] macOS Intel cached CI
- [x] macOS ARM cached CI
- [x] Windows cached CI
- [x] Linux cached CI
- [x] Other cached CI


Lgt2x:
Well I guess pipelines should somehow rebuild their VTK with the added option

mwestphal:
> Well I guess pipelines should somehow rebuild their VTK with the added option

Indeed, you need to increase cache value in `cache-vtk` step in `.github/actions/vtk-install-dep/action.yml `
I'll take care of the android/wasm stuff.

testing/data/blob.vtkhdf
mwestphal, line 1: I supose this is an unstructured grid ? What happens if you load a partitioned data or a partioned dataset collection ? What about a HTG ?

Its ok to not support them, but we should not crash on them.

Lgt2x:
> Indeed, you need to increase cache value in `cache-vtk` step in `.github/actions/vtk-install-dep/action.yml `

We may want to document that somewhere

Lgt2x:
mh CI is still not happy about it

testing/data/blob.vtkhdf
mwestphal: Perfect, thanks for checking. Maybe we should add support for pdc but its not for today.

mwestphal:
> > Indeed, you need to increase cache value in `cache-vtk` step in `.github/actions/vtk-install-dep/action.yml `
> 
> We may want to document that somewhere

The CI workflow is basically not documented from a dev POV, There are more critical things to do imo.

mwestphal:
> mh CI is still not happy about it

Our CI tests against VTK 9.2.6, which is not compatible with VTKHDF.
Lets not add VTKHDF reader when VTK_VERSION is <= 9.3.0

If you can find the exact data it started working because you know the MR lets do that, but its not critical and you can cut off at 9.3.0

Of course you need to put that in the native plugin CMakeLists.txt and in the testing logic as well.

plugins/native/CMakeLists.txt
```diff
@@ -176,6 +176,15 @@ f3d_plugin_declare_reader(
   FORMAT_DESCRIPTION "VTK XML MultiBlock"
 )
 
+f3d_plugin_declare_reader(
+  NAME VTKHDF
+  SCORE 99
+  EXTENSIONS vtkhdf
+  MIMETYPES application/vnd.vtkhdf
```
mwestphal, line 183: missing the mimetype xml declaration

plugins/native/CMakeLists.txt
```diff
@@ -176,6 +176,15 @@ f3d_plugin_declare_reader(
   FORMAT_DESCRIPTION "VTK XML MultiBlock"
 )
 
+f3d_plugin_declare_reader(
+  NAME VTKHDF
```
mwestphal, line 180: add doc in SUPPORTED_FORMATS.md

mwestphal:
Since hdf5 cannot be enabled on android, I think this should go into its own `hdf` or `vtkhdf` plugin. What do you think @Meakk ?

Meakk:
> Since hdf5 cannot be enabled on android, I think this should go into its own `hdf` or `vtkhdf` plugin. What do you think @Meakk ?

I think we moved exodus to its own plugin because of hdf. Maybe we can put it there and rename the `exodus` plugin to `extra` or `hdf`?

mwestphal:
That'd be possible indeed, but we need to deprecate the exodus plugin manually in the libf3d. That is doable though.

Lgt2x:
> I think we moved exodus to its own plugin because of hdf. Maybe we can put it there and rename the exodus plugin to extra or hdf?

Exodus does not depend on HDF as far as I'm aware

Meakk:
> Exodus does not depend on HDF as far as I'm aware

I think it does: https://github.com/Kitware/VTK/blob/b4f4b646ff1f4761d2003fec94a08d0debf00c7e/ThirdParty/exodusII/vtk.module#L10

Lgt2x:
Oh right, I can make a `hdf` plugin then. I'm not familiar with f3d's deprecation mechanism, do you have an example?

mwestphal:
>  I'm not familiar with f3d's deprecation mechanism, do you have an example?

There is no dedicated deprecation mechanism for option _values_, so it will have to be handled manually.

The best way is probably to handle that in `engine::loadPlugin` with a deprecated name to new name conversion and a warn.

That can wait, you can start by creating the plugin first and add support for vtkHDFReader in it, then move exodus into it.

Lgt2x:
Moved the reader to a `hdf` plugin as suggested, exodus reader to be moved there

mwestphal:
That was fast ^^, let me know if you need help with the deprecation of the exodus plugin.

Lgt2x:
Deprec and testing done, let me know if it is what you had in mind

doc/user/SUPPORTED_FORMATS.md
```diff
@@ -6,14 +6,15 @@ F3D supports the following file formats:
 | ----------------------------------------- | ---------------------------------------------- | ---------- | -------------------- | --------- |
 | Legacy VTK                                | `.vtk`                                         | No         | No                   | `native`  |
 | VTK XML                                   | `.vtp`, `.vtu`, `.vtr`, `.vti`, `.vts`, `.vtm` | No         | No                   | `native`  |
+| VTKHDF                                    | `.vtkhdf`                                      | No         | No                   | `hdf`     |
```
mwestphal, line 9: I'd assume animated vtkhdf should work ?

library/src/engine.cxx
```diff
@@ -270,6 +270,12 @@ void engine::loadPlugin(const std::string& pathOrName, const std::vector<fs::pat
     return;
   }
 
+  // For easier removal when removing deprecation: F3D_DEPRECATED
+  if (pathOrName == "exodus") {
+    f3d::log::warn("The 'exodus' plugin is deprecated, load 'hdf' instead");
+    return f3d::engine::loadPlugin("hdf");
```
mwestphal, line 276: it may be better to copy pathOrName into a non const local variable and edit it here so that the whole logic can work with `hdf` instead of force loading it ?

testing/data/blob.vtkhdf
mwestphal, line 1: what is the license of this file ? Was it copied from VTK or did you produce it yourself ?

mwestphal (commented):
nice, some small changes needed

Please note this will also require changes in the f3d-superbuild

application/testing/CMakeLists.txt
```diff
@@ -758,11 +760,13 @@ if(F3D_PLUGIN_BUILD_EXODUS)
   f3d_test(NAME TestTimeRangeLessThanZeroNegativeAnimationTime DATA negative_range_animated.e ARGS -s --load-plugins=exodus --animation-time=-3)
```
mwestphal, line 760:
```suggestion
  f3d_test(NAME TestTimeRangeLessThanZeroNegativeAnimationTime DATA negative_range_animated.e ARGS -s --load-plugins=hdf --animation-time=-3)
```

Many other like that in this file

library/src/engine.cxx
Lgt2x: I thought using a recursive call was cleaner than using another temporary variable, but as you wish. I also notice that the call is missing the second parameter

testing/data/blob.vtkhdf
Lgt2x: homemade in ParaView as said in PR description

library/src/engine.cxx
mwestphal: ha, its a recursive call! I missed that. Current impl is fine.

testing/data/blob.vtkhdf
mwestphal: thanks!

doc/user/SUPPORTED_FORMATS.md
Lgt2x: yes, fixed

Lgt2x:
> Please note this will also require changes in the f3d-superbuild

About that, I have 2 possibilities in mind:
 - Use the existing `ENABLE_EXODUS` option to activate the new `hdf` f3d plugin, that does not contain only exodus anymore.
 - Create a new dummy project (`f3dhdf` for instance) that depends on exodus and hdf5 and activates building the plugin

What is the preferred way?

library/src/engine.cxx
```diff
@@ -265,32 +265,40 @@ std::map<std::string, bool> engine::getRenderingBackendList()
...
+
+  if (localPathOrName.empty())
   {
     return;
   }
 
+  // For easier removal when removing deprecation: F3D_DEPRECATED
+  if (localPathOrName == "exodus") {
+    f3d::log::warn("The 'exodus' plugin is deprecated, loading 'hdf' instead");
+    localPathOrName = "hdf";
```
mwestphal, line 278: Sorry about that, recursive call was fine after all, I misunderstood what you implemented initially.

.github/actions/wasm-ci/action.yml
```diff
@@ -16,7 +16,7 @@ runs:
         -DF3D_MODULE_UI=OFF
         -DF3D_PLUGIN_BUILD_ASSIMP=ON
         -DF3D_PLUGIN_BUILD_DRACO=ON
-        -DF3D_PLUGIN_BUILD_EXODUS=OFF
+        -DF3D_PLUGIN_BUILD_HDF=OFF
```
mwestphal, line 19: turns out wasm supports hdf, so stay tuned, you should be able to turn that on once I rebuild the image (see https://github.com/f3d-app/f3d-docker-images/pull/40)

mwestphal: (and add the related extension to supported F3DWeb stuff)

mwestphal (changes requested):
Please resolve discussions when you adress them :)

library/src/engine.cxx
Lgt2x: reverted last change

.github/actions/wasm-ci/action.yml
Lgt2x: nice!

mwestphal: alright, you should now be able to enable HDF plugin on wasm after a rebase on master

.github/actions/wasm-ci/action.yml
```diff
@@ -16,7 +16,7 @@ runs:
         -DF3D_MODULE_UI=OFF
         -DF3D_PLUGIN_BUILD_ASSIMP=ON
         -DF3D_PLUGIN_BUILD_DRACO=ON
-        -DF3D_PLUGIN_BUILD_EXODUS=OFF
+        -DF3D_PLUGIN_BUILD_HDF=ON
```
mwestphal, line 19: can you add vtkhdf and exodus extensions to the F3DWeb app ?

mwestphal (changes requested):
I've edited the PR descr for the new CI workflow template. Please check the checkboxes for CI when you want.

application/testing/CMakeLists.txt
```diff
@@ -721,57 +721,61 @@ if(F3D_PLUGIN_BUILD_DRACO)
 
 endif()
 
-if(F3D_PLUGIN_BUILD_EXODUS)
-  f3d_test(NAME TestExodus DATA disk_out_ref.ex2 ARGS --load-plugins=exodus -s --camera-position=-11,-2,-49)
+if(F3D_PLUGIN_BUILD_HDF)
+  f3d_test(NAME TestVTKHDF DATA blob.vtkhdf ARGS --load-plugins=hdf -s)
```
mwestphal, line 725: you need to put this test behind a VTK_VERSION check 9.3 or 9.4 I think. Better if you know the exact date.

webassembly/app.html
```diff
@@ -58,7 +58,7 @@ <h1 class="title">F3D Web</h1>
                 class="file-input"
                 type="file"
                 id="file-selector"
-                accept=".gml,.gltf,.glb,.obj,.ply,.pts,.stl,.vtk,.vtp,.vtu,.3ds,.wrl,.vrml,.fbx,.off,.x,.dae,.stp,.step,.igs,.iges,.brep,.xbf,.drc"
+                accept=".gml,.gltf,.glb,.obj,.ply,.pts,.stl,.vtk,.vtp,.vtu,.vtkhdf,,.3ds,.wrl,.vrml,.fbx,.off,.x,.dae,.ex2,.stp,.step,.igs,.iges,.brep,.xbf,.drc"
```
mwestphal, line 61:
```suggestion
                accept=".gml,.gltf,.glb,.obj,.ply,.pts,.stl,.vtk,.vtp,.vtu,.vtkhdf,.3ds,.wrl,.vrml,.fbx,.off,.x,.dae,.ex2,.stp,.step,.igs,.iges,.brep,.xbf,.drc"
```

mwestphal, line 61:
```suggestion
                accept=".gml,.gltf,.glb,.obj,.ply,.pts,.stl,.vtk,.vtp,.vtu,.vtkhdf,.3ds,.wrl,.vrml,.fbx,.off,.x,.dae,.ex2,.exo,.e,.stp,.step,.igs,.iges,.brep,.xbf,.drc"
```

application/testing/CMakeLists.txt
Lgt2x: waiting for CI to tell me if 9.3 is fine

testing/configs/exodus.json
mwestphal, line 1: it would be nice to rename that file imo.

mwestphal (changes requested):
One last  comment

testing/configs/exodus.json
Lgt2x: it is still only matching exodus files though, so renaming to `hdf` it would also mean changing its content, which is not useful for the tests

mwestphal (approved):

Meakk (approved):

PR #2193, opened 2025-04-28
add `dir` multi-file mode

snoyer:
### Describe your changes

Add `dir` multi-file mode to group files according to their parent directory.

I refactored the grouping logic to use a group key computed from the file path according to the mode instead of having separate logic for each.

The test case is kind of ugly and has a warning but it does hit the recursive case for file addition with existing data. It can be improved when I add a proper dataset to test regex grouping in a future PR.

- [x] If it is a modifying the default versions, I have updated timestamp
- [x] I have added [documentation](https://f3d.app/doc/dev/TESTING.html) for new features

### Continuous integration

- [x] Style checks
- [x] Fast CI
- [x] Coverage cached CI
- [x] Analysis cached CI
- [x] WASM CI
- [x] Android CI
- [x] macOS Intel cached CI
- [x] macOS ARM cached CI
- [x] Windows cached CI
- [x] Linux cached CI
- [x] Other cached CI


application/testing/CMakeLists.txt
```diff
@@ -230,6 +230,7 @@ f3d_test(NAME TestTextures DATA WaterBottle.glb ARGS --texture-material=${F3D_SO
 f3d_test(NAME TestMetaDataImporter DATA BoxAnimated.gltf ARGS -m UI)
 f3d_test(NAME TestMultiblockMetaData DATA mb.vtm ARGS -m UI)
 f3d_test(NAME TestMultiFileMetaData DATA mb/recursive ARGS -m --multi-file-mode=all UI)
+f3d_test(NAME TestMultiFileModeDir DATA mb ARGS -n --multi-file-mode=dir UI)
```
mwestphal, line 233: do we need a "group-index" CLI option ? Currently there is no way to load a specific group on startup.

snoyer: The group that shows on startup should be the one that holds the latest of the input files. Or at least that's what I assumed based on #756.

However it looks like `AddFile`'s return value is ignored in places. Once that is fixed the there will be a way to select a specific group without needing an index option

mwestphal: > The group that shows on startup should be the one that holds the latest of the input files.

I agree

> without needing an index option

What do you mean ?

snoyer: you would be able to do `f3d foo/ foo/bar.xyz` and be sure that the group containing `foo/bar.xyz` would be shown regardless of what the grouping logic is and what the position of `foo/bar.xyz` is in the recursive walk of the directory

snoyer: I was mistaken, the files passed as argument are supposed to be shown in order with the _first_ one displaying on startup.

So, without changing that behavior, it is possible to load a specific group by picking the _first_ of the arguments, for example:
- `f3d --multi-file-mode=dir testing/data/mb/mb_0_0.vtu testing/data/mb -n` f3d-1
- `f3d --multi-file-mode=dir testing/data/mb/recursive/mb_0_0.vtu testing/data/mb -n` f3d-2

mwestphal: Ok! As long as its consistent fine by me! My original question (do we need a CLI option to select the group to load?) is still valid but its out of scope of this PR entirely anyway :)

mwestphal, line 233: Not a big fan of the fact that the warning makes this a UI test, but I agree that adding a new folder hierachy just to test this would not be great either.

mwestphal (approved):

Issue #1746, opened 2024-12-03
Add glob support for config file match


mwestphal:
**Is your feature request related to a problem? Please describe.**
F3D support being configured using config file, each config block applies to the files currently being loaded.
The pattern match is based on regex, which may not be very user friendly.

Lets improve that by adding other match type, especially `glob`

**Describe the solution you'd like**
 - Add a glob -> regex implementation in `utils`
 - Read an optional `match_type` json arg in F3DConfigFilesTools.cxx, eg `glob`, `regex`, `exact`
 - Implement the pattern matching and convert it to regex for the actual config logic
 - Optionnaly, support proper logging of the match and match-type

**Additional context**
@snoyer Have a partial implement of glob -> regex somewhere

skusel:
I would be interested in working on this.

mwestphal:
Of course! 

Please feel free to join the discord to discuss the issue :)

https://discord.f3d.app

mwestphal:
Fixed by #2281

PR #2281, opened 2025-05-29
Config glob matching

skusel:
### Describe your changes
Adds glob support for config file match

#1746

- [x] If it is a modifying the `.github/workflows/versions.json`, I have updated `timestamp`

### Continuous integration

- [x] Style checks
- [x] Fast CI
- [x] Coverage cached CI
- [x] Analysis cached CI
- [x] WASM docker CI
- [x] Android docker CI
- [x] macOS Intel cached CI
- [x] macOS ARM cached CI
- [x] Windows cached CI
- [x] Linux cached CI
- [x] Other cached CI


library/public/utils.h
```diff
@@ -68,6 +68,20 @@ class F3D_EXPORT utils
...
+  /**
+   * Converts a glob expression to a regular expression
+   *
+   * It handles the following glob features:
+   * - '*': Matches any number of characters (except path separators)
+   * - '?': Matches exactly one character (except path separators)
+   * - '[...]': Character classes
+   * - '[!...] or '[^...]': Negated character classes
+   * - '**': Matches any number of characters including path separators if enabled
+   * - '{a,b,c}': Alternation (matches any of the comma-separated patterns)
```
snoyer, line 80: backticks for IDEs that support Markdown
image
 
```suggestion
   * - `*`: Matches any number of characters (except path separators)
   * - `?`: Matches exactly one character (except path separators)
   * - `[...]`: Character classes
   * - `[!...]` or `[^...]`: Negated character classes
   * - `**`: Matches any number of characters including path separators if enabled
   * - `{a,b,c}`: Alternation (matches any of the comma-separated patterns)
```

library/src/utils.cxx
```diff
@@ -119,9 +119,191 @@ fs::path utils::collapsePath(const fs::path& path, const fs::path& baseDirectory
   }
 }
 
+//----------------------------------------------------------------------------
+std::string utils::globToRegex(std::string_view glob, bool fullMatch, bool supportGlobStars)
```
snoyer, line 123: IMO the `fullMatch` param is unnecessary; Callers should be able to achieve the same thing by using `regex_match` vs `regex_search`, or they can concat the `^` and `$` themselves.
Unless it makes some more subtle differences I'm not seeing?

library/src/utils.cxx
```diff
@@ -119,9 +119,191 @@ fs::path utils::collapsePath(const fs::path& path, const fs::path& baseDirectory
...
+            if (starCount > 1 && (prevChar == '/' || prevChar == '\0') &&
+              (nextChar == '/' || !nextChar))
+            {
+              result += "(?:[^/]*(?:/|$))*";
+              i++; // Eat nextChar
+            }
+            else
+            {
+              result += "[^/]*";
+            }
```
snoyer, line 168: maybe needs to handle `\` and `/` interchangeably to support Windows paths?

python/F3DPythonBindings.cxx
```diff
@@ -220,7 +220,8 @@ PYBIND11_MODULE(pyf3d, module)
 
   utils //
     .def_static("text_distance", &f3d::utils::textDistance)
-    .def_static("collapse_path", &f3d::utils::collapsePath);
+    .def_static("collapse_path", &f3d::utils::collapsePath)
+    .def_static("glob_to_regex", &f3d::utils::globToRegex);
```
snoyer, line 224: add matching default/named args
```suggestion
    .def_static("glob_to_regex", &f3d::utils::globToRegex, py::arg("glob"),
      py::arg("full_match") = true, py::arg("support_glob_stars") = true);
```

python/testing/test_utils.py
```diff
@@ -9,3 +9,7 @@ def test_text_distance():
 
 def test_collapse_path():
     assert f3d.Utils.collapse_path("/folder/../file.ext", ".") == Path("/file.ext")
+
+
+def test_glob_to_regex():
+    assert f3d.Utils.glob_to_regex("*vt?", False, False) == ".*vt."
```
snoyer, line 15: need to test with default/named args too

snoyer (commented):
Nice!

Few comments on the core code. I haven't looked at the app side yet, maybe @mwestphal will get to that (and do a better job than I would) before I come back to it

mwestphal:
> Nice!
> 
> Few comments on the core code. I haven't looked at the app side yet, maybe @mwestphal will get to that (and do a better job than I would) before I come back to it

I won't be able to look at it before June I'm afraid.

library/public/utils.h
skusel: Added backticks

library/src/utils.cxx
skusel: Removing it makes sense to me. It should be removed in my latest commit.

library/src/utils.cxx
skusel: I added support for the preferred separator on Windows. I also added a `useGenericSeparator` parameter to the `f3d::utils::globToRegex` function, which is `false` by default, in case a caller doesn't want to use the OS's preferred path separator for whatever reason.

testing/baselines/TestMatchFirstConfigGlobOptionBlock.png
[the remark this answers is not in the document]
skusel: Thanks! Instead of showing a grid, I changed the background color. More tests passed, but I am still seeing a few failures. I'll mess around with the config/test a little more and see if I can resolve those failures.

library/src/utils.cxx
snoyer: My idea of handling both separators interchangeably may indeed not have been the best way to go indeed and I agree that leaving the choice to the callers is good.

However it looks like using `fs::path::preferred_separator` internally makes the behavior of the function platform specific when `useGenericSeparator=true`, which means it can't be fully unit-tested on a single platform?
Would using a `pathSeparator='/'` parameter work instead? This way you could unit-test both types of paths in a platform-independent way by setting the param to `/` or `\\`, and the app/lib could call with `pathSeparator=fs::path::preferred_separator` IRL

Alternatively maybe we could also go back to the version that only handles forward slashes but make sure the paths that we match are formatted with `path.generic_string()` instead of `path.string()`, but then we'd also have to run the globs through `generic_string()` so that a Windows user can use globs with backslashes...

Really not sure what the best solution is here. @Meakk, @mwestphal, opinions?

skusel: I think `char pathSeparator='/'` as the third parameter instead of `bool useGenericSeparator=true` makes a lot of sense and makes testing the function a whole lot nicer. I changed it to take `pathSeparator` instead, cleaned up the unit tests, and added a few more tests specifically for the Windows preferred separator. The application passes `fs::path::preferred_separator` as you suggested.

I am happy to revert things and pass `path.generic_string()` in the application code instead if there is consensus that that is the better route.

doc/user/CONFIGURATION_FILE.md
```diff
@@ -176,6 +176,54 @@ keyboard layout, eg. on a `QWERTY` layout, the bind `Exclam` is unreachable whil
 We tried to make the default bindings of F3D to be a keyboard layout independent as possible, but feel free to redefined them
 for your needs.
 
+## Glob and Exact Matching
+
+While regex is the default match type, glob and exact match types are also supported. Below is an example using glob and exact matching.
```
mwestphal, line 182: maybe add a mention of the matching possibilities in the first section of the doc and a link to this section ?

application/F3DStarter.cxx
```diff
@@ -417,12 +417,65 @@ class F3DStarter::F3DInternals
       std::to_string(maxNumberingAttempts) + " attempts");
   }
 
+  static bool PatternMatched(const std::string& source, const std::string& patternType,
+    const std::string& pattern, const std::string& inputFile)
```
mwestphal: maybe keep naming consistent : `matchType`, `match` ?

I know it was inconsistent before, just a suggestion :)

library/public/utils.h
```diff
@@ -68,6 +68,20 @@ class F3D_EXPORT utils
...
+  /**
+   * Converts a glob expression to a regular expression
+   *
+   * It handles the following glob features:
+   * - `*`: Matches any number of characters (except path separators)
+   * - `?`: Matches exactly one character (except path separators)
+   * - `[...]`: Character classes
+   * - `[!...]` or `[^...]`: Negated character classes
+   * - `**`: Matches any number of characters including path separators if enabled
+   * - `{a,b,c}`: Alternation (matches any of the comma-separated patterns)
```
mwestphal: please precise exception behavior in doc

mwestphal: and arg doc as well

application/F3DStarter.cxx
```diff
@@ -417,12 +417,65 @@ class F3DStarter::F3DInternals
...
+    if (patternType == "exact")
+    {
+      return pattern == inputFile;
+    }
+    else
+    {
+      try
+      {
+        std::regex re(patternType == "glob"
+            ? f3d::utils::globToRegex(pattern, false, fs::path::preferred_separator)
```
mwestphal: should we give user control of the globstar behavior ?

application/F3DStarter.cxx
```diff
@@ -417,12 +417,65 @@ class F3DStarter::F3DInternals
       std::to_string(maxNumberingAttempts) + " attempts");
   }
 
+  static bool PatternMatched(const std::string& source, const std::string& patternType,
+    const std::string& pattern, const std::string& inputFile)
+  {
+    static std::set<std::string> erroredPatterns;
+    if (patternType == "exact")
+    {
+      return pattern == inputFile;
```
mwestphal: I do not see  a test for this ?

mwestphal (changes requested):
excelent work!
I've left a review :)

application/F3DStarter.cxx
mwestphal: And somehow coverage dost not report it as missed though ?

skusel: The old match_types.json config file had an exact block, and I believe that was causing this block to be hit in the tests, even though it was not really testing it. I added some improved exact tests.

application/F3DStarter.cxx
skusel: Sure, I made some changes that try to be smart about when to enable it. I added a function to check if there is a globstar in the expression, and if there is, I enable globstar support. I updated the docs to communicate this behavior to users. I think this should work well because I wouldn't expect to see a globstar in their expression unless they were trying to use it.

library/public/utils.h
skusel: Added some documentation on the args and exception scenarios.

application/F3DStarter.cxx
skusel: Made them consistent.

doc/user/CONFIGURATION_FILE.md
skusel: Updated this doc to mention the matching possibilities in the first section and linked this section.

application/F3DStarter.cxx
mwestphal: sounds fine by me, wdyt @snoyer ?

application/F3DStarter.cxx
```diff
@@ -417,12 +417,65 @@ class F3DStarter::F3DInternals
...
+        origin = match;
+      }
+      else
+      {
+        origin = source;
+        origin += ":`";
+        origin += match;
+        origin += "` (";
+        origin += matchType;
+        origin += ")";
```
mwestphal, line 477:
```suggestion
        origin = source + ":`" + match + "` (" + matchType +  ")";
```
?

application/F3DStarter.cxx
```diff
@@ -457,13 +510,11 @@ class F3DStarter::F3DInternals
...
         // For each entry (eg: different config files)
-        for (auto const& [conf, source, pattern] : entries)
+        for (auto const& [conf, source, matchType, match] : entries)
         {
-          std::regex re(pattern, std::regex_constants::icase);
-          std::smatch matches;
           // If the source is empty, there is no pattern, all options applies
           // Note: An empty inputFile matches with ".*"
-          if (source.empty() || std::regex_match(inputFile, matches, re))
+          if (source.empty() || PatternMatched(source, matchType, match, inputFile))
```
mwestphal:
```suggestion
          if (source.empty() || F3DInternals::PatternMatched(source, matchType, match, inputFile))
```

I know its not required but more explicit this way

application/F3DStarter.cxx
```diff
@@ -536,15 +587,27 @@ class F3DStarter::F3DInternals
...
+                    origin = match;
+                  }
+                  else
+                  {
+                    origin = source;
+                    origin += ":`";
+                    origin += match;
+                    origin += "` (";
+                    origin += matchType;
+                    origin += ")";
```
mwestphal, line 610: same as above

application/F3DStarter.cxx
```diff
@@ -553,8 +616,20 @@ class F3DStarter::F3DInternals
...
```
mwestphal, line 633: same as above

application/F3DStarter.cxx
```diff
@@ -777,14 +852,11 @@ class F3DStarter::F3DInternals
...
-        for (auto const& [bindings, source, pattern] : this->ConfigBindingsEntries)
+        for (auto const& [bindings, source, matchType, match] : this->ConfigBindingsEntries)
         {
-          std::regex re(pattern, std::regex_constants::icase);
-          std::smatch matches;
-
           // If the source is empty, there is no pattern, all bindings applies
           // Note: An empty inputFile matches with ".*"
-          if (source.empty() || std::regex_match(inputFile, matches, re))
+          if (source.empty() || PatternMatched(source, matchType, match, inputFile))
```
mwestphal: same as above

application/F3DStarter.cxx
snoyer: looks prettier but I believe `clang-tidy` won't like that, worth checking

application/testing/CMakeLists.txt
```diff
@@ -1238,6 +1238,37 @@ f3d_test(NAME TestNoConfigWithConfig DATA f3d.glb ARGS --no-config CONFIG ${F3D_
...
+  set(GLOBSTAR_EXPR [=[**\\\\*stl]=])
+else()
+  set(GLOBSTAR_EXPR "**/*stl")
+endif()
+configure_file("${F3D_SOURCE_DIR}/testing/configs/glob.json.in" "${CMAKE_BINARY_DIR}/glob.json")
+f3d_test(NAME TestMatchFirstConfigGlobOptionBlock DATA dragon.vtu CONFIG ${CMAKE_BINARY_DIR}/glob.json)
+f3d_test(NAME TestMatchSecondConfigGlobOptionBlock DATA suzanne.stl CONFIG ${CMAKE_BINARY_DIR}/glob.json)
+
+# Test exact matching
+cmake_path(APPEND FIRST_EXACT_PATH "${F3D_SOURCE_DIR}" "testing" "data" "suzanne.obj")
```
mwestphal: I dont see the point of APPEND here?

```suggestion
cmake_path(SET FIRST_EXACT_PATH "${F3D_SOURCE_DIR}/testing/data/suzanne.obj")
```

application/testing/CMakeLists.txt
```diff
@@ -1238,6 +1238,37 @@ f3d_test(NAME TestNoConfigWithConfig DATA f3d.glb ARGS --no-config CONFIG ${F3D_
...
+cmake_path(NATIVE_PATH FIRST_EXACT_PATH NATIVE_FIRST_EXACT_PATH)
+cmake_path(APPEND SECOND_EXACT_PATH "${F3D_SOURCE_DIR}" "testing" "data" "suzanne.ply")
```
mwestphal: same as above

application/testing/CMakeLists.txt
```diff
@@ -1238,6 +1238,37 @@ f3d_test(NAME TestNoConfigWithConfig DATA f3d.glb ARGS --no-config CONFIG ${F3D_
...
+cmake_path(NATIVE_PATH SECOND_EXACT_PATH NATIVE_SECOND_EXACT_PATH)
+if(WIN32)
+  string(REPLACE "\\" "\\\\" NATIVE_FIRST_EXACT_PATH "${NATIVE_FIRST_EXACT_PATH}")
+  string(REPLACE "\\" "\\\\" NATIVE_SECOND_EXACT_PATH "${NATIVE_SECOND_EXACT_PATH}")
```
mwestphal, line 1267: I wish there was a better way, but I dont see any.

doc/user/CONFIGURATION_FILE.md
```diff
@@ -176,6 +177,56 @@ keyboard layout, eg. on a `QWERTY` layout, the bind `Exclam` is unreachable whil
...
+    "match-type": "exact",
+    "match": "/path/to/file.mhd",
+    "options": {
+      "volume": true
+    }
+  }
+]
+```
+
+Globstar support will only be enabled when using the glob match type and the pattern contains a globstar (`**`). By default, globstar support is disabled. When globstar support is enabled `*` and `?` will not match file path separators.
```
mwestphal: from a user POV the enabling/disabling of the globstar doesn't matter imo. Just mention that globstar is supported ?

application/F3DStarter.cxx
snoyer: is that just an optimization or does it change the behavior? I mean is checking for globstars first different from running with `supportGlobStars=true` always?

library/public/utils.h
```diff
@@ -68,6 +68,30 @@ class F3D_EXPORT utils
   [[nodiscard]] static std::filesystem::path collapsePath(
     const std::filesystem::path& path, const std::filesystem::path& baseDirectory = {});
 
+  /**
+   * Returns true if the glob contains a globstar (`**`).
+   */
+  [[nodiscard]] static bool globContainsGlobstar(std::string_view glob);
```
mwestphal: from libf3d pov the globstar detection could be put into globToRegex in order to avoid exposing another method in utils.h.

But maybe the API you suggest here is better and would be more expected by libf3d users, wdyt ?

application/F3DStarter.cxx
skusel: Checking for globstars first vs always setting `supportGlobStars=true` does change behavior. When I wrote my first glob expression in a config, I naturally wanted to add a match like `*vtp` because I didn't have to think about the file path, only the file extension I was trying to match, and it's not too different from the regex, `.*vtp`, that it translates to. If globstars were always supported, `*vtp` would *most likely* not match anything because the `*` would not match path separators. Instead, the user would have to use the glob `**/*vtp`. This is nice for users who care about the matching against other parts of the file path, but not for users who only want to match against the file name or extension. My solution attempts to make both types of users happy.

Another option would be to allow the users to state what type of match they want. One way to go about this would be to expand the match type options to include "glob" and "glob-with-globstars". Then the user would be in complete control of enabling/disabling globstar support.

application/F3DStarter.cxx
skusel: I made this multiple lines because `clang-tidy` was reporting the error "String concatenation results in allocation of unnecessary temporary strings; consider using 'operator+=' or 'string::append()' instead [[performance-inefficient-string-concatenation](https://clang.llvm.org/extra/clang-tidy/checks/performance/inefficient-string-concatenation.html)]". 

I could tell `clang-tidy` not to show this error here, so we can keep it to a single line. (Or technically 2 with the comment.)

```c++
// NOLINTNEXTLINE(performance-inefficient-string-concatenation)
origin = source + ":`" + match + "` (" + matchType + ")";
```

Let me know if I should keep it as is or if I should update it with the clang-tidy error disabled.

application/F3DStarter.cxx
skusel: Done

application/F3DStarter.cxx
skusel: Done

application/testing/CMakeLists.txt
skusel: Makes sense, replaced `APPEND` with `SET` in both locations.

application/testing/CMakeLists.txt
skusel: I do too, but I don't know of any.

doc/user/CONFIGURATION_FILE.md
skusel: Sure, simplified the docs by letting users know globstar is supported.

library/public/utils.h
skusel: If I were an outside caller of `globToRegex`, I would want to be able to specify whether globstars are supported, or have `globToRegex` be consistent about whether it supports them. I agree that `globContainsGlobstar` feels a little like clutter in the public interface of libf3d, but I think the `globToRegex` function has a nicer interface when it has a parameter that can be used to turn on/off globstar support. For that reason, I would lean towards keeping `globContainsGlobstar` and `globToRegex` as separate methods. 

However, I am happy to combine them if you would prefer to expose fewer methods. Or if we decide to go a different direction with enabling globstars in the f3d application, like making the user explicitly tell us they want to use globstars in their config via the match-type (as mentioned in https://github.com/f3d-app/f3d/pull/2281#discussion_r2124940265), then I could remove the `globContainsGlobstar` method altogether.

application/F3DStarter.cxx
snoyer: > My solution attempts to make both types of users happy.

Ok I get it know. Makes perfect sense to me.

> Another option would be to allow the users to state what type of match they want. One way to go about this would be to expand the match type options to include "glob" and "glob-with-globstars"

I don't think that's necessary, if a user is that far down the matching rabbit hole that  globs (with auto/smart globstar support) is too limiting for them they're probably better off switching to regex IMO

application/F3DStarter.cxx
snoyer: if the goal is to keep in one line it'd be better to use chained `.append()` calls than to bypass `clang-tidy` (even tho it's not critical, the warning is sensible)
example: https://github.com/f3d-app/f3d/blob/c117e5b76ed366eb11b313b34529f8e7d6e2c663/library/src/interactor_impl.cxx#L820

now is a bunch of `.append()` on a single line more readable than a bunch of `+=` on multiple lines is another question...

library/public/utils.h
snoyer: > If I were an outside caller of globToRegex, I would want to be able to specify whether globstars are supported

In the general sense that is true and your impl with the `supportGlobStars` would be great in a standalone library. However here the only "outside" caller will effectively be F3D so we could reduce the scope to only what's needed by the app. The real-life use-case being offering a more user-friendly alternative to regex to match file names maybe it doesn't need to be _that_ flexible?

You could go as far as making the `supportGlobStars=globContainsGlobstar()` logic internal to `globToRegex()` and dropping the `bool supportGlobStars` param altogether.  You wouldn't even need to change the documentation: 

> `*`: Matches any number of characters (except path separators when using globstars)

you meant "using globstar" as in "passing `supportGlobStars=true`" but it could also be read as "the glob also contains `**`"

Are there other realistic/planned use-cases where we'd explicitly need `supportGlobStars=false`?

application/F3DStarter.cxx
mwestphal: fine to keep as is then ^^.
You can add a comment above mentionning using std::format once c++20 is supported :)

library/public/utils.h
mwestphal: I agree with @snoyer here :)

mwestphal (commented):
LGTM, just need to figure out the right API in regard to globStar.

I also defer the regex review to @snoyer :)

library/public/utils.h
skusel: Cool, made `globContainsGlobstar` and an internal part of `globToRegex`.

application/F3DStarter.cxx
skusel: Sounds good, added a comment to switch to `std::format` once C++20 is supported.

mwestphal (approved):
Approved, but also needs @snoyer approval :)

library/public/utils.h
```diff
@@ -68,6 +68,23 @@ class F3D_EXPORT utils
...
+   * Converts a glob expression to a regular expression. When a glob contains globstars,
+   * the `pathSeparator` (`/` by default) is used to help generate the regular expression.
+   *
+   * It handles the following glob features:
+   * - `*`: Matches any number of characters (except path separators when using globstars)
+   * - `?`: Matches exactly one character (except path separators when using globstars)
+   * - `[...]`: Character classes
+   * - `[!...]` or `[^...]`: Negated character classes
+   * - `**`: Matches any number of characters including path separators when using globstars
+   * - `{a,b,c}`: Alternation (matches any of the comma-separated patterns)
```
snoyer:
```suggestion
   * - `*`: Matches zero or more characters (except path separators when also using `**`)
   * - `?`: Matches exactly one character (except path separators when also using `**`)
   * - `[abc]`: Character class, matches any of the given characters
   * - `[!abc]` or `[^abc]`: Negated character class, matches none of the given characters
   * - `{foo,bar,baz}`: Alternation, matches any of the given comma-separated patterns
   * - `**`: Matches any number of characters including path separators
```
Should the "globstar" term be dropped now that the parameter is gone? alternatively keep it but have `- `**`: globstar, matches any number of characters including path separators`?

library/src/utils.cxx
```diff
@@ -119,9 +119,240 @@ fs::path utils::collapsePath(const fs::path& path, const fs::path& baseDirectory
...
+      if (c == '\\')
+      {
+        if (escaped)
+        {
+          escaped = false;
+        }
+        else
+        {
+          escaped = true;
+        }
```
snoyer:
```suggestion
        escaped = !escaped;
```
simple enough to be made more concise?

library/src/utils.cxx
```diff
@@ -119,9 +119,240 @@ fs::path utils::collapsePath(const fs::path& path, const fs::path& baseDirectory
...
+      }
+
+      if (!escaped && c == '*')
+      {
+        starCount++;
+      }
+      else
+      {
+        starCount = 0;
+      }
```
snoyer:
```suggestion
starCount = !escaped && c == '*' ? starCount + 1 : 0;
```
because Allman style taking 8 lines to for a conditional assignation is just too damn much :D but feel free to ignore if you deem the longer one more readable

library/src/utils.cxx
```diff
@@ -119,9 +119,240 @@ fs::path utils::collapsePath(const fs::path& path, const fs::path& baseDirectory
...
+
+      if (starCount == 2)
```
snoyer:
```suggestion
      if (starCount >= 2)
```
Just in case if we hit a cosmic ray and miss a loop somehow :scream:

library/public/utils.h
skusel: Updated

library/src/utils.cxx
skusel: Updated

library/src/utils.cxx
skusel: Updated to be more concise

library/src/utils.cxx
skusel: Updated :milky_way:

mwestphal:
Thanks for the contribution @skusel

PR #2296, opened 2025-06-12
Fix collinear camera direction and up vector bug

### LINKED ISSUE #2048: Bug with collinear `--up` and `--camera-direction`
**Describe the bug**
The initial camera setup doesn't work when `--up` and `--camera-direction` values are collinear.

**To Reproduce**

1.  Open a file with `--up` and `--camera-direction` set to collinear values, for example:
  - `f3d testing/data/dragon.vtu --up=0,1,0 --camera-direction=0,2,0`
  - `f3d testing/data/dragon.vtu --up=z --camera-direction=-z`
2.  model isn't shown, or maybe the camera is inside of the model?

**F3D Information**
Happens on current master but already present in `3.0`

**Additional context**
Offsetting the directions ever so slightly avoids the problem (eg.`--up=0,1,0 --camera-direction=0,1,0.000000001`) so it's probably just a matter of adding a check to avoid exact collinearity

skusel:
### Describe your changes
Fixes an issue where the model would not appear if the camera position to focal point vector and up vector are collinear.

Fixes a bug where the model would disappear after pressing '7' for the top view and then attempting to change the perspective by clicking and dragging the mouse.

Fix #2048

### Continuous integration

- [x] Style checks
- [x] Fast CI
- [x] Coverage cached CI
- [x] Analysis cached CI
- [x] WASM docker CI
- [x] Android docker CI
- [x] macOS Intel cached CI
- [x] macOS ARM cached CI
- [x] Windows cached CI
- [x] Linux cached CI
- [x] Other cached CI


library/src/camera_impl.cxx
```diff
@@ -29,7 +29,7 @@ camera& camera_impl::setPosition(const point3_t& pos)
 {
   vtkCamera* cam = this->GetVTKCamera();
   cam->SetPosition(pos.data());
-  cam->OrthogonalizeViewUp();
+  OrthogonalizeViewUp();
```
snoyer: I'm pretty sure I've been told to use explicit `this->` before but I don't see it in the style guide... @mwestphal, is that a thing?

library/src/interactor_impl.cxx
```diff
@@ -175,7 +175,9 @@ class interactor_impl::internals
         axis = { +1, 0, 0 };
         break;
       case ViewType::VT_TOP:
-        axis = { 0, 0, +1 };
+        // The model is prone to disappear when clicked and dragged to change perspective if axis =
+        // { 0, 0, +1 }
+        axis = { 0, +0.000001, +0.999999 };
```
snoyer: I'm not convinced by this...
- this only handles the top case, the same bug (model disappearing when dragging) will still happen when pressing the `1` key with `--up=X` for example
- isn't this vector nudging what this PR implements already? it feels like a call to the new method after the switch would be the way (which I assume you tried a per the added blank line in the diff?)

Is this actually a different (but closely related) bug than the one this PR addresses?

library/src/camera_impl.cxx
skusel: This is a different bug. I noticed it happens when the camera position to focal point vector falls on the Y-axis. When I put `--camera-direction=0,1,0` for example, I notice the same issue even when up is not collinear. I didn't try to fix that case because I figured the user explicitly asked for that positioning, but I could nudge it to address the issue.

The bug originally mentioned in #2048 happens when the camera position to focal point vector and up vector are collinear. In the case of the top view (i.e. '7' is pressed), the camera position to focal point vector and up vector were perpendicular and the model disappearing when clicking and dragging still appeared until I moved the camera position to up vector slightly off the Y-axis. Nudging the up vector didn't help. 

It might be better if I undo this fix and only address the original issue mentioned in #2048, which is fixed by the nudging in `camera_impl.*`. It might be worth creating another issue for this particular bug, so it can be addressed separately. Let me know what you think!

library/src/interactor_impl.cxx
skusel: This is a different bug. I noticed it happens when the camera position to focal point vector falls on the Y-axis. When I put `--camera-direction=0,1,0` for example, I notice the same issue even when the up vector is not collinear. I didn't try to fix that case because I figured the user explicitly asked for that positioning, but I could nudge it to address the issue.

The bug originally mentioned in #2048 happens when the camera position to focal point vector and up vector are collinear. That is fixed by the changes in `camera_impl.*`. In the case of the top view (i.e. '7' is pressed) bug, the camera position to focal point vector and up vector were perpendicular and the model disappearing when clicking and dragging still appeared until I moved the camera position to up vector slightly off the Y-axis. Nudging the up vector didn't help. 

It might be better if I undo this fix and only address the original issue mentioned in #2048, and creating another issue for this particular bug, so it can be addressed separately. Let me know what you think!

library/src/camera_impl.cxx
mwestphal: Its missing for the guide but yes, please use `this` whenever possible

library/src/interactor_impl.cxx
mwestphal: > It might be better if I undo this fix and only address the original issue mentioned in https://github.com/f3d-app/f3d/issues/2048, and creating another issue for this particular bug, so it can be addressed separately. Let me know what you think!

Let's do that :)

library/src/interactor_impl.cxx
```diff
@@ -193,6 +195,7 @@ class interactor_impl::internals
     /* set camera coordinates back */
     cam.setPosition(newPos);
     cam.setViewUp(up);
+
```
mwestphal: change is not needed

library/src/camera_impl.cxx
```diff
@@ -295,4 +295,27 @@ vtkCamera* camera_impl::GetVTKCamera()
 {
   return this->Internals->VTKRenderer->GetActiveCamera();
 }
+
+//----------------------------------------------------------------------------
+void camera_impl::OrthogonalizeViewUp()
```
mwestphal: Should this be part of `f3d::camera` API ? I'm not sure about this, merely asking.

If not, then I think this should simply be a static method with a vtkCamera* arg.

library/src/camera_impl.cxx
skusel: I updated `OrthogonalizeViewUp` to be a static method in `camera_impl::internals`, so calls now look like `this->Internals->OrthogonalizeViewUp(cam);`

library/src/camera_impl.cxx
skusel: I think it makes more sense as a `static` method in `camera_impl::internals` that takes a `vtkCamera*` arg. That change has been made.

library/src/interactor_impl.cxx
skusel: Removed new line.

library/src/interactor_impl.cxx
skusel: I did a little more investigation into this because it was bugging me, and I noticed that [vtkext/private/module/vtkF3DInteractorStyle.cxx](https://github.com/f3d-app/f3d/pull/2296/files#diff-f9be083ab744e1ee5845ac73f243758dbb8a53ef598ee9e1c14198f1252f190b) was using the environment's up vector in `vtkF3DInteractorStyle::Rotate`. Unless the `--up` argument is passed as something different, the environment's up vector will be `0,1,0`, which is collinear to the camera direction vector anytime the camera direction vector falls on the y-axis (which happens when viewing the model from the top). This is why the model disappears after navigating to the top view and clicking and dragging.

I believe it makes more sense to use the camera's up vector instead, although I may be thinking about things incorrectly. After making this change, `InterpolateTemporaryUp` didn't seem to be necessary. Again, I may be thinking about this wrong. It was first implemented in #1879 to address #1741, however, when using the camera's up vector instead of the environment's up vector, all the jerkiness is gone without any of the `InterpolateTemporaryUp` logic, so I removed it.

These changes are self-contained in the 3rd commit of this PR, so I can easily discard them if you would still like me to address this bug separately, but I added them here because the underlying issue is the same: the camera direction and up vector (in this case the environment's up vector) were collinear, causing the model to disappear.

snoyer: if only if were that simple... ;) here's 2 things to keep in mind for this one:

> I noticed that [vtkext/private/module/vtkF3DInteractorStyle.cxx](https://github.com/f3d-app/f3d/pull/2296/files#diff-f9be083ab744e1ee5845ac73f243758dbb8a53ef598ee9e1c14198f1252f190b) was using the environment's up vector in `vtkF3DInteractorStyle::Rotate`. 

This is by design, there's currently 2 interactor modes in F3D: "trackball" and... "not trackball?", and you can press `K` key to switch between them. Trackball lets you spin the object freely, otherwise it's more like walking around the object with the floor (that you can visualize by toggling the grid with `--grid` or the `G` key) always being level basically (hence using the environment's up vector for the spin axis).

> Unless the `--up` argument is passed as something different, the environment's up vector will be `0,1,0`

the up vector is `Y` by default in the app but in nominal use it will change to match the conventions of various file formats. So anything other that `0,1,0` isn't just an exotic corner case, for example a user dealing with CAD or 3D printing workflow will have a 100% `--up=z` usage by default.

library/src/camera_impl.cxx
mwestphal: nice thanks

mwestphal:
Nice work @skusel , however I'll let you reach consensus with @snoyer before approving anything :)

library/src/interactor_impl.cxx
skusel: Got it, I reverted the 3rd commit and added a 5th commit that takes another approach. Now, if the interpolated up vector and the camera direction vector are collinear, I set the temporary up vector to the camera's up vector and interpolate the up vector again. This seems to be working well. I also added an interaction test to cover this scenario.

I had to reduce the `maxAbsElevation` a little so that the user cannot rotate the model into the range where it is considered collinear with the environment's up vector.

mwestphal:
@snoyer please review :)

snoyer (approved):
Looks good now :+1:
From what I could try the `7` bug is indeed fixed too.

There's still some snapping if you use `--up=x` then press `1` then drag but I think that has to do with the `set_camera` command so it's unrelated to this PR and should be investigated/addressed later.

mwestphal (approved):

Issue #2312, opened 2025-06-18
Add a feature to skip a file


mwestphal:
**Is your feature request related to a problem? Please describe.**
When loading a folder, F3D will try to load all files, even files that are not supported. This is intended because a plugin could provide a reader for any file. However, it also means that folders with unopenable files will pollute F3D. Having a way to specify that a file should NOT be opened at all and not added to the file groups would be very practical

**Describe the solution you'd like**
 - Distinguish between unsupported and invalid files
 - Add a `--skip-unsupported-files` CLI option and add it to the default config
 - Add a test opening a folder containing a file to skip with a config file to skip that file based on the extension

**Additional context**
While this may be a bit confusing, this will be very practical for interactive use of F3D and should be trivial to implement

wyattbiker:
I would suggest an option to skip-files-except:.

E.g. `--skip-files-except: ".*(gltf|fbx)"`

I know this can be done from command line but not drag and drop a folder: E.g.

`~/Assets$ f3d gltf/*.gltf fbx/*.fbx`

AoGao-Kedoka:
Hi kedoka from discord. I would like to work on this one.

mwestphal:
> I would suggest an option to skip-files-except:.
> 
> E.g. `--skip-files-except: ".*(gltf|fbx)"`
> 
> I know this can be done from command line but not drag and drop a folder: E.g.
> 
> `~/Assets$ f3d gltf/*.gltf fbx/*.fbx`

Interesting idea, however this will likely be done differently.

mwestphal:
Unassigned as asked by @AoGao-Kedoka

MayankKumarSingh1:
I'd like to be assigned to this issue. 
But I want to clarify where is the issue in the code and is there a utility exists which distinctions between the supported and invalid files?

mwestphal:
> I'd like to be assigned to this issue. But I want to clarify where is the issue in the code and is there a utility exists which distinctions between the supported and invalid files?

Of course. Can you join the discord so I can help you more easily ?

https://discord.f3d.app

mwestphal:
This is being implemented slightly differently

PR #2326, opened 2025-06-24
Add option skip-unsupported-files

AoGao-Kedoka:
### Describe your changes
- Add option skip-unsupported-files
- Add serveral test cases:
    - Static tests:
        -  `TestMultiFileSkipUnsupportedFiles`, **input**: [unsupported file, supported file], **result**: renders supported file
        - `TestMultiFileNoSkipUnsupportedFiles`,  **input**: [unsupported, supported file], **result**: empty UI
    - Interation tests:
        - `TestInteractionDirectoryLoopNoSkipUnsuppored`: old behavior
        - `TestInteractionDirectoryLoopSkipUnsuppored`: one unsuppored file skipped
        - `TestInteractionSkipUnsupported`, **input**: [unsupported file, supported file]; right arrow* 9, **result**: no loop (skips the unsupported), renders normally
        - `TestInteractionNoSkipUnsupported`, **input**: [unsupported file, supported file]; right arrow * 9, **result**: empty UI

https://github.com/f3d-app/f3d/issues/2312

### Continuous integration

- [x] Style checks
- [x] Fast CI
- [ ] Coverage cached CI
- [ ] Analysis cached CI
- [ ] WASM docker CI
- [ ] Android docker CI
- [ ] macOS Intel cached CI
- [ ] macOS ARM cached CI
- [ ] Windows cached CI
- [ ] Linux cached CI
- [ ] Other cached CI


application/F3DStarter.cxx
```diff
@@ -1369,35 +1375,54 @@ void F3DStarter::LoadFileGroup(int index, bool relativeIndex, bool forceClear)
...
+    }
+
+    if (loadingResult == 1)
+    {
+      break;
+    }
+    groupIndex += skipDirection;
+
+    maximumLoadingCounter++;
+  } while (maximumLoadingCounter < this->Internals->FilesGroups.size());
```
mwestphal, line 1425: I'm confused by this implementation, what does it do ?

To handle unsupported file you can simply do it in: `void F3DStarter::LoadFileGroup(const std::vector<fs::path>& paths, bool clear, const std::string& groupIdx)`. There we check if file are supported are not, just do not add them in that case.

mwestphal (changes requested):
impl doesnt look as expected

application/F3DStarter.cxx
AoGao-Kedoka: It skips to the next file index if the current loading is unsupported, and `maximumLoadingCounter` is to prevent only unsupported files are loaded to have infinite loop.

Files are all added before they are loaded. If we have an unsupported file (and we can only know when it's being loaded), we either have to remove it while iterating from the list or skip it.

mwestphal: Haaa, I understand, you want to avoid adding them to FileGroups. Actually we dont care, you can keep them in file groups, just skip them later when actually adding them to the scene. Should be much simpler.

AoGao-Kedoka: now I'm confused :), i'm not avoiding adding to FileGroups and skipping is actually what I implemented

Just to double check
`F3DStarter::LoadFileGroup(const std::vector<fs::path>& paths, bool clear, const std::string& groupIdx)`: though the method is called LoadFileGroup, it only loads single file path from what I see from debugger (except --multi-file-mode set)
image
That's why the skipping is handled in `void F3DStarter::LoadFileGroup(int index, bool relativeIndex, bool forceClear)`, it get the loading result `CurrentFileLoadingResult` (which will be set by 
`F3DStarter::LoadFileGroup(const std::vector<fs::path>& paths, bool clear, const std::string& groupIdx)`) and  if it's unsupported, it skips based on the skipDirection and try to load the next model.

mwestphal: Indeed, a file group can contain a single file path, that depends on multi-file mode but it should not matter.

We still need to check if each entry in the file group, even if it contain only one. This check is done here https://github.com/f3d-app/f3d/blob/master/application/F3DStarter.cxx#L1500.

After this check, you should just not add the file to the scene here: https://github.com/f3d-app/f3d/blob/master/application/F3DStarter.cxx#L1535

AoGao-Kedoka:
Update: not working on it any more cause I'm having segfault issue with invalid files on vcpkg vtk version.

Issue #2349, opened 2025-06-29
Add a `print_config_info` command


mwestphal:
**Is your feature request related to a problem? Please describe.**
When running F3D **not** from terminal, it can be hard to figure out where are config files located.
Add a simple `print_config_info` that would print the potential location of these file and if F3D found each of these would be a nice addition.

**Describe the solution you'd like**
- Add a `dryRun` boolean and `logLevel` parameter to the `F3DConfigFileTools::ReadConfigFiles` method and implement them
 - Add a applicative command `print_config_info`
 - Make it run `F3DConfigFileTools::ReadConfigFiles` with dryRun to true and logLevel to info
 - Add testing and doc

NaniNoni:
Hello,

I'd love to help and be assigned this issue.

miki-dev-5618:
✨ Hii! I’m new to open source and really excited to make my first contribution 💖
This issue looks like a great starting point — could you please assign it to me? 🌱
Looking forward to learning and helping out! 🚀

NaniNoni:
Hi,

I'm still working on the issue. The WIP PR is linked above.

mwestphal:
Indeed, please pick another issue @blah-bleh-hmm :)

mwestphal:
@NaniNoni still working on this ?

mwestphal:
Need help moving forward @NaniNoni ?

mwestphal:
No feedback from @NaniNoni , unassigning

NaniNoni:
Hi, so sorry about this. I'm applying to universities and have had no time for OSS or side projects in recent months. I should've predicted this and not signed up and wasted your time. I'm even more sorry for not communicating properly. I was quite ashamed, and didn't know how to say this.

mwestphal:
> Hi, so sorry about this. I'm applying to universities and have had no time for OSS or side projects in recent months. I should've predicted this and not signed up and wasted your time. I'm even more sorry for not communicating properly. I was quite ashamed, and didn't know how to say this.

No need to be ashamed! Sometimes, life gets in the way and thats perfectly understandable :)
Indeed, do not hesitate to communicate when possible. Good luck with your studies, we hope to see you again :)

atharvRsharma:
Hi, could i be assigned this issue?

PR #2382, opened 2025-07-09
Implement print_config_info application command

NaniNoni:
### Describe your changes
I implemented the `print_config_info` command to allow for easily checking which config files are loaded. This is especially useful when f3d is ran not from the command line.

f3d_print_config_info

#2349

- [ ] I have performed a [self-review](https://f3d.app/doc/dev/CODING_STYLE.html) of my code
- [ ] I have added [tests](https://f3d.app/doc/dev/TESTING.html) for new features and bugfixes

### Continuous integration

- [ ] Style checks
- [ ] Fast CI
- [ ] Coverage cached CI
- [ ] Analysis cached CI
- [ ] WASM docker CI
- [ ] Android docker CI
- [ ] macOS Intel cached CI
- [ ] macOS ARM cached CI
- [ ] Windows cached CI
- [ ] Linux cached CI
- [ ] Other cached CI

### DIFF
--- a/application/F3DConfigFileTools.cxx
+++ b/application/F3DConfigFileTools.cxx
@@ -86,7 +86,7 @@ std::vector<fs::path> GetConfigPaths(const std::string& configSearch)
 
 //----------------------------------------------------------------------------
 F3DConfigFileTools::ParsedConfigFiles F3DConfigFileTools::ReadConfigFiles(
-  const std::string& userConfig)
+  const std::string& userConfig, bool dryRun, f3d::log::VerboseLevel logLevel)
 {
   // Default config directory name
   std::string configSearch = "config";
@@ -136,7 +136,7 @@ F3DConfigFileTools::ParsedConfigFiles F3DConfigFileTools::ReadConfigFiles(
     // Recover all config files if needed in directories
     if (fs::is_directory(configPath))
     {
-      f3d::log::debug("Using config directory ", configPath.string());
+      f3d::log::print(logLevel, "Using config directory ", configPath.string());
       const size_t oldSize = actualConfigFilePaths.size();
       auto dirIter = fs::directory_iterator(configPath);
       std::copy(std::filesystem::begin(dirIter), std::filesystem::end(dirIter),
@@ -146,7 +146,7 @@ F3DConfigFileTools::ParsedConfigFiles F3DConfigFileTools::ReadConfigFiles(
     }
     else
     {
-      f3d::log::debug("Using config file ", configPath.string());
+      f3d::log::print(logLevel, "Using config file ", configPath.string());
       actualConfigFilePaths.emplace_back(configPath);
     }
   }
@@ -154,7 +154,13 @@ F3DConfigFileTools::ParsedConfigFiles F3DConfigFileTools::ReadConfigFiles(
   // If we used a configSearch but did not find any, inform the user
   if (!configSearch.empty() && actualConfigFilePaths.empty())
   {
-    f3d::log::info("Configuration file for \"", configSearch, "\" could not be found");
+    f3d::log::print(logLevel, "Configuration file for \"", configSearch, "\" could not be found");
+  }
+
+  // Clear config file paths to avoid reading them if dryRun is set
+  if (dryRun)
+  {
+    actualConfigFilePaths.clear();
   }
 
   // Read config files
--- a/application/F3DConfigFileTools.h
+++ b/application/F3DConfigFileTools.h
@@ -5,6 +5,7 @@
  * @brief   A namespace to parse config files
  */
 #include "F3DOptionsTools.h"
+#include "log.h"
 
 #include <string>
 
@@ -25,7 +26,8 @@ struct ParsedConfigFiles
  * Read config files using userConfig if any, return a ParsedConfigFiles
  * containing ordered optionDict, ordered imperative optionDict and ordered bindingsEntries
  */
-ParsedConfigFiles ReadConfigFiles(const std::string& userConfig);
+ParsedConfigFiles ReadConfigFiles(
+  const std::string& userConfig, bool dryRun, f3d::log::VerboseLevel logLevel);
 }
 
 #endif
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -1041,7 +1041,7 @@ int F3DStarter::Start(int argc, char** argv)
   if (!noConfig)
   {
     F3DConfigFileTools::ParsedConfigFiles parsedConfigFiles =
-      F3DConfigFileTools::ReadConfigFiles(config);
+      F3DConfigFileTools::ReadConfigFiles(config, false, f3d::log::getVerboseLevel());
     this->Internals->ConfigOptionsEntries = parsedConfigFiles.Options;
     this->Internals->ImperativeConfigOptionsEntries = parsedConfigFiles.ImperativeOptions;
     this->Internals->ConfigBindingsEntries = parsedConfigFiles.Bindings;
@@ -1865,6 +1865,9 @@ void F3DStarter::AddCommands()
     // parsing_exception is caught within interactor implementation
     return f3d::options::parse<bool>(args[0]);
   };
+
+  interactor.addCommand("print_config_info", [this](const std::vector<std::string>&)
+    { F3DConfigFileTools::ReadConfigFiles(std::string(), true, f3d::log::VerboseLevel::INFO); });
 
   interactor.addCommand("remove_file_groups",
     [this](const std::vector<std::string>&)


application/F3DConfigFileTools.cxx
```diff
@@ -146,15 +146,21 @@ F3DConfigFileTools::ParsedConfigFiles F3DConfigFileTools::ReadConfigFiles(
...
   if (!configSearch.empty() && actualConfigFilePaths.empty())
   {
-    f3d::log::info("Configuration file for \"", configSearch, "\" could not be found");
+    f3d::log::print(logLevel, "Configuration file for \"", configSearch, "\" could not be found");
+  }
+
+  // Clear config file paths to avoid reading them if dryRun is set
+  if (dryRun)
+  {
+    actualConfigFilePaths.clear();
```
mwestphal: insteaed you may want to just return ?

application/F3DConfigFileTools.h
```diff
@@ -25,7 +26,8 @@ struct ParsedConfigFiles
  * Read config files using userConfig if any, return a ParsedConfigFiles
  * containing ordered optionDict, ordered imperative optionDict and ordered bindingsEntries
  */
-ParsedConfigFiles ReadConfigFiles(const std::string& userConfig);
+ParsedConfigFiles ReadConfigFiles(
+  const std::string& userConfig, bool dryRun, f3d::log::VerboseLevel logLevel);
```
mwestphal: please document the new parameters

mwestphal (changes requested):
looks good! 

Please add a test though, you can look at

application/testing/CMakeLists.txt:1101

`TestCommandScriptPrintScene` and copy the logic.

snoyer:
Nice! thanks @NaniNoni

As it is, from a user point of view, this could answer "where are the current config files?" but not "where should I _create_ my config file?" as only the existing files are listed?

How about removing the `fs::exists` check from `GetConfigPaths` https://github.com/f3d-app/f3d/blob/3bec77e7be4867cd62fe3df7416b8204d38e16ab/application/F3DConfigFileTools.cxx#L66 and handling it in `ReadConfigFiles` where both existing and non-existing files could be logged accordingly?

application/F3DConfigFileTools.cxx
NaniNoni: Yes I will do that

NaniNoni:
> looks good!
> 
> Please add a test though, you can look at
> 
> application/testing/CMakeLists.txt:1101
> 
> `TestCommandScriptPrintScene` and copy the logic.

Thank you, will do

NaniNoni:
I fixed/implemented the requested changes. However, in the loop over canonical paths, I ended up using `debug` instead of `print(logLevel, ...)` due to the large amount of unwanted output when the config files were actually being read when f3d is launched. I think the level of output is appropriate now, but it just makes the `logLevel` parameter slightly misleading since it no longer applies to all internal log calls.

```c++
// Recover an absolute canonical path to config file
try
{
  configPath = fs::canonical(fs::path(configPath)).string();
}
catch (const fs::filesystem_error&)
{
  // This (file not found) error is expected in dry run mode
  if (dryRun)
  {
    f3d::log::print(logLevel, "Found available config location ", configPath.string());
  }
  else
  {
    f3d::log::debug(
      "Configuration file does not exist: ", configPath.string(), " , ignoring it");
  }
  continue;
}
```

application/F3DConfigFileTools.h
NaniNoni: :+1:

application/F3DConfigFileTools.cxx
```diff
@@ -62,16 +62,9 @@ std::vector<fs::path> GetConfigPaths(const std::string& configSearch)
...
-          f3d::log::debug("Config file found: ", configPath.string());
-          paths.emplace_back(configPath);
-        }
-        else
-        {
-          f3d::log::debug("Candidate config file not found: ", configPath.string());
-        }
+        configPath = dir / configName;
+        paths.emplace_back(configPath);
+        f3d::log::debug("Found potential config file: ", configPath.string());
```
mwestphal: So you changed the logic a bit so that GetConfigPath does not check existence.

TBH I dont think this is needed. You can control the logLevel instead so it doesnt debug only.

snoyer:
Would it be better to improve the separation of concerns between `GetConfigPaths` and `ReadConfigFiles` instead of trying to include the user-facing logging into `ReadConfigFiles` with a dry-run param?

The idea would be:
- `GetConfigPaths` returns both the candidate config files and search directories (whether they exist or not)
- `ReadConfigFiles` iterates over the result of `GetConfigPaths`, reading the items that are file files and recursing into the items that are dirs
- the `print_config_info` command uses only `GetConfigPaths` to print "found this files, would have used that file, would check that dir, ..." without having to do a full dry run of `ReadConfigFiles`

NaniNoni:
> Would it be better to improve the separation of concerns between `GetConfigPaths` and `ReadConfigFiles` instead of trying to include the user-facing logging into `ReadConfigFiles` with a dry-run param?
> 
> The idea would be:
> 
>     * `GetConfigPaths` returns both the candidate config files and search directories (whether they exist or not)
> 
>     * `ReadConfigFiles` iterates over the result of `GetConfigPaths`, reading the items that are file files and recursing into the items that are dirs
> 
>     * the `print_config_info` command uses only `GetConfigPaths` to print "found this files, would have used that file, would check that dir, ..." without having to do a full dry run of `ReadConfigFiles`

I like this idea a lot

NaniNoni:
I implemented the requested changes. The only downside I'm aware of is that the `print_config_info` command considers `"config"` as the only `configSearch` query for `GetConfigPaths`. I think this may cause an issue for different config locations.

```c++
std::vector<std::filesystem::path> availableConfigPaths =
        F3DConfigFileTools::GetConfigPaths("config");
```

snoyer:
> The only downside I'm aware of is that the `print_config_info` command considers `"config"` as the only `configSearch` query for `GetConfigPaths`

You'd have to track down where the `configSearch` param comes from for the real `GetConfigPaths` call and store that somewhere to reuse it in `print_config_info`. I don't know what the proper way to do it would but I'm sure @mwestphal will

application/F3DConfigFileTools.cxx
```diff
@@ -110,7 +95,7 @@ F3DConfigFileTools::ParsedConfigFiles F3DConfigFileTools::ReadConfigFiles(
   std::vector<fs::path> configPaths;
   if (!configSearch.empty())
   {
-    configPaths = ::GetConfigPaths(configSearch);
+    configPaths = GetConfigPaths(configSearch);
```
mwestphal:
```suggestion
    configPaths = F3DConfigFileTools::GetConfigPaths(configSearch);
```

application/F3DConfigFileTools.cxx
```diff
@@ -122,17 +107,14 @@ F3DConfigFileTools::ParsedConfigFiles F3DConfigFileTools::ReadConfigFiles(
   std::vector<fs::path> actualConfigFilePaths;
   for (auto configPath : configPaths)
   {
-    // Recover an absolute canonical path to config file
-    try
-    {
-      configPath = fs::canonical(fs::path(configPath)).string();
-    }
-    catch (const fs::filesystem_error&)
+    if (!fs::exists(configPath))
```
mwestphal: you need a try catch around this call

NaniNoni:
> I'm confused by this implementation. It looks to me this will not output anything about config files now ? We want the same or at least a similar output when `--verbose` is used.

Yes, you're right. That's another problem. Previously to these changes, the `GetConfigPaths` function checked for file existence and logged in `debug` if they didn't (because that was expected). Now, the existence check moved to `ReadConfigFiles`. `ReadConfigFiles` also had an existence check, but logged the results in `error` verbosity, since the files checked was supposed to exist at that point.

Now, the existence check in `ReadConfigFiles` cannot be an error, since it's sort of equivalent to the one we had before in `GetConfigPaths`. What do you think?

NaniNoni:
> > The only downside I'm aware of is that the `print_config_info` command considers `"config"` as the only `configSearch` query for `GetConfigPaths`
> 
> You'd have to track down where the `configSearch` param comes from for the real `GetConfigPaths` call and store that somewhere to reuse it in `print_config_info`. I don't know what the proper way to do it would but I'm sure @mwestphal will

Yes I agree. @mwestphal how would you like this to work?

mwestphal:
> > > The only downside I'm aware of is that the `print_config_info` command considers `"config"` as the only `configSearch` query for `GetConfigPaths`
> > 
> > 
> > You'd have to track down where the `configSearch` param comes from for the real `GetConfigPaths` call and store that somewhere to reuse it in `print_config_info`. I don't know what the proper way to do it would but I'm sure @mwestphal will
> 
> Yes I agree. @mwestphal how would you like this to work?

Alright, actually your implementation is not far off!

How about adding another method:

```
F3DConfigFileTools::PrintConfigInfo(const std::vector<fs::path>&)
```

that prints output using log::info like this by checking the file/dir existence:

```
Candidate config file not found: /etc/f3d/config.json
Candidate config file not found: /etc/f3d/config.d
Candidate config file not found: /usr/share/f3d/configs/config.json
Candidate config file not found: /usr/share/f3d/configs/config.d
Candidate config file not found: /home/glow/dev/f3d/others/F3D-3.2.0-RC2-Linux-x86_64-raytracing/share/f3d/confi
Config file found: /home/glow/dev/f3d/others/F3D-3.2.0-RC2-Linux-x86_64-raytracing/share/f3d/configs/config.d
Candidate config file not found: /home/glow/.config/f3d/config.json
Candidate config file not found: /home/glow/.config/f3d/config.d
```

And use it in `print_config_info` always  and in `F3DConfigFileTools::ReadConfigFiles` if `verbose == debug`.

mwestphal:
Hi @NaniNoni , how is it going ? Do you need any help ?

mwestphal:
Hi @NaniNoni , how is it going ? Do you need any help ?

NaniNoni:
Hi, sorry for the delay. I've been away and then unexpectedly very busy, but I've implemented the changes and it should hopefully be closer to the desired implementation. However, the quoted comment below is still unresolved. How would you like this to work?

> > The only downside I'm aware of is that the `print_config_info` command considers `"config"` as the only `configSearch` query for `GetConfigPaths`
> 
> You'd have to track down where the `configSearch` param comes from for the real `GetConfigPaths` call and store that somewhere to reuse it in `print_config_info`. I don't know what the proper way to do it would but I'm sure @mwestphal will

application/F3DConfigFileTools.h
```diff
@@ -21,6 +22,16 @@ struct ParsedConfigFiles
   BindingsEntries Bindings;
 };
 
+/**
+ * Recover a OS-specific vector of potential config file directories
+ */
+std::vector<std::filesystem::path> GetConfigPaths(const std::string& configSearch);
```
mwestphal, line 28:
```suggestion
std::vector<fs::path> GetConfigPaths(const std::string& configSearch);
```

application/F3DConfigFileTools.cxx
```diff
@@ -82,6 +68,21 @@ std::vector<fs::path> GetConfigPaths(const std::string& configSearch)
...
+  {
+    if (fs::exists(path))
+    {
+      f3d::log::info("Config file found: ", path);
+    }
+    else
+    {
+      f3d::log::info("Candidate config file not found: ", path);
+    }
+  }
```
snoyer, line 85: I'd say we can get rid of this method and log specifically in `ReadConfigFiles` and the `print_config_info` command

snoyer: I mean doing something like
```cpp
      for (const fs::path& path : F3DConfigFileTools::GetConfigPaths("config"))
      {
        f3d::log::info("- ", path, fs::exists(path) ? "(found)" : "(not found)");
      }
```
in the command and logging in the existing `if (!fs::exists(configPath))` branches in `ReadConfigFiles()` (and it may not need to be `info` level there?)

mwestphal: I dont think that works ?

mwestphal: ha right, sure, that works too :)

mwestphal:
@NaniNoni you need to rebase on latest master

mwestphal:
Hi @NaniNoni , need any help moving forward ?

mwestphal:
Hi @NaniNoni , need any help moving forward ?

mwestphal:
Hi @NaniNoni , need any help moving forward ?

PR #2355, opened 2025-07-03
[Feat] Add support for group skins from QuakMDL files

yashnator:
### Describe your changes

- Add support for group skins from QuakeMDL format
- Minor fixes and changes to only load the selected skins. Earlier, all the skins were loaded in memory even if they were discarded later

### Issue ticket number and link if any

Fixes #2137

### Checklist for finalizing the PR

- [x] I have performed a [self-review](https://f3d.app/doc/dev/CODING_STYLE.html) of my code
- [x] I have added [tests](https://f3d.app/doc/dev/TESTING.html) for new features and bugfixes
- [x] I have added [documentation](https://f3d.app/) for new features
- [ ] If it is a modifying the libf3d API, I have updated bindings
- [ ] If it is a modifying the `.github/workflows/versions.json`, I have updated `timestamp`

### Continuous integration

Please check the checkbox of the CI you want to run, then push again on your branch.

- [x] Style checks
- [x] Fast CI
- [x] Coverage cached CI
- [x] Analysis cached CI
- [x] WASM docker CI
- [x] Android docker CI
- [x] macOS Intel cached CI
- [x] macOS ARM cached CI
- [x] Windows cached CI
- [x] Linux cached CI
- [x] Other cached CI

PR #2397, opened 2025-07-11
[Fix] Render last frame of QuakeMDL animations

yashnator:
### Describe your changes

Extends animation range to render the last frame correctly. Earlier, the last frame was skipped since the animation range didn't account for its rendering time.

To be merged before PR #2355

- [X] I have performed a [self-review](https://f3d.app/doc/dev/CODING_STYLE.html) of my code
- [X] I have added [tests](https://f3d.app/doc/dev/TESTING.html) for new features and bugfixes

### Continuous integration

- [X] Style checks
- [X] Fast CI
- [X] Coverage cached CI
- [X] Analysis cached CI
- [X] WASM docker CI
- [X] Android docker CI
- [X] macOS Intel cached CI
- [X] macOS ARM cached CI
- [X] Windows cached CI
- [X] Linux cached CI
- [X] Other cached CI


plugins/native/module/vtkF3DQuakeMDLImporter.cxx
```diff
@@ -336,11 +336,8 @@ struct vtkF3DQuakeMDLImporter::vtkInternals
         {
           times.emplace_back(0.0);
         }
-        else
-        {
-          // Single frames are 10 fps
-          times.emplace_back(times.back() + 0.1);
-        }
+        // Single frames are 10 fps
+        times.emplace_back(times.back() + 0.1);
```
mwestphal, line 340: I'm not sure I understand this change. This means that when we a one single frame, it goes from 0 to 0.1  ? Whats the point ? there is not animation anyway ?

plugins/native/module/vtkF3DQuakeMDLImporter.cxx
```diff
@@ -354,6 +351,7 @@ struct vtkF3DQuakeMDLImporter::vtkInternals
         std::vector<double> times;
         std::vector<vtkSmartPointer<vtkPolyData>> meshes;
 
+        times.emplace_back(0.0);
```
mwestphal, line 355: why is this needed ? Not following.

plugins/native/module/vtkF3DQuakeMDLImporter.cxx
```diff
@@ -471,12 +469,14 @@ bool vtkF3DQuakeMDLImporter::UpdateAtTimeValue(double timeValue)
...
+    // First time >= value, excluding the last element as it only represents finish time of the last
+    // frame
+    const auto found = std::lower_bound(times.begin(), times.end() - 1, timeValue);
     // If none, select last, if found, select distance
-    const size_t i = found == times.end() ? times.size() - 1 : std::distance(times.begin(), found);
+    const size_t i =
+      (found == times.end() - 1) ? times.size() - 2 : std::distance(times.begin(), found);
     // If found time > timeValue, the the previous one
-    const size_t frameIndex = *found > timeValue && i > 0 ? i - 1 : i;
+    const size_t frameIndex = times[i] > timeValue && i > 0 ? i - 1 : i;
```
mwestphal, line 482: can you add more comments explaining the logic here ? I'm not following :)

plugins/native/module/vtkF3DQuakeMDLImporter.cxx
yashnator: Agreed. I missed single frames here. I can adjust this in `GetTemporalInformation` because we need to do this for animations with >1 frames to give some rendering time to the last frame.

plugins/native/module/vtkF3DQuakeMDLImporter.cxx
yashnator: The animation's duration points to the animation's end time. So, currently, the renderer takes the interval `[End time of first animation, End time of last animation]` instead of `[Start time of first animation, End time of last animation]`.

For example, in `flame_mixed.mdl`, with 0.0 at front we render:

https://github.com/user-attachments/assets/18268990-87b0-4e66-9319-ffff0cf8ddb9

And without 0.0 at the front, we render:

https://github.com/user-attachments/assets/b7bcd817-9b0f-400a-90df-c6cb4b922bde

mwestphal: ha got it.

plugins/native/module/vtkF3DQuakeMDLImporter.cxx
yashnator: We don't have a frame for the last time stamp; it's only there to show the end time of the last frame. We should ideally wrap around at end time of animation since frames are rendered for time ranges like `[L, R)`, with R excluded. However, I don't know how to represent open interval for animation range, so doing `*found` will compare to wrong value in the end.

Example- consider flame animation from `[0, 0.7]`. At t=0.65, we have`i=6`, `*found=0.7`(at `i + 1`). If I compare `*found > timeValue`, we subtract 1 from i again and get `i=5`, rendering the wrong frame.

Log if we compare with `*found`:
```
Rendered the frame: 6 with i: 6 at timeValue: 0.6
Rendered the frame: 5 with i: 6 at timeValue: 0.633333
Rendered the frame: 5 with i: 6 at timeValue: 0.666667
Rendered the frame: 6 with i: 6 at timeValue: 0.7
```

An alternative would be to set timeValue = 0 if input timeValue = end time of animation (or use `std::fmod`) to wrap around. I will add some comments

mwestphal: sounds good, lets add some of that in the comment for future us :)

yashnator: I have added some comments and made the changes. Let me know if something else needs to be modified as well

yashnator:
I have updated the `testing/baselines/TestQuakeMDLAnimationGroupFrame.png` for `TestQuakeMDLAnimationGroupFrame` test because the start times were incorrect earlier (the flame example above)

plugins/native/module/vtkF3DQuakeMDLImporter.cxx
mwestphal: I still dont get why you need to take this out of the else?

mwestphal (changes requested):
still a question

plugins/native/module/vtkF3DQuakeMDLImporter.cxx
```diff
@@ -531,6 +533,7 @@ bool vtkF3DQuakeMDLImporter::GetTemporalInformation(vtkIdType animationIndex,
   const std::vector<double>& times = this->Internals->AnimationTimes[animationIndex];
   // F3D does not care about timesteps, only set time range
   timeRange[0] = times.front();
-  timeRange[1] = times.back();
+  // If single frame, keep animation duration = 0
+  timeRange[1] = times.size() == 2 ? times.front() : times.back();
```
mwestphal, line 538: I dont think this is right. 

What if an animation has two frames ? It can be something else than a single frame.

yashnator: While parsing, we add `n+1` time stamps for an `n` frame animation because we added 0.0 initially on our own. So a two-frame animation will have 3 elements in `AnimationTimes`

yashnator: We still add 2 time entries for a single frame. I only added this behaviour since you suggested making the animation duration of single frames 0. Are there cases that I am missing?

mwestphal: ok I got it now, but it sounds a bit overengineered, ill think about it.

plugins/native/module/vtkF3DQuakeMDLImporter.cxx
mwestphal, line 355:
```suggestion
        // groupFrames always starts at 0.0
        times.emplace_back(0.0);
```

plugins/native/module/vtkF3DQuakeMDLImporter.cxx
mwestphal: ok I really got it now!

mwestphal (changes requested):
one small comment for clarity but I finally managed to understand what I did wrong innitially, and you fix is indeed the right way to go!

plugins/native/module/vtkF3DQuakeMDLImporter.cxx
yashnator: Done!

mwestphal:
Thanks for the contribution!

Issue #2101, opened 2025-03-28
comments and colors are not compatible in commands


mwestphal:
**Describe the bug**
It is supposed to be possible to set colors using a dedicated formating using a command
`set color #000000`

However `#` is the char for comments so the command above doesnt work and it is required to use:
`set color "#000000"` or `set color \#000000` to be able to set the color as expected.

**To Reproduce**

1.  Open `f3d`
2.  Press `Esc` to open the console
3.  `set render.background.color #000000`
4. Error

**Expected behavior**
No Error and background change

**Suggested Implementation**

A way to control the tokenizer and being able to disable comments or control quotes behavior would work, as this could be used in interactive console only.

mwestphal:
Fixed by #2469

PR #2469, opened 2025-09-13
Set HEX Color from command without required to escape it

Ni-g-3l:
### Describe your changes

Hi ! This patch should fix the #2101 issue. I added some tests in order to cover more test case. Let me know If you have other tests to add.

Feel free to comment ! 

* [comments and colors are not compatible in commands #2101
](https://github.com/f3d-app/f3d/issues/2101)

### Continuous integration

See [here](https://f3d.app/CONTRIBUTING.html#continuous-integration) for more info.

image

### DIFF
--- a/doc/user/COMMANDS.md
+++ b/doc/user/COMMANDS.md
@@ -148,6 +148,8 @@ Command syntax is similar to bash, as in they will be split by "token" to be pro
 - Quotes inside quotes are supported as well, eg: `set render.hdri.file "/path/to/file'with'quotes.png"`.
 - Quotes and spaces can be escaped, eg: `set render.hdri.file /path/to/file\ with\ spaces\ and\ \'quotes\".png`.
 - Comment are supported using `#`, Any character after will be ignored. Use `\#` to add it verbatim.
+> Comments are only supported in command script, in interactive console `#` and all characters after will be handled
+as standard character.
 - Escapes can be escaped too: eg: `set render.hdri.file C:\\path\\to\\windows\\file.png`.
 - Other escaped character will be processed as if the escape was not present, eg: `set scene.up.direction +\Z`
 - Unfinished quoted section is invalid, eg: `set scene.up.direction "+Z`
--- a/doc/user/COMMANDS.md
+++ b/doc/user/COMMANDS.md
@@ -148,8 +148,8 @@ Command syntax is similar to bash, as in they will be split by "token" to be pro
 - Quotes inside quotes are supported as well, eg: `set render.hdri.file "/path/to/file'with'quotes.png"`.
 - Quotes and spaces can be escaped, eg: `set render.hdri.file /path/to/file\ with\ spaces\ and\ \'quotes\".png`.
 - Comment are supported using `#`, Any character after will be ignored. Use `\#` to add it verbatim.
-> Comments are only supported in command script, in interactive console `#` and all characters after will be handled
-as standard character.
+  > Comments are only supported in command script, in interactive console `#` and all characters after will be handled
+  > as standard character.
 - Escapes can be escaped too: eg: `set render.hdri.file C:\\path\\to\\windows\\file.png`.
 - Other escaped character will be processed as if the escape was not present, eg: `set scene.up.direction +\Z`
 - Unfinished quoted section is invalid, eg: `set scene.up.direction "+Z`
--- a/library/public/interactor.h
+++ b/library/public/interactor.h
@@ -104,8 +104,8 @@ public:
    *
    * If the command fails, it prints a debug log explaining why.
    *
-   * When keepComments argument is true, comments are supported with `#`, any characters after are ignored
-   * otherwise '#' and any characters after will be handled as standard character
+   * When keepComments argument is true, comments are supported with `#`, any characters after are
+   * ignored otherwise '#' and any characters after will be handled as standard character
    *
    * Return true if the command succeeded, false otherwise.
    * Throw an interactor::command_runtime_exception if the command callback
--- a/library/public/interactor.h
+++ b/library/public/interactor.h
@@ -110,7 +110,7 @@ public:
    * Note that default commands will never throw this exception, but adding commands
    * without exception catching may trigger this behavior.
    */
-  virtual bool triggerCommand(std::string_view command) = 0;
+  virtual bool triggerCommand(std::string_view command, bool skipComment=false) = 0;
   ///@}
 
   ///@{ @name Bindings
--- a/library/public/interactor.h
+++ b/library/public/interactor.h
@@ -110,7 +110,7 @@ public:
    * Note that default commands will never throw this exception, but adding commands
    * without exception catching may trigger this behavior.
    */
-  virtual bool triggerCommand(std::string_view command, bool skipComment=false) = 0;
+  virtual bool triggerCommand(std::string_view command, bool skipComment = false) = 0;
   ///@}
 
   ///@{ @name Bindings
--- a/library/public/interactor.h
+++ b/library/public/interactor.h
@@ -104,13 +104,16 @@ public:
    *
    * If the command fails, it prints a debug log explaining why.
    *
+   * When keepComments argument is true, comments are supported with `#`, any characters after are ignored
+   * otherwise '#' and any characters after will be handled as standard character
+   *
    * Return true if the command succeeded, false otherwise.
    * Throw an interactor::command_runtime_exception if the command callback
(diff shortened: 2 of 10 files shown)


Ni-g-3l:
\ci fast

snoyer:
Not having to escape when entering colors would be really nice but I'm not sure about adding such a special case all the way down into the tokenization...
for example consider the following commands/script:
```
set value 1 #good
set value "1" #bad
```
obviously this is made up and contrived on purpose but the second "comment" is a valid hex string so it wouldn't actually be tokenized as a comment, right?

Are there alternative solutions in terms of design/impl to achieve the original goal?

If we have to change the tokenization maybe have something less specific like "comments start with `# ` (hash sign followed by space)" instead of just `#`? in the example above neither line would have a comment, and the proper commented form would be
```
set value 1 # good
set value "1" # bad
```

Or could we just change the color parser so that the `#` is optional and `ff00ff` is parsed the same as `#ff00ff`?

mwestphal:
I definitely agree, the fix should not be color specific here.

library/public/interactor.h
```diff
@@ -110,7 +110,7 @@ class F3D_EXPORT interactor
    * Note that default commands will never throw this exception, but adding commands
    * without exception catching may trigger this behavior.
    */
-  virtual bool triggerCommand(std::string_view command) = 0;
+  virtual bool triggerCommand(std::string_view command, bool skipComment = false) = 0;
```
mwestphal: needs a method doc update

library/public/utils.h
```diff
@@ -54,7 +54,7 @@ class F3D_EXPORT utils
    * `set scene.up.direction "+Z` -> tokenize_exception
    * `set scene.up.direction +Z\` -> tokenize_exception
    */
-  [[nodiscard]] static std::vector<std::string> tokenize(std::string_view str);
+  [[nodiscard]] static std::vector<std::string> tokenize(std::string_view str, bool skipComment=false);
```
mwestphal: needs a method doc update

library/src/utils.cxx
```diff
@@ -82,7 +82,7 @@ std::vector<std::string> utils::tokenize(std::string_view str)
         escaped = false;
         break;
       case '#':
-        if (!escaped && !quoted)
+        if (!escaped && !quoted && !skipComment)
```
mwestphal: Needs a doc update in COMMANDS.md

library/src/utils.cxx
```diff
@@ -25,7 +25,7 @@ unsigned int utils::textDistance(std::string_view strA, std::string_view strB)
 }
 
 //----------------------------------------------------------------------------
-std::vector<std::string> utils::tokenize(std::string_view str)
+std::vector<std::string> utils::tokenize(std::string_view str, bool skipComment)
```
snoyer: so `skipComment` skips the skipping of comments?
Should it be `keepComments` or `enableComments` instead?

mwestphal: yes, I agree.

doc/user/COMMANDS.md
```diff
@@ -148,6 +148,8 @@ Command syntax is similar to bash, as in they will be split by "token" to be pro
 - Quotes inside quotes are supported as well, eg: `set render.hdri.file "/path/to/file'with'quotes.png"`.
 - Quotes and spaces can be escaped, eg: `set render.hdri.file /path/to/file\ with\ spaces\ and\ \'quotes\".png`.
 - Comment are supported using `#`, Any character after will be ignored. Use `\#` to add it verbatim.
+  > Comments are only supported in command script, in interactive console `#` and all characters after will be handled
```
mwestphal, line 151: >  cite

mwestphal: I'm not sure about the use of `>` for that doc though

Ni-g-3l: Idk at all, I thought we should keep the list but add something like little "note" on comment. If you have a better idea no problem I can make the change

mwestphal: Actually the may look nice :), lets keep as is.

library/public/interactor.h
```diff
@@ -104,13 +104,16 @@ class F3D_EXPORT interactor
...
+   * ignored otherwise '#' and any characters after will be handled as standard character
+   *
    * Return true if the command succeeded, false otherwise.
    * Throw an interactor::command_runtime_exception if the command callback
    * throw an unrecognized exception.
    * Note that default commands will never throw this exception, but adding commands
    * without exception catching may trigger this behavior.
    */
-  virtual bool triggerCommand(std::string_view command) = 0;
+  virtual bool triggerCommand(std::string_view command, bool keepComments = true) = 0;
```
mwestphal, line 116: please ensure python bindings for this optional arg are working by adding a smoke test in `test_interactor.py`

library/public/utils.h
```diff
@@ -54,7 +55,7 @@ class F3D_EXPORT utils
    * `set scene.up.direction "+Z` -> tokenize_exception
    * `set scene.up.direction +Z\` -> tokenize_exception
    */
-  [[nodiscard]] static std::vector<std::string> tokenize(std::string_view str);
+  [[nodiscard]] static std::vector<std::string> tokenize(std::string_view str, bool keepComments = true);
```
mwestphal, line 58: looks like we forgot to bind tokenize in python, lets add it and add a test :)

doc/user/COMMANDS.md
snoyer: idk if that's somewhat standard makdown that jekyll/docusaurus would understand or if it's github specific but...

```md
> [!NOTE]   
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action.
```

> [!NOTE]   
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action.

mwestphal: Note looks nice.

library/public/utils.h
Ni-g-3l: tokenize added with dedicated test

doc/user/COMMANDS.md
Ni-g-3l: Oh yeah however it look like that it is not available on jekyll

https://www.markdownguide.org/tools/jekyll/

python/F3DPythonBindings.cxx
```diff
@@ -281,7 +282,8 @@ PYBIND11_MODULE(pyf3d, module)
       py::arg("callback"), py::arg("doc") = std::nullopt, py::arg("completionCallback") = nullptr)
     .def("remove_command", &f3d::interactor::removeCommand, "Remove a command")
     .def("get_command_actions", &f3d::interactor::getCommandActions, "Get all command actions")
-    .def("trigger_command", &f3d::interactor::triggerCommand, "Trigger a command")
+    .def("trigger_command", &f3d::interactor::triggerCommand, "Trigger a command",
+      py::arg("command"), py::arg("keepComments") = true)
```
snoyer:
```suggestion
      py::arg("command"), py::arg("keep_comments") = true)
```
:snake:

Ni-g-3l:
\ci full

library/public/utils.h
```diff
@@ -37,7 +37,8 @@ class F3D_EXPORT utils
    *  - Split by quoted section and remove the quotes
    *  - Supported quotes are: '"`
    *  - Use escaped \ quotes, spaces and escape to add them verbatim
-   *  - Comments are supported with `#`, any characters after are ignored
+   *  - When keepComments argument is true, comments are supported with `#`, any characters after are ignored
+   * otherwise '#' and any characters after will be handled as standard character
    *  - Use escaped \# to add it verbatim
```
mwestphal: this think this comment needs updating ?

mwestphal:
```suggestion
   *  - Use escaped \# to add it verbatim when using keepComments = true
```

?

mwestphal (approved):
nice!

snoyer (approved):

PR #2548, opened 2025-10-20
Fix more wasm issues

Meakk:
### Describe your changes

- Make `f3d::camera` accessors const
- Fix and add test for `f3d::engine` class api
- Fix and add test for `f3d::scene` class api
- Fix and add test for `f3d::camera` class api

### Continuous integration

### DIFF
--- a/webassembly/testing/utils.js
+++ b/webassembly/testing/utils.js
@@ -1,4 +1,4 @@
-import f3d from "../../dist/f3d.js";
+import f3d from "/src/dist/f3d.js";
 
 const utils = {
   assert: (condition, description) => {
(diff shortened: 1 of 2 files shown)


library/private/camera_impl.h
```diff
@@ -36,20 +36,20 @@ class camera_impl : public camera
    * Documented public API
    */
   camera& setPosition(const point3_t& pos) override;
-  point3_t getPosition() override;
-  void getPosition(point3_t& pos) override;
+  point3_t getPosition() const override;
```
Meakk, line 39: @mwestphal not sure how we miss that but that needs to be fixed.  
I think it's backward compatible.

mwestphal: it is indeed backward compat.

webassembly/testing/utils.js
```diff
@@ -20,19 +32,19 @@ const utils = {
...
 
     f3d(settings)
       .then(async (Module) => {
         // write in the wasm filesystem
-        const modelFile = await fetch("../../../testing/data/" + dataPath).then(
-          (b) => b.arrayBuffer(),
-        );
-        Module.FS.writeFile(dataPath, new Uint8Array(modelFile));
+        const modelFile = await fetch(
+          "../../../testing/data/" + args.data,
```
mwestphal: should that relative path be cmake configured instead ?

Meakk: Made it absolute path. It doesn't need to be configured because the path is known.

Issue #1803, opened 2024-12-18
Add CPU gaussian sorting


Meakk:
**Is your feature request related to a problem? Please describe.**
When changing the camera orientation, it is required to sort the gaussians by depth in order to have a correct result.
Sorting is currently executed on the GPU using compute shaders.  
However it's not supported on macOS and the GPU sorting can be slow on low-end GPUs.
Let's consider CPU sorting using [vtkDepthSortPolyData](https://vtk.org/doc/nightly/html/classvtkDepthSortPolyData.html)

exbluesbreaker:
I would like to work on this issue. @Meakk can you provide a bit more context in which scenarios this sorting is happening?

mwestphal:
Just visualize any splat file with gaussian splatting, eg:

```
 ./bin/f3d  ../src/testing/data/small.splat --point-sprites --point-sprites-type=gaussian --scalar-coloring --coloring-array=color --coloring-component=-2 --up=-Y
```

Meakk:
https://github.com/f3d-app/f3d/blob/9994d18147bcad77d351c3ce936d8f1ecb0043cd/vtkext/private/module/vtkF3DPointSplatMapper.cxx#L92
The gaussians are sorted by depth in camera direction in order to have proper alpha blending.  
It's done in a compute shader using a bitonic sort for performance reason. Unfortunately it doesn't work on macOS, and OpenGL ES backends (web and android), so we need a fallback cpu method even if it's slower.

exbluesbreaker:
@mwestphal @Meakk thanks a lot, I think it should be sufficient, I will start on implementation in near time

mwestphal:
Hi @exbluesbreaker 

As far as I understand you may not have time to work on this, should I unassign it ? :)

exbluesbreaker:
> Hi [@exbluesbreaker](https://github.com/exbluesbreaker)
> 
> As far as I understand you may not have time to work on this, should I unassign it ? :)

Yes, I think is better to unassign. Apart from me currently being too busy to work on issues, this particular issues definitely required some refinement of OpenGL/computer graphics behavior to understand processing here from my side. So somebody more knowledable is a better candidate to work on it.

mwestphal:
No worries on that regard, feel free to comment on it whenever you want to pick it up again :)

AoGao-Kedoka:
@exbluesbreaker are you still interested in this one? I would like to take a look into this issue if you are not interested in this one anymore

[...]

PR #2250, opened 2025-05-10
WIP: stochastic transparency support

Meakk:
### Describe your changes

Add support for stochastic transparency.
It's an alternative to the expensive depth peeling.
While it could be quite noisy, it's also way more efficient, and the noise can be reduce by SSAA or by the upcoming TAA.

### Issue ticket number and link if any

#2236 

### Checklist for finalizing the PR

- [ ] I have performed a [self-review](https://f3d.app/doc/dev/CODING_STYLE.html) of my code
- [ ] I have added [tests](https://f3d.app/doc/dev/TESTING.html) for new features and bugfixes
- [ ] I have added [documentation](https://f3d.app/) for new features
- [ ] If it is a modifying the libf3d API, I have updated bindings
- [ ] If it is a modifying the `.github/workflows/versions.json`, I have updated `timestamp`

### Continuous integration

Please check the checkbox of the CI you want to run, then push again on your branch.

- [ ] Style checks
- [ ] Fast CI
- [ ] Coverage cached CI
- [ ] Analysis cached CI
- [ ] WASM docker CI
- [ ] Android docker CI
- [ ] macOS Intel cached CI
- [ ] macOS ARM cached CI
- [ ] Windows cached CI
- [ ] Linux cached CI
- [ ] Other cached CI

PR #2593, opened 2025-11-05
CPU sorting pipeline

avaer:
This PR implements a cpu-based sorting pipeline. Fixes https://github.com/f3d-app/f3d/issues/1803.

I don't intend this to be merged, but hopefully it is useful code to someone if they need this feature.

### DIFF
--- a/vtkext/private/module/vtkF3DPointSplatMapper.cxx
+++ b/vtkext/private/module/vtkF3DPointSplatMapper.cxx
@@ -28,6 +28,8 @@
 #endif
 
 #include <sstream>
+#include <algorithm>
+#include <vector>
 
 //----------------------------------------------------------------------------
 class vtkF3DSplatMapperHelper : public vtkOpenGLPointGaussianMapperHelper
@@ -62,6 +64,7 @@ protected:
 
 private:
   void SortSplats(vtkRenderer* ren);
+  void SortSplatsCPU(vtkRenderer* ren);
 
   vtkNew<vtkShader> DepthComputeShader;
   vtkNew<vtkShaderProgram> DepthProgram;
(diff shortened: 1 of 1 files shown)


mwestphal:
> I don't intend this to be merged

Why not though ? :)

avaer:
@mwestphal This was written as a hack, I just needed to get correct splatting results on MacOS CLI.

Tbh I have no clue about the f3d codebase, or how this PR interacts with the rest of the app surface. I'm just a graphics person fixing my own use case.

Replacing the compute shader with an implementation that reads from GL and runs sort on CPU is going to be slow regardless, maybe not up to par for this project. Plus this is using STL with alloc on every sort/frame, not what I would do in a RT rendering pipeline. Do we want to multithread or desynchronize the render? I don't know.

 And I don't really have the bandwidth to do this PR correctly to the standards I would expect for my own projects :P.
 
This is such a tight change though, I think anyone who knows the codebase should be able to put this code in the right place and test to make sure it works.

mwestphal:
@avaer understood, thanks for sharing!

By the way we have a discord and a tight community of devs and users, you are welcome to join!
https://discord.f3d.app

@Meakk what do you think of this impl? Wanna finish it?

Meakk:
Thanks @avaer I will take a look.  
I'm also currently implementing stochastic transparency that works pretty well for gaussian splatting and is compatible with macOS since sorting isn't required anymore.  
See #2250 in case you want to track the feature.

mwestphal:
@Meakk should we keep this one open or not ?

Meakk:
Yes, it's better in order to not lose it.

Meakk:
Superseded by https://github.com/f3d-app/f3d/pull/2844

PR #2597, opened 2025-11-06
Improve TAA implementation

Meakk:
### Describe your changes

- Use observers to reset the history texture instead of inspecting the camera manually
- Properly accumulate the result instead of using 10% blending all the time
- Reset history texture when the window is resized

The accumulation fix is needed for the upcoming stochastic transparency feature.

### Continuous integration

### DIFF
--- a/vtkext/private/module/vtkF3DTAAResolvePass.cxx
+++ b/vtkext/private/module/vtkF3DTAAResolvePass.cxx
@@ -49,7 +49,7 @@ void vtkF3DTAAResolvePass::Render(const vtkRenderState* state)
     this->HistoryTexture->SetWrapT(vtkTextureObject::ClampToEdge);
     renderer->GetTiledSizeAndOrigin(&size[0], &size[1], &pos[0], &pos[1]);
     this->HistoryTexture->Allocate2D(size[0], size[1], 4, VTK_FLOAT);
-    this->HistoryIteration = 0;
+    this->ResetIterations();
   }
   this->HistoryTexture->Resize(size[0], size[1]);
 


Meakk:
@AoGao-Kedoka @mwestphal please review

vtkext/private/module/vtkF3DTAAResolvePass.cxx
```diff
@@ -119,47 +116,11 @@ void vtkF3DTAAResolvePass::Render(const vtkRenderState* state)
   this->ColorTexture->Deactivate();
   this->HistoryTexture->Deactivate();
   this->HistoryTexture->CopyFromFrameBuffer(pos[0], pos[1], size[0], size[1], size[0], size[1]);
-  this->HistoryInitialized = true;
+  this->HistoryIteration++;
```
AoGao-Kedoka: i think this part of the code is running per frame? I would clamp it to a relatively large value to avoid overflow. 1024/1025 = 0.9990 is already small enough.
```suggestion
this->HistoryIteration = std::min(this->HistoryIteration + 1, 1024);
```

Meakk: That makes sense! Alternatively, maybe we can stop rendering after 1024 iterations?

Meakk: I leave the stop of rendering after N iterations for future work, I think the current approach is fine.

AoGao-Kedoka: yes, sounds valid :thinking: , and we can theoretically reduce the iteration when velocity buffer is not 0 (changing blending factor) and keep the rendering going for future implementation. 

it would require `vtkF3DRenderer::Render` know the internal iteration status of `vtkF3DTaaResolvePass` right? Cause we want to also stop the jittering once the iteration reached.

AoGao-Kedoka:
LGTM!

vtkext/private/module/vtkF3DTAAResolvePass.cxx
```diff
@@ -119,47 +116,11 @@ void vtkF3DTAAResolvePass::Render(const vtkRenderState* state)
   this->ColorTexture->Deactivate();
   this->HistoryTexture->Deactivate();
   this->HistoryTexture->CopyFromFrameBuffer(pos[0], pos[1], size[0], size[1], size[0], size[1]);
-  this->HistoryInitialized = true;
+  this->HistoryIteration = std::min(this->HistoryIteration + 1, 1024);
```
mwestphal, line 125: why min ?

Meakk: See https://github.com/f3d-app/f3d/pull/2597#discussion_r2499867482

vtkext/private/module/vtkF3DTAAResolvePass.cxx
Meakk: Yes we have to find a way to get the iteration count in `vtkF3DRenderer` to only remove `RENDER_UI_ONLY` key if the iteration count is below the threshold.

mwestphal (approved):

Issue #2539, opened 2025-10-19
Open file dialog doesn't have focus on macOS


Meakk:
**Describe the bug**
When opening the open file dialog on macOS, the dialog doesn't have the focus and the user has to click on it.

**To Reproduce**

1.  Open F3D
2.  Press `CMD+O` (or `CTRL+O`)
3.  Press escape
4. The file dialog doesn't close
5. Close it by clicking on the close button
6. The console is opened (because escape is caught afterward by F3D instead of the dialog)

**Expected behavior**
At step 3, the file dialog should be closed.

**System Information:**

- OS: macOS
- GPU and GPU driver:

**F3D Information**
3.3.0-RC1

**Additional context**

PR #2603, opened 2025-11-08
App: fixed behaviour of focus for open file window

arcticsixxx:
### Describe your changes
- Added a macOS-specific file dialog implementation using NSOpenPanel to ensure the dialog receives focus and handles the Escape key correctly.
- Ensured the dialog properly closes on Escape and prevents the console from opening unintentionally after canceling the dialog.
- Linked the necessary macOS frameworks (UniformTypeIdentifiers) in CMake.

Fix https://github.com/f3d-app/f3d/issues/2539

### Continuous integration
\ci full

### DIFF
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -2250,7 +2250,7 @@ void F3DStarter::AddCommands()
       {
         for (const auto& ext : info.Extensions)
         {
-#if defined(MAC_OS_X_VERSION_MAX_ALLOWED) && MAC_OS_X_VERSION_MAX_ALLOWED >= 1090
+#ifdef __APPLE__
           filters.push_back(ext);
 #else
           filters.push_back("*." + ext);
@@ -2268,7 +2268,7 @@ void F3DStarter::AddCommands()
       std::optional<std::string> file = f3d::utils::getEnv("CTEST_OPEN_DIALOG_FILE");
       if (!file.has_value())
       {
-#if defined(MAC_OS_X_VERSION_MAX_ALLOWED) && MAC_OS_X_VERSION_MAX_ALLOWED >= 1090
+#ifdef __APPLE__
         F3DNSDelegate::ShowOpenFileDialog(cstrings.data(), cstrings.size());
 #else
         char* ptr = tinyfd_openFileDialog("Open File", nullptr, static_cast<int>(cstrings.size()),


mwestphal:
Hi @arcticsixxx 

That doesnt seem related to #2539 at all I'm afraid.

arcticsixxx:
Hi @mwestphal
Why do you think so? 
I have reproduced this bug and write fix or wdym?

arcticsixxx:
\ci fast

arcticsixxx:
\ci full

mwestphal:
Your implementation is macos specific. The point of using tinyfiledialog is to avoid having to implement custom version of file dialog.

In any case, I defer to @Meakk as I do not know much about MacOS platform.

mwestphal:
@arcticsixxx btw we have a discord if you want :)

https://discord.f3d.app

arcticsixxx:
@mwestphal The issue was marked as macOS-specific, so I wrote macOS-specific code. :)
I tried patching tinyfiledialogs, but it uses the old osascript API for creating dialog windows, which doesn’t work on macOS versions > 9, and also fails to handle window focus correctly.

mwestphal:
> @mwestphal The issue was marked as macOS-specific, so I wrote macOS-specific code. :) I tried patching tinyfiledialogs, but it uses the old osascript API for creating dialog windows, which doesn’t work on macOS versions > 9, and also fails to handle window focus correctly.

Got it! Please join the discord so we can discuss how to best adress this issue :)

application/CMakeLists.txt
```diff
@@ -157,6 +157,7 @@ function(f3d_create_exec target_name)
 
   if(APPLE)
     target_link_libraries(${target_name} PRIVATE "-framework Cocoa")
+    target_link_libraries(${target_name} PRIVATE "-framework UniformTypeIdentifiers")
```
Meakk: You can merge these two lines

application/F3DStarter.cxx
```diff
@@ -2250,7 +2250,11 @@ void F3DStarter::AddCommands()
       {
         for (const auto& ext : info.Extensions)
         {
+#if defined(MAC_OS_X_VERSION_MAX_ALLOWED) && MAC_OS_X_VERSION_MAX_ALLOWED >= 101900
```
Meakk: `101900` means macOS 10.19? How did you choose that value?

application/F3DNSDelegate.mm
```diff
@@ -125,13 +127,66 @@ - (void)applicationDidFinishLaunching:(NSNotification *)notification
...
+      }
+      self.Starter->Render();
+    }
+  }
+}
+
 @end
 
 // ----------------------------------------------------------------------------
+static F3DNSDelegateInternal* delegate = nullptr;
```
Meakk: do we really need a static variable here?  
We are setting the delegate with `[NSApp setDelegate:delegate];`, can't we just get it from NSApp?

application/F3DStarter.cxx
arcticsixxx: my bad, it should be 1090, not 101900
you can find this value in tinyfiledialogs sources in function
`static int osx9orBetter(void)`

application/CMakeLists.txt
arcticsixxx: fixed

application/F3DNSDelegate.mm
arcticsixxx: yep, fixed

application/F3DStarter.cxx
Meakk: I think our official support is macOS >= 10.15 so that means this test is not required, right?

arcticsixxx: fixed

PR #2621, opened 2025-11-13
Python typing improvements

snoyer:
### Describe your changes

we've been letting `pybind11-stubgen` generate stubs for the `__init__.py` file but, while it's happy to do it, it's not very good at it. 
Run `pybind11-stubgen` only on the `f3d.pyf3d` submodule which contains the bindings give better results.
While we're at it we can also avoid exposing local variables from `__init__.py` and do some minor style/typing fixes in the tests.

There still some improvements to be made with the post-processing of the stubs but I don't think it can be done with the current regex approach so it will be a separate PR for deeper refactoring

### Continuous integration

### DIFF
--- a/python/testing/test_utils.py
+++ b/python/testing/test_utils.py
@@ -10,7 +10,11 @@ def test_text_distance():
 
 def test_collapse_path():
     assert f3d.Utils.collapse_path("/folder/../file.ext", ".") == Path("/file.ext")
-    assert f3d.Utils.collapse_path("/folder/../file.ext", ".").is_absolute()
+
+    # ensure we're getting a `Path` back
+    assert isinstance(f3d.Utils.collapse_path("/folder/../file.ext", Path(".")), Path)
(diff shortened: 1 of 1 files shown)


snoyer:
\ci full

python/testing/test_utils.py
```diff
@@ -30,11 +35,11 @@ def test_glob_to_regex():
 
 
 def test_get_env():
-    assert f3d.Utils.get_env("F3D_TEST_ENV_EMPTY") == None
+    assert f3d.Utils.get_env("F3D_TEST_ENV_EMPTY") is None
```
mwestphal, line 38: is that different ?

Meakk: probably just more pythonic

snoyer: short answer: same thing but more pythonic (and typeckers nudge you into doing it that way)

longer answer: it's always faster, and in some esoteric cases it may not be the same thing (classes can override `__eq__` and make it so `x == None` even if `x is not None`)

PR #2629, opened 2025-11-15
use AST transformer to fix generated stubs

snoyer:
Currently we generate the stubs for the Python bindings using `pybind11-stubgen` and automatically apply some corrections by regexping through the code. This works for the most part but there are some cases where we need finer control.

For example `pybind-stubgen` hints `fs::path` types as `PathLike`, which we substitute to `PathLike[str]` to make it more correct but we still get type-checking errors on valid code, see the following test which runs correctly but is full of squiggles in my IDE:
image 
~~this is because `pybind` will convert `fs::path` _to_ `pathlib.Path` but _from_ either `PathLike` or `str`. So we want to change `pybind-stubgen`'s `PathLike` to `PathLike | str` for arguments but to `Path` for returns. Currently we just change it to `PathLike` everywhere, hence the above warnings~~ I was running an old version of `pybind`, tne new version F3D uses does better here, it just needs `PathLike` changed to `PathLike[str]`.

Similarly, `pybind` will convert `std::vector<T>` _to_ `list[T]` but will also convert it _from_ `Sequence[T]` so if we have a `std::vector<int> -> std::vector<int>` C++ function we want its binding to be typed `Sequence[int] -> list[int]` because both `f([1,2])` and `f((1,2))` work (list input vs tuple input).

In order to be able to specify different substitutions for argument and return types we can use Python's built-in AST parser to narrow down the bits we apply the regexes to.

---

I've also added some tests, they were mostly for my own debugging and research but they should be a good enough illustration of what `generate_stubs.py` does.
They are not in `python/testing` because the don't apply to the package we ship ~~and so they currently aren't ran by the CI. They probably should be ran but I'm not sure when/how~~.

### DIFF
--- a/python/test_generate_stubs.py
+++ b/python/test_generate_stubs.py
@@ -50,6 +50,7 @@ def test_transformer_f3d_lists():
     """
     assert postprocess_generated_stub(dedent(src)) == parse_unparse(expected)
 
+
 def test_transformer_f3d_void_callbacks():
     src = """
     class A:


snoyer:
\ci full

python/generate_stubs.py
```diff
@@ -104,5 +183,5 @@ def retrieve_changed_files(directory: Path, files_glob: str):
     )
 
 
-if __name__ == "__main__":
+if __name__ == "__main__":  # pragma: nocover
```
mwestphal, line 194: what does that pragma do ?

snoyer: it excludes the block from the test coverage stats so we don't get a miss from the following `   main()` line.
The `main()` function is tested and covered explicitly be we can't test this call (at least I don't think we can without doing some hacky stuff)

mwestphal (approved):

snoyer:
diffs look good now, and `basedpyright` is happy locally

mwestphal (approved):

PR #2633, opened 2025-11-15
Fix: Reject non-ASCII filenames to prevent Windows parsing issues 

nikhil2004-blip:


snoyer:
This PR:
- claims to "avoid Windows API parsing issues" and links to a Linux-related display issue
- says it's "using std::all_of" and the code has no `all_of`
- has random formatting changes including removing all comments from an unrelated piece of code
- pushes a random `.exe`

what is the deal?

---

Now, giving it the benefit of the doubt and ignoring the above points...

Not loading the file at all because the filename wouldn't display well is a bit harsh, isn't it?
According to the linked issue, this is a visual text display bug not a critical file loading bug. Part of the problem is likely in VTK, and the other part has to do with the font selection.

IMO, _if_ anything should be done in F3D, it would be better to just remove or replace the non-ascii characters before setting the window title

mwestphal:
I was told LLM were better than me at code, turns out not yet ^^

@nikhil2004-blip If you truly want to contribute to f3d, please join our discord, we can help, but do not open LLM generated PRs or you will get banned.

PR #2634, opened 2025-11-16
CLIOptions: Rework logic for parsing custom options

mwestphal:
### Describe your changes
 - Handle custom mapped options properly so that interactive options still work with it
 - Add tests
 - Fix a small vtk.module issue

### Continuous integration

### DIFF
--- a/application/testing/CMakeLists.txt
+++ b/application/testing/CMakeLists.txt
@@ -328,6 +328,7 @@ f3d_test(NAME TestAnimationInputChangePointSprites DATA v_rock2.mdl ARGS --point
 f3d_test(NAME TestBackdropOpacityOpaque DATA suzanne.ply ARGS -n --backdrop-opacity=1.0 UI)
 f3d_test(NAME TestBackdropOpacityTransparent DATA suzanne.ply ARGS -n --backdrop-opacity=0.0 UI)
 f3d_test(NAME TestBackdropOpacityMedium DATA suzanne.ply ARGS -n --backdrop-opacity=0.5 UI)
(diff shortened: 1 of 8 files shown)


application/testing/CMakeLists.txt
```diff
@@ -1033,6 +1034,7 @@ f3d_test(NAME TestInteractionFocalPointPickingPoints DATA pointsCloud.vtp INTERA
 f3d_test(NAME TestInteractionLightIntensity DATA dragon.vtu INTERACTION LONG_TIMEOUT)
 f3d_test(NAME TestInteractionMultiFileColoring DATA mb/recursive ARGS --multi-file-mode=all INTERACTION) #SSSB
 f3d_test(NAME TestInteractionOpacity DATA dragon.vtu INTERACTION)
+f3d_test(NAME TestInteractionAndCLIBlending ARGS --blending DATA suzanne.stl red_translucent_monkey.gltf INTERACTION)#P;Right : use "sort" then switch file and show "sort" is still used
```
Meakk: `sort` has no effect on polygon based geometries so if switching file resets to `none`, you won't see the difference. Let's add one more `P` to toggle `stochastic`.

mwestphal: except `stochastic` looks a lot like `ddp` which means that if pressing `p` has no effect, CI wont detect it.

Using `sort` was volontary here.

Alternatively I could `--blending=none` and press `p` once.

Meakk: Or you can use a model with a more obvious opacity like a format with no material and use `--opacity=0.5`.  
Anyhow, we should clearly see what's the blending mode used in the baseline.

Issue #2009, opened 2025-02-22
Create dedicated framework integration examples


mwestphal:
**Context**
the libf3d can be used for a quick and easy 3d viewer, but how can it be integrating into existing framework ?
Let's create example showing how to.

**Suggested solution**
Create dedicated framework integration examples, demonstrating how to integrate the libf3d with well known open source frameworks such as Gnome, KDE or Unity

There is an existing code, Exhibit, that integrates libf3d into Gnome using python: https://github.com/Nokse22/Exhibit

Related to: https://github.com/f3d-app/f3d/issues/1670

mwestphal:
This is fixed! Qt6, QML and fltk!

PR #2664, opened 2025-11-24
libf3d + Qt6 integration example

noclone:
### Describe your changes

- Add qt6 integration example in c++ and python with event forwarding

https://github.com/f3d-app/f3d/issues/2009

### Continuous integration

### DIFF
--- a/examples/libf3d/cpp/CMakeLists.txt
+++ b/examples/libf3d/cpp/CMakeLists.txt
@@ -20,7 +20,6 @@ if (F3D_EXAMPLES_EXTERNAL_QT)
   add_subdirectory(qt6)
 endif ()
 
-
 if(BUILD_TESTING)
   add_subdirectory(tests)
 endif()
--- a/examples/libf3d/cpp/external-glfw/main.cxx
+++ b/examples/libf3d/cpp/external-glfw/main.cxx
@@ -4,6 +4,7 @@
 
 #include <GLFW/glfw3.h>
 
+#include <chrono>
 #include <iostream>
 
 f3d::interactor::MouseButton translateButton(int button)
@@ -308,6 +309,13 @@ int main(int argc, char* argv[])
     return EXIT_FAILURE;
   }
 
+  int timeout = 0;
+  if (argc > 2)
+  {
+    timeout = std::atoi(argv[2]);
+  }
+  auto startTime = std::chrono::steady_clock::now();
+
   // setup glfw window
   if (!glfwInit())
   {
@@ -334,77 +342,93 @@ int main(int argc, char* argv[])
 
   glfwMakeContextCurrent(window);
 
-  f3d::engine::autoloadPlugins();
+  {
+    f3d::engine::autoloadPlugins();
 
-  f3d::engine eng = f3d::engine::createExternal(glfwGetProcAddress);
-  eng.getWindow().setSize(defaultSize[0], defaultSize[1]);
+    f3d::engine eng = f3d::engine::createExternal(glfwGetProcAddress);
+    eng.getWindow().setSize(defaultSize[0], defaultSize[1]);
 
-  glfwSetWindowUserPointer(window, &eng);
+    glfwSetWindowUserPointer(window, &eng);
 
-  // resize callback
-  glfwSetFramebufferSizeCallback(window,
-    [](GLFWwindow* window, int width, int height)
-    {
-      f3d::engine* pe = static_cast<f3d::engine*>(glfwGetWindowUserPointer(window));
-      pe->getWindow().setSize(width, height);
-    });
-
-  // key callback
-  glfwSetKeyCallback(window,
-    [](GLFWwindow* window, int key, int scancode, int action, int mods)
-    {
-      f3d::engine* pe = static_cast<f3d::engine*>(glfwGetWindowUserPointer(window));
-      pe->getInteractor().triggerModUpdate(translateMod(mods));
-      pe->getInteractor().triggerKeyboardKey(translateAction(action), translateKey(key));
-    });
-
-  glfwSetCharCallback(window,
-    [](GLFWwindow* window, unsigned int codepoint)
-    {
-      f3d::engine* pe = static_cast<f3d::engine*>(glfwGetWindowUserPointer(window));
-      pe->getInteractor().triggerTextCharacter(codepoint);
-    });
-
-  // mouse callback
-  glfwSetMouseButtonCallback(window,
-    [](GLFWwindow* window, int button, int action, int mods)
-    {
-      f3d::engine* pe = static_cast<f3d::engine*>(glfwGetWindowUserPointer(window));
-      pe->getInteractor().triggerModUpdate(translateMod(mods));
-      pe->getInteractor().triggerMouseButton(translateAction(action), translateButton(button));
-    });
-
-  glfwSetCursorPosCallback(window,
-    [](GLFWwindow* window, double xpos, double ypos)
-    {
-      f3d::engine* pe = static_cast<f3d::engine*>(glfwGetWindowUserPointer(window));
-      pe->getInteractor().triggerMousePosition(xpos, ypos);
-      pe->getWindow().render();
-      glfwSwapBuffers(window);
-    });
-
-  glfwSetScrollCallback(window,
-    [](GLFWwindow* window, double xoffset, double yoffset)
-    {
-      if (yoffset != 0.0)
+    // resize callback
+    glfwSetFramebufferSizeCallback(window,
+      [](GLFWwindow* window, int width, int height)
       {
         f3d::engine* pe = static_cast<f3d::engine*>(glfwGetWindowUserPointer(window));
-        pe->getInteractor().triggerMouseWheel(yoffset > 0
-            ? f3d::interactor::WheelDirection::FORWARD
-            : f3d::interactor::WheelDirection::BACKWARD);
+        pe->getWindow().setSize(width, height);
+      });
+
+    // key callback
+    glfwSetKeyCallback(window,
+      [](GLFWwindow* window, int key, int scancode, int action, int mods)
+      {
+        f3d::engine* pe = static_cast<f3d::engine*>(glfwGetWindowUserPointer(window));
+        pe->getInteractor().triggerModUpdate(translateMod(mods));
+        pe->getInteractor().triggerKeyboardKey(translateAction(action), translateKey(key));
+      });
+
+    glfwSetCharCallback(window,
+      [](GLFWwindow* window, unsigned int codepoint)
+      {
+        f3d::engine* pe = static_cast<f3d::engine*>(glfwGetWindowUserPointer(window));
+        pe->getInteractor().triggerTextCharacter(codepoint);
+      });
+
+    // mouse callback
+    glfwSetMouseButtonCallback(window,
+      [](GLFWwindow* window, int button, int action, int mods)
+      {
+        f3d::engine* pe = static_cast<f3d::engine*>(glfwGetWindowUserPointer(window));
+        pe->getInteractor().triggerModUpdate(translateMod(mods));
+        pe->getInteractor().triggerMouseButton(translateAction(action), translateButton(button));
+      });
+
+    glfwSetCursorPosCallback(window,
+      [](GLFWwindow* window, double xpos, double ypos)
+      {
+        f3d::engine* pe = static_cast<f3d::engine*>(glfwGetWindowUserPointer(window));
+        pe->getInteractor().triggerMousePosition(xpos, ypos);
+        pe->getWindow().render();
         glfwSwapBuffers(window);
+      });
+
+    glfwSetScrollCallback(window,
+      [](GLFWwindow* window, double xoffset, double yoffset)
+      {
+        if (yoffset != 0.0)
+        {
+          f3d::engine* pe = static_cast<f3d::engine*>(glfwGetWindowUserPointer(window));
+          pe->getInteractor().triggerMouseWheel(yoffset > 0
+              ? f3d::interactor::WheelDirection::FORWARD
+              : f3d::interactor::WheelDirection::BACKWARD);
+          glfwSwapBuffers(window);
+        }
+      });
+
+    eng.getScene().add(argv[1]);
+
+    while (!glfwWindowShouldClose(window))
+    {
+      // For testing purposes only, shutdown the example after `timeout` seconds
+      if (timeout > 0)
+      {
+        auto now = std::chrono::steady_clock::now();
+        auto elapsed = std::chrono::duration_cast<std::chrono::seconds>(now - startTime).count();
+
+        if (elapsed >= timeout)
+        {
+          glfwSetWindowShouldClose(window, GLFW_TRUE);
+          break;
+        }
       }
-    });
 
-  eng.getScene().add(argv[1]);
-
-  while (!glfwWindowShouldClose(window))
-  {
-    eng.getWindow().render();
-    glfwSwapBuffers(window);
-    glfwWaitEvents();
+      eng.getWindow().render();
+      glfwSwapBuffers(window);
+      glfwPollEvents();
+    }
   }
 
+  glfwDestroyWindow(window);
   glfwTerminate();
 
   return EXIT_SUCCESS;
--- a/examples/libf3d/cpp/qt6/CMakeLists.txt
+++ b/examples/libf3d/cpp/qt6/CMakeLists.txt
@@ -5,6 +5,8 @@ project(minimal_qt6)
 find_package(f3d REQUIRED COMPONENTS library)
 find_package(Qt6 REQUIRED COMPONENTS Widgets OpenGLWidgets)
 
+# This example only works if libf3d was built against a VTK >= 9.3.20240914
+
 add_executable(minimal_qt6 main.cxx)
 target_link_libraries(minimal_qt6 f3d::libf3d Qt6::Widgets Qt6::OpenGLWidgets)
 set_target_properties(minimal_qt6 PROPERTIES CXX_STANDARD 17)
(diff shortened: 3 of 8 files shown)


noclone:
\ci fast

examples/libf3d/cpp/qt6/main.cxx
```diff
@@ -0,0 +1,446 @@
...
+      this->Engine = std::make_unique<f3d::engine>(f3d::engine::createExternalGLX());
+    }
+#elif defined(Q_OS_WIN)
+    this->Engine = std::make_unique<f3d::engine>(f3d::engine::createExternalWGL());
+#elif defined(Q_OS_MACOS)
+    this->Engine = std::make_unique<f3d::engine>(f3d::engine::createExternalCOCOA());
+#else
+    std::cerr << "Unsupported platform for this example" << '\n';
+    return;
+#endif
```
Meakk: Is it possible to use QT abstraction directly?
```suggestion
    f3d::context::function loadFunc = [this](const char* name) {
      return this->context()->getProcAddress(name);
    };

    this->Engine = std::make_unique<f3d::engine>(f3d::engine::createExternal(loadFunc));
```

noclone: indeed, however createExternal is not in the python bindings, is that on purpose ?

Meakk: Is there something equivalent to `this->context()->getProcAddress` for Python?

noclone: I think self.context().get_proc_address exists

library/public/interactor.h
```diff
@@ -352,6 +352,16 @@ class F3D_EXPORT interactor
   virtual interactor& triggerTextCharacter(unsigned int codepoint) = 0;
   ///@}
 
+  /**
+   * Manually configure the event loop with a given refresh rate.
+   */
+  virtual interactor& configureEventLoop(double loopTime) = 0;
```
mwestphal: I dont get the usecase for this method

noclone: Since we're not calling `StartEventLoop` from the Qt example, we need a method to configure the delta time of the UI and animation

Meakk: If the user has to call `triggerEventLoop`, then you do not need a delta time since it's managed by the user directly? Am I missing something?

noclone: I'm not sure to understand what you said.
Basically the way I did it, the user creates its own timer with its own delta time Qt-side. Then the timer calls `triggerEventLoop` every `delta time` seconds. However, the animation manager in libf3d has its own delta time, which we need to set to the same `delta time` as the one the user chose.

examples/libf3d/cpp/qt6/main.cxx
```diff
@@ -0,0 +1,431 @@
...
+protected:
+  // ------------------------------------------------------------------
+  // OpenGL / F3D setup
+  // ------------------------------------------------------------------
+
+  void initializeGL() override
+  {
+    f3d::engine::autoloadPlugins();
+
+    const QString platform = QGuiApplication::platformName().toLower();
```
Meakk: not needed anymore

noclone:
\ci full

library/src/animationManager.cxx
```diff
@@ -166,11 +159,11 @@ void animationManager::ToggleAnimation()
 }
 
 //----------------------------------------------------------------------------
-void animationManager::Tick()
+void animationManager::Tick(double loopTime)
 {
   if (this->Playing)
   {
-    this->CurrentTime += this->DeltaTime * this->SpeedFactor;
```
Meakk, line 173: you probably want to remove `DeltaTime` from the header too.

examples/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 3.19)
 project(examples)
 
 find_package(f3d REQUIRED)
+find_package(VTK REQUIRED)
```
mwestphal: I dont think this should be needed ?

examples/libf3d/cpp/tests/CMakeLists.txt
```diff
@@ -45,3 +45,8 @@ set_tests_properties(test_use-options-struct PROPERTIES
 add_test(NAME test_use-options-variant COMMAND "$<TARGET_FILE:use-options-variant>" "${CMAKE_CURRENT_SOURCE_DIR}/data/cow.vtp" 1)
 set_tests_properties(test_use-options-variant PROPERTIES
         PASS_REGULAR_EXPRESSION "Number of points: 2903")
+
+# qt6
+if (VTK_VERSION VERSION_GREATER_EQUAL 9.3.20240914 AND F3D_EXAMPLES_EXTERNAL_QT)
+  add_test(NAME test_minimal_qt6 COMMAND "$<TARGET_FILE:minimal_qt6>" "${CMAKE_CURRENT_SOURCE_DIR}/data/cow.vtp" "--timeout" 1)
+endif ()
```
mwestphal, line 57: I dont see the glfw example here, is it missing ?

library/private/animationManager.h
```diff
@@ -80,15 +80,10 @@ class animationManager
...
-   */
-  void SetDeltaTime(double deltaTime);
-
-  /**
-   * Advance animationTime of DeltaTime and call loadAtTime accordingly
+   * Advance animationTime of loopTime and call loadAtTime accordingly
    * Do nothing if IsPlaying is false
    */
-  void Tick();
+  void Tick(double loopTime);
```
mwestphal:
```suggestion
  void Tick(double deltaTime);
```

mwestphal: everywhere

library/src/interactor_impl.cxx
```diff
@@ -1255,6 +1255,14 @@ std::vector<std::string> interactor_impl::getCommandActions() const
   return actions;
 }
 
+//----------------------------------------------------------------------------
+interactor& interactor_impl::triggerEventLoop(double loopTime)
+{
+  this->Internals->DeltaTime = loopTime;
```
mwestphal: we could even avoid this member by passing deltaTime as an arg of EventLoop

library/src/animationManager.cxx
```diff
@@ -14,7 +14,6 @@
 #include <vtkProgressBarRepresentation.h>
 #include <vtkRenderWindow.h>
 #include <vtkRendererCollection.h>
-#include <vtkVersion.h>
```
mwestphal, line 17: needed

mwestphal (changes requested):
changes needed

library/src/interactor_impl.cxx
Meakk: I thought about that, but I'm not sure it's possible because you have to retrieve the value from the callback when using `.start()`  
If possible that'd be better indeed.

examples/libf3d/cpp/CMakeLists.txt
```diff
@@ -15,6 +15,12 @@ if(F3D_EXAMPLES_EXTERNAL_GLFW)
   add_subdirectory(external-glfw)
 endif()
 
+option(F3D_EXAMPLES_EXTERNAL_QT "Enable external-qt example (requires qt)" OFF)
+if (F3D_EXAMPLES_EXTERNAL_QT)
+  add_subdirectory(qt6)
+endif ()
+
+
```
Meakk: remove one blank line here

library/src/animationManager.cxx
Meakk: if the CI is green I suppose it's not. It's indirectly including by `vtkF3DRenderer.h` I think

library/src/interactor_impl.cxx
noclone: indeed, I didn't find a way to retrieve the value from the callback

examples/CMakeLists.txt
noclone: it brings the VTK_VERSION cmake variable, it there another way ?

examples/libf3d/cpp/tests/CMakeLists.txt
noclone: indeed, i'll add it

examples/CMakeLists.txt
mwestphal: lets not use VTK_VERSION but add a comment in the example in the CMakeLists mentionning the limitation and only enable F3D_EXAMPLES_EXTERNAL_QT in the right CI jobs.

library/src/interactor_impl.cxx
mwestphal: Indeed, you cant, however, slighly better would be to do somthing like this:

```
    this->CallbackDeltaTime = deltaTime;
    vtkNew<vtkCallbackCommand> timerCallBack;
    timerCallBack->SetCallback(
      [](vtkObject*, unsigned long, void* clientData, void*)
      {    
        internals* that = static_cast<internals*>(clientData);
        that->EventLoop(that->CallbackDeltaTime);
      });
```

This way its cleat that this member is only used for transferring through the callback API.

library/src/animationManager.cxx
mwestphal: yes, and indirect include are not what we want, we (try to) include what we use

examples/libf3d/python/qt6/test_minimal_qt6.py
```diff
@@ -0,0 +1,17 @@
+import os
+import sys
+from pathlib import Path
+
+import pytest
+
+from minimal_qt6 import main
+
+pytestmark = pytest.mark.skipif(
+    os.environ.get("F3D_EXAMPLES_EXTERNAL_QT", "").upper() != "ON",
```
mwestphal: I dont see this being defined in the env ?
Also is there a bettee way to do this ?

@snoyer

library/public/interactor.h
```diff
@@ -352,6 +352,11 @@ class F3D_EXPORT interactor
   virtual interactor& triggerTextCharacter(unsigned int codepoint) = 0;
   ///@}
 
+  /**
+   * Manually trigger the event loop.
```
mwestphal, line 356: please doc `deltaTime`

library/src/interactor_impl.cxx
```diff
@@ -600,7 +594,11 @@ class interactor_impl::internals
       this->CommandBuffer.reset();
     }
 
-    this->AnimationManager->Tick();
+    this->AnimationManager->Tick(deltaTime);
+
+    vtkRenderWindow* renWin = this->Window.GetRenderWindow();
+    vtkF3DRenderer* ren = vtkF3DRenderer::SafeDownCast(renWin->GetRenderers()->GetFirstRenderer());
+    ren->SetUIDeltaTime(deltaTime);
```
mwestphal, line 601: hum, I wonder if this may have performance impact. I suppose the setter check the value is actually different ?

wdyt @Meakk ?

examples/libf3d/python/qt6/test_minimal_qt6.py
noclone: right, not sure what I was thinking, F3D_EXAMPLES_EXTERNAL_QT is not defined in the env.
Not sure what do here, the purpose of this check is to avoid running the this test in offscreen environments

library/src/interactor_impl.cxx
noclone: it doesn't, but I get we could add it
```cpp
void vtkF3DImguiActor::SetDeltaTime(double time)
{
  ImGuiIO& io = ImGui::GetIO();
  io.DeltaTime = time;
}

Comment (Meakk) [reply library/src/interactor_impl.cxx:601]: It's specified in imgui documentation that it's fine to change the value every frame.

File: examples/libf3d/python/qt6/test_minimal_qt6.py (hunk 11 above)
Comment (noclone) [reply examples/libf3d/python/qt6/test_minimal_qt6.py]: maybe we could add something like this in the CI so that we can use it in pytest ?
```
- name: Set offscreen flag for pytest
      if: inputs.rendering_backend != 'auto'
      shell: bash
      run: echo "F3D_OFFSCREEN=1" >> "$GITHUB_ENV"

snoyer: what happens when running the test with no offscreen support, crash? exception?

noclone: it generates
```
Fatal Python error: Aborted Current thread 0x00007fe7c3a78b80 (most recent call first): File "/__w/f3d/f3d/source/examples/libf3d/python/qt6/minimal_qt6.py", line 27 in main File 
...
"/__t/Python/3.10.19/x64/lib/python3.10/runpy.py", line 196 in _run_module_as_main Extension modules: PyQt6.QtCore, PyQt6.QtGui, PyQt6.QtWidgets, PyQt6.QtOpenGL, PyQt6.QtOpenGLWidgets (total: 5) /__w/_temp/2eaa5043-13f8-4cfb-87a1-2b28a54ae609.sh: line 1: 2075 Aborted (core dumped) python -m pytest examples/libf3d/python/qt6/test_minimal_qt6.py Error: Process completed with exit code 134.

Comment (snoyer) [reply examples/libf3d/python/qt6/test_minimal_qt6.py]: not great, I guess we're not catching that at runtime then :(

Instead of env variables, would it be simpler to have the CI not install `pyside` where we know there won't be offscreen support and then do a `pytest.importorskip` like for the `tkinter` example test?

File: examples/libf3d/python/qt6/minimal_qt6.py [hunk 14]
```diff
@@ -0,0 +1,315 @@
...
+            Qt.Key.Key_Space: "Space",
+            Qt.Key.Key_Tab: "Tab",
+        }
+        if key in special:
+            return special[key]
+
+        text = event.text()
+        if text:
+            return text
+        return ""
```
snoyer:
```suggestion
        try:
            return special[key]
        except KeyError:
            return event.text()
```

examples/libf3d/cpp/qt6/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+cmake_minimum_required(VERSION 3.10)
+
+project(minimal_qt6)
+
+find_package(f3d REQUIRED COMPONENTS library)
+find_package(Qt6 REQUIRED COMPONENTS Widgets OpenGLWidgets)
+
+# This example only works with VTK >= 9.3.20240914
```
mwestphal:
```suggestion
# This example only works if libf3d was built against a VTK >= 9.3.20240914
```

examples/libf3d/cpp/tests/CMakeLists.txt
```diff
@@ -45,3 +45,13 @@ set_tests_properties(test_use-options-struct PROPERTIES
 add_test(NAME test_use-options-variant COMMAND "$<TARGET_FILE:use-options-variant>" "${CMAKE_CURRENT_SOURCE_DIR}/data/cow.vtp" 1)
 set_tests_properties(test_use-options-variant PROPERTIES
         PASS_REGULAR_EXPRESSION "Number of points: 2903")
+
+# glfw -- only works with VTK >= 9.3.20240914
```
mwestphal:
```suggestion
# glfw -- only works if libf3d was built against VTK >= 9.3.20240914
```

examples/libf3d/cpp/tests/CMakeLists.txt
```diff
@@ -45,3 +45,13 @@ set_tests_properties(test_use-options-struct PROPERTIES
...
+if (F3D_EXAMPLES_EXTERNAL_GLFW)
+  add_test(NAME test_external-glfw COMMAND "$<TARGET_FILE:external-glfw>" "${CMAKE_CURRENT_SOURCE_DIR}/data/cow.vtp" 1)
+endif ()
+
+# qt6 -- only works with VTK >= 9.3.20240914
```
mwestphal:
```suggestion
# qt6 -- only works if libf3d was built against VTK >= 9.3.20240914
```

examples/libf3d/python/qt6/test_minimal_qt6.py
```diff
@@ -0,0 +1,14 @@
...
+import pytest
+from pathlib import Path
+
+from minimal_qt6 import main
+
+
+@pytest.mark.skipif(
+    sys.platform != "win32",
+    reason="Python CI not working on non-windows https://github.com/f3d-app/f3d/issues/2670",
+)
```
mwestphal: Please use CI logic to disable this test, not python logic.

library/public/interactor.h
```diff
@@ -352,6 +352,12 @@ class F3D_EXPORT interactor
...
+   * The event loop will be triggered every `deltaTime` in seconds.
```
mwestphal: this doc seems incorrect though, it's more about advancing the internal timer of deltaTime ?

mwestphal (changes requested):
some changes

library/public/interactor.h
noclone: updated

examples/libf3d/cpp/external-glfw/main.cxx
```diff
@@ -5,6 +5,7 @@
 #include <GLFW/glfw3.h>
 
 #include <iostream>
+#include <thread>
```
mwestphal:
```suggestion
#include <chrono>
```

examples/libf3d/python/qt6/minimal_qt6.py
```diff
@@ -0,0 +1,312 @@
+import sys
+from argparse import ArgumentParser
+from pathlib import Path
+
+import f3d
+from PySide6.QtGui import QKeyEvent, QMouseEvent, QWheelEvent
+
```
mwestphal:
```suggestion
```

mwestphal (approved):

noclone:
> style check

didn't the style check pass ?

Meakk (approved):
Nice work!!

mwestphal:
> > style check
> 
> didn't the style check pass ?

probably a mistake on my part

PR #2688, opened 2025-12-02
python: improve type handling in `options.__setitem__`

snoyer:
### Describe your changes

Currently the Python bindings won't allow setting setting a `double` options using an `int` value, or a `vector<double>` options using a sequence of integers. 
```py
    options["scene.animation.speed_factor"] = 3.0  # ok
    options["scene.animation.speed_factor"] = 3    # not ok :(

    options["scene.up_direction"] = 1.0, 2.0, 3.0  # ok
    options["scene.up_direction"] = 1, 2, 3        # not ok :(
```
This appears to be due to how `pybind` matches types, and we can work around it in the implementation of `options.__setitem__`.

Also, `__setitem__` currently raises `AttributeError` when trying to set a valid option using the wrong type. Raising `TypeError` makes more sense here, see https://docs.python.org/3/library/exceptions.html#TypeError:

> Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3/library/stdtypes.html#list) when an [int](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).

### Continuous integration

### DIFF
--- a/python/F3DPythonBindings.cxx
+++ b/python/F3DPythonBindings.cxx
@@ -171,19 +171,19 @@ PYBIND11_MODULE(pyf3d, module)
         }
         catch (const f3d::options::incompatible_exception&)
         {
-          // failed to set an `int` try again as `double`
+          // failed to set an `int`, try again as `double`
           if (std::holds_alternative<int>(value))
           {
             opts.set(key, static_cast<double>(std::get<int>(value)));
           }
-          // failed to set an `vector<int>` try again as `vector<double>`
+          // failed to set a `vector<int>`, try again as `vector<double>`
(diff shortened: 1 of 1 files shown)


mwestphal (approved):

mwestphal:
\ci full

mwestphal:
go for merge @snoyer ?

snoyer:
> go for merge @snoyer ?

now that I've actually proofread my comments it should be good

Issue #2699, opened 2025-12-08
Export GetAnimationName function to libf3d


iommu:
**Is your feature request related to a problem? Please describe.**
The primary [`AnimationManager`](https://github.com/f3d-app/f3d/blob/4ced3dc7c81ebcbbd4d9edcbb30bd88d6b365558/library/private/animationManager.h#L72) class supports a method `GetAnimationName -> std::string` which returns the name of the currently playing animation. This is used by the f3d GUI, but is not available by using libf3d.

**Describe the solution you'd like**
Add a method of libf3d users accessing this data. 
Options include :
1. export a `getAnimationName -> std::string` function in the Interaction class [similar to other animation controls](https://github.com/f3d-app/f3d/blob/4ced3dc7c81ebcbbd4d9edcbb30bd88d6b365558/library/public/interactor.h#L252) 
2. export the name in the `f3d::options` data type [similar to the animation indices option](https://f3d.app/docs/api-libf3d/structf3d_1_1options_1_1scene_1_1animation/#structf3d_1_1options_1_1scene_1_1animation_1ae3270de15c7fbc48104bb5a89baa8881)
note : this would presumable be an immutable option which would be a bit of a bad design pattern. 

**Describe alternatives you've considered**
There is a possible alternative of manually loading the parser plugins, parsing the current file again and then getting the animation name from the plugins function [e.g.](https://github.com/f3d-app/f3d/blob/4ced3dc7c81ebcbbd4d9edcbb30bd88d6b365558/plugins/assimp/module/vtkF3DAssimpImporter.h#L49)

**Additional context**
The animation handling in libf3d is possible due for an overhaul anyway. The current method of pausing/playing/setting/scrubbing animations feels a bit inconsistent jumping between the interactor and the options data and maybe a seperate `f3d::engine.animator` may be an options

**Contribution**
I would be happy to contribute this code :)

mwestphal:
Hi @iommu 

I agree it is missing, but it should be added not as an immutable option but as an API, eg:

`interactor::getCurrentAnimationName` , wdyt ?

FYI @Ni-g-3l

Ni-g-3l:
Yes I think it's good idea :) 

I'm adding it to the meta animation issue.

If your are interested in contributing you can create a PR and ping me inside (or on discord) @iommu if you are looking for help.

iommu:
Absolutely, I can make a PR for this :) @Ni-g-3l

iommu:
Completed with #2702

PR #2702, opened 2025-12-09
Export get current animation name

iommu:
### Describe your changes
Exported the internal `AnimationManager::GetAnimationName` as `interactor::getAnimationName` so that libf3d can access animation name data.
#2699 

- [X] I have performed a [self-review](https://f3d.app/dev/CODING_STYLE) of my code
- [X] I have added [tests](https://f3d.app/dev/TESTING) for new features and bugfixes ***(see notes)**
- [X] If it is a modifying the libf3d API, I have updated bindings ***(see notes)**

Notes:
- Java bindings do not exist for any interactor code ans so have not been added for this change
- Python tests do not exist for any interactor animation code (But I would be happy to contribute that if that would be accepted)

### Continuous integration

### DIFF
--- a/python/testing/test_scene.py
+++ b/python/testing/test_scene.py
@@ -44,6 +44,9 @@ def test_scene():
     engine.scene.add(sphere1)
     engine.scene.add([sphere2, cube])
 
+    assert engine.scene.animation_time_range() == (0.0, 4.0)
+    engine.scene.load_animation_time(2)
+
     img = engine.window.render_to_image()
     img.save(output)
 


iommu:
Question regarding the docs : the checklist specifically mentions the QUICKSTART docs. So I'm assuming that, as this feature doesn't really fit being exposed as a CLI option, this doesn't need any changes to that?

library/public/interactor.h
```diff
@@ -252,6 +252,10 @@ class F3D_EXPORT interactor
   virtual interactor& startAnimation() = 0;
   virtual interactor& stopAnimation() = 0;
   [[nodiscard]] virtual bool isPlayingAnimation() = 0;
+  /**
+   * Access animation data.
+   */
+  [[nodiscard]] virtual std::string getAnimationName() = 0;
```
mwestphal: looking back at the current implementation we can see that there is a:

`scene::availableAnimations` method.

I wonder if it should be better to get all available animation names with a 

`scene::availableAnimationNames` and then use the animation_indices option to recover the one that are enabled.

Btw such discussion may be better on discord :)

https://discord.f3d.app

mwestphal (changes requested):
A suggested API change

mwestphal:
Hey @iommu 

Need any help moving forward ?

iommu:
Hi, sorry yes I was a bit unsure what you meant by your last comment of 
`getAnimationName` vs `availableAnimationNames` is that a possible future change or is that something that you would like done for this merge request?

library/public/interactor.h
iommu: Similar to this proposed `getAnimationName` method is the `animationTimeRange` method which also accesses animation data. Should this one possibly be changes to this style too? 
Possibly both functions could default to accessing the current animation data
i.e.
```
fn getAnimationName(index = -1)
  if index == -1 {
    return getAnimationName(current animation index)
  } 
  return animation name at this index
}
```

mwestphal: I agree, but lets discuss on discord for faster turnaround.

mwestphal:
> Hi, sorry yes I was a bit unsure what you meant by your last comment of `getAnimationName` vs `availableAnimationNames` is that a possible future change or is that something that you would like done for this merge request?

Lets try to do it in this PR :)

mwestphal:
> Question regarding the docs : the checklist specifically mentions the QUICKSTART docs. So I'm assuming that, as this feature doesn't really fit being exposed as a CLI option, this doesn't need any changes to that?

yes, unrelated to QUICKSTART indeed.

library/public/interactor.h
```diff
@@ -263,6 +264,11 @@ class F3D_EXPORT interactor
     AnimationDirection direction = AnimationDirection::FORWARD) = 0;
   virtual interactor& stopAnimation() = 0;
   [[nodiscard]] virtual bool isPlayingAnimation() = 0;
+  /**
+   * Access animation data.
+   */
+  [[nodiscard]] virtual std::string getAnimationName(int indices = INT_MIN) = 0;
+  [[nodiscard]] virtual std::vector<std::string> getAnimationNames() = 0;
```
mwestphal: imo this should be in `scene` not in `interactor`

mwestphal (changes requested):
api change

library/public/scene.h
```diff
@@ -148,6 +149,19 @@ class F3D_EXPORT scene
    */
   [[nodiscard]] virtual unsigned int availableAnimations() const = 0;
 
+  /**
+   * Return the animation name of a given animation indices, if any.
+   * Will return current animation if indices is INT_MIN (default)
```
mwestphal: I'd rather use -1 than INT_MIN, wdyt ?

mwestphal (changes requested):
a question

mwestphal:
\ci fast

library/public/scene.h
iommu: Is there ever a case where the index can be out of the range 0..number_animations?
If not then yes I'm ok with that.

mwestphal: no, it can only be 0..n

iommu: Easy, I'm happy with that then. Changed

library/public/scene.h
```diff
@@ -148,6 +148,19 @@ class F3D_EXPORT scene
    */
   [[nodiscard]] virtual unsigned int availableAnimations() const = 0;
 
+  /**
+   * Return the animation name of a given animation indices, if any.
+   * Will return current animation if indices is -1 (default)
```
mwestphal: clarify the behavior when multiple animations (or none) are enabled

iommu:
```c++
/**
* Return the animation name of a given animation indices, if any.
*
* Specific animation (0..availableAnimations): Returns the name of the animation at that index
* Current animation (-1):
*   - Returns the name of the current animation
*   - Returns "Multi animations" if more than one animation is current
*   - Returns "All animations" if all animations are current
*   - Returns "No animations" if no animations are current
* Fallback: Returns "No animation" for out-of-bounds requests.
*
* Can be called before initialization safely
*/
```
thoughts?

mwestphal: perfect!

iommu: Done, I've also updated the `getAnimationNames` comment to specify that a vector of 0 length is returned if no animations

iommu:
\ci full

mwestphal:
Thanks for the contribution @iommu !

PR #2752, opened 2025-12-24
Do not throw if cache path is empty

Meakk:
### Describe your changes

On Android, caches are local to app, so it's the app responsibility to set the cache path using `f3d::engine::setCachePath` right after engine initialization.  
Do not throw if the cache path cannot be retrieved using env vars.

Supersedes #2024 

### Continuous integration

### DIFF
--- a/application/testing/CMakeLists.txt
+++ b/application/testing/CMakeLists.txt
@@ -1648,7 +1648,7 @@ set_tests_properties(f3d::TestNoFileFileNameTemplate PROPERTIES PASS_REGULAR_EXP
 set_tests_properties(f3d::TestNoFileFileNameTemplate PROPERTIES ENVIRONMENT "CTEST_F3D_NO_DATA_FORCE_RENDER=1")
 
 if(NOT WIN32)
-  f3d_test(NAME TestInvalidCache ENV "HOME=" REGEXP "Could not use default cache directory" NO_BASELINE)
+  f3d_test(NAME TestHDRINoCachePath ENV "HOME=" DATA dragon.vtu ARGS -f --hdri-file=${F3D_SOURCE_DIR}/testing/data/palermo_park_1k.hdr REGEXP "Cannot cache .* as no cache path has been set." NO_BASELINE LONG_TIMEOUT)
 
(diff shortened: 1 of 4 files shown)


application/testing/CMakeLists.txt
```diff
@@ -1632,8 +1632,6 @@ set_tests_properties(f3d::TestNoFileFileNameTemplate PROPERTIES PASS_REGULAR_EXP
 set_tests_properties(f3d::TestNoFileFileNameTemplate PROPERTIES ENVIRONMENT "CTEST_F3D_NO_DATA_FORCE_RENDER=1")
 
 if(NOT WIN32)
-  f3d_test(NAME TestInvalidCache ENV "HOME=" REGEXP "Could not use default cache directory" NO_BASELINE)
```
mwestphal, line 1651: why not keeping the test ? Its a good test, its just the result of the test that changes

Meakk: It's not testing anything now tbh.  
Maybe I should log a warning if we are trying to cache an HDRI but the cache path is not set.

mwestphal: Definitely

library/src/engine.cxx
```diff
@@ -84,19 +84,17 @@ engine::engine(
...
 
   this->Internals->Window =
     std::make_unique<detail::window_impl>(*this->Internals->Options, windowType, offscreen, loader);
-  this->Internals->Window->SetCachePath(cachePath);
+
+  if (!cachePath.empty())
+  {
+    cachePath /= "f3d";
+    this->Internals->Window->SetCachePath(cachePath);
+  }
```
mwestphal, line 97: the warnings are not very useful for user:           "Cannot cache HDRI LUT texture as no cache path has been set.");

Maybe add a:


```suggestion
  }
  else
  {
          F3DLog::Print(F3DLog::Severity::Warning,
          "Could not setup cache, please set the appropriate environment variable.");
  }
```
?

mwestphal: And with this warning, do you need the other warnings then ?

Meakk: No I don't think it's a good idea to put it there since it's in the engine constructor. Even if I call `setCachePath` later I won't be able to hide the warning.

mwestphal: ok I understand now. got it.

PR #2724, opened 2025-12-17
Add getters for camera parameters #1419

Dtsitos:
### Describe your changes

This PR adds read-only getters to the `camera` API to retrieve derived world-space
camera parameters (azimuth, elevation, and distance).

The values are computed from the camera position and focal point and do not
modify camera state. This enables external applications and bindings to query
camera orientation consistently without duplicating math.

No existing behavior is changed.

### Issue ticket number and link if any

Fixes #1419
https://github.com/f3d-app/f3d/issues/1419

### Checklist for finalizing the PR

- [x] I have performed a self-review of my code
- [ ] I have added tests for new features and bugfixes
- [ ] I have added documentation for new features
- [ ] If it is a modifying the libf3d API, I have updated bindings
- [ ] If it is a modifying the `.github/workflows/versions.json`, I have updated `docker_timestamp`

### Continuous integration

\ci fast

PR #2767, opened 2026-01-02
Add getters for camera parameters #1419

### LINKED ISSUE #1419: Add getters for camera parameters
When using libf3d and an external window you will  need to use yaw, azimuth, elevation to move but to get them in degree you would need to calculate them from the camera position.

It could be useful to have getters for these parameters. I know they are not actually parameters, but ways to manipulate the camera, but a getter for yaw/elevation that returns the angle of the camera to the model can be useful, also distance from the model etc.

**Suggested implementation**
In `camera.h` and `camera.cxx` , add `camera::getYaw`, `camera::getAzimuth`,  `camera::getElevation`. Implement them to be computed on the fly.

Dtsitos:
### Describe your changes
This PR adds read-only getters to the camera API to retrieve derived world-space
camera parameters (azimuth, elevation, and distance).

The values are computed from the camera position and focal point and do not
modify camera state. This enables external applications and bindings to query
camera orientation consistently without duplicating math.

No existing behavior is changed.

Original PR: https://github.com/f3d-app/f3d/pull/2724

Fixes https://github.com/f3d-app/f3d/issues/1419

- [X] I have performed a [self-review](https://f3d.app/dev/CODING_STYLE) of my code
- [X] I have added [tests](https://f3d.app/dev/TESTING) for new features and bugfixes
- [X] I have added [documentation](https://f3d.app/docs/next/user/QUICKSTART) for new features

### Continuous integration


Dtsitos:
I think this should be correct now, will work on better testing ASAP

library/src/camera_impl.cxx
```diff
@@ -103,6 +105,74 @@ void camera_impl::getFocalPoint(point3_t& foc) const
...
+  vtkMath::Normalize(view.data());
+
+  vtkRenderer* ren = this->Internals->VTKRenderer;
+  double* up = ren->GetEnvironmentUp();
+
+  // Elevation is angle above the horizontal plane
+  double dot = vtkMath::Dot(view.data(), up);
+  dot = std::clamp(dot, -1.0, 1.0);
+
+  return vtkMath::DegreesFromRadians(std::asin(dot));
```
snoyer: Would be nice to check if that can be done with `AngleBetweenVectors` to avoid the raw trig. Maybe something like `90° - angle_between(view, up)`?

library/src/camera_impl.cxx
```diff
@@ -103,6 +105,74 @@ void camera_impl::getFocalPoint(point3_t& foc) const
...
+  vector3_t horizontal;
+  vtkMath::Subtract(view.data(), projUp.data(), horizontal.data());
+
+  static constexpr double EPS = 128 * std::numeric_limits<double>::epsilon();
+  if (vtkMath::Norm(horizontal.data()) < EPS)
+  {
+    return 0.0;
+  }
+
+  vtkMath::Normalize(horizontal.data());
```
snoyer, line 144: Once you've got the tests you can double check if this is actually needed

mwestphal:
please let us know when its ready for review @Dtsitos

Dtsitos:
@mwestphal Hello there, I have implemented all the bindings and some basic testing for them. Full disclosure, I have never worked with bindings again and mostly deduced what to do based on the patterns from the existing implementations. Thank you for your patience

mwestphal:
> @mwestphal Hello there, I have implemented all the bindings and some basic testing for them. Full disclosure, I have never worked with bindings again and mostly deduced what to do based on the patterns from the existing implementations. Thank you for your patience

Thats perfectly fine and it looks good to me :)

mwestphal:
@Dtsitos can you adress and resolve the suggestions made by @snoyer above ?
If you did adress them already, please resolve the discussions.

mwestphal:
\ci full

Dtsitos:
\ci full

library/src/camera_impl.cxx
Dtsitos: I have updated getWorldElevation() to use vtkMath::AngleBetweenVectors instead of using asin() directly, expressing elevation as 90° - angle_between(view, up). This avoids the raw trig mentioned.

mwestphal:
not ready for full CI yet :)

library/src/camera_impl.cxx
mwestphal: so, is it needed @Dtsitos ?

mwestphal:
do you need help with the style check @Dtsitos ?

Dtsitos:
> do you need help with the style check @Dtsitos ?

Yes please, I am really not sure what is wrong. Sorry for not replying earlier but exam season is starting and I am not having fun with that 🙃

mwestphal:
> > do you need help with the style check @Dtsitos ?
> 
> Yes please, I am really not sure what is wrong. Sorry for not replying earlier but exam season is starting and I am not having fun with that 🙃

 - copy the comment above
 - paste in a file name `file.patch`
 - apply the patch on your code `patch -p1 < file.patch`
 - commit and push the change

library/src/camera_impl.cxx
Dtsitos: I believe it is necessary because after projecting the view vector onto the plane orthogonal to the environment up vector, the resulting vector could be really close to zero when the camera is looking parallel or nearly parallel to EnvironmentUp. In that case azimuth isn't defined and should return 0 instead of an arbitrary value

Dtsitos:
@mwestphal Hello there. Thank you for showing me how to fix styling. I changed the tests for the new getters radically because after careful consideration (bashing my head against the keyboard) I realised that the tests produced wrong values not because the math is wrong but because I cannot play around with custom environment vectors as the test file does not have access to the renderer methods to change the environmental vectors. The tests work now locally. I will push with the proper styling and then it should be good to go.

snoyer:
The environment vectors are indeed not accessible from the camera, however you can set the `up` vector through the engine's options, see for example https://github.com/f3d-app/f3d/blob/ae5488a66c7735417bb047662c50e4573049751f/library/testing/TestSDKDynamicUpDirection.cxx#L53-L57

you'll need to set `options.scene.up_direction` and call `engine.render()` before setting the camera state and performing your checks but that should let you test any case you need

Dtsitos:
Alas, my testing and styling is proper. What should I do now to proceed with the PR? I am guessing I need a CI full check first.

mwestphal:
> What should I do now to proceed with the PR? 

Ask for a review from a maintainer, top right corner of the PR :)

library/src/camera_impl.cxx
mwestphal:
```suggestion
#include <vtkCamera.h>
#include <vtkMatrix4x4.h>
#include <vtkRenderWindow.h>
#include <vtkRenderer.h>
#include <vtkVersion.h>

#include <cmath>
```

mwestphal: by the way what do you need cmath for ?

python/testing/test_camera.py
```diff
@@ -117,6 +117,10 @@ def test_moves():
     camera.elevation(angle)
     camera.pitch(angle)
 
+    camera.get_world_azimuth()
+    camera.get_world_elevation()
+    camera.get_distance()
```
mwestphal, line 122: why does some method are `_world` method, and not others ?

mwestphal (changes requested):
some questions

python/testing/test_camera.py
Dtsitos: Because azimuth and elevation in my implementation are dependant on the environment vectors, I think it is best that we use the "world" qualifier as a distinction because azimuth and elevation are already declared in the camera implementation but modify the state of the camera. Distance on the other hand does not need this distinction because the world frame does not change it's value.

library/testing/TestSDKCamera.cxx
```diff
@@ -312,5 +312,118 @@ int TestSDKCamera([[maybe_unused]] int argc, [[maybe_unused]] char* argv[])
     return EXIT_FAILURE;
   }
 
+  // test world azimuth, elevation and distance
```
mwestphal: can you add a test with more "random" value, more a typical usage if you know what I mean, like adding Azimuth/elevation first.

library/src/camera_impl.cxx
```diff
@@ -6,6 +6,8 @@
 #include <vtkRenderer.h>
 #include <vtkVersion.h>
 
+#include <cmath>
```
mwestphal: what do you need cmath for ?

mwestphal (changes requested):
small questions and suggestions

library/src/camera_impl.cxx
```diff
@@ -103,6 +105,76 @@ void camera_impl::getFocalPoint(point3_t& foc) const
...
+  {
+    return 0.0;
+  }
+
+  vtkMath::Normalize(horizontal.data());
+
+  // Signed angle between right and horizontal projection
+  double angleRad = vtkMath::SignedAngleBetweenVectors(right, horizontal.data(), up);
+
+  return vtkMath::DegreesFromRadians(angleRad);
```
snoyer: should the angle be between the flat view direction and the environment's forward vector rather than its right vector? that is 90° more than what that currently does? (assuming the default state is the camera looking forward)

library/testing/TestSDKCamera.cxx
Dtsitos: Ok, I think the new test case 6 should be good for that.

library/src/camera_impl.cxx
Dtsitos: To be totally honest I needed it for a prior implementation and while I noticed it was no longer necessary your previous comment about changing their rotation made me think it was needed and I couldn't tell why so I just kept it to be sure 😸. I will remove it right away

library/src/camera_impl.cxx
Dtsitos: The environment right vector provides a fixed world reference, while forward depends on the camera orientation. Switching to forward would only shift the convention by 90°, but would make the value depend on camera rotations rather than the world, which goes again the logic that these are "World" getters that should only care about the environment.

Dtsitos:
So from what I can tell the testing procedures changed a lot. I fixed my part of the tests but the program stil won't compile because other tests are in the old format. Any suggestions about what I should do? I could attempt at fixing all the tests

mwestphal:
> So from what I can tell the testing procedures changed a lot. I fixed my part of the tests but the program stil won't compile because other tests are in the old format. Any suggestions about what I should do? I could attempt at fixing all the tests

Correct, we just revamped all the test and you need to rebase. I can help you though, if needed, let me know :)

mwestphal:
> @Dtsitos Merge branch 'master' into feature/getters-for-camera 

Avoid merging, prefer rebase:

https://f3d.app/dev/CONTRIBUTING#f3d-development-workflow

library/src/camera_impl.cxx
snoyer: > The environment right vector provides a fixed world reference

yes, and the _environment's_ forward vector is fixed the same way

> Switching to forward would only shift the convention by 90°

also yes, and I understand this is kind of arbitrary, but my point is that, unless I'm mistaken, F3D's default state for the camera is pointing along the _environment's_ forward vector and the environment's right vector pointing to the right of the screen, and in that case I would assume the azimuth should compute to 0 but with your code as it is it would compute to -90°

Basically I would expect the getters to return angles so that if you applied them to a camera in the default state you would end up pointing in the same direction again.

I don't think the problem is your code/math but the vagueness of the issue, we'll need to hear from @mwestphal and @Meakk to confirm/explain the requirements

mwestphal: That what I meant on discord. I do agree that the issue is vague and current implementation can be argued.

But at least it does something and I think what it does still provide enough info to fix the issue, I think.

Wdyt @Meakk ?

Dtsitos:
So I am not sure if force pushing was a nuclear move but it seemed necessary for a proper rebase. The new tests integrate properly with the new format and work now, so I think it is ready for review again

mwestphal:
> So I am not sure if force pushing was a nuclear move but it seemed necessary for a proper rebase. The new tests integrate properly with the new format and work now, so I think it is ready for review again

Force push is indeed part of the normal rebase workflow.

library/testing/TestSDKCamera.cxx
```diff
@@ -163,5 +163,75 @@ int TestSDKCamera([[maybe_unused]] int argc, [[maybe_unused]] char* argv[])
...
+  const double initialDistance = cam.getDistance();
+
+  cam.azimuth(37.0);
+  cam.elevation(-22.0);
+
+  test("getDistance invariant after azimuth/elevation", cam.getDistance(), approx(initialDistance));
+
+  test("getWorldAzimuth finite after camera operations", std::isfinite(cam.getWorldAzimuth()));
+
+  test("getWorldElevation finite after camera operations", std::isfinite(cam.getWorldElevation()));
```
mwestphal: why do you test isfinite ? test just expected value ?

mwestphal (changes requested):
changes needed

library/src/camera_impl.cxx
```diff
@@ -103,6 +103,76 @@ void camera_impl::getFocalPoint(point3_t& foc) const
...
 
+//----------------------------------------------------------------------------
+void camera_impl::getPositionToFocalVector(vector3_t& vec) const
+{
+  point3_t pos, focal;
+  this->getPosition(pos);
+  this->getFocalPoint(focal);
+
+  vtkMath::Subtract(pos.data(), focal.data(), vec.data());
+}
```
Meakk, line 114: (pos - focal) is not "position to focal" vector, it's the opposite. Do you mean `vtkMath::Subtract(focal.data(), pos.data(), vec.data());` or name `getFocalToPositionVector`?

library/src/camera_impl.cxx
```diff
@@ -103,6 +103,76 @@ void camera_impl::getFocalPoint(point3_t& foc) const
...
+{
+  vector3_t view;
+  this->getPositionToFocalVector(view);
+  vtkMath::Normalize(view.data());
+
+  vtkRenderer* ren = this->Internals->VTKRenderer;
+  double* up = ren->GetEnvironmentUp();
+  double* right = ren->GetEnvironmentRight();
+
+  // Project view vector onto plane orthogonal to up
```
Meakk: The comment is not correct. You are not projecting the view vector to the plane orthogonal to up, you are projecting the view vector to the up direction.

library/src/camera_impl.cxx
```diff
@@ -103,6 +103,76 @@ void camera_impl::getFocalPoint(point3_t& foc) const
...
+  static constexpr double EPS = 128 * std::numeric_limits<double>::epsilon();
+  if (vtkMath::Norm(horizontal.data()) < EPS)
+  {
+    return 0.0;
+  }
+
+  vtkMath::Normalize(horizontal.data());
+
+  // Signed angle between right and horizontal projection
+  double angleRad = vtkMath::SignedAngleBetweenVectors(right, horizontal.data(), up);
```
Meakk: From my comments above, the horizontal direction computed is pointing behind the camera. Make sure it gives the same result here as I expect the sign to be flipped here.

library/src/camera_impl.cxx
Meakk: I agree with @snoyer. If we just query the azimuth after opening a file without changing the camera, I'd expect 0 to be returned.

library/testing/TestSDKCamera.cxx
Dtsitos: Due to the other changes requested I changed testing drastically, mostly to improve me own understanding of the math and to be more thorough

library/src/camera_impl.cxx
Dtsitos: To avoid causing problems downstream (with bindings and such) I changed the subtract order

library/src/camera_impl.cxx
Dtsitos: I changed the azimuth implementation based on the convention you ask for in a comment further down so that is resolved.

library/src/camera_impl.cxx
Dtsitos: Will do!

Dtsitos:
I changed the testing quite a lot to make it more thorough (edge case and custom environment for each getter) and changed my world azimuth convention to align with expectations from an above discussion (about azimuth being zero when looking forward). I hope my code is more up to spec now

mwestphal:
Ill let @Meakk review this one :)

library/src/camera_impl.cxx
```diff
@@ -103,6 +103,79 @@ void camera_impl::getFocalPoint(point3_t& foc) const
...
+{
+  vector3_t view;
+  this->getPositionToFocalVector(view);
+  vtkMath::Normalize(view.data());
+
+  vtkRenderer* ren = this->Internals->VTKRenderer;
+  const double* up = ren->GetEnvironmentUp();
+  const double* right = ren->GetEnvironmentRight();
+
+  // Derive environment forward = up × right
```
Meakk, line 127: It looks like `×` is not an ascii character. I don't think it's necessary a problem but just wanted to make sure we agree with that @mwestphal

library/src/camera_impl.cxx
```diff
@@ -103,6 +103,79 @@ void camera_impl::getFocalPoint(point3_t& foc) const
...
+  double angleRad = vtkMath::AngleBetweenVectors(view.data(), up);
+  return 90.0 - vtkMath::DegreesFromRadians(angleRad);
+}
+
+//----------------------------------------------------------------------------
+double camera_impl::getDistance() const
+{
+  vector3_t v;
+  this->getPositionToFocalVector(v);
+  return vtkMath::Norm(v.data());
```
snoyer, line 176:
```suggestion
  return this->GetVTKCamera()->GetDistance();
```
is that equivalent?

library/testing/TestSDKCamera.cxx
```diff
@@ -163,5 +163,110 @@ int TestSDKCamera([[maybe_unused]] int argc, [[maybe_unused]] char* argv[])
...
   test("up when cross product of pos->foc and up is 0 - test 4", cam.getViewUp(), { 0, 1, 0 });
 
+  // test distance getter
+  cam.setPosition({ 0., 0., 0. });
+  cam.setFocalPoint({ 0., 0., 1. });
+  test("Distance: 1", cam.getDistance(), approx(1.0));
+
+  cam.setPosition({ 0., -10., 0. });
+  cam.setFocalPoint({ 0., 0., 0 });
+  test("Distance: 1", cam.getDistance(), approx(10.0));
```
Meakk, line 173:
```suggestion
  test("Distance: 10", cam.getDistance(), approx(10.0));
```

Meakk (approved):
Typo and question but LGTM

library/src/camera_impl.cxx
mwestphal: lol yeah, please use ASCII @Dtsitos

library/testing/TestSDKCamera.cxx
```diff
@@ -163,5 +163,110 @@ int TestSDKCamera([[maybe_unused]] int argc, [[maybe_unused]] char* argv[])
...
+  cam.setViewUp({ 0., 0., 1. });
+  test("Azimuth: vertical view", cam.getWorldAzimuth(), approx(0.0));
+
+  opt.scene.up_direction = { 0, 1, 0 };
+  eng.setOptions(opt);
+  win.render();
+  cam.setPosition({ -10., 0., 0. });
+  cam.setFocalPoint({ 0., 0., 0. });
+  cam.setViewUp({ 0., 1., 0 });
+  test("Azimuth (Y-up): environment forward (+X)", cam.getWorldAzimuth(), approx(90.0));
```
Meakk, line 270: Let's add the code below to test setter/getter equivalence (credits to @snoyer):

```cpp
  std::vector<f3d::direction_t> up_directions = {
    { 0, 0, +1 },
    { 0, +1, 0 },
    { +1, 0, 0 },
    { 0, 0, -1 },
    { 0, -1, 0 },
    { -1, 0, 0 },
    { -1, +2, +3 },
    { +4, -5, -6 },
  };

  const std::vector<std::pair<double, double>> azimuths_elevations = {
    { 0, 0 },
    { +12, +34 },
    { +12, -34 },
    { -12, +34 },
    { -12, -34 },
  };

  for (const auto up : up_directions)
  {
    for (auto [a, e] : azimuths_elevations)
    {
      f3d::engine eng = f3d::engine::create(true);
      f3d::window& win = eng.getWindow();
      f3d::camera& cam = win.getCamera();
      opt.scene.up_direction = up;
      win.render();

      cam.azimuth(a).elevation(e);
      const std::string title = " after .azimuth(" + f3d::options::format(a) + ").elevation(" +
        f3d::options::format(e) + ") with up = " + f3d::options::format(up);
      test("azimuth  " + title, cam.getWorldAzimuth(), approx(a, 1e-10));
      test("elevation" + title, cam.getWorldElevation(), approx(e, 1e-10));
    }
  }
```

java/F3DCameraBindings.cxx
```diff
@@ -38,6 +38,21 @@ extern "C"
     return ret;
   }
 
+  JNIEXPORT jdouble JAVA_BIND(Camera, getWorldAzimuth)(JNIEnv* env, jobject self)
```
Meakk, line 41: Missing functions in `Camera.java`

Meakk:
Here's a suggested patch by @snoyer on discord to fix the flipped signed:

```patch
diff --git a/library/src/camera_impl.cxx b/library/src/camera_impl.cxx
index 2d4bd562b..5b47f0d68 100644
--- a/library/src/camera_impl.cxx
+++ b/library/src/camera_impl.cxx
@@ -2,6 +2,7 @@
 
 #include <vtkCamera.h>
 #include <vtkMatrix4x4.h>
+#include <vtkPlane.h>
 #include <vtkRenderWindow.h>
 #include <vtkRenderer.h>
 #include <vtkVersion.h>
@@ -116,56 +117,39 @@ void camera_impl::getPositionToFocalVector(vector3_t& vec) const
 double camera_impl::getWorldAzimuth() const
 {
-  vector3_t view;
-  this->getPositionToFocalVector(view);
-  vtkMath::Normalize(view.data());
-
   vtkRenderer* ren = this->Internals->VTKRenderer;
   const double* up = ren->GetEnvironmentUp();
-  const double* right = ren->GetEnvironmentRight();
-
-  // Derive environment forward = up × right
-  vector3_t forward;
-  vtkMath::Cross(up, right, forward.data());
-  vtkMath::Normalize(forward.data());
-
-  vector3_t projUp;
-  vtkMath::ProjectVector(view.data(), up, projUp.data());
 
-  vector3_t horizontal;
-  vtkMath::Subtract(view.data(), projUp.data(), horizontal.data());
+  double projectedView[3];
+  vtkPlane::ProjectVector(this->GetVTKCamera()->GetDirectionOfProjection(),
+    this->getFocalPoint().data(), up, projectedView);
 
   static constexpr double EPS = 128 * std::numeric_limits<double>::epsilon();
-  if (vtkMath::Norm(horizontal.data()) < EPS)
+  if (vtkMath::Norm(projectedView) < EPS)
   {
     return 0.0;
   }
 
-  vtkMath::Normalize(horizontal.data());
+  const double angleRad =
+    vtkMath::SignedAngleBetweenVectors(ren->GetEnvironmentRight(), projectedView, up);
 
-  double angleRad = vtkMath::SignedAngleBetweenVectors(horizontal.data(), forward.data(), up);
-
-  return vtkMath::DegreesFromRadians(angleRad);
+  return vtkMath::DegreesFromRadians(angleRad) - 90.0;
 }
 
 double camera_impl::getWorldElevation() const
 {
-  vector3_t view;
-  this->getPositionToFocalVector(view);
+  double* view = this->GetVTKCamera()->GetDirectionOfProjection();
 
   static constexpr double EPS = 128 * std::numeric_limits<double>::epsilon();
-  if (vtkMath::Norm(view.data()) < EPS)
+  if (vtkMath::Norm(view) < EPS)
   {
     return 0.0;
   }
-  vtkMath::Normalize(view.data());
 
   vtkRenderer* ren = this->Internals->VTKRenderer;
-  double* up = ren->GetEnvironmentUp();
-
-  double angleRad = vtkMath::AngleBetweenVectors(view.data(), up);
-  return 90.0 - vtkMath::DegreesFromRadians(angleRad);
+  const double angleRad = vtkMath::AngleBetweenVectors(ren->GetEnvironmentUp(), view);
+  return vtkMath::DegreesFromRadians(angleRad) - 90.0;
 }
 
```

mwestphal:
Still working on this @Dtsitos ?

mwestphal:
Still working on this @Dtsitos ?

PR #2785, opened 2026-01-06
Add support for saving multiple frames when using animation

noclone:
### Describe your changes

Add support for saving multiple frames when using animation with `--output` and `frame` template filename

### Issue ticket number and link if any

https://github.com/f3d-app/f3d/issues/46

### Checklist for finalizing the PR

- [x] I have performed a [self-review](https://f3d.app/dev/CODING_STYLE) of my code
- [x] I have added [tests](https://f3d.app/dev/TESTING) for new features and bugfixes
- [x] I have added [documentation](https://f3d.app/docs/next/user/QUICKSTART) for new features
- [ ] If it is a modifying the libf3d API, I have updated bindings
- [ ] If it is a modifying the `.github/workflows/versions.json`, I have updated `docker_timestamp`

### Continuous integration

Please write a comment to run CI, eg: `\ci fast`.
See [here](https://f3d.app/dev/CONTRIBUTING#continuous-integration) for more info.

PR #2794, opened 2026-01-08
add function to check for variables in string template

snoyer:
### Describe your changes

Add `string_template::hasVariable(regex)` to check if a string template contains a variable.
This is refactoring an existing feature currently implemented as a lambda in the app because it had only one usecase but it looks like it would be useful in other cases (#2785 for example)

### Continuous integration

### DIFF
--- a/library/public/utils.h
+++ b/library/public/utils.h
@@ -288,14 +288,9 @@ inline std::vector<std::string> utils::string_template::variables() const
 //------------------------------------------------------------------------------
 inline bool utils::string_template::hasVariable(const std::regex& pattern) const
 {
-  for (const auto& variable : this->variables())
-  {
-    if (std::regex_match(variable, pattern))
-    {
-      return true;
-    }
-  }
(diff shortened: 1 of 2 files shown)


snoyer:
\ci full

library/public/utils.h
```diff
@@ -165,6 +165,9 @@ class F3D_EXPORT utils
     /** List the remaining un-substituted variables. */
     [[nodiscard]] std::vector<std::string> variables() const;
 
+    /** Check if the template contains a variable matching the regex pattern. */
+    [[nodiscard]] bool hasVariable(const std::regex& pattern) const;
```
mwestphal, line 169: needs bindings

snoyer: do we have any bindings for `string_template`?

mwestphal: looks like we dont, maybe we should, but out of scope anyway.

Issue #4, opened 2021-11-12
Implement normal rendering mode


mwestphal:

mwestphal:
@Meakk could you add some context to this issue ? It could then be added a the good-first-issue.

Meakk:
My idea was to use glyphs (simple lines oriented in the normal direction). It may be a bit difficult for a first issue.

ArchiePayne:
@mwestphal Happy to take a look into this one if it's still available 👍

mwestphal:
Of course!

ArchiePayne:
Hey there, 

Shift + N binding is taken by the HDRI Filename option, is there an alternative key you would prefer?

mwestphal:
Ctrl+N ?

mwestphal:
Fixed by https://github.com/f3d-app/f3d/pull/2958

PR #2811, opened 2026-01-11
Initial Implementation of the Normal rendering mode.

ArchiePayne:
### Describe your changes

Implement the Normal Rendering mode, displaying arrows on top of the geometry to depict vertex normals.

Added option `--normal-glyphs` with binding `ctrl+n` to toggle on/off. 

Added auto-test for the functionality.

[Issue #4](https://github.com/f3d-app/f3d/issues/4)

### Continuous integration

### DIFF
--- a/vtkext/private/module/vtkF3DRenderer.cxx
+++ b/vtkext/private/module/vtkF3DRenderer.cxx
@@ -108,8 +108,6 @@
 #include <chrono>
 #include <sstream>
 
-
-
 namespace
 {
 std::string DeprecatedCollapsePath(const fs::path& path)
@@ -2913,6 +2911,7 @@ void vtkF3DRenderer::SetUsePointSprites(bool use)
   if (this->UsePointSprites != use)
   {
     this->UsePointSprites = use;
+    this->CheatSheetConfigured = false;
     this->ColoringConfigured = false;
     this->PointSpritesConfigured = false;
   }


ArchiePayne:
\ci fast

library/src/interactor_impl.cxx
```diff
@@ -1210,6 +1210,16 @@ interactor& interactor_impl::initCommands()
     std::bind(complNames, std::placeholders::_1,
       std::vector<std::string>{ "front", "top", "right", "isometric" }));
 
+  this->addCommand(
+    "toggle_normal_glyphs",
```
mwestphal: not needed

library/src/interactor_impl.cxx
```diff
@@ -1611,6 +1621,7 @@ interactor& interactor_impl::initBindings()
 #endif
   this->addBinding({mod_t::NONE, "V"}, "toggle_volume_rendering","Scene", std::bind(docTgl, "Volume rendering", std::cref(opts.model.volume.enable)), f3d::interactor::BindingType::TOGGLE);
   this->addBinding({mod_t::NONE, "I"}, "toggle model.volume.inverse","Scene", std::bind(docTgl, "Inverse volume opacity", std::cref(opts.model.volume.inverse)), f3d::interactor::BindingType::TOGGLE);
+  this->addBinding({mod_t::CTRL, "N"}, "toggle_normal_glyphs","Scene", std::bind(docTgl, "Normal glyphs", std::cref(opts.model.normal_glyphs.enable)), f3d::interactor::BindingType::TOGGLE);
```
mwestphal: use a standard toggle

testing/baselines/TestNormalGlyphsEnable.png
mwestphal, line 1: Should glyph be colored too @Meakk ?

library/src/interactor_impl.cxx
ArchiePayne: Thanks, copied the wrong example 👍

testing/baselines/TestNormalGlyphsEnable.png
Meakk: No, having the same color makes it harder to read I think. We should have an option to tweak the color though. A CLI option is not really necessary, but a libf3d option makes sense to me.

library/options.json
```diff
@@ -383,6 +383,12 @@
         "default_value": "false"
       }
     },
+    "normal_glyphs": {
+      "enable": {
```
Meakk, line 405: Can be done later but additional options would be great (color, shape, scale/size...)

vtkext/private/module/vtkF3DMetaImporter.cxx
```diff
@@ -253,18 +262,37 @@ bool vtkF3DMetaImporter::Update()
...
+      vtkF3DMetaImporter::NormalGlyphsStruct& ngs =
+        this->Pimpl->NormalGlyphsActorsAndMappers.back();
+
+      vtkNew<vtkArrowSource> arrowSource;
+
+      ngs.glyph3D->SetSourceConnection(arrowSource->GetOutputPort());
+      ngs.glyph3D->SetVectorModeToUseNormal();
+      ngs.glyph3D->SetInputData(points);
+      ngs.glyph3D->SetScaleFactor(.2);
+      ngs.glyph3D->Update();
```
Meakk: We should use the GPU instancing mapper instead (`vtkGlyph3DMapper`).

vtkext/private/module/vtkF3DMetaImporter.cxx
```diff
@@ -253,18 +263,45 @@ bool vtkF3DMetaImporter::Update()
...
         points = genericImporter->GetImportedPoints(actorIndex);
       }
+
+      // Create and configure normal glyph actors
+      this->Pimpl->NormalGlyphsActorsAndMappers.emplace_back(
+        vtkF3DMetaImporter::NormalGlyphsStruct(actor, importer));
+      vtkF3DMetaImporter::NormalGlyphsStruct& ngs =
+        this->Pimpl->NormalGlyphsActorsAndMappers.back();
+
+      vtkNew<vtkPolyDataNormals> normals;
```
Meakk: I don't think we should recompute the normals. This feature is actually useful to debug current model normals so we should display current normals.

vtkext/private/module/vtkF3DMetaImporter.cxx
```diff
@@ -253,18 +263,45 @@ bool vtkF3DMetaImporter::Update()
...
+      normals->ComputePointNormalsOn();
+      normals->ComputeCellNormalsOff();
+      normals->Update();
+
+      vtkNew<vtkArrowSource> arrowSource;
+
+      ngs.GlyphMapper->SetInputData(normals->GetOutput());
+      ngs.GlyphMapper->SetSourceConnection(arrowSource->GetOutputPort());
+      ngs.GlyphMapper->SetOrientationModeToDirection();
+      ngs.GlyphMapper->SetOrientationArray("Normals");
```
Meakk: If not using `vtkPolyDataNormals`, the normal array isn't necessarily called "Normals".  
There is a second API that should work better:

```suggestion
      ngs.GlyphMapper->SetOrientationArray(vtkDataSetAttributes::NORMALS);
```

application/testing/CMakeLists.txt
```diff
@@ -1,14 +1,863 @@
...
-include(tests.native.cmake)
-include(tests.features.cmake)
-include(tests.backends.cmake)
-include(tests.custom.cmake)
-include(tests.home.cmake)
-include(tests.screenshot.cmake)
-include(tests.gaussian.cmake)
-include(tests.interaction.cmake)
-include(tests.plugins.cmake)
+# F3D Testing
```
Meakk: The rebase conflict is incorrect, please just move your new tests in the appropriate `test.*.cmake` file

vtkext/private/module/vtkF3DMetaImporter.cxx
Meakk: Also we should probably log a warning when trying to display normal glyphs but the model has no normal array.

Meakk (changes requested):

application/testing/CMakeLists.txt
ArchiePayne: Thanks, sorted this.

vtkext/private/module/vtkF3DMetaImporter.cxx
ArchiePayne: Ammended this

vtkext/private/module/vtkF3DMetaImporter.cxx
ArchiePayne: @Meakk Thanks for the suggestions, could you kindly check the wording for the log I have added as well as verify I'm using the correct api to get whether there are a valid set of normals or not, currently I'm using  `points->GetPointData()->GetNormals()`  and just checking if they are not null.

Meakk:
That looks good. I've run the main CI but I think you will have to add a test to cover the warning. Let's see what the coverage check says.

vtkext/private/module/vtkF3DMetaImporter.cxx
```diff
@@ -253,18 +264,40 @@ bool vtkF3DMetaImporter::Update()
...
+        this->Pimpl->NormalGlyphsActorsAndMappers.back();
+
+      ngs.InputDataHasNormals = points->GetPointData()->GetNormals() != nullptr;
+
+      vtkNew<vtkArrowSource> arrowSource;
+      ngs.GlyphMapper->SetInputData(points);
+      ngs.GlyphMapper->SetSourceConnection(arrowSource->GetOutputPort());
+      ngs.GlyphMapper->SetOrientationModeToDirection();
+      ngs.GlyphMapper->SetOrientationArray(vtkDataSetAttributes::NORMALS);
+      ngs.GlyphMapper->SetScaleFactor(0.2);
```
mwestphal: should that depends on the bounds of  the data ?

vtkext/private/module/vtkF3DMetaImporter.cxx
```diff
@@ -253,18 +264,40 @@ bool vtkF3DMetaImporter::Update()
...
+      // Create and configure normal glyph actors
+      this->Pimpl->NormalGlyphsActorsAndMappers.emplace_back(
+        vtkF3DMetaImporter::NormalGlyphsStruct(actor, importer));
+      vtkF3DMetaImporter::NormalGlyphsStruct& ngs =
+        this->Pimpl->NormalGlyphsActorsAndMappers.back();
+
+      ngs.InputDataHasNormals = points->GetPointData()->GetNormals() != nullptr;
+
+      vtkNew<vtkArrowSource> arrowSource;
+      ngs.GlyphMapper->SetInputData(points);
```
mwestphal: should you use masking ?

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -2955,6 +2968,21 @@ void vtkF3DRenderer::ConfigureColoring()
     this->ColoringPointSpritesMappersConfigured = true;
   }
 
+  //// Handle Normal Glyphs
```
mwestphal:
```suggestion
  // Handle Normal Glyphs
```

vtkext/private/module/vtkF3DRenderer.h
```diff
@@ -338,6 +338,12 @@ class vtkF3DRenderer : public vtkOpenGLRenderer
    */
   void ShowScalarBar(bool show);
 
+  /**
+   * Set the visibility of the normal glyphs actor.
+   * It will only be shown if raytracing and volume, and Point Sprites are not enabled
```
mwestphal: this comment is not super clear

vtkext/private/module/vtkF3DMetaImporter.cxx
ArchiePayne: @mwestphal sorry, I'm not familiar with what you mean here, assuming masking is a vtk concept in which you can 'mask' which points are with in the mapper. What would the ideal outcome of this be?

vtkext/private/module/vtkF3DRenderer.h
ArchiePayne: I've tried to make the comment more clear now.

vtkext/private/module/vtkF3DMetaImporter.cxx
mwestphal: I was refering to this: https://vtk.org/doc/nightly/html/classvtkGlyph3DMapper.html#a393dd3c17ef15a051fb59da4ca2f7bec

But afer discussing it with @Meakk , we decided that this should not be needed and displaying an arrow for each point is fine.

vtkext/private/module/vtkF3DMetaImporter.cxx
ArchiePayne: I have now implemented this, tested against various glb/objs in the testing data and looks good to me.

vtkext/private/module/vtkF3DMetaImporter.cxx
```diff
@@ -253,18 +264,49 @@ bool vtkF3DMetaImporter::Update()
...
+      ngs.InputDataHasNormals = points->GetPointData()->GetNormals() != nullptr;
+
+      double actorBounds[6];
+      actor->GetBounds(actorBounds);
+
+      double dx = actorBounds[1] - actorBounds[0];
+      double dy = actorBounds[3] - actorBounds[2];
+      double dz = actorBounds[5] - actorBounds[4];
+      double diagonal = std::sqrt(dx * dx + dy * dy + dz * dz);
+      double glyphScale = 0.05 * diagonal;
```
mwestphal: lets make 0.05 a constexpr on top of the file ?

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -3012,6 +3025,21 @@ void vtkF3DRenderer::ConfigureColoring()
     this->ColoringPointSpritesMappersConfigured = true;
   }
 
+  // Handle Normal Glyphs
```
mwestphal: why is this in ConfigureColoring ? normals are unrelated to coloring.

mwestphal: Create your own  `ConfigureNormalGlyphs` to handle it.

testing/baselines/TestNormalGlyphsNoNormalsAvailable.png
mwestphal, line 1: Should normals be generated on the fly ? I think its out of scope of this PR though.

mwestphal: @Meakk

Meakk: Definitely not. If there's no normal, the user expect flat shading et showing normals would be confusing.

vtkext/private/module/vtkF3DMetaImporter.cxx
snoyer: how expensive would it be to change the scaling factor on render so that the glyphs size is more constant in screen space?

ArchiePayne: I agree that constant screenspace would be a nice solution, if that's what we want I may need pointing to the shaders that need modifying as I haven't looked at that area before.

snoyer: very quick dirty proof of concept based on the existing implementation for illustration purposes
```diff
diff --git a/vtkext/private/module/vtkF3DRenderer.cxx b/vtkext/private/module/vtkF3DRenderer.cxx
index e0ab98ece..4e8fef744 100644
--- a/vtkext/private/module/vtkF3DRenderer.cxx
+++ b/vtkext/private/module/vtkF3DRenderer.cxx
@@ -2115,6 +2115,11 @@ void vtkF3DRenderer::UpdateActors()
 //----------------------------------------------------------------------------
 void vtkF3DRenderer::Render()
 {
+  if (this->UseNormalGlyphs)
+  {
+    ConfigureNormalGlyphs();
+  }
+
   if (!this->TimerVisible)
   {
     this->Superclass::Render();
@@ -2764,6 +2769,7 @@ void vtkF3DRenderer::ConfigureNormalGlyphs()
       continue;
     }
 
+    normalGlyph.GlyphMapper->SetScaleFactor(GetActiveCamera()->GetDistance() * 0.02);
     normalGlyph.Actor->SetVisibility(normalGlyphsVisible);
   }
```

https://github.com/user-attachments/assets/97812288-5c89-491c-94fc-26234b3168ef

https://github.com/user-attachments/assets/650ca64b-5ccd-4cfa-b40f-34d942b6800e

I'll have to defer to the VTK experts for insights regarding the _right_ way to do it

mwestphal: lgtm but I defer to @Meakk

Meakk: Looks nice to me :)

snoyer: > Looks nice to me :)

the visual result or the impl? is it reasonable to call `GlyphMapper->SetScaleFactor(...)` every frame?

Meakk: Both. I think the scale factor is only updating an uniform value in the shader so it should be efficient.

snoyer: nice. Is there a proper way to compute the scaling factor? I used the camera distance to test it out but I don't expect that to work with an orthographic camera

ArchiePayne: I've updated the implementation of scaling to match this for now

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -2852,16 +2888,29 @@ void vtkF3DRenderer::ShowScalarBar(bool show)
...
   if (this->UsePointSprites != use)
   {
     this->UsePointSprites = use;
-    this->CheatSheetConfigured = false;
     this->ColoringConfigured = false;
     this->PointSpritesConfigured = false;
   }
+
+  // Need to update the state of the normal glyphs if point sprites state updates
+  this->NormalGlyphsConfigured = false;
```
mwestphal: why ?

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -2852,16 +2888,29 @@ void vtkF3DRenderer::ShowScalarBar(bool show)
...
+  }
+}
+
 //----------------------------------------------------------------------------
 void vtkF3DRenderer::SetUsePointSprites(bool use)
 {
   if (this->UsePointSprites != use)
   {
     this->UsePointSprites = use;
-    this->CheatSheetConfigured = false;
```
mwestphal, line 2861: this looks incorrect

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -2841,6 +2856,27 @@ void vtkF3DRenderer::ConfigurePointSprites()
   }
 }
 
+void vtkF3DRenderer::ConfigureNormalGlyphs()
+{
+  bool normalGlyphsVisible =
+    !this->UseRaytracing && !this->UsePointSprites && this->UseNormalGlyphs;
```
mwestphal: you can should normal glyphs on point sprites, why not ?

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -107,6 +109,8 @@
 #include <chrono>
 #include <sstream>
 
+static constexpr float NormalGlyphScaleMultiplier = 0.02;
```
mwestphal: put in namespace

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -28,6 +29,7 @@
 #include <vtkCullerCollection.h>
 #include <vtkDiscretizableColorTransferFunction.h>
 #include <vtkFloatArray.h>
+#include <vtkGlyph3D.h>
```
mwestphal: not used ?

vtkext/private/module/vtkF3DMetaImporter.cxx
```diff
@@ -288,18 +299,39 @@ bool vtkF3DMetaImporter::Update()
...
+      ngs.GlyphMapper->SetInputData(points);
+      ngs.GlyphMapper->SetSourceConnection(arrowSource->GetOutputPort());
+      ngs.GlyphMapper->SetOrientationModeToDirection();
+      ngs.GlyphMapper->SetOrientationArray(vtkDataSetAttributes::NORMALS);
+      ngs.GlyphMapper->ScalingOn();
+
+      ngs.Actor->SetMapper(ngs.GlyphMapper);
+
+      this->Renderer->AddActor(ngs.Actor);
+      ngs.Actor->VisibilityOff();
```
mwestphal: you can skip that if there is no normals right ?

vtkext/private/module/vtkF3DRenderer.cxx
```diff
@@ -2841,6 +2856,27 @@ void vtkF3DRenderer::ConfigurePointSprites()
...
+  {
+    if (normalGlyphsVisible && !normalGlyph.InputDataHasNormals)
+    {
+      F3DLog::Print(F3DLog::Severity::Warning,
+        "Data does not contain any normals to display the normal glyphs with");
+      continue;
+    }
+
+    normalGlyph.GlyphMapper->SetScaleFactor(
+      GetActiveCamera()->GetDistance() * NormalGlyphScaleMultiplier);
```
snoyer: @Meakk, do you know a better way to compute the scale for the arrows? some magic trick with the camera's projection matrix maybe?

Meakk: Nothin comes to mind. But this formula looks perfectly fine to me. Have you noticed something wrong with it?

snoyer: it won't work with an orthographic camera and it doesn't take into account the view angle (which changes when zooming if I remember correctly), so I assume computing something from the final projection rather than some properties of the VTK camera would be more robust. I don't have anything to back any of that up tho :/

vtkext/private/module/vtkF3DRenderer.cxx
ArchiePayne: It was just an assumption that users wouldn't want both, but this could work of course. I've altered the code to allow both Point Sprites and Normal Glyphs.

vtkext/private/module/vtkF3DRenderer.cxx
Meakk: Let's use this formula for now, maybe someone smarter than us will come with a better approach in the future.

snoyer: taking the problem by the other end we could make it work for orthographic camera first using the camera's `parallelScale` property instead of the distance (which is undefined in that case)

Then for the perspective camera case we can use the same formula we already use here https://github.com/f3d-app/f3d/blob/a0fac0410dd5cc7b5aac3368fd5eb38911b46d86/vtkext/private/module/vtkF3DRenderer.cxx#L1999-L2000 to approximate the equivalent parallel scale

snoyer:
```diff
diff --git a/vtkext/private/module/vtkF3DRenderer.cxx b/vtkext/private/module/vtkF3DRenderer.cxx
index 97b969689..a538ee690 100644
--- a/vtkext/private/module/vtkF3DRenderer.cxx
+++ b/vtkext/private/module/vtkF3DRenderer.cxx
@@ -236,7 +236,7 @@ void ExecFuncOnAllPolyDataUniforms(vtkActorCollection* actors, F&& func)
   }
 }
 
-static constexpr float NormalGlyphScaleMultiplier = 0.02;
+static constexpr float NormalGlyphScaleMultiplier = 0.15; // probably needs tweaking
 }
 
 //----------------------------------------------------------------------------
@@ -2876,10 +2876,24 @@ void vtkF3DRenderer::ConfigureNormalGlyphs()
 
 void vtkF3DRenderer::UpdateNormalGlyphsScale()
 {
+  const auto getScale = [](vtkCamera* camera)
+  {
+    if (camera->GetParallelProjection())
+    {
+      return camera->GetParallelScale();
+    }
+    else
+    {
+      const double angle = vtkMath::RadiansFromDegrees(camera->GetViewAngle());
+      const double distance = camera->GetDistance();
+      return distance * tan(angle / 2);
+    }
+  };
+
+  const double scaleFactor = getScale(GetActiveCamera()) * NormalGlyphScaleMultiplier;
   for (const auto& normalGlyph : this->Importer->GetNormalGlyphsActorsAndMappers())
   {
-    normalGlyph.GlyphMapper->SetScaleFactor(
-      GetActiveCamera()->GetDistance() * NormalGlyphScaleMultiplier);
+    normalGlyph.GlyphMapper->SetScaleFactor(scaleFactor);
   }
 }
```

Meakk:
I suggest to move forward with this one and log an issue to improve scaling with zoom factor and orthographic camera later. We should also consider other improvements like specifying a color for the glyphs, changing the arrow discretization, the glyph type, etc...

snoyer:
> I suggest to move forward with this one and log an issue to improve scaling with zoom factor and orthographic camera later.

@Meakk see https://github.com/f3d-app/f3d/pull/2811#discussion_r2877722808 for a suggestion that uses the orthographic camera's parallel scale as the base case and a conversion from distance and view angle for the perspective camera (based on the existing camera switcher code)

vtkext/private/module/vtkF3DRenderer.cxx
Meakk: It's great and it makes sense to me. However, that would need more tests (with orthographic camera, and different FOV) to validate and cover that. If @ArchiePayne wants to take care of it now, that's fine, but it can be done later too.

Meakk:
@ArchiePayne Can you rebase and update `TestInteractionCheatsheetCentered.png`?
Let us know if you have time to address @snoyer's suggestion, otherwise we'll do it in a follow-up PR.

Meakk:
I'm taking care of the rebase and adding @snoyer's suggestion in #2958

Meakk:
Closing this one. Thanks for the prototype @ArchiePayne :+1:

Issue #2641, opened 2025-11-17
Search bar in cheatsheet


noclone:
**Is your feature request related to a problem? Please describe.**
There are a lot of bindings displayed in the cheatsheet, it can be hard to find a specific option or bind.

**Describe the solution you'd like**
A search bar at the top of the cheatsheet to filter the rows would be great, with an option to filter by keybind instead of word.
This would allow to quickly find the option you are looking for, or to check if it exists. The "filter by keybind" option would allow to quickly check if a key is already taken or if it is available to bind to a new option.

mwestphal:
Interesting idea, thats true that its growing more and more.

RealEdwardS:
Hi, can I get assigned to this?

mwestphal:
Sure!

PR #2883, opened 2026-02-14
Add a search bar on top of the cheat sheet to filter rows

noclone:
### Describe your changes

Add a search bar on top of the cheat sheet to filter rows.
The user can select between two buttons that can either filter between description or keybind.

based on the work done in https://github.com/f3d-app/f3d/pull/2750

https://github.com/f3d-app/f3d/issues/2641

### Continuous integration


noclone:
\ci fast

noclone:
@mwestphal please review

noclone:
\ci full

testing/baselines/TestSDKTriggerInteractionsCheatSheetSearch.png
mwestphal, line 1: searching by keybind ? is that useful ?

noclone: personnally I used it a lot in other softwares/video games, so I figured it makes sense here to.
It allows to very quickly find if a keybind is already assigned, in the event that you want to add a new one or change them.

mwestphal: right, we cant change bind interactively yet, but I guess that makes sense.

snoyer: does it need the toggle tho? could it search in both description and keybind by default to keep the UI simple?

noclone: it kind of defeats the purpose imo, if you search "e" in both for example, you'll have plenty of results, while you may have wanted to find the keybind "e"

snoyer: that does make sense in theory, although in practice searching `e` in bindings doesn't exactly return a single item either :D
image

noclone: right, but it's still less than if we searched in both ^^

vtkext/private/module/vtkF3DImguiActor.cxx
```diff
@@ -273,6 +275,10 @@ struct vtkF3DImguiActor::Internals
   vtkSmartPointer<vtkOpenGLBufferObject> IndexBuffer;
   vtkSmartPointer<vtkShaderProgram> Program;
   vtkSmartPointer<vtkTextureObject> LogoTexture;
+
+  std::array<char, 256> SearchFilter = {};
+  int SearchMode = 0;
```
mwestphal: use a proper enum ?

vtkext/private/module/vtkF3DImguiActor.cxx
```diff
@@ -273,6 +275,10 @@ struct vtkF3DImguiActor::Internals
   vtkSmartPointer<vtkOpenGLBufferObject> IndexBuffer;
   vtkSmartPointer<vtkShaderProgram> Program;
   vtkSmartPointer<vtkTextureObject> LogoTexture;
+
+  std::array<char, 256> SearchFilter = {};
```
mwestphal, line 286: std::string ?

library/testing/TestSDKTriggerInteractions.cxx
```diff
@@ -117,6 +117,40 @@ int TestSDKTriggerInteractions([[maybe_unused]] int argc, [[maybe_unused]] char*
       f3d::interactor::InputAction::RELEASE, f3d::interactor::MouseButton::RIGHT);
   });
 
+  // Trigger cheatsheet search filtering
+  helper("TestSDKTriggerInteractionsCheatSheetSearch", [](f3d::engine& engine) {
+    engine.getInteractor().triggerKeyboardKey(f3d::interactor::InputAction::PRESS, "H");
+    engine.getInteractor().triggerKeyboardKey(f3d::interactor::InputAction::RELEASE, "H");
+    engine.getInteractor().triggerMousePosition(80, 30);
```
mwestphal, line 124: would it make sense to have a short cut to search ?
Im not sur it does tbh

vtkext/private/module/vtkF3DImguiActor.cxx
noclone: ImGui::InputTextWithHint needs a char* to write in the buffer

vtkext/private/module/vtkF3DImguiActor.cxx
noclone: I added an enum for the value, but the type has to remain int, since ImGui::RadioButton's API needs an int

library/testing/TestSDKTriggerInteractions.cxx
noclone: I think it would add complexity for a small gain, not sure it's worth it

vtkext/private/module/vtkF3DImguiActor.cxx
mwestphal: std::string::data ? ^^

noclone: the issue is that ImGui expects a preallocated buffer with a fixed buf_size. With a std::string, we'd have to .resize(256). At that point it's just a worse std::array<char, 256>, isn't it ?

mwestphal: yup, you are right.

vtkext/private/module/vtkF3DImguiActor.cxx
mwestphal: got it.

noclone:
@mwestphal I'm not sure what to do about the static analysis error, since the enum has to use `int`, because that's the type required by the ImGui API

mwestphal:
> @mwestphal I'm not sure what to do about the static analysis error, since the enum has to use `int`, because that's the type required by the ImGui API

Add an exception

https://github.com/f3d-app/f3d/blob/master/.cppcheck.supp

Meakk:
Alternatively, there's another API:
```cpp
RadioButton(const char* label, bool active);                    // use with e.g. if (RadioButton("one", my_value==1)) { my_value = 1; }
```
That should allow you to use `enum class` with any internal value storage format.

noclone:
> Alternatively, there's another API:
> 
> ```c++
> RadioButton(const char* label, bool active);                    // use with e.g. if (RadioButton("one", my_value==1)) { my_value = 1; }
> ```
> 
> That should allow you to use `enum class` with any internal value storage format.

Indeed thanks, updated

testing/baselines/TestInteractionCheatsheetAnimationName.png
Meakk, line 1: The blue disk is not the F3D blue, can we change that?

testing/baselines/TestInteractionCheatsheetCentered.png
Meakk, line 1: incorrect baseline, wrong blue

Issue #2578, opened 2025-10-28
binding feedback on interaction


mwestphal:
**Is your feature request related to a problem? Please describe.**
When interacting with F3D and pressing binds, there is not much feedback except on the cheatsheet or obviously visually.

Let's add a visual feedback that disapear after a while.

**How**
 - Add an ImGui window dedicated to display interactive binds, probably bottom right
 - When pressing a bind, display a reference to that interaction there for a few seconds
 - When pressing multiple binds, they should stack vertically
 - Backdrop should be present, extending but disapear too if nothing is shown

snoyer:
Are there any events other than key presses that should trigger one of these notifications? and if so does it affect the design for the implementation?

I'm thinking of when a file is reloaded with `--watch` for example

mwestphal:
> Are there any events other than key presses that should trigger one of these notifications? and if so does it affect the design for the implementation?
> 
> I'm thinking of when a file is reloaded with `--watch` for example

Hum, I did not think of that but I suppose, yes.

davidwuluetang:
Does that mean when a file reloaded, actived binds should trigger notifications? And options relate to binds in config file also trigger notifications after app launched?

mwestphal:
no and no.

davidwuluetang:
How does `--watch` related to binding? and what is the expected behavior for notification when in use?

mwestphal:
`--watch` doesnt related to binding, but it would be useful to inform user interactively that their fiule have been reloaded.

Just something to display in screen, some kind of status bar somewhere that would display information for a time.

davidwuluetang:
So that file reload is not trigger by command, now I see why those questions were asked.

mwestphal:
> So that file reload is not trigger by command, now I see why those questions were asked.

nice :)

davidwuluetang:
Hi @mwestphal, I would like to work this.

mwestphal:
Assigned :)

mwestphal:
Fixed by https://github.com/f3d-app/f3d/pull/2970

PR #2893, opened 2026-02-17
Implement notification window and binding feedback on interaction

davidwuluetang:
### Describe your changes
Implement binding notification window to bottom left
- Added APIs to libf3d.interactor
- Modified `interactor_impl::TriggerBinding()`
- Added `RenderNotifications()` on `vtkF3DImguiActor`

  #2578 

- [ ] I have performed a [self-review](https://f3d.app/dev/CODING_STYLE) of my code

### Continuous integration

### Screenshot
scene_14

### DIFF
--- a/library/src/interactor_impl.cxx
+++ b/library/src/interactor_impl.cxx
@@ -33,6 +33,7 @@
 #include <vtkStringArray.h>
 #include <vtkVersion.h>
 #include <vtksys/SystemTools.hxx>
+#include <vtkTimerLog.h>
 
 #include <algorithm>
 #include <chrono>
@@ -571,7 +572,7 @@ public:
   //----------------------------------------------------------------------------
   bool StartEventLoop(double deltaTime, std::function<void()> userCallBack)
   {
-    if (this->EventLoopObserverId != -1)
+    if (this->EventLoopObserverId[0] != -1)
     {
       log::info("Interaction: event loop has already been started");
       return false;
@@ -594,10 +595,15 @@ public:
       [](vtkObject*, unsigned long, void* clientData, void*)
       {
         internals* that = static_cast<internals*>(clientData);
-        that->EventLoop(that->CallbackDeltaTime);
+        double now = vtkTimerLog::GetUniversalTime();
+        double dt = now - that->LastTime;
+        that->LastTime = now;
+        that->EventLoop(dt);
       });
-    this->EventLoopObserverId =
+    this->EventLoopObserverId[0] =
       this->VTKInteractor->AddObserver(vtkCommand::TimerEvent, timerCallBack);
+    this->EventLoopObserverId[1] =
+      this->VTKInteractor->AddObserver(vtkCommand::InteractionEvent, timerCallBack);
     timerCallBack->SetClientData(this);
     return true;
   }
@@ -605,14 +611,16 @@ public:
   //----------------------------------------------------------------------------
   bool StopEventLoop()
   {
-    if (this->EventLoopObserverId == -1)
+    if (this->EventLoopObserverId[0] == -1)
     {
       log::info("Interaction: event loop has not been started hence cannot be stopped");
       return false;
     }
-    this->VTKInteractor->RemoveObserver(this->EventLoopObserverId);
+    this->VTKInteractor->RemoveObserver(this->EventLoopObserverId[0]);
+    this->VTKInteractor->RemoveObserver(this->EventLoopObserverId[1]);
     this->VTKInteractor->DestroyTimer(this->EventLoopTimerId);
-    this->EventLoopObserverId = -1;
+    this->EventLoopObserverId[0] = -1;
+    this->EventLoopObserverId[1] = -1;
     this->EventLoopTimerId = 0;
     return true;
   }
@@ -708,11 +716,12 @@ public:
 
   std::function<void()> EventLoopUserCallBack = nullptr;
   unsigned long EventLoopTimerId = 0;
-  int EventLoopObserverId = -1;
+  int EventLoopObserverId[2] = { -1, -1 };
   std::atomic<bool> RenderRequested = false;
   std::atomic<bool> StopRequested = false;
 
   double CallbackDeltaTime = 1.0 / 30; /* Default DeltaTime (30fps) */
+  double LastTime = vtkTimerLog::GetUniversalTime();
 };
 
 //----------------------------------------------------------------------------


mwestphal (changes requested):
please share a screenshot of how it looks.

mwestphal:
Still working on this @davidwuluetang ?

davidwuluetang:
For the feature itself, the work is basically done. For the event loop freeze and slow down problem, I tried some workaround. Which adding another observer to `VTKInteractor` watch for `InteractionEvent`, but this require a timer feeding real delta time to event loop (not longer use a fix delta time). So, Idk this is a good solution or not.

mwestphal:
> For the feature itself, the work is basically done. For the event loop freeze and slow down problem, I tried some workaround. Which adding another observer to `VTKInteractor` watch for `InteractionEvent`, but this require a timer feeding real delta time to event loop (not longer use a fix delta time). So, Idk this is a good solution or not.

Ok, let us know if you need help with that.

davidwuluetang:
I just push all code, maybe you guys can try it out, it work but Idk how this solution can effect other things. If this is not a acceptable solution, you can unassign me from this issue, cus I really don't think I can come up a solution for this problem.

mwestphal:
> I just push all code, maybe you guys can try it out, it work but Idk how this solution can effect other things. If this is not a acceptable solution, you can unassign me from this issue, cus I really don't think I can come up a solution for this problem.

Got it, we will take a look.

application/F3DStarter.cxx
```diff
@@ -977,6 +976,8 @@ class F3DStarter::F3DInternals
       interactor.addBinding({mod_t::CTRL, "Q"}, "exit", "Others", std::bind(docString, "Quit"));
       // clang-format on
 
+      interactor.initBindNotificationMap();
```
Meakk, line 979: The API name is too implementation oriented. We should use `Map` because the user doesn't care and we may change impl details.  
What about `initNotificationSystem()`? Or just initialize it automatically with the interactor if possible?

library/private/interactor_impl.h
```diff
@@ -132,6 +136,19 @@ class interactor_impl : public interactor
    */
   void SetCommandBuffer(const char* command);
 
+  /**
+   * Implementation only API.
+   * Add or modify binding documentation callback to binding notification map using command string
+   * as key.
+   */
+  void setBindNotiCallback(std::string command, documentation_callback_t doc_callback);
```
Meakk, line 144: implementation only functions starts with an upper case letter.

vtkext/private/module/vtkF3DImguiActor.cxx
```diff
@@ -968,3 +886,169 @@ void vtkF3DImguiActor::SetDeltaTime(double time)
...
+
+  for (auto it = this->Notifications.begin(); it != this->Notifications.end();)
+  {
+    const auto& [desc, value, bind, duration, start_time ] = *it;
+    const double elapsed = this->TotalTime - start_time;
+
+    if (elapsed > duration)
+    {
+      it = Notifications.erase(it);
+    }
```
Meakk, line 905: Should we move the logic of notification cleanup to `vtkF3DUIActor::SetTotalTime`? That way you ensure that the map is not growing infinitely if imgui is not enabled.

library/src/interactor_impl.cxx
```diff
@@ -557,25 +595,32 @@ class interactor_impl::internals
...
-        that->EventLoop(that->CallbackDeltaTime);
+        double now = vtkTimerLog::GetUniversalTime();
+        double dt = now - that->LastTime;
+        that->LastTime = now;
+        that->EventLoop(dt);
       });
-    this->EventLoopObserverId =
+    this->EventLoopObserverId[0] =
       this->VTKInteractor->AddObserver(vtkCommand::TimerEvent, timerCallBack);
+    this->EventLoopObserverId[1] =
```
Meakk, line 605: Is it needed? If the goal is to update the time when pressing a binding, we could just update the time in `TriggerBinding` function

library/src/interactor_impl.cxx
```diff
@@ -557,25 +595,32 @@ class interactor_impl::internals
       [](vtkObject*, unsigned long, void* clientData, void*)
       {
         internals* that = static_cast<internals*>(clientData);
-        that->EventLoop(that->CallbackDeltaTime);
+        double now = vtkTimerLog::GetUniversalTime();
+        double dt = now - that->LastTime;
```
Meakk, line 599: Can you use `deltaTime` instead of computing `dt` here?

library/src/interactor_impl.cxx
```diff
@@ -557,25 +595,32 @@ class interactor_impl::internals
       [](vtkObject*, unsigned long, void* clientData, void*)
       {
         internals* that = static_cast<internals*>(clientData);
-        that->EventLoop(that->CallbackDeltaTime);
+        double now = vtkTimerLog::GetUniversalTime();
```
Meakk, line 598: That's wrong I think:
> Returns the elapsed number of **seconds** since 00:00:00 Coordinated Universal Time (UTC), Thursday, 1 January 1970. 

We need a millisecond accurate time

Meakk: nevermind, there are second fraction too.

application/F3DStarter.cxx
davidwuluetang: I don't have opinion with name change.

> Or just initialize it automatically with the interactor if possible?

This function do more than init, it clear the map and reload the callback as well. So, if users make change on the binding naming or callback, they need to call this function after the change in order for the change reflect on binding notifications. That why I make it public.

library/src/interactor_impl.cxx
davidwuluetang: This part is the workaround I mention for event loop freeze issue on Windows system when interacting with camera, not really relate to the binding notification feature itself. It freeze the animation and the `UITotalTime` update, make the notification stay on screen if you interacting the camera.

library/src/interactor_impl.cxx
davidwuluetang: It will work as well, but animation will look slightly faster when `deltaTime` is `1/30`. I tested on my machine it stay constant on 36-37 loops per second when interacting.

davidwuluetang:
So, you guys decided to take out the work from here?

Meakk:
Yes, see https://github.com/f3d-app/f3d/pull/2970

Meakk:
Closing this one in favor of https://github.com/f3d-app/f3d/pull/2970  
I'll keep you as co-author on this one.  
Please feel free to make comments on it if I missed something important.

PR #2984, opened 2026-03-29
add fade-in animation to notifications

snoyer:
### Describe your changes

Add slide-up and fade-in animations to notifications to (hopefully) make them visually smoother and easier to track.

Also implements re-using the last notification when appropriate, that is when it has the same description but a different value. This should make it easier to visually parse:
- notifications cycling the same value (AA, light intensity, ...) by not spamming fade-in animations
- notifications that have the same description and value (or no value at all, "play/pause animation", file reloaded with `--watch`) by having the animation to hightlight that there has indeed been a new notification even tho it looks exactly the same 
 
This is currently a draft PR because the fade-in breaks all of the relevant tests as you need to wait a few frames to see the new notifications.

### DIFF
--- a/application/testing/tests.interaction.cmake
+++ b/application/testing/tests.interaction.cmake
@@ -39,6 +39,7 @@ f3d_test(NAME TestInteractionSimpleExit DATA cow.vtp REGEXP "Interactor has been
 # Needs some event loop fix that's not in 9.3
 if(VTK_VERSION VERSION_GREATER_EQUAL 9.4.0)
   f3d_test(NAME TestInteractionNotifications DATA cow.vtp ARGS --notifications INTERACTION UI) #E;
+  f3d_test(NAME TestInteractionNotificationsUpdate DATA cow.vtp ARGS --notifications INTERACTION UI) #E;E;Up;Up
   f3d_test(NAME TestInteractionNotificationsBindings DATA cow.vtp ARGS --notifications -Dui.notifications.show_bindings=ON INTERACTION UI) #E;
   f3d_test(NAME TestInteractionNotificationsBindingsModifier DATA cow.vtp ARGS --notifications -Dui.notifications.show_bindings=ON RESOLUTION 400,300 INTERACTION UI) #SHIFT+L;
   f3d_test(NAME TestInteractionToggleNotifications DATA cow.vtp INTERACTION UI) #CTRL+K;
--- a/testing/baselines/TestInteractionNotificationsUpdate.png
+++ b/testing/baselines/TestInteractionNotificationsUpdate.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:19152329a1092bfb0eefa31616ff6494c29c1a15b16883db2988411bc186b533
+size 25785
--- a/testing/recordings/TestInteractionNotificationsUpdate.log
+++ b/testing/recordings/TestInteractionNotificationsUpdate.log
@@ -1,0 +1,36 @@
+# StreamVersion 1.2
+ExposeEvent 0 299 0 0 0 0 0
+RenderEvent 0 299 0 0 0 0 0
+
+# Press E to show a notification
+KeyPressEvent -357 -370 0 101 1 e 0
+CharEvent -357 -370 0 101 1 e 0
+KeyReleaseEvent -357 -370 0 101 1 e 0
+TimerEvent -357 -370 0 101 1 e 0
+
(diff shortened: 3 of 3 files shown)


Meakk:
I'm fine with that. For the test issue, I suppose we could fix it by adding libf3d options to configure fade in time and use 0 in the application tests? We could test the fade in in the SDK test.

snoyer:
> I suppose we could fix it by adding libf3d options to configure fade in time and use 0 in the application tests?

We could do that, or maybe just have a boolean to enable/disable animations altogether, but IMO that should only be done if there's a good UX reason. Doing it for testing only doesn't make much sense to me because the behavior would be too different from the real use case that the passing tests would be meaningless.

On the other hand, could we use interaction tests? I believe that's how you tested the TAA which also needs a delay to take effect?

snoyer:
\ci fast

snoyer:
\ci main

snoyer:
\ci -fast

snoyer:
\ci -main

snoyer:
\ci full

snoyer:
failed job is `f3d::TestNormalGlyphsPerspectiveEnable (Timeout)`, unrelated?

mwestphal:
> TestNormalGlyphsPerspectiveEnable

I've rerun, we will see.

testing/baselines/TestSDKNotificationFading.png
Meakk, line 1: Is this change required because it's pushed 0.2s ahead in time?

testing/recordings/TestInteractionNotificationsBindings.log
```diff
@@ -7,3 +7,12 @@ KeyPressEvent -357 -370 0 101 1 e 0
 CharEvent -357 -370 0 101 1 e 0
 KeyReleaseEvent -357 -370 0 101 1 e 0
 TimerEvent -357 -370 0 101 1 e 0
+
+# Allow fade-in
+TimerEvent 0 299 0 0 0 0 0
```
Meakk, line 12: Exactly 3 renders are needed because it's 0.1s so exactly the fade-in and slide-up time?

vtkext/private/module/vtkF3DUIActor.cxx
```diff
@@ -280,8 +280,19 @@ int vtkF3DUIActor::RenderOverlay(vtkViewport* vp)
...
-void vtkF3DUIActor::AddNotification(
-  const std::string& desc, const std::string& value, const std::string& bind, double stopTime)
+void vtkF3DUIActor::AddNotification(const std::string& desc, const std::string& value,
+  const std::string& bind, double startTime, double duration)
 {
-  this->Notifications.emplace_front(Notification{ desc, value, bind, stopTime });
+  if (!this->Notifications.empty())
+  {
+    Notification& last = this->Notifications.front();
+    if (last.desc == desc && last.value != value)
```
Meakk, line 289: We'll need a new test to cover that. Like pressing twice a toggle binding and capture that a single notification is visible.

Meakk: Or add an additional grid toggle at the end of `TestSDKNotification` and change the baseline (same number of notifications but the last one should show the grid is enabled)

Meakk: Nevermind, relying on the grid will cause macOS issues. Maybe call it 3 times and don't change the baseline ^^

testing/baselines/TestSDKNotificationFading.png
snoyer: yes, the sub-tests in `TestSDKNotification.cxx` aren't independent and the wait I added for the first one has a side effect on the second one

testing/recordings/TestInteractionNotificationsBindings.log
snoyer: I didn't calculate it ahead of time TBH, I started with more of them and removed until the tests didn't pass anymore

vtkext/private/module/vtkF3DUIActor.cxx
snoyer: yes, also having a non-value/same-value notification twice and checking that it does _not_ merge these. I'll see if I can make it happen

testing/recordings/TestInteractionNotificationsBindings.log
Meakk: F3D is 30fps by default so 3 ticks is exactly 0.1 so I assume it's correct.

Issue #2974, opened 2026-03-26
Add support for specifying default liboptions through parsing


mwestphal:
**Is your feature request related to a problem? Please describe.**
libf3d options currently require to be constructable, but we should be able to use native parsing instead.
eg colors are set using RGB instead of their names.

**Describe the solution you'd like**
Add cmake logic to parse instead of using constructor

mwestphal:
Related to https://github.com/f3d-app/f3d/pull/2967

Prospect138:
Its related to 2967, so I wanna try do this

mwestphal:
It is not as easy, you will need to write some CMake code, let me know :)

mwestphal:
Fixed by https://github.com/f3d-app/f3d/pull/2999

PR #2999, opened 2026-03-31
Feature/liboptions parsing

Prospect138:
### Describe your changes
Changed CMake code generation function, it now generate parsing of default_value of options instead of constructing type with it

#2974

### Continuous integration


cmake/f3dOptions.cmake
```diff
@@ -182,11 +162,11 @@ function(_parse_json_option _top_json)
 
        if(_default_value_error STREQUAL "NOTFOUND")
          # Use default_value
-         set(_optional_default_value_initialize "${_option_explicit_constr}${_option_default_value_start}${_option_default_value}${_option_default_value_end}")
+         set(_optional_default_value_initialize "parse<${_option_actual_type}>(\"${_option_default_value}\")")
```
mwestphal:
```suggestion
         set(_optional_default_value_initialize "options_tools::parse<${_option_actual_type}>(\"${_option_default_value}\")")
```
?

cmake/f3dOptions.cmake
```diff
@@ -182,11 +162,11 @@ function(_parse_json_option _top_json)
...
          string(APPEND _options_struct "${_option_indent}  ${_option_deprecated_string}${_option_actual_type} ${_member_name} = ${_optional_default_value_initialize};\n")
          set(_optional_getter "")
          list(APPEND _options_is_optional "if (name == \"${_option_name}\") return false")
-         list(APPEND _options_reset "if (name == \"${_option_name}\") opt.${_option_name} = ${_optional_default_value_initialize}")
+         list(APPEND _options_reset "if (name == \"${_option_name}\") opt.${_option_name} = options_tools::parse<${_option_actual_type}>(\"${_option_default_value}\")")
```
mwestphal: i dont think this change is needed ?

cmake/f3dOptions.cmake
Prospect138: its an options::parse that wraps options_tools::parse in options.cxx, so not sure it compiles. Do i need to include options_tools in options.h instead? Seems like it would make circular dependency...

mwestphal: sure, just make it work first :)

Prospect138:
\ci fast

Prospect138:
it works as it is

cmake/f3dOptions.cmake
mwestphal:
```suggestion
         set(_optional_default_value_initialize "options::parse<${_option_actual_type}>(\"${_option_default_value}\")")
```

cmake/f3dOptions.cmake
mwestphal: this change is not needed


```suggestion
         list(APPEND _options_reset "if (name == \"${_option_name}\") opt.${_option_name} = ${_optional_default_value_initialize}")
```

library/options.json
```diff
@@ -191,15 +191,15 @@
...
       "type": "color",
-      "default_value": "0.47843137254901963,1.0,0.47843137254901963"
+      "default_value": "f3d_green"
     },
     "z_color": {
       "type": "color",
-      "default_value": "0.47058823529411764,0.5450980392156862,1.0"
+      "default_value": "f3d_blue"
     },
     "scalar_bar": {
```
mwestphal, line 204: - font_color should be f3d_white
 - animation_progress_color could now have a default value of f3d_blue
 - loader_progress_color could now have a default value of f3d_yellow

mwestphal (changes requested):
changes needed

library/src/animationManager.cxx
```diff
@@ -69,15 +70,16 @@ void animationManager::Initialize()
     progressRep->SetPosition(0.0, 0.0);
     progressRep->SetPosition2(1.0, 0.0);
     progressRep->SetMinimumSize(0, 5);
+    std::string animationProgress = "ui.animation_progress_color";
     f3d::color_t color;
-    if (!this->Options.ui.animation_progress_color.has_value())
+    if (!this->Options.isOptional(animationProgress) && std::holds_alternative<std::vector<double>>(this->Options.get(animationProgress)))
```
mwestphal: no need to check isOptional, with the default_value, you know it is not optional, just use the value.

Prospect138: what if we leave options.json like:
```
    "animation_progress_color": {
      "type": "color"
    },
```
with no default value, or it is unexpected now?

mwestphal: thats the point, now the animation_progress_color default behavior is directly in options.json, so no need to check for optional anymore.

Please come on discord to ask questions, it will be more fluid :)

mwestphal:
\ci main

library/options.json
```diff
@@ -275,21 +275,23 @@
...
       "type": "bool",
       "default_value": "false"
     },
     "animation_progress_color": {
-      "type": "color"
+      "type": "color",
+      "default_value": "f3d_blue"
     },
     "dpi_aware": {
       "type": "bool",
```
mwestphal, line 297: nice please update `doc/libf3d/03-OPTIONS.md` and `doc/user/03-OPTIONS.md` doc accordingly

mwestphal:
\ci full

snoyer:
Does that make the initialization of every struct member go through parsing? I mean does it now generate
```cpp
struct foo {
  bool bar = parse<bool>("true");
  int baz = parse<int>("1");
}
```
where it used to do
```cpp
struct foo {
  bool bar = true;
  int baz = 1;
}
```
And if so is that enough of an problem that it should be avoided somehow? (have a separate way to specify defaults that need to be parsed? make the parsing functions constexpr so it the parsing can be done at compile time and doesn't matter?)

mwestphal:
> Does that make the initialization of every struct member go through parsing? I mean does it now generate
> 
> ```c++
> struct foo {
>   bool bar = parse<bool>("true");
>   int baz = parse<int>("1");
> }
> ```
> 
> where it used to do
> 
> ```c++
> struct foo {
>   bool bar = true;
>   int baz = 1;
> }
> ```
> 
> And if so is that enough of an problem that it should be avoided somehow? (have a separate way to specify defaults that need to be parsed? make the parsing functions constexpr so it the parsing can be done at compile time and doesn't matter?)

Definitely does, not sure if this is an issue or not. Wdyt @Meakk ?

Prospect138:
> Does that make the initialization of every struct member go through parsing? I mean does it now generate
> 
> ```c++
> struct foo {
>   bool bar = parse<bool>("true");
>   int baz = parse<int>("1");
> }
> ```
> 
> where it used to do
> 
> ```c++
> struct foo {
>   bool bar = true;
>   int baz = 1;
> }
> ```
> 
> And if so is that enough of an problem that it should be avoided somehow? (have a separate way to specify defaults that need to be parsed? make the parsing functions constexpr so it the parsing can be done at compile time and doesn't matter?)

yes, i was thinking about it, but it takes std::string as arg and e g color_t parse() makes lot of dynamic stuff inside, so even if i mark this as constexpr it could not be calculated at compile time on C++17 anyway. So if we want to pass options like this, runtime parsing is the way.

Meakk:
It's fine but I think we need `constexpr` parse function alternatives taking `string_view` as argument.

But it's only called when initializing the `f3d::options` class, right? It's neglectable.

mwestphal:
> It's fine but I think we need constexpr parse function alternatives taking string_view as argument.

Ok lets do that. @Prospect138 do you know how to add that ?

> But it's only called when initializing the f3d::options class, right? It's neglectable.

Indeed

Prospect138:
\ci full

mwestphal:
thanks for the contribution @Prospect138 !

PR #2749, opened 2025-12-24
Draft: Added Support for TimeSteps in vtkF3DGenericImporter

vishnukoushik143:
### Describe your changes
Added Two more members to store the TimeSteps and NumberOfTimeSteps in vtkF3DGenericImporter class's Internals struct.
Modified GetTemporalInformation and UpdateTemporalInformation methods of the class accordingly.

### Issue ticket number and link if any
#2733 

### Checklist for finalizing the PR

- [x] I have performed a [self-review](https://f3d.app/dev/CODING_STYLE) of my code
- [ ] I have added [tests](https://f3d.app/dev/TESTING) for new features and bugfixes
- [ ] I have added [documentation](https://f3d.app/docs/next/user/QUICKSTART) for new features
- [ ] If it is a modifying the libf3d API, I have updated bindings
- [ ] If it is a modifying the `.github/workflows/versions.json`, I have updated `docker_timestamp`

### Continuous integration

Please write a comment to run CI, eg: `\ci fast`.
See [here](https://f3d.app/dev/CONTRIBUTING#continuous-integration) for more info.

Issue #2733, opened 2025-12-21
Add support for recovering keyframes/timesteps in vtkF3DGenericImporter


mwestphal:
**Is your feature request related to a problem? Please describe.**
F3D support playing animation, but assumes interpolation between frames, we will soon add support for jumping between keyframes/timesteps, but not all readers/importers supports that.

Lets add keyframes/timesteps recovery to the vtkF3DGenericImporter

**Describe the solution you'd like**
 - Implement GetTemporalInformation so that it also return the list of timesteps and not just the timerange
 - Add a simple SDK test for it

vishnukoushik143:
Hi @mwestphal , this is VK from Discord. Can I work on this issue?

mwestphal:
Some more details.

vtkF3DGenericImporter::GetTemporalInformation currently does not fill up the timesteps. To fix this issue is should fill them.
You can recover the timesteps in `UpdateTemporalInformation` by recovering the `TIME_STEPS` information key.

See https://vtk.org/doc/nightly/html/group__InformationKeys.html#gad1634046d0539f9ea928b3479a5f8b8a

with something like that:

```
    this->NumberOfTimeSteps = inInfo->Length(vtkStreamingDemandDrivenPipeline::TIME_STEPS());
    this->TimeValues.resize(this->NumberOfTimeSteps, 0.0);
    auto timeValues = inInfo->Get(vtkStreamingDemandDrivenPipeline::TIME_STEPS());
```

mwestphal:
Unassigned as requested.

CPerry26:
@mwestphal Heya, this is CP26 from Discord. May I take this issue?

mwestphal:
Of course, assigned :)

Meakk:
Fixed by #3043

PR #3043, opened 2026-04-17
Added support for time steps in vtkF3DGenericImporter

CPerry26:
### Describe your changes
Add support for time steps when updating temporal information in `vtkF3DGenericImporter.cxx`. Also added a test to ensure the values are set correctly when there's animation. I wasn't sure if there was existing test data to be used specifically for this, so I copy/pasted some data from another test in the generic importer tests and then validated the temporal information existed as expected. I don't know if this is sufficient testing, happy to add more!

I would also like to note I didn't change much from the original draft PR on this issue: https://github.com/f3d-app/f3d/pull/2749, credit going to @vishnukoushik143. Based on that PR, the code was pretty much there, it just needed some testing. I made sure all the original PR feedback was included and the style matched. Please let me know if this should be adapted further.

Addresses [ISSUE-2733](https://github.com/f3d-app/f3d/issues/2733)

### AI Disclosure

- [x] I did not use AI to generate any of the content of that pull request
- [ ] I used AI to generate code in that pull request, if yes [please disclose](https://f3d.app/dev/AI_POLICY) which part of the code was generated and with which model.
- ...

### Continuous integration


CPerry26:
\ci fast

mwestphal:
thanks! ill review in details when I can. TO fix the style, just grab the diff from the comment above.

vtkext/private/module/Testing/TestF3DGenericImporter.cxx
```diff
@@ -441,5 +442,36 @@ int TestF3DGenericImporter(int argc, char* argv[])
     }
   }
 
+  // Test time step temporal information
+  {
+    vtkNew<vtkGLTFReader> reader;
+    std::string filename = std::string(argv[1]) + "data/BoxAnimated.gltf";
```
mwestphal: I dont think using the GLTFReader is the best choice here. Better to use a reader used within F3D, like the vtkHDFReader

mwestphal (changes requested):
looks good, just need the style and test fixed

vtkext/private/module/vtk.module
```diff
@@ -31,3 +31,4 @@ TEST_DEPENDS
   VTK::IOPLY
   VTK::IOXML
   VTK::TestingCore
+  VTK::IOHDF
```
mwestphal, line 35: so there is a small problem with this. VTK::IOHDF is an optional module, as in it may not be present in VTK and we want to keep it that way.

The solution is to use 

`TEST_OPTIONAL_DEPENDS` instead ot `TEST_DEPENDS`. But it also means the the code you added in TestF3DGenericImporter.cxx should be put in its own test ( TestF3DGenericImporterTimeSteps.cxx for examples ) and this test should be added only if this module in available, which can be done by adding a check in Testing/CMakeLists.txt like this:

```
if(TARGET VTK::IOHDF)
```

Let me know if you need more help!

mwestphal (changes requested):
one change about testing needed

vtkext/private/module/Testing/TestF3DGenericImporterTimeSteps.cxx
```diff
@@ -0,0 +1,43 @@
...
+
+    if (nbTimeSteps < 1 || nbTimeSteps != timeSteps->GetNumberOfTuples())
+    {
+      std::cerr << "Unexpected number of timesteps\n";
+      return EXIT_FAILURE;
+    }
+  }
+
+  return EXIT_SUCCESS;
+}
```
mwestphal: missing end of line

vtkext/private/module/vtk.module
```diff
@@ -31,3 +31,5 @@ TEST_DEPENDS
   VTK::IOPLY
   VTK::IOXML
   VTK::TestingCore
+TEST_OPTIONAL_DEPENDS
+  VTK::IOHDF
```
mwestphal: missing end of line

mwestphal (changes requested):
small changes, lets run full ci

mwestphal:
\ci full

vtkext/private/module/Testing/TestF3DGenericImporter.cxx
```diff
@@ -44,6 +44,7 @@ int TestF3DGenericImporter(int argc, char* argv[])
     int nbTimeSteps;
     double timeRange[2];
     vtkNew<vtkDoubleArray> timeSteps;
+
```
mwestphal: non necesary change

vtkext/private/module/Testing/TestF3DGenericImporterTimeSteps.cxx
```diff
@@ -0,0 +1,43 @@
+#include <vtkDoubleArray.h>
+#include <vtkHDFReader.h>
+#include <vtkNew.h>
+
+#include "vtkF3DGenericImporter.h"
+
```
mwestphal:
```suggestion
#include "vtkF3DGenericImporter.h"

#include <vtkDoubleArray.h>
#include <vtkHDFReader.h>
#include <vtkNew.h>
```

vtkext/private/module/Testing/TestF3DGenericImporterTimeSteps.cxx
```diff
@@ -0,0 +1,43 @@
...
+    bool temporalInfoExists =
+      importer->GetTemporalInformation(0, timeRange, nbTimeSteps, timeSteps);
+    if (!temporalInfoExists)
+    {
+      std::cerr << "Unexpected return value with GetTemporalInformation\n";
+      return EXIT_FAILURE;
+    }
+
+    if (nbTimeSteps < 1 || nbTimeSteps != timeSteps->GetNumberOfTuples())
```
mwestphal: You should check the exact number of timesteps recovered, right now this test is not testing much. 

Ideally also test the range of `timesSteps` is indeed the same as timeRange.

mwestphal (changes requested):
small changes needed

vtkext/private/module/Testing/TestF3DGenericImporterTimeSteps.cxx
```diff
@@ -0,0 +1,42 @@
+#include "vtkF3DGenericImporter.h"
+
+#include <vtkDoubleArray.h>
+#include <vtkHDFReader.h>
+#include <vtkNew.h>
+#include <iostream>
```
mwestphal, line 9:
```suggestion
#include <vtkNew.h>

#include <iostream>
```

mwestphal (approved):

vtkext/private/module/Testing/TestF3DGenericImporterTimeSteps.cxx
```diff
@@ -0,0 +1,57 @@
...
+#include <vtkNew.h>
+
+#include <iostream>
+
+int TestF3DGenericImporterTimeSteps(int argc, char* argv[])
+{
+  // Test time step temporal information
+  {
+    vtkNew<vtkHDFReader> reader;
+    std::string filename = std::string(argv[1]) + "data/blob.vtkhdf";
```
lknknm: I would suggest instead:

```suggestion
    const std::string filename = std::format("{}data/blob.vtkhdf", argv[1]);
```
This way you ensure `const` correctness and the `std::format` guarantees a single allocation is done, instead of allocating 2 buffers and copying the contents in order to concatenate the second into the first. It is also a good hint that in the future we can use std::format with `constexpr std::string` capabilities, as it's the current direction C++20 up is taking.

For reference, please check:
- https://isocpp.org/wiki/faq/const-correctness
- https://en.cppreference.com/cpp/utility/format/format

vtkext/private/module/Testing/TestF3DGenericImporterTimeSteps.cxx
```diff
@@ -0,0 +1,57 @@
...
+    reader->SetFileName(filename.c_str());
+    reader->UpdateInformation();
+
+    vtkNew<vtkF3DGenericImporter> importer;
+    importer->SetInternalReader(reader);
+    importer->Update();
+    importer->Print(std::cout);
+
+    int nbTimeSteps;
+    double timeRange[2];
```
lknknm, line 26: Using an `std::array<T, n>` gives you .size(), participates in structured bindings, and is value-initialized with {} (zeroing the doubles), unlike the raw C-array which is default-initialized. Needless to say it's more standard compliant and can be easily passed to `std::span` if we ever need it.

```suggestion
    std::array<double, 2> timeRange{};
```

For reference, please check:
- https://en.cppreference.com/cpp/container/array
- https://en.cppreference.com/cpp/language/aggregate_initialization

vtkext/private/module/Testing/TestF3DGenericImporterTimeSteps.cxx
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    const double* timeStepsRange = timeSteps->GetRange();
+    if (timeStepsRange[0] != timeRange[0] || timeStepsRange[1] != timeRange[1])
+    {
+      std::cerr << "Unexpected time step range\n";
+      return EXIT_FAILURE;
+    }
+  }
+
```
lknknm: Here it's possible to use the `std::ranges::all_of`
As far as I could consult, the `vtk::DataArrayValueRange` exposes a raw iterator, so you can use it in order to return a conditional value from the `all_of` algorithm.

It's more C++ idiomatic, but I'd consult @mwestphal to understand if this is compliant to both F3D and vtk idioms. I might also point out it should be tested on MSVC, because the way it deals with [raw] iterators is slightly different from how Clang implements it.

I might also point out that I had to ask Claude to understand how to use it with vtk::DataArrayValueRange().

Anyway, for reference:
- https://en.cppreference.com/cpp/algorithm/ranges/all_any_none_of
- https://en.cppreference.com/cpp/ranges/range

```suggestion
    const auto values = vtk::DataArrayValueRange(timeSteps);
    const bool allInRange = std::ranges::all_of(values, [&](const double t)
    {
      return t >= timeRange[0] && t <= timeRange[1];
    });
    
    if (!allInRange)
    {
      std::cerr << "Unexpected time step in time range\n";
      return EXIT_FAILURE;
    }
```

vtkext/private/module/Testing/TestF3DGenericImporterTimeSteps.cxx
```diff
@@ -0,0 +1,57 @@
...
+    // 11 timesteps in blob.vtkhdf
+    if (nbTimeSteps != reader->GetNumberOfSteps() || nbTimeSteps != 11)
+    {
+      std::cerr << "Unexpected time steps recovered\n";
+      return EXIT_FAILURE;
+    }
+
+    const double* timeStepsRange = timeSteps->GetRange();
+    if (timeStepsRange[0] != timeRange[0] || timeStepsRange[1] != timeRange[1])
```
lknknm: I feel there are a bunch of scattered considitons that could be grouped together and then analysed afterwards.
```cpp
const bool validStepCount { 
  nbTimeSteps >= 1
  && nbTimeSteps == timeSteps->GetNumberOfTuples()
  && nbTimeSteps == reader->GetNumberOfSteps()
  && nbTimeSteps == 11
};
```

This could be implemented to the detriment of having specific indifivual conditions, so it's up to you about which one is better for debuggability and overall readability.

vtkext/private/module/vtkF3DGenericImporter.cxx
```diff
@@ -77,6 +79,20 @@ void vtkF3DGenericImporter::UpdateTemporalInformation()
...
+    const double* readerTimeSteps = readerInfo->Get(vtkStreamingDemandDrivenPipeline::TIME_STEPS());
+    const int readerNumberOfTimeSteps =
+      readerInfo->Length(vtkStreamingDemandDrivenPipeline::TIME_STEPS());
+
+    this->Pimpl->TimeSteps->SetNumberOfTuples(readerNumberOfTimeSteps);
+
+    for (vtkIdType idx = 0; idx < readerNumberOfTimeSteps; idx++)
+    {
+      this->Pimpl->TimeSteps->SetValue(idx, readerTimeSteps[idx]);
+    }
```
lknknm, line 94: I believe here it's possible to use `std::ranges::copy`, but I'm not sure:
```cpp
  std::ranges::copy(
    std::span<const double>(readerTimeSteps, readerNumberOfTimeSteps),
    this->Pimpl->TimeSteps->GetPointer(0) // is this possible to get this as a `begin` in the range?
  );
```

The `std::ranges::copy` algorithm is [guaranteed to be N assignments complexity by the standard](https://eel.is/c++draft/alg.copy#5) and it's often implemented as a single memmove by libstdc++ and libc++ as an optimization. This might potentially eliminate the `SetValue()` indirection, although I'm not 100% sure.

lknknm (commented):
Comments done on my side!
I kindly ask @mwestphal to double check because I'm not 100% sure on the bigger picture for the whole PR. 
Thank you for your contribution!

vtkext/private/module/vtkF3DGenericImporter.cxx
mwestphal: using GetPointer() is not a best idea in the context of VTK, it can least to weirness. Since this code is not specifically performance oriented (number of timesteps is not big), i think current impl is fine.

vtkext/private/module/Testing/TestF3DGenericImporterTimeSteps.cxx
mwestphal: your suggestion looks fine. Not specially critical in the context of a test but if it works, lets do it.

vtkext/private/module/vtkF3DGenericImporter.cxx
lknknm: It makes sense, thanks!

vtkext/private/module/Testing/TestF3DGenericImporterTimeSteps.cxx
CPerry26: Good catch, I'm going to go with your suggestion. It's a pretty simple test, so if it were to fail I don't think it would be too hard to figure out why.

vtkext/private/module/Testing/TestF3DGenericImporterTimeSteps.cxx
CPerry26: I have no problem with this change, it makes sense to me. @mwestphal `GetTemporalInformation` expects `double * timeRange`. Is it safe to do something like the following here?
`std::array<double, 2> timeRange{};
...GetTemporalInformation(0, timeRange.data(), nbTimeSteps, timeSteps);
`

vtkext/private/module/Testing/TestF3DGenericImporterTimeSteps.cxx
CPerry26: Working for me, had a quick question on Discord since I only have access to Clang

vtkext/private/module/Testing/TestF3DGenericImporterTimeSteps.cxx
mwestphal: yes, that works and is indeed more modern.

mwestphal (approved):

mwestphal (approved):

lknknm (approved):
LGTM, thanks for your contribution!

PR #3195, opened 2026-05-28
Support for OpenUSD static library

Meakk:
### Describe your changes

- Add support for static library by linking with `WHOLE_ARCHIVE`.
- Avoid process exit when OpenUSD reports a fatal error (which happens when `plugInfo.json` is not found).

### AI Disclosure

- [x] I did not use AI to generate any of the content of that pull request
- [ ] I used AI to generate code in that pull request, if yes [please disclose](https://f3d.app/dev/AI_POLICY) which part of the code was generated and with which model.
- ...

### Continuous integration

### DIFF
--- a/plugins/usd/CMakeLists.txt
+++ b/plugins/usd/CMakeLists.txt
@@ -29,15 +29,11 @@ f3d_plugin_declare_reader(
   FORMAT_DESCRIPTION "Universal Scene Descriptor"
 )
 
-set(rpaths "")
-list(GET PXR_LIBRARIES 0 first_lib)
-get_target_property(target_type ${first_lib} TYPE)
-if(target_type STREQUAL SHARED_LIBRARY)
-  list(APPEND rpaths "$<TARGET_FILE_DIR:${first_lib}>")
-else()
-  # When built as a static library, OpenUSD has to be linked with the WHOLE_ARCHIVE link option.
-  # It's not done currently, so just exit with a fatal error.
-  message(FATAL_ERROR "OpenUSD seems to have been built as static libraries. This is currently not supported.")
+set(_additional_rpath "")
+list(GET PXR_LIBRARIES 0 _first_lib)
+get_target_property(_target_type ${_first_lib} TYPE)
+if(_target_type STREQUAL SHARED_LIBRARY)
+  list(APPEND _additional_rpath "$<TARGET_FILE_DIR:${_first_lib}>")
 endif ()
 
 f3d_plugin_build(
@@ -45,7 +41,7 @@ f3d_plugin_build(
   VERSION 1.0
   DESCRIPTION "USD support"
   VTK_MODULES IOImage
-  ADDITIONAL_RPATHS ${rpaths}
+  ADDITIONAL_RPATHS ${_additional_rpath}
   MIMETYPE_XML_FILES "${CMAKE_CURRENT_SOURCE_DIR}/f3d-usd-formats.xml"
   CONFIGURATION_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/configs/config.d" "${CMAKE_CURRENT_SOURCE_DIR}/configs/thumbnail.d"
   FREEDESKTOP
--- a/plugins/usd/module/CMakeLists.txt
+++ b/plugins/usd/module/CMakeLists.txt
@@ -6,7 +6,13 @@ vtk_module_add_module(f3d::vtkextUSD
   FORCE_STATIC
   CLASSES ${classes})
 
-vtk_module_link(f3d::vtkextUSD PRIVATE ${PXR_LIBRARIES})
+list(GET PXR_LIBRARIES 0 _first_lib)
+get_target_property(_target_type ${_first_lib} TYPE)
+if(_target_type STREQUAL "STATIC_LIBRARY")
+  vtk_module_link(f3d::vtkextUSD PRIVATE $<LINK_LIBRARY:WHOLE_ARCHIVE,${PXR_LIBRARIES}>)
+else()
+  vtk_module_link(f3d::vtkextUSD PRIVATE ${PXR_LIBRARIES})
+endif()
 
 vtk_module_include(f3d::vtkextUSD PRIVATE ${PXR_INCLUDE_DIRS})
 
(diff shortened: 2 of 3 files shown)


plugins/usd/module/CMakeLists.txt
```diff
@@ -6,7 +6,13 @@ vtk_module_add_module(f3d::vtkextUSD
   FORCE_STATIC
   CLASSES ${classes})
 
-vtk_module_link(f3d::vtkextUSD PRIVATE ${PXR_LIBRARIES})
+list(GET PXR_LIBRARIES 0 _first_lib)
+get_target_property(_target_type ${_first_lib} TYPE)
+if(_target_type STREQUAL "STATIC_LIBRARY")
+  vtk_module_link(f3d::vtkextUSD PRIVATE $<LINK_LIBRARY:WHOLE_ARCHIVE,${PXR_LIBRARIES}>)
```
mwestphal, line 12: it works, but its huge, can you use the same technic as pdal ?

Meakk: I'll try but if it's not trivial I'll open an issue and leave it for future work.

mwestphal: ok, please note it means static build for openusd in superbuild for binary is not on the table for now.

Meakk: symbols are versioned (e.g. `_ZN35pxrInternal_v0_25_11__pxrReserved__L16_Tf_RegistryAdd0Ev`) and couldn't make it work when picking related symbols.

Meakk: > ok, please note it means static build for openusd in superbuild for binary is not on the table for now.

Why though? Linking with whole archive is not worst than shipping the shared lib.

mwestphal: Indeed, you are right.

plugins/usd/module/CMakeLists.txt
```diff
@@ -6,7 +6,13 @@ vtk_module_add_module(f3d::vtkextUSD
   FORCE_STATIC
   CLASSES ${classes})
 
-vtk_module_link(f3d::vtkextUSD PRIVATE ${PXR_LIBRARIES})
+list(GET PXR_LIBRARIES 0 _first_lib)
```
mwestphal, line 9: can this contain multiple libs ?

plugins/usd/module/vtkF3DUSDImporter.cxx
```diff
@@ -1363,7 +1363,9 @@ class vtkF3DUSDImporter::vtkInternals
 
     void IssueFatalError(const pxr::TfCallContext&, const std::string& msg) override
     {
-      vtkErrorWithObjectMacro(this->Parent, << msg);
+      // if we do not throw here, OpenUSD just exit(1) the process
+      // we catch this exception upstream and only report an importer failure
+      throw std::runtime_error(msg);
```
mwestphal, line 1368: dedicated exceptions would be better?

plugins/usd/CMakeLists.txt
```diff
@@ -29,23 +29,19 @@ f3d_plugin_declare_reader(
...
-list(GET PXR_LIBRARIES 0 first_lib)
-get_target_property(target_type ${first_lib} TYPE)
-if(target_type STREQUAL SHARED_LIBRARY)
-  list(APPEND rpaths "$<TARGET_FILE_DIR:${first_lib}>")
-else()
-  # When built as a static library, OpenUSD has to be linked with the WHOLE_ARCHIVE link option.
-  # It's not done currently, so just exit with a fatal error.
-  message(FATAL_ERROR "OpenUSD seems to have been built as static libraries. This is currently not supported.")
+set(_additional_rpath "")
+list(GET PXR_LIBRARIES 0 _first_lib)
```
mwestphal, line 33: same question, if there is always one lib, maybe call it `_single_lib` ?

plugins/usd/module/CMakeLists.txt
Meakk: if not monolithic, yes

plugins/usd/module/vtkF3DUSDImporter.cxx
Meakk: Doesn't feel necessary to me. Keep in mind we are not in libf3d here.

plugins/usd/module/CMakeLists.txt
mwestphal: then this logic cannot work in that case ?

plugins/usd/module/vtkF3DUSDImporter.cxx
mwestphal: indeed, but this mean you would catch any exception coming from usd, not only this one. Ok to skip if you think its not needed.

plugins/usd/module/CMakeLists.txt
Meakk: It does. All usd libraries have the same type. So just pick the first one and check the type so it doesn't matter if it's monolithic or not.

plugins/usd/CMakeLists.txt
Meakk: answered on the other thread.

plugins/usd/module/vtkF3DUSDImporter.cxx
Meakk: Yeah, let's keep that simple for now. Even if we catch an unexpected exception, it doesn't hurt.

mwestphal: it may hide a genuine error, but fine by me

plugins/usd/module/CMakeLists.txt
mwestphal: right, my mistake.

plugins/usd/module/vtkF3DUSDImporter.cxx
Meakk: the error is reported and will be printed with `vtkErrorMacro` so visible when running with `--verbose`

mwestphal (approved):

Issue #2284, opened 2025-06-03
OSMesa has been removed


Meakk:
**Describe the bug**
It seems like OSMesa library has been removed from Mesa recent versions (starting from 25.1).  
See https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/33836

Let's decide what we should do with the osmesa backend:
- Continue supporting it as long as VTK has a OSMesa render window available
- Deprecate and remove `osmesa` backend in 4.0

In a comment, a dev suggest to use EGL with `EGL_MESA_device_software` instead.
We should also find a way to let the user select software rasterizer with EGL somehow.

**To Reproduce**

1.  `f3d example.glb --rendering-backend=osmesa`

**Expected behavior**
We should find a way to do software rendering on recent Mesa versions

**System Information:**

- OS: Linux with Mesa >= 25.1
- GPU and GPU driver:

**Additional context**
https://www.reddit.com/r/EndeavourOS/comments/1ksq2to/libosmesaso8_missing_since_mesa_update/

mwestphal:
not a F3D issue imo, we should keep the osmesa backend until we stop supporting a VTK version that supports it, so for quite some time.

Using EGL with mesa is as simple as installing mesa, nothing to do on F3D side.

Meakk:
If we decide to keep the backend, we should at least document/recommend to not use it, and we should have a way to enable software rendering in EGL somehow.

mwestphal:
> If we decide to keep the backend, we should at least document/recommend to not use it, and we should have a way to enable software rendering in EGL somehow.

Its not F3D nor VTK responsability to enable software rendering in EGL, it depends on the libegl being used.

mwestphal:
We also probably need to disable/remove TestSDKExternalWindowOSMesa

Meakk:
Posted a message on Mesa merge request: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/33836

Meakk:
No workaround for now. OSMesa is now used in the CI on all platforms. See https://github.com/f3d-app/f3d/pull/3226

mwestphal:
stalled is for PR, what do you mean by that for an issue ?

Meakk:
I meant there is nothing we can do about it now. I felt that `wont-fix` is too definitive, but feel free to change the labels.

mwestphal:
ok thats fair.

Issue #1670, opened 2024-10-22
External window EGL/QT/GLFW test are failing without X server 


Meakk:
Failure with QT is expected because it uses `xcb` platform by default.
Using an alternative platform like `eglfs` should work, but I couldn't make it work yet.

Failure with GLFW is expected too and it seems like it's possible to use EGL or OSMesa, but only if GLFW support for these platforms is enabled at compile-time.

Regarding EGL failure, it's passing locally on my side, so it needs more investigation.

**To Reproduce**
 1. Stops the X server
 2. Run tests `TestSDKExternalWindowGLFW`, TestSDKExternalWindowQT` and `TestSDKExternalWindowEGL`

mwestphal:
@Meakk do you have more insight about this issue ?

Meakk:
Yes EGL problem is fixed by #3226.
About QT I need to try `eglfs` again it should work.
GLFW wont' work except if we build it ourselves but I don't think we want that.

PR #3226, opened 2026-06-09
Add OSMesa dependency and many related fixes

### LINKED ISSUE #792: Cell coloring fails on macOS

### LINKED ISSUE #1276: CI: Grid related test fails on macos arm64 CI because of a grid artifact
**Describe the bug**
Tests involving the grid fails on macos arm64 CI.
It does not fail locally, this was tested on two different macos arm64 computer.

In order to still run macos arm64 CI, Grid tests have been excluded.

Failure example:

|Test|Result|Diff|
|--|--|--|
|TestConfigStemBuild|TestConfigStemBuild|TestConfigStemBuild diff|
|TestDXF|TestDXF|TestDXF diff|
|TestGridX|TestGridX|TestGridX diff|
|TestGridZ|TestGridZ|TestGridZ diff|

### LINKED ISSUE #1760: F3D Not creating  images in new windows servers
While trying to use f3d.exe it is not generating the images. It is not throwing any errors either.

There is an article in the f3d website to download three dll files. libgallium_wgl.dll is one of them. After doing it, it now records an entry into the event viewer in the windows server.

It is not very clear what is the problem and how to fix it, Any help is appreciated.

Faulting application name: f3d.exe, version: 0.0.0.0, time stamp: 0x66831ffd
Faulting module name: libgallium_wgl.dll, version: 24.0.9.0, time stamp: 0x666295d4
Exception code: 0xc0000005
Fault offset: 0x0000000000d6716a
Faulting process ID: 0x229c
Faulting application start time: 0x01db471c33429ad0
Faulting application path: C:\Program Files\F3D\bin\f3d.exe
Faulting module path: C:\Program Files\F3D\bin\libgallium_wgl.dll
Report ID: c16cac79-ca11-48de-809b-d360f5109638
Faulting package full name:
Faulting package-relative application ID:

### LINKED ISSUE #2884: `--rendering-backend` does not work as expected without dev packages
**Describe the bug**
According to debian policy: https://www.debian.org/doc/debian-policy/ch-sharedlibs.html#development-files
the `.so` version of the libraries should only provided by `-dev` packages.

This is indeed the case in many distributions such as debian, flatpak and alpine linux.

eg:
 - mesa-egl: https://pkgs.alpinelinux.org/package/edge/main/x86_64/mesa-egl
 - mesa-dev: https://pkgs.alpinelinux.org/package/edge/main/x86_64/mesa-dev

The side effects of this is that `--rendering-backend` does not work as expected because it looks only for `libEGL.so` or `libOSMesa.so`:

```
  gladLoaderLoadEGL(EGL_NO_DISPLAY);
  return getSymbol("EGL", "eglGetProcAddress");
```

which looks only for `libEGL.so` or `EGL.dll` on Windows.

Interestingly when not specifying the rendering backend, egl can be found without issue by the underlying glag/vtk layer.

And indeed, looking at glad and VTK we find:

egl:
```
static void* glad_egl_dlopen_handle(void) {
#if GLAD_PLATFORM_APPLE
    static const char *NAMES[] = {"libEGL.dylib"};
#elif GLAD_PLATFORM_WIN32
    static const char *NAMES[] = {"libEGL.dll", "EGL.dll"};
#else
    static const char *NAMES[] = {"libEGL.so.1", "libEGL.so"};
#endif

    if (_egl_handle == NULL) {
        _egl_handle = glad_get_dlopen_handle(NAMES, sizeof(NAMES) / sizeof(NAMES[0]));
    }    
```

gl:
```
#if GLAD_PLATFORM_APPLE
    static const char *NAMES[] = {
        "../Frameworks/OpenGL.framework/OpenGL",
        "/Library/Frameworks/OpenGL.framework/OpenGL",
        "/System/Library/Frameworks/OpenGL.framework/OpenGL",
        "/System/Library/Frameworks/OpenGL.framework/Versions/Current/OpenGL"
    };   
#elif GLAD_PLATFORM_WIN32
    static const char *NAMES[] = {"opengl32.dll"};
#else
    static const char *NAMES[] = {
  #if defined(__CYGWIN__)
        "libGL-1.so",
  #endif
        "libGL.so.1",
        "libGL.so"
    };   
#endif

    if (_glad_GL_loader_handle == NULL) {
        _glad_GL_loader_handle = glad_get_dlopen_handle(NAMES, sizeof(NAMES) / sizeof(NAMES[0]));
    }    

    return _glad_GL_loader_handle;
}
```

osmesa:

```
#if defined(_WIN32)
  OSMesaLibraryHandle = LoadLibraryA("osmesa.dll");
  if (OSMesaLibraryHandle == nullptr)
  {
    vtkGenericWarningMacro(
      << "osmesa.dll not found. It appears that OSMesa is not installed in "
         "your system. Please install the OSMesa library. You can obtain pre-built binaries for "
         "Windows from https://github.com/pal1000/mesa-dist-win. Ensure that osmesa.dll is "
         "available in PATH.");
  }
#elif defined(__linux__)
  const std::vector<std::string> libNamesToTry = { "libOSMesa.so.8", "libOSMesa.so.6",
    "libOSMesa.so" };
  for (const auto& libName : libNamesToTry)
  {
    OSMesaLibraryHandle = dlopen(libName.c_str(), RTLD_LAZY | RTLD_GLOBAL);
    if (OSMesaLibraryHandle)
    {   
      break;
    }   
  }
```

for glx its even weirder, but there is a dedicated glx.c in glad that takes care of it.

We should generlize our `context.cxx` implementation to look for variant of the libs at least

**To Reproduce**

1. Fresh install without `mesa-dev` on debian
2.  Open the file using `f3d --no-config example.glb --output=a.png --rendering-backend=egl`
3.  Fail to find libEGL.so.1

**Expected behavior**
EGL being used

**System Information:**

- OS: Linux (but the problem is generic)
- GPU and GPU driver: Any

**F3D Information**
Paste the content of `f3d --version` or `f3d-console.exe --version` on Windows: master

### LINKED ISSUE #3047: TestInteractionTAA* tests are flaky on macOS
**Describe the bug**
TestInteractionTAA* tests are flaky on macOS

**To Reproduce**

1.  Run CI without the exception
2.  See the occasional failure

Image

**Expected behavior**

Image

**System Information:**

- OS: macOS (intel and arm)
- GPU and GPU driver: Emulated Graphical in virtual machines

**F3D Information**
Paste the content of `f3d --version` or `f3d-console.exe --version` on Windows: any

**Additional context**

- https://github.com/f3d-app/f3d/actions/runs/24342727127/job/71075946037
- https://github.com/f3d-app/f3d/actions/runs/24385608217/job/71218846605?pr=3038

### LINKED ISSUE #3085: Create a CI action to build and cache mesa on all platforms for reliability
**Context**
F3D CI currently rely on different graphical stack on each platform:
 - Linux: xvfb
 - macOS: native
 - Windows: binary mesa

Lets unify them by building and caching a specific version of mesa in order to avoid rendering inconsistencies, especially on macOS.

**Suggested solution**
 - Add mesa as a classic CI dependency
 - Use it in our CI

Meakk:
### Describe your changes

- Build OSMesa with llvmpipe driver in the CI for all platforms
- Bump VTK to add support for OSMesa on macOS
- Enable OSMesa SDK test on all platforms
- Add `F3D_TESTING_ENABLE_WGL_TESTS` to disable WGL related tests (useful if remote rendering with no Mesa)
- Use `install-mesa` action only for regular CI, other jobs like Python or examples will use OSMesa on Windows
- `f3d::context::getSymbol` now expect a full library name (with prefix and extension)
- Use `LoadLibraryA` on Windows (it was working only with full path before)
- Fixed `f3d::context::egl()` and `f3d::context::osmesa()` to look for versioned libraries
- Automatic backend on Windows fallback on OSMesa if OpenGL 3.2 is not available
- Fixed `TestSDKExternalWindowOSMesa` to support macOS and Windows too
- Fixed `getDPIScale()` returning 0 on Windows with OSMesa
- Updated documentation to reflect the changes

- Fixes https://github.com/f3d-app/f3d/issues/2884
- Fixes https://github.com/f3d-app/f3d/issues/3085
- Fixes https://github.com/f3d-app/f3d/issues/1760
- Fixes https://github.com/f3d-app/f3d/issues/1276
- Fixes https://github.com/f3d-app/f3d/issues/792
- Fixes https://github.com/f3d-app/f3d/issues/3047
- Related: https://github.com/f3d-app/f3d/issues/2284
- Related: https://github.com/f3d-app/f3d/issues/1670

### AI Disclosure

- [ ] I did not use AI to generate any of the content of that pull request
- [x] I used AI to generate code in that pull request, if yes [please disclose](https://f3d.app/dev/AI_POLICY) which part of the code was generated and with which model.

I used VSCode Copilot AI autocompletion.

### Continuous integration


.github/actions/install-mesa/action.yml
mwestphal, line 1: please note you will probably need to put that in a dedicated repo.

Meakk: of course, just experimenting at the moment

library/public/context.h
```diff
@@ -66,10 +66,15 @@ class F3D_EXPORT context
...
 
   /**
-   * Create a context function from a library name and a function name.
-   * The library name must be specified without its prefix and extension.
-   * For example, `getSymbol("EGL", "eglGetProcAddress")` looks for the symbol
-   * `eglGetProcAddress` in the library `libEGL.so` on Linux.
+   * Add the appropriate prefix and extension to a library name for the current platform.
+   * For example, `addLibraryDecoration("EGL")` returns `libEGL.so` on Linux.
+   */
+  [[nodiscard]] static std::string addLibraryDecoration(std::string_view lib);
```
mwestphal: how about fixing this too: https://github.com/f3d-app/f3d/issues/2884 ?

.github/actions/generic-ci/action.yml
```diff
@@ -195,13 +195,14 @@ runs:
         -DF3D_PLUGIN_BUILD_VDB=${{ inputs.optdeps_label == 'optdeps' && 'ON' || 'OFF' }}
         -DF3D_PLUGIN_OCCT_COLORING_SUPPORT=${{ inputs.mindeps_label != 'mindeps' && 'ON' || 'OFF' }}
         -DF3D_STRICT_BUILD=ON
+        -DF3D_TESTING_ENABLE_WGL_TESTS=${{ runner.os == 'Windows' && 'ON' || 'OFF' }}
```
mwestphal, line 192: was that forgotten before ?

.github/actions/generic-ci/action.yml
```diff
@@ -220,25 +221,14 @@ runs:
       shell: bash
       run: Xvfb $DISPLAY -screen 0 1280x1024x24 &
 
+    # We install a proper WGL Mesa driver for regular tests
+    # This is needed because F3D_TESTING_ENABLE_WGL_TESTS options is enabled
```
mwestphal, line 219: not following why we need both the binary version and compiled version

.github/actions/generic-ci/action.yml
Meakk: it's a new option

.github/actions/generic-ci/action.yml
```diff
@@ -220,25 +221,14 @@ runs:
...
       if: runner.os == 'Windows'
       uses: f3d-app/install-mesa-windows-action@v2
       with:
         path: ${{github.workspace}}\build\bin_Release
 
-    - name: Install Mesa Windows Python
-      if: runner.os == 'Windows'
-      uses: f3d-app/install-mesa-windows-action@v2
-      with:
-        path: ${{github.workspace}}\build\Release\f3d
```
mwestphal, line 233: how do you force the runner to use the compiled mesa in install instead of the faulty opengl in C:/Windows ?

.github/actions/mesa-install-dep/action.yml
```diff
@@ -0,0 +1,141 @@
...
+    - name: Remove Git link.exe (Windows)
+      if: steps.cache-mesa.outputs.cache-hit != 'true' && runner.os == 'Windows'
+      shell: bash
+      run: rm $(which link.exe)
+
+    - name: Set up Python
+      if: steps.cache-mesa.outputs.cache-hit != 'true'
+      uses: actions/setup-python@v6
+      with:
+        python-version: "3.14"
```
mwestphal: I do not like that, we already install a specific version of python in our CI, reuse it ?

Why does mesa needs python anyway ?

.github/actions/mesa-install-dep/action.yml
```diff
@@ -0,0 +1,141 @@
...
+      if: steps.cache-mesa.outputs.cache-hit != 'true'
+      shell: bash
+      run: python -m pip install meson mako packaging PyYAML
+
+    - name: Install Flex/Bison (Windows)
+      if: steps.cache-mesa.outputs.cache-hit != 'true' && runner.os == 'Windows'
+      shell: bash
+      run: choco install winflexbison3 --no-progress -y
+
+    - name: Install Flex/Bison (Linux)
```
mwestphal, line 70: what about macos ?

.github/actions/mesa-install-dep/action.yml
```diff
@@ -0,0 +1,141 @@
...
+      if: steps.cache-mesa.outputs.cache-hit != 'true' && runner.os == 'Linux'
+      shell: bash
+      run: apt-get install -y bison byacc flex pkg-config ninja-build libzstd-dev
+
+    - name: Checkout mesa
```
mwestphal, line 75: why not using checkout action ?

.github/actions/mesa-install-dep/action.yml
```diff
@@ -0,0 +1,141 @@
...
+
+    - name: Setup mesa
+      if: steps.cache-mesa.outputs.cache-hit != 'true'
+      working-directory: ${{github.workspace}}/dependencies
+      shell: bash
+      run: |
+        mkdir mesa_build
+        mkdir mesa_install
+
+    - name: Add LLVM to PATH to find llvm-config
```
mwestphal, line 88: no better way to do that ?

library/public/engine.h
```diff
@@ -101,7 +101,8 @@ class F3D_EXPORT engine
 
   /**
    * Create an engine with an offscreen OSMesa window.
-   * VTK >= 9.4 required.
+   * On Windows and Linux, VTK >= 9.4 required.
+   * On macOS, VTK >= 9.6.20260630 is required.
```
mwestphal, line 102: you need the yet to be merged fix, no ?

library/src/context.cxx
```diff
@@ -18,11 +18,12 @@ namespace f3d
 //----------------------------------------------------------------------------
 context::function context::getSymbol(std::string_view lib, std::string_view func)
 {
-  std::string libName = vtksys::DynamicLoader::LibPrefix();
-  libName += lib;
-  libName += vtksys::DynamicLoader::LibExtension();
-
-  vtksys::DynamicLoader::LibraryHandle handle = vtksys::DynamicLoader::OpenLibrary(libName);
+#ifdef _WIN32
+  // on Windows vtksys::DynamicLoader::OpenLibrary behaves differently (it expects a full path)
```
mwestphal, line 22: I think this should be a VTK issue

library/src/context.cxx
```diff
@@ -91,7 +92,22 @@ context::function context::egl()
 {
 #if defined(VTK_OPENGL_HAS_EGL)
   gladLoaderLoadEGL(EGL_NO_DISPLAY);
-  return getSymbol("EGL", "eglGetProcAddress");
+
+#ifdef _WIN32
+  for (const auto& lib : { "libEGL.dll", "EGL.dll" })
+#else
+  for (const auto& lib : { "libEGL.so.1", "libEGL.so" })
```
mwestphal, line 99: what about macos , should we throw ?

mwestphal (changes requested):

.github/actions/generic-ci/action.yml
Meakk: What we install here is the binary version of Mesa with WGL surface. It's basically `opengl32.dll` with Mesa.  
What I compile is only `osmesa.dll` which is very different (even if both have their own llvmpipe driver internally)

.github/actions/generic-ci/action.yml
Meakk: I'm not forcing anything here. We fallback to OSMesa automatically because we detect that WGL (opengl32.dll) does not provide OpenGL 3.2.

.github/actions/mesa-install-dep/action.yml
Meakk: Mesa uses Python for compilation. The installation is local to the Mesa build and do not bleed to the main action.

.github/actions/mesa-install-dep/action.yml
Meakk: the runner already ships it

.github/actions/mesa-install-dep/action.yml
Meakk: because it's not on github

.github/actions/mesa-install-dep/action.yml
Meakk: I couldn't find a way to tell `meson` where to find llvm-config. Maybe it's possible I'll double check but I already spend some time on this.

library/public/engine.h
Meakk: I'm currently NOT using OSMesa in the CI for macOS yet. I'll take care of it tomorrow, and I will need the fix indeed. When it's done I should be able to remove all exclusions and bye bye flaky tests! (it will be much faster too)

library/src/context.cxx
Meakk: read about it here: https://gitlab.kitware.com/vtk/vtk/-/merge_requests/13382#note_1833495

library/src/context.cxx
Meakk: that's already the case since `VTK_OPENGL_HAS_EGL` is always false on macos.

.github/actions/generic-ci/action.yml
mwestphal: got it.

.github/actions/generic-ci/action.yml
mwestphal: got it

.github/actions/mesa-install-dep/action.yml
mwestphal: still, there is a version string that will get outdated.
Is it possible to use the version we install for everyone ?
if not,  can we at least use the same version ?

.github/actions/mesa-install-dep/action.yml
mwestphal: as long as you checked, fine by me

.github/actions/mesa-install-dep/action.yml
mwestphal, line 1: both this and llvm should be moved to their own repository, because we need them in superbuild too i think

mwestphal (changes requested):
Raised an issue about reusing this for the superbuild.

.github/actions/mesa-install-dep/action.yml
Meakk: Both llvm and mesa are already part of common-superbuild.
I also think we should include osmesa library to the packages to guarantee offscreen capabilities everywhere.

mwestphal: ok, we will study this in the future.

.github/actions/mesa-install-dep/action.yml
Meakk: The alternative is to create a `native-config.ini` containing:
```ini
[binaries]
llvm-config = '$(pwd)/install/bin/llvm-config'
```
And configure meson with `--native-file=native-config.ini`.  
It feels like the PATH approach is better, wdyt?

.github/actions/mesa-install-dep/action.yml
Meakk: Sure we can forward the python version to the action input. I'll do that.

.github/actions/mesa-install-dep/action.yml
mwestphal: yeah thats fine

library/testing/TestSDKExternalWindowEGL.cxx
```diff
@@ -36,6 +36,13 @@ int TestSDKExternalWindowEGL([[maybe_unused]] int argc, [[maybe_unused]] char* a
   EGLContext eglCtx = eglCreateContext(eglDpy, eglCfg, EGL_NO_CONTEXT, nullptr);
   eglMakeCurrent(eglDpy, eglSurf, eglSurf, eglCtx);
 
+  f3d::log::setVerboseLevel(f3d::log::VerboseLevel::DEBUG);
+
+  std::cout << "eglGetProcAddress: " << (void*)eglGetProcAddress << '\n';
+  std::cout << "egl(): " << (void*)f3d::context::egl()("eglGetProcAddress") << '\n';
+  std::cout << "glGetString: " << (void*)eglGetProcAddress("glGetString") << '\n';
+  std::cout << "glGetString 2: " << (void*)f3d::context::egl()("glGetString") << '\n';
+
```
Meakk: To remove when EGL crash is solved.

library/testing/TestSDKHelpers.h
```diff
@@ -10,6 +11,20 @@
 
 namespace TestSDKHelpers
 {
+inline f3d::engine CreateOffscreenEngine(const std::string& renderingBackend)
+{
+  if (renderingBackend == "osmesa")
+  {
+    return f3d::engine::createOSMesa();
```
mwestphal, line 18: reachable with a dedicated test ?

library/testing/TestSDKHelpers.h
```diff
@@ -10,6 +11,20 @@
...
+  }
+
+  if (renderingBackend == "egl")
+  {
+    return f3d::engine::createEGL();
```
mwestphal, line 23: reachable with a dedicated test ?

library/src/context.cxx
```diff
@@ -91,7 +92,22 @@ context::function context::egl()
...
+  {
+    try
+    {
+      return getSymbol(lib, "eglGetProcAddress");
+    }
+    catch (const loading_exception&)
+    {
+    }
+  }
+  throw loading_exception("Cannot find EGL library");
```
mwestphal, line 110:
```suggestion
  // Unreachable in CI
  throw loading_exception("Cannot find EGL library");
```

library/src/context.cxx
```diff
@@ -100,10 +116,24 @@ context::function context::egl()
...
+  {
+    try
+    {
+      return getSymbol(lib, "OSMesaGetProcAddress");
+    }
+    catch (const loading_exception&)
+    {
+    }
+  }
+  throw loading_exception("Cannot find OSMesa library");
```
mwestphal, line 137:
```suggestion
  // Unreachable in CI
  throw loading_exception("Cannot find OSMesa library");
```

.github/actions/generic-ci/action.yml
```diff
@@ -220,52 +221,37 @@ runs:
...
       shell: bash
       run: echo "F3D_CTEST_EXCEPTIONS=(test_scene_zero_copy)" >> $GITHUB_ENV
 
-    # Certain tests are failing on macOS with GPU pass through
-    # https://github.com/f3d-app/f3d/issues/792
-    # https://github.com/f3d-app/f3d/issues/1534
-    # https://github.com/f3d-app/f3d/issues/2268
-    # https://github.com/f3d-app/f3d/issues/3047
+    # Certain tests are failing on macOS software rendering
+    # OSMesa is used for VTK > 9.6, remove when it's dropped
```
mwestphal, line 233: so check VTK version too ?

.github/actions/generic-ci/action.yml
```diff
@@ -220,52 +221,37 @@ runs:
...
       shell: bash
       run: echo "F3D_CTEST_EXCEPTIONS=(TestDXF)|(TestPipedDXF)|(TestScalarsCell)|(TestXCAFColors)|(TestInteractionAnimationSlow)|(TestInteractionFocalPointPickingPoints)|(TestInteractionCycleAnimation)|(TestInteractionCycleComp)|(TestGLTFRigArmature)|(TestInteractionNoModelScrollWheel)|(TestInteractionNoModelScrollBar)|(TestInteractionTAA)|(TestInteractionPointCloudTAA)|(TestInteractionStochasticTAA)" >> $GITHUB_ENV
 
-    # Certain tests are failing on macOS arm64 for unknown reasons
-    # https://github.com/f3d-app/f3d/issues/1276
-    # https://github.com/f3d-app/f3d/issues/792
-    # https://github.com/f3d-app/f3d/issues/1534
-    # https://github.com/f3d-app/f3d/issues/3047
+    # Certain tests are failing on macOS software rendering
+    # OSMesa is used for VTK > 9.6, remove when it's dropped
```
mwestphal, line 243: so check VTK version too ?

.github/actions/mesa-install-dep/action.yml
```diff
@@ -0,0 +1,149 @@
...
+
+    # https://github.com/ilammy/msvc-dev-cmd#caveats
+    - name: Remove Git link.exe (Windows)
+      if: steps.cache-mesa.outputs.cache-hit != 'true' && runner.os == 'Windows'
+      shell: bash
+      run: rm $(which link.exe)
+
+    - name: Set up Python
+      if: steps.cache-mesa.outputs.cache-hit != 'true'
+      uses: actions/setup-python@v6
```
mwestphal: how do you make sure this python install is not at the same location as the other python install ?

.github/workflows/ci.yml
```diff
@@ -271,6 +271,7 @@ jobs:
           vtk_version: ${{matrix.vtk_version}}
           cpu: x86_64
           lfs_sha: ${{ needs.cache_lfs.outputs.lfs_sha}}
+          rendering_backend: osmesa
```
mwestphal: I'm not so sure about that, why do you need it ? it should use it automatically, no ?

.github/workflows/ci.yml
```diff
@@ -315,6 +316,7 @@ jobs:
           vtk_version: ${{matrix.vtk_version}}
           bundle_label: ${{matrix.bundle_label}}
           static_label: ${{matrix.static_label}}
+          rendering_backend: ${{matrix.vtk_version == 'commit' && 'osmesa' || 'auto'}}
```
mwestphal, line 323: why a different logic between intel and arm ?

.github/workflows/versions.json
```diff
@@ -47,6 +47,14 @@
     "version": "26",
     "min_version": "17"
   },
+  "llvm": {
+    "version": "22.1.8",
+    "min_version": "22.1.8"
+  },
+  "mesa": {
```
mwestphal, line 55: needs a comment explaining that we cant update it

.github/workflows/versions.json
```diff
@@ -47,6 +47,14 @@
...
+    "version": "24.3.4",
```
mwestphal: why not the latest containing osmesa ?

application/testing/tests.custom.cmake
```diff
@@ -31,3 +31,14 @@ if (APPLE)
...
+# Test EGL/OSMesa load failure
+if (UNIX AND NOT APPLE)
+  add_test(NAME f3d::TestEGLLoadFailure COMMAND $<TARGET_FILE:f3d> --output=${CMAKE_BINARY_DIR}/Testing/Temporary/egl.png --rendering-backend=egl --verbose)
+  set_tests_properties(f3d::TestEGLLoadFailure PROPERTIES PASS_REGULAR_EXPRESSION "Cannot find EGL library")
+  set_tests_properties(f3d::TestEGLLoadFailure PROPERTIES ENVIRONMENT "LD_LIBRARY_PATH=${F3D_SOURCE_DIR}/testing/data")
+
+  add_test(NAME f3d::TestOSMesaLoadFailure COMMAND $<TARGET_FILE:f3d> --output=${CMAKE_BINARY_DIR}/Testing/Temporary/osmesa.png --rendering-backend=osmesa --verbose)
+  set_tests_properties(f3d::TestOSMesaLoadFailure PROPERTIES PASS_REGULAR_EXPRESSION "Cannot find OSMesa library")
+  set_tests_properties(f3d::TestOSMesaLoadFailure PROPERTIES ENVIRONMENT "LD_LIBRARY_PATH=${F3D_SOURCE_DIR}/testing/data")
+endif ()
```
mwestphal: missing eol

application/testing/tests.custom.cmake
```diff
@@ -31,3 +31,14 @@ if (APPLE)
   add_test(NAME f3d::TestAppleCmdMod COMMAND $<TARGET_FILE:f3d> --output=${CMAKE_BINARY_DIR}/Testing/Temporary/TestAppleCmdMod.png --reference=${F3D_SOURCE_DIR}/testing/baselines/TestAppleCmdMod.png --resolution=400,400)
   set_tests_properties(f3d::TestAppleCmdMod PROPERTIES ENVIRONMENT "CTEST_F3D_NO_DATA_FORCE_RENDER=1")
 endif ()
+
+# Test EGL/OSMesa load failure
+if (UNIX AND NOT APPLE)
+  add_test(NAME f3d::TestEGLLoadFailure COMMAND $<TARGET_FILE:f3d> --output=${CMAKE_BINARY_DIR}/Testing/Temporary/egl.png --rendering-backend=egl --verbose)
+  set_tests_properties(f3d::TestEGLLoadFailure PROPERTIES PASS_REGULAR_EXPRESSION "Cannot find EGL library")
+  set_tests_properties(f3d::TestEGLLoadFailure PROPERTIES ENVIRONMENT "LD_LIBRARY_PATH=${F3D_SOURCE_DIR}/testing/data")
+
```
mwestphal, line 40: this test will fail is egl is in the system, does it not ?

application/testing/tests.custom.cmake
```diff
@@ -31,3 +31,14 @@ if (APPLE)
...
+  add_test(NAME f3d::TestOSMesaLoadFailure COMMAND $<TARGET_FILE:f3d> --output=${CMAKE_BINARY_DIR}/Testing/Temporary/osmesa.png --rendering-backend=osmesa --verbose)
+  set_tests_properties(f3d::TestOSMesaLoadFailure PROPERTIES PASS_REGULAR_EXPRESSION "Cannot find OSMesa library")
+  set_tests_properties(f3d::TestOSMesaLoadFailure PROPERTIES ENVIRONMENT "LD_LIBRARY_PATH=${F3D_SOURCE_DIR}/testing/data")
```
mwestphal, line 43: same ?

application/testing/tests.features.cmake
```diff
@@ -320,7 +320,7 @@ if(VTK_VERSION VERSION_GREATER_EQUAL 9.5.20251001)
 endif()
 
 ## Skinning
-if(APPLE) # MacOS does not support OpenGL 4.3
+if(APPLE AND NOT F3D_TESTING_FORCE_RENDERING_BACKEND STREQUAL "osmesa") # MacOS driver does not support OpenGL 4.3
```
mwestphal, line 323: why not osmesa ?

doc/libf3d/06-MIGRATION.md
```diff
@@ -51,3 +51,8 @@ The `jump_to_frame` and `jump_to_keyframe` commands no longer take a second bool
 ## ui.animation_progress
 
 The `ui.animation_progress` option (CLI `--animation-progress`) was a boolean toggling a progress bar during animation playback. It is now a string selecting the progress bar mode: `none` (hidden), `default` (the progress bar alone) or `advanced` (the progress bar with time range, animation name, current time labels and keyframe markers). Replace `true` with `default` (or `advanced`) and `false` with `none`.
+
+## Context symbol
+
+The function `f3d::context::getSymbol` function is expecting a library full path or a filename.
```
mwestphal:
```suggestion
The function `f3d::context::getSymbol` function is now expecting a library full path or a filename.
```

library/public/engine.h
```diff
@@ -41,8 +41,8 @@ class F3D_EXPORT engine
   /**
    * Create an engine with an automatic window.
    * Optionally, the window can be hidden by setting offscreen to true.
-   * Linux: Try GLX, then EGL, then OSMesa.
-   * Windows: Try Win32, then EGL, then OSMesa.
+   * Linux: GLX if a X server is running, otherwise EGL if available, otherwise OSMesa.
+   * Windows: Win32 if it supports OpenGL >= 3.2, otherwise EGL if available, otherwise OSMesa.
    * macOS: Always use Cocoa.
```
mwestphal, line 46: does it ? :)

library/src/window_impl.cxx
```diff
@@ -72,7 +72,17 @@ class window_impl::internals
   {
     // Override VTK logic
 #ifdef _WIN32
-    return vtkSmartPointer<vtkF3DWGLRenderWindow>::New();
+    vtkSmartPointer<vtkOpenGLRenderWindow> wglRenWin =
+      vtkSmartPointer<vtkF3DWGLRenderWindow>::New();
+    if (!wglRenWin->SupportsOpenGL())
```
mwestphal, line 77: should we check that on all platforms ?

library/testing/CMakeLists.txt
```diff
@@ -163,7 +166,7 @@ list(APPEND libf3dSDKTestsNoRender_list
 # Add all the ADD_TEST for each test
 foreach (test ${libf3dSDKTests_list})
   get_filename_component (TName ${test} NAME_WE)
-  add_test (NAME libf3d::${TName} COMMAND libf3dSDKTests ${TName}  "${F3D_SOURCE_DIR}/testing/" "${CMAKE_BINARY_DIR}/Testing/Temporary/" "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
+  add_test (NAME libf3d::${TName} COMMAND libf3dSDKTests ${TName} "${F3D_SOURCE_DIR}/testing/" "${CMAKE_BINARY_DIR}/Testing/Temporary/" "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}" "${F3D_TESTING_FORCE_RENDERING_BACKEND}")
```
mwestphal, line 171: maybe im missing something but i dont see where this is used

library/testing/TestTestSDKHelpers.cxx
```diff
@@ -12,6 +12,18 @@ int TestTestSDKHelpers([[maybe_unused]] int argc, [[maybe_unused]] char* argv[])
 {
   PseudoUnitTest test;
 
+  // Coverage of CreateOffscreenEngine
+  for (auto backend : { "osmesa", "egl", "auto" })
```
mwestphal, line 16: this doesnt seem to work ?

application/testing/tests.custom.cmake
```diff
@@ -31,3 +31,14 @@ if (APPLE)
   add_test(NAME f3d::TestAppleCmdMod COMMAND $<TARGET_FILE:f3d> --output=${CMAKE_BINARY_DIR}/Testing/Temporary/TestAppleCmdMod.png --reference=${F3D_SOURCE_DIR}/testing/baselines/TestAppleCmdMod.png --resolution=400,400)
   set_tests_properties(f3d::TestAppleCmdMod PROPERTIES ENVIRONMENT "CTEST_F3D_NO_DATA_FORCE_RENDER=1")
 endif ()
+
+# Test EGL/OSMesa load failure
```
mwestphal: add some comment explaining that testing/data contains invalid libEGL.so and so on

mwestphal (changes requested):
questions and changes

.github/workflows/ci.yml
Meakk: Intel CI only runs `commit` with `ci: main` so I thought it wasn't necessary. I'll fix it.

.github/workflows/ci.yml
Meakk: It's NOT automatically on macOS because Apple's driver is bugged but it manages to generate something and is compatible with OpenGL 3.2 on paper, as opposed to Microsoft's remote driver.

.github/workflows/versions.json
Meakk: I'll try the last version recommended by the Mesa maintainer. I'll leave the minimum version as it is, I'm not sure it's worth investigating what is the oldest one compatible.

application/testing/tests.features.cmake
Meakk: Because `osmesa` supports OpenGL 4.3 (needed here to support many bones)

library/public/engine.h
Meakk: It does :)

library/src/window_impl.cxx
Meakk: I don't think it's needed, all other systems are supporting OpenGL 3.2 for decades.

library/testing/CMakeLists.txt
Meakk: It's used in `TestSDKNotification`

library/testing/TestTestSDKHelpers.cxx
Meakk: wdym?

library/testing/TestSDKHelpers.h
Meakk: tested in `TestTestSDKHelpers`

library/src/context.cxx
Meakk: tested by `TestEGLLoadFailure`

library/src/context.cxx
Meakk: tested by `TestOSMesaLoadFailure`

.github/actions/mesa-install-dep/action.yml
Meakk: Let me check closer but I think Python installed here is uninstalled automatically at the end of the action.

.github/workflows/ci.yml
Meakk: you are right this was needed.

library/public/engine.h
mwestphal: Well, you added osmesa, didnt you ?

.github/workflows/versions.json
mwestphal: we agree on the oldest version.

Lets update to the latest one.

library/testing/TestTestSDKHelpers.cxx
mwestphal: codecov reports not reached.

library/testing/TestSDKHelpers.h
mwestphal: codecov reports not reached

library/src/context.cxx
mwestphal: codecov reports not reached

library/src/context.cxx
mwestphal: codecov reports not reached

.github/actions/mesa-install-dep/action.yml
mwestphal: I dont think so.

.github/workflows/ci.yml
mwestphal: ci: main is not supposed to run macos intel

library/testing/TestSDKHelpers.h
Meakk: reload the page

library/src/context.cxx
Meakk: reload the page

library/src/context.cxx
Meakk: reload the page

.github/actions/mesa-install-dep/action.yml
Meakk: ok let me confirm that.

.github/workflows/ci.yml
Meakk: it runs `commit` only

library/public/engine.h
Meakk: Yes but `create()` always instantiate cocoa since it's always available.

library/testing/TestTestSDKHelpers.cxx
Meakk: reload the page

.github/workflows/ci.yml
mwestphal: Its not what I see, can you point me where you see that ?

doc/dev/06-TESTING.md
mwestphal, line 1: 12-DESKTOP_INTEGRATION.md needs updating.

library/public/engine.h
mwestphal: got it, does `--list-rendering-backends` works as expected ?

mwestphal: alright, it seems to be tested on macOS :)

.github/workflows/ci.yml
mwestphal: got it. please add a comment explaining why we use osmesa here

library/testing/TestTestSDKHelpers.cxx
```diff
@@ -12,6 +12,18 @@ int TestTestSDKHelpers([[maybe_unused]] int argc, [[maybe_unused]] char* argv[])
...
+  {
+    try
+    {
+      std::ignore = TestSDKHelpers::CreateOffscreenEngine(backend);
+    }
+    catch (...)
```
mwestphal, line 22: not reachable ? ^^

library/testing/CMakeLists.txt
mwestphal: ok, but should it be used in everry single TestSDK ?

Currently we are testing with `auto` even when we force at global CI level.

.github/workflows/ci.yml
Meakk: my bad, I'm wrong

mwestphal (commented):
More questions.

library/testing/TestTestSDKHelpers.cxx
Meakk: lol no

library/testing/CMakeLists.txt
Meakk: Ideally yes, but some test have their own engine creation logic so I cannot really generalize. `TestSDKNotification` doesn't care about the engine type though.

.github/actions/mesa-install-dep/action.yml
Meakk: Here's a job setting Python for Mesa here: https://github.com/f3d-app/f3d/actions/runs/28663327444/job/85012073175#step:3:2647
And then Python reinstalled here: https://github.com/f3d-app/f3d/actions/runs/28663327444/job/85012073175#step:3:7572

I can't tell what it does, but it works

library/testing/CMakeLists.txt
mwestphal: >  but some test have their own engine creation logic 

Indeed, and for those thats ok, but isnt there a "classic generic" case we could handle ?

.github/actions/mesa-install-dep/action.yml
mwestphal: my guess is that the second one is useless. We should add python install stuff in the generic depencies.yml so it happens before mesa, and do it only once.

library/testing/CMakeLists.txt
Meakk: done

.github/actions/mesa-install-dep/action.yml
Meakk: Got it.
I think it means we'll install Python in many places just in case Mesa needs to be built but we have no choice.

.github/actions/coverage-ci/action.yml
```diff
@@ -28,10 +28,19 @@ runs:
...
         cache_postfix: cache-0
 
+    # Python is needed in case Mesa must be built
+    - name: Extract python version
+      id: python_ver
+      uses: f3d-app/extract-version-action@main
+      with:
+        file: ${{inputs.versions_file}}
+        name: python
+
```
mwestphal, line 38: should be done just before the install, do not add an input param to generic-dependencies.yml

.github/actions/mesa-install-dep/action.yml
```diff
@@ -0,0 +1,145 @@
...
+
+    - name: Extract mesa version
+      id: mesa_ver
+      uses: f3d-app/extract-version-action@main
+      with:
+        file: ${{inputs.versions_file}}
+        name: mesa
+        minver_label: ${{inputs.mindeps_label == 'mindeps' && 'minver' || ''}}
+
+    - name: Extract python version
```
mwestphal: not needed

.github/actions/generic-ci/action.yml
mwestphal: lets install pip only once

.github/actions/generic-dependencies/action.yml
```diff
@@ -25,6 +29,11 @@ runs:
       run: |
         [[ "${{ inputs.versions_file }}" ]] || { echo "versions_file is empty" ; exit 1; }
 
+    - name: Set up Python
+      uses: actions/setup-python@v6
+      with:
+        python-version: ${{ inputs.python_version }}
```
mwestphal, line 35: tbh it may be better to have an action dedicated to installing python and all pip installed deps

doc/user/12-DESKTOP_INTEGRATION.md
```diff
@@ -29,12 +29,17 @@ F3D supports rendering in environments with limited graphical contexts, includin
...
-  - `libGLX.so`
-  - `libEGL.so`
-  - `libOSMesa.so`
+- If EGL is available, you can use: `--rendering-backend=egl`. The following libraries are attempted to be loaded dynamically (in that order):
+  - `libEGL.so.1`, `libEGL.so` on Linux
+  - `libEGL.dll`, `EGL.dll` on Windows
+- If OSMesa is available, you can use: `--rendering-backend=osmesa`. The following libraries are attempted to be loaded dynamically (in that order):
+  - `libOSMesa.so.8`, `libOSMesa.so.6`, `libOSMesa.so` on Linux
+  - `osmesa.dll` on Windows
+  - `libOSMesa.8.dylib`, `libOSMesa.6.dylib`, `libOSMesa.dylib` on macOS
```
mwestphal, line 62: you are currently in `##Linux` subpart. This should be reworked imo.

library/testing/TestSDKScene.cxx
mwestphal: missing change

.github/actions/generic-ci/action.yml
mwestphal: not adressed

.github/actions/generic-ci/action.yml
mwestphal: not addressed

mwestphal (changes requested):
more changes

library/testing/TestSDKScene.cxx
Meakk: No the change broke the test so I revert

doc/user/12-DESKTOP_INTEGRATION.md
Meakk: ah right I missed that

.github/actions/generic-dependencies/action.yml
Meakk: I disagree because pip dependencies are not the same depending on the action.

.github/actions/coverage-ci/action.yml
Meakk: I don't think we can because `python-ci` requires a custom python installation not related to the versions file.

doc/user/12-DESKTOP_INTEGRATION.md
Meakk: I created a section for the rendering backends, let me know if it's better.

library/testing/TestSDKScene.cxx
mwestphal: its intersting that it broke it though, there may be an underlying issue ?

.github/actions/coverage-ci/action.yml
mwestphal: ha right, did not think of that

mwestphal: no sorry i dont get it, python-ci doesnt do anything special ?

Just install python once, as well as all pip needed dependencies , why not ?

.github/actions/generic-ci/action.yml
```diff
@@ -220,52 +215,37 @@ runs:
...
       shell: bash
       run: echo "F3D_CTEST_EXCEPTIONS=(test_scene_zero_copy)" >> $GITHUB_ENV
 
-    # Certain tests are failing on macOS with GPU pass through
-    # https://github.com/f3d-app/f3d/issues/792
-    # https://github.com/f3d-app/f3d/issues/1534
-    # https://github.com/f3d-app/f3d/issues/2268
-    # https://github.com/f3d-app/f3d/issues/3047
+    # Certain tests are failing on macOS software rendering
+    # OSMesa is used for VTK > 9.6, remove when it's dropped
```
mwestphal, line 233: so check the VTK version ?

mwestphal (changes requested):
still some questions

.github/actions/coverage-ci/action.yml
Meakk: That's what I did. It's just the extract python version that must be done at *-ci level.

.github/actions/generic-dependencies/action.yml
mwestphal: imo its not a big deal to install the pip deps all the time for everyone, at least everything is handled in the same location. We can just document each one. But ok to keep separate if you prefer.

.github/actions/coverage-ci/action.yml
mwestphal: indeed, my bad.

.github/actions/generic-dependencies/action.yml
Meakk: it's suboptimal, some dependencies can take a while to install (e.g. numpy). I prefer to keep separate.

.github/actions/generic-ci/action.yml
Meakk: I think it's better to check for the rendering backend rather than the VTK version because it's the important condition to exclude the tests or not.

library/testing/TestSDKScene.cxx
Meakk: That's because in the test a second engine is created with COCOA (auto) backend. The failure was happening on macOS only but I don't really want to investigate further why mixed rendering backends could fail.  
I changed the two backends in the test.

doc/user/12-DESKTOP_INTEGRATION.md
```diff
@@ -75,6 +47,37 @@ regsvr32 /u F3DShellExtension.dll
 
 There is no support for thumbnails on MacOS, the .dmg binary release provides automatic file openings.
 
+## Rendering backends
+
+F3D supports rendering in environments with limited graphical contexts, including sandboxes or without Xorg running.
```
mwestphal:
```suggestion
F3D supports rendering in environments with limited graphical contexts, including sandboxes or without Xorg running on Linux.
```

doc/user/12-DESKTOP_INTEGRATION.md
```diff
@@ -75,6 +47,37 @@ regsvr32 /u F3DShellExtension.dll
...
+  - `libEGL.so.1`, `libEGL.so` on Linux
+  - `libEGL.dll`, `EGL.dll` on Windows
+- If OSMesa is available, you can use: `--rendering-backend=osmesa`. The following libraries are attempted to be loaded dynamically (in that order):
+  - `libOSMesa.so.8`, `libOSMesa.so.6`, `libOSMesa.so` on Linux
+  - `osmesa.dll` on Windows
+  - `libOSMesa.8.dylib`, `libOSMesa.6.dylib`, `libOSMesa.dylib` on macOS
+- When using `--rendering-backend=auto`, F3D have the following logic:
+  - On Linux, check if a X server is running, in which case GLX is used. Otherwise, try to load EGL library and fallback on OSMesa if not available.
+  - On Windows, check if the OpenGL driver supports OpenGL 3.2, in which case WGL is used. Otherwise, try to load EGL library and fallback on OSMesa if not available.
+  - On macOS, always use COCOA.
```
mwestphal:
```suggestion
  - On macOS, always use COCOA. To use OSMesa, you must use `--rendering-backend=osmesa`.
```

Maybe ?

library/public/engine.h
mwestphal, line 90:
```suggestion
   * `VTK_DEFAULT_EGL_DEVICE_INDEX` allows its selection.
   * Not supported on macOS.
```

mwestphal, line 85: there is no `createCOCOA` ?

mwestphal (changes requested):
small changes

library/public/engine.h
Meakk: Not really useful because it would be equivalent to `create()`

mwestphal (approved):

PR #3411, opened 2026-07-24
Fix crash when no backend is available

### LINKED ISSUE #2522: Running f3d without glx, elg or osmesa trigger a segfault
**Describe the bug**
Run f3d without a valid rendering backend and it will fail with a segfault.
Trivial to reproduce with docker

**To Reproduce**

1.  run docker `ubuntu:24.04`
2. install nightly .deb package, do not install suggested packages
2. run f3d
3.  segfault

```
========== Configuring engine ==========
Generic Warning: In vtkOSOpenGLRenderWindow.cxx, line 149
libOSMesa not found. It appears that OSMesa is not installed in your system. Please install the OSMesa library from your distribution's package manager.


VTK window class type is vtkOSOpenGLRenderWindow
Segmentation fault (core dumped)
```

**Expected behavior**
A failure but no segfault

**System Information:**

- OS: Ubuntu 24.04
- GPU and GPU driver: None

**F3D Information**
Paste the content of `f3d --version` or `f3d-console.exe --version` on Windows: Nightly

**Additional context**

Meakk:
### Describe your changes

Automatic backend returns null is OSMesa is not available.

Fix https://github.com/f3d-app/f3d/issues/2522

### AI Disclosure

...

### Continuous integration

### DIFF
--- a/library/src/window_impl.cxx
+++ b/library/src/window_impl.cxx
@@ -106,7 +106,9 @@ public:
 #endif
 
     // OSMesa
-    return vtkSmartPointer<vtkOSOpenGLRenderWindow>::New();
+    vtkSmartPointer<vtkRenderWindow> osmesaRenWin = vtkSmartPointer<vtkOSOpenGLRenderWindow>::New();
+    osmesaRenWin->Initialize();
+    return osmesaRenWin->GetInitialized() ? osmesaRenWin : nullptr;
 #else
     // fallback on VTK logic for other systems
     return vtkSmartPointer<vtkRenderWindow>::New();


library/src/window_impl.cxx
mwestphal, line 105: shouldnt this warn that there is not return in the else case ?

Meakk: wdym?

mwestphal: its unrelated to your changes but it looks like:

```
    // try EGL
    vtkSmartPointer<vtkRenderWindow> eglRenWin = vtkSmartPointer<vtkF3DEGLRenderWindow>::New();
    if (eglRenWin)
    {   
      return eglRenWin;
    } 
```

Doesnt have a else, which should warn because there is no return ?

mwestphal (approved):
Does this still segfault when selecting

library/src/window_impl.cxx
Meakk: Why should we warn? If EGL is not available it should silently go to the next step (OSMesa)

Meakk:
> Does this still segfault when selecting

?

mwestphal:
> > Does this still segfault when selecting
> 
> ?

sorry, never meant to send this.

library/src/window_impl.cxx
mwestphal: right, i got confused in the defines

mwestphal (approved):

mwestphal:
Technically coverable by adding a dedicated test like I did in generic CI. So its already covered, just not reprorted as covered, as you see fit.

Meakk:
Are you sure? This would be covered only if GLX and EGL are not available but I think EGL is always available?

mwestphal:
> Are you sure? This would be covered only if GLX and EGL are not available but I think EGL is always available?

Indeed, you are right. not coverable. Technically covered in the Windows CI though ^^

PR #3424, opened 2026-07-29
Document coverage exclusion

Meakk:

### AI Disclosure

...

### Continuous integration

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,7 +108,8 @@ Please add only the labels required to work on your feature, in order to avoid u
 
 F3D continuous integration will also check the coverage as it is a good way to evaluate if new features are being tested or not.
 When adding code to F3D, always try to cover it by adding/modifying [tests](doc/dev/06-TESTING.md).
-Only if no coverage is possible, use `// LCOV_EXCL_START` and `// LCOV_EXCL_STOP` comments to exclude unreachable lines.
+Only if no coverage is possible, use `// LCOV_EXCL_START` and `// LCOV_EXCL_STOP` comments to exclude unreachable lines,
+and add another comment to explain why it cannot be reached by the coverage CI.
 
 F3D continuous integration also checks formatting using clang-format and other tools and will inform you if changes needs to be made.
 Some [formatting rules](doc/dev/09-CODING_STYLE.md) are not enforced by clang-format and will be checked during the review process.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,6 +108,7 @@ Please add only the labels required to work on your feature, in order to avoid u
 
 F3D continuous integration will also check the coverage as it is a good way to evaluate if new features are being tested or not.
 When adding code to F3D, always try to cover it by adding/modifying [tests](doc/dev/06-TESTING.md).
+Only if no coverage is possible, use `// LCOV_EXCL_START` and `// LCOV_EXCL_STOP` comments to exclude unreachable lines.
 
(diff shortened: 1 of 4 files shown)


mwestphal:
There are many other place where this should be put, do you plan to do it in another PR ?

Meakk:
> There are many other place where this should be put, do you plan to do it in another PR ?

Ideally at some point, but I wanted to add the documentation and some examples in the code.
Determining if an uncovered line is a miss or unreachable will take time.

mwestphal:
> > There are many other place where this should be put, do you plan to do it in another PR ?
> 
> Ideally at some point, but I wanted to add the documentation and some examples in the code. Determining if an uncovered line is a miss or unreachable will take time.

Thats fair.

Meakk:
0.1% coverage increase, what a PR!

CONTRIBUTING.md
```diff
@@ -108,6 +108,7 @@ Please add only the labels required to work on your feature, in order to avoid u
 
 F3D continuous integration will also check the coverage as it is a good way to evaluate if new features are being tested or not.
 When adding code to F3D, always try to cover it by adding/modifying [tests](doc/dev/06-TESTING.md).
+Only if no coverage is possible, use `// LCOV_EXCL_START` and `// LCOV_EXCL_STOP` comments to exclude unreachable lines.
```
snoyer: Should it be required to add a comment explaining why the code is unreachable (or what it would take to reach it, eg. "we'd need clipboard support in CI to cover this") when excluding sections?

Meakk: Fair point. Wdyt @mwestphal?

mwestphal: in my opinion, if there is no comments, it means this is not reachable **ever**. If there is a way, then a comment should explain how.

snoyer: Is there code that's objectively never reachable?
It feels like "unreachable in testing" would always be about some limit of the testing infrastructure, no?
Like if I commit some code I can't/won't reach I expect I would have to sell it to the maintainers that it's actually unreachable, so whatever that explanation is should probably be recorded as a comment instead of lost to time in a Github/Discord thread

Meakk: I tend to agree with @snoyer, it's still useful to explain why it's not coverable.

mwestphal: fair.

Meakk: Ok I'll add the comments and specify it's needed in the documentation.

mwestphal (changes requested):

library/src/engine.cxx
```diff
@@ -396,15 +396,17 @@ engine::state engine::state::fromClipboard()
   {
     if (!clip::get_text(content))
     {
-      // Unreachable in testing
+      // LCOV_EXCL_START
```
Meakk, line 400: @noclone can you remind me why these lines couldn't be covered?

noclone: it only happens in case of clipboard manager failure, which we can't produce in the CI

Meakk: but it can fail in real use case?

mwestphal:
@Meakk finish this ?

library/src/engine.cxx
Meakk: > but it can fail in real use case?

Ping @noclone

When can the clipboard manager fail?

noclone: sorry
get_text fails whenever there's no text to paste: empty clipboard, non-text content, or on X11 without a clipboard manager

mwestphal (approved):

Issue #856, opened 2023-06-15
Keep window on last closed position


Da3n0n:
**Context**
When F3D opens, the window is position is not set at all and left to the window manager to position it, unless `--position` is set.
It would be nice to be able automatically save the position the window was the last time and restore it

Same for size, aka `resolution`.

**Comments**
This require to actually save the info somewhere on disk. We do already save a cache so this could be added to a generic cache where we put info in a .json format, but more discussions needed.

**Notes**

Original issue below

these are obviously not necessity, but these two small things would make the experience so much better.

1.Dark Title bar Implementation. 

 https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/apply-windows-themes#enable-a-dark-mode-title-bar-for-win32-applications

2. Open F3D Window in Last closed position or Open in Center of Current Monitor.

Meakk:
1 should be fairly easy to implement, but it has to be done in VTK.  
2 would be interesting and that's not the first time someone ask for it.

mwestphal:
Shouldn't 2 be handled by the window manager already ? It seems implementable but i'd rather not if window manager provide the feature.

Meakk:
Definitely! We're always open to new ideas :)  
Please share @Da3n0n

Da3n0n:
well i don't want to just keep pumping you with features and make it overwhelming, but here's a few more off the top of the head Add View options to the viewport for orthographic views. 

drag and drop hdri into viewport to load it in the viewport would be really amazing or load HDRI option in the viewport would be nice either.

i will jsut leave these for now but i cna keep going 

I love this project and  even if yous dony implement these features will be still using f3d  and supporting it .if I had more experience I would go ahead and try implement these features myself.

mwestphal:
>    drag and drop hdri into viewport to load it in the viewport

Done in 2.1.0 release

mwestphal:
Rework this issue into the keep window on last closed position

Da3n0n:
Amazing Stuff Guys Really well down such a Fast Implementation of the HDRI Drag and Drop feature  Honestly Amazing i Love this Project So much.

mwestphal:
Size should be saved as well.

brendevr:
This would be especially useful because I am having issues on CachyOS with F3D opening on my secondary monitor in the bottom left. This issue would fix it.

mwestphal:
[...]

PR #3457, opened 2026-08-26
Record window geometry while window is still alive

noclone:
### Describe your changes

Record window geometry while window is still alive

https://github.com/f3d-app/f3d/issues/856

### AI Disclosure

...

### Continuous integration

### DIFF
--- a/application/F3DStarter.cxx
+++ b/application/F3DStarter.cxx
@@ -1303,7 +1303,18 @@ public:
     return geometry;
   }
 
-  /* Store the current window geometry in the cache so that the next run can restore it */
+  /* Record the current window geometry, which can only be queried while the window is still
+   * alive.
+   */
+  void RecordWindowGeometry()
+  {
+    const f3d::window& window = this->Engine->getWindow();
+    const auto [width, height] = window.getSize();
+    const auto [left, top] = window.getPosition();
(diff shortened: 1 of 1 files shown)


noclone:
\ci fast

mwestphal (approved):

snoyer:
Does this make WM calls on every frame of the event loop?
Would it be possible to only do it when the window is moved or resized ~~by listening to VTK's `ConfigureEvent`~~?

edit: VTK only fires `ConfigureEvent` on resize :( (even if on Linux for example it does see an `XConfigureEvent` on move)

noclone:
> Does this make WM calls on every frame of the event loop? Would it be possible to only do it when the window is moved or resized by listening to VTK's `ConfigureEvent`?

It does call the window manager every frame, but the call is very light, so I don't think that's a problem, and I don't see another way to do it anyway.

I think ConfigureEvent only covers resizing, not moving

snoyer:
I was double checking and editing my comment as you replied :)
I'll trust you on the cost of the calls, just wanted to double check as it _looks_ like a lot of work.

But yeah I agree there's not much choice if VTK doesn't expose an event. Unless it's done in the event loop but not _every_ frame? may not be worth the effort if the calls are cheap I guess. Oh well carry on then

Meakk (approved):

PR #3496, opened 2026-09-20
Fix pytest with optional modules

### LINKED ISSUE #3495: Running `pytest` fails when some modules are not enabled
**Describe the bug**
When running Python tests with `ctest`, test exclusion is working but not when testing wheels with `pytest`.

**To Reproduce**

From your virtual environment, and root of the project
1.  Run `pip install .`
2.  Run `pytest python/testing`

Some tests are failing:
```
FAILED python/testing/test_video.py::test_video_api - RuntimeError: Video encoder is not supported, please build f3d with FFMPEG support
```

**Expected behavior**
Module-dependent tests must be excluded at runtime when a module is disabled.
For that a `conftest.py` must be created in `python/testing` directory and fill it depending on `Engine.get_lib_info().modules`

**System Information:**

- OS:
- GPU and GPU driver:

**F3D Information**
Paste the content of `f3d --version` or `f3d-console.exe --version` on Windows:

**Additional context**

Meakk:
### Describe your changes

Add a CI for python wheels with no optional modules/plugins.
Add a `conftest.py` file to filter tests using inactive modules.

Fix https://github.com/f3d-app/f3d/issues/3495

### AI Disclosure

...

### Continuous integration


mwestphal (approved):

python/testing/conftest.py
```diff
@@ -0,0 +1,10 @@
+import f3d
+
+collect_ignore = []
+modules = f3d.Engine.get_lib_info().modules
+
+if "FFmpeg" not in modules:
+    collect_ignore.append("test_video.py")
+
+if "ImGui" not in modules:
+    collect_ignore.append("test_scene_zero_copy.py")
```
snoyer: `modules` is a `dict[str, bool]` so you want to check that the value is true not just that the key exists

```suggestion
if not modules.get("FFmpeg"):
    collect_ignore.append("test_video.py")

if not modules.get("ImGui"):
    collect_ignore.append("test_scene_zero_copy.py")
```

`not modules.get("FFmpeg")` will be false if either the value isn't true or if the key is missing altogether.
You can use `not modules["FFmpeg"]` instead if you want to crash on missing key as a way to make sure you're not testing for an invalid module (assuming all possible modules are always listed)

snoyer:
```suggestion
```

Alternatively you can skip the tests instead of preventing the collection with something like this in `test_video.py`
```py
import pytest

needs_FFmpeg = pytest.mark.skipif(
    not f3d.Engine.get_lib_info().modules.get("FFmpeg"), reason="no FFmpeg module"
)


@needs_FFmpeg
def test_video_api():
   ...
```

This way the fix would be in the test(s) own file(s) and the test would still show as skipped in the report instead of being ignored

Meakk: Much better, it avoids duplication with CMake! Thanks :)

Meakk:
Can you take another look @snoyer?

snoyer:
I can't speak for the CMake side but the pytest stuff looks nicer now :+1:
image

snoyer (approved):

mwestphal (approved):

